PR #5, opened 2015-06-03
Justin/win32 support

justinmchase:
Unfortunately I'm not familiar with eslint and am not sure how to get it to run so I haven't run the tests yet. I thought I'd throw it out there for now while I try to figure it out.


stefanjudis:
Nice!!! @justinmchase 

Thanks a lot!

I will check it out tomorrow morning at work ( I'm in Berlin/Germany ) and will set up a `contributing.md` with further instructions on how to deal with tests and stuff.

lib/win.js
```diff
@@ -49,6 +49,24 @@ var WinBuilder = {
       options.log('makensis error: ' + data);
     });
 
+    makensis.on( 'error', function ( err ) {
+      if (err.code === 'ENOENT') {
+        options.log( '- makensis not found.');
+        options.log( '  Please install "nullsoft scriptable install system"' );
+        switch(process.platform) {
+          case 'darwin':
```
stefanjudis: Can we add this 

```
brew install wine makensis
```

here?

lib/win.js
```diff
@@ -92,13 +110,16 @@ function _copyAssetsToTmpFolder( options ) {
 
 /**
  * Replace / with \ to write file
+ * also, ensure drive character is upper case when run on windows platforms
  * paths for windows
```
stefanjudis: :+1:

lib/macos.js
```diff
@@ -45,6 +48,7 @@ var MacosBuilder = {
     options.log( 'Wrote temporary ´appdmg.json´' );
     options.log( 'Kicking off ´appdmg´' );
 
+    var appdmg = require( 'appdmg' );
     var ee = appdmg( {
```
stefanjudis: May I ask why you moved the `require` call down?

stefanjudis:
@justinmchase

Had a quick look without testing it. 
Big thanks man! :+1: 

I left some notes. And also an update of the "Pre-requisites" section in the readme would be great to cover also the other platforms.

Will check it tomorrow in detail then. 

And I'm usually picky regarding coding style, so please don't take it personally when I mark some more stuff tomorrow ( I like spaces ;) ). :)

lib/macos.js
justinmchase: Sure, this package does not install successfully on windows, which is why I moved it into optionalDependencies. So when you require it on windows/linux it will throw. This module always gets loaded so it's not safe to require it always, only attempt to require it when you are building for osx and then you don't get a throw.

justinmchase:
@stefanjudis NP! Be picky I'll fix whatever you want. Also, I just confirmed that this works on linux as well (for building windows setup).

You have to apt-get wine and nsis, I'll add that in the pre-reqs page like you asked.

lib/macos.js
stefanjudis: And the assert is preventing the `require`, correct?

lib/win.js
```diff
@@ -49,6 +49,24 @@ var WinBuilder = {
...
+        options.log( '  Please install "nullsoft scriptable install system"' );
+        switch(process.platform) {
+          case 'darwin':
+          case 'win32':
+            options.log( '  http://nsis.sourceforge.net/Main_Page' );
+            break;
+          case 'linux':
+            options.log( '  $ sudo apt-get nsis' );
+        }
+        options.log();
```
stefanjudis: Empty log?

stefanjudis:
Thanks man!

You'll hear from me tomorrow then. :)

PS. Currently `npm.debug.log` is checked in. ;)

lib/macos.js
justinmchase: Yeah the assert is just stating that you can only attempt to build an osx installer when actually run from osx, it won't even get to the require below. Actually, now that I am more familiar with the code I should probably change that to: `if (process.platform != 'darwin') return callback(new Error('Invalid platform'));`

The assert will just throw and it is probably not very graceful.

lib/win.js
justinmchase: Done. Added apt-get instructions for linux too.

lib/win.js
justinmchase: I added the log so there was a space between the info printed here and the error that gets printed out next. The actual output looks like this:

```
> electron-builder ./release/win32/electron-updater-sample-win32 --platform=win32 --config=config.json --out=./release/win32

- Starting build for ´win32´ -
- makensis not found.
  Please install "nullsoft scriptable install system"
  http://nsis.sourceforge.net/Download

c:\Users\justin\projects\github\electron-builder\cli.js:30
    throw error;
          ^
```

Your call, I can leave it or remove it however you want.

lib/macos.js
stefanjudis: > `if (process.platform != 'darwin') return callback(new Error('Invalid platform'));`

Totally agree on that. :)

justinmchase:
I think I've figured out a way to make the mac .dmg file from linux also. I will test that tomorrow and probably make a new pull request for it.

If it works then this means you'll be able to create installers for mac and windows from a travis-ci build, which will be really nice.

stefanjudis:
@justinmchase Man, this sounds awesome. I'm setting up the contributing.md right now. :)

stefanjudis:
@justinmchase I pushed some updates to the readme and tweaked the `npm test` command a bit more to include also coding style rules. 

So after a quick merge master and `npm test` check we're ready to merge this PR.

Thanks!!! :tada:

justinmchase:
Nice. I'll do the linux build work in another branch now. Thanks for this :D

justinmchase:
I merged and ran eslint successfully.

stefanjudis:
@justinmchase Alright. Then let's do it! 

Thanks again for your effort. :bowtie:

lib/win.js
```diff
@@ -92,13 +115,16 @@ function _copyAssetsToTmpFolder( options ) {
...
+ * also, ensure drive character is upper case when run on windows platforms
  * paths for windows
  *
  * @param  {String} path path
  * @return {String}      windows path
  */
-function _windowsify( path ) {
-  return path.replace( /\//g, '\\' );
+function _windowsify( p ) {
+  p = path.win32.normalize(p);
```
stefanjudis: Hmm. Just discovered this fails on macos...

`path.win32 // undefined`

pyro2927: What version of node are you using? It relies on v0.12:
https://nodejs.org/api/path.html

stefanjudis: Ahhh. I was on v0.11.

Good call! Thanks. Will add it to readme also. :)

stefanjudis:
It's out.

https://www.npmjs.com/package/electron-builder

:sunglasses: 

My test for macos worked fine. If you could confirm, that'd be great. :blush:

PR #24, opened 2015-07-23
Concurrent build causes race condition

tengyifei:
Use unique installer.nsi name


stefanjudis:
Hui - sorry I missed this one. 

Looking fine for me. thanks. :)

PR #26, opened 2015-08-12
Feature(fileName): Add additional option to include a file name separate from the title.

jbavari:
First off, thank you very much for developing this tool. It's worked great.

I've thrown together a quick PR to add the option of passing a fileName to name the Mac OSX DMG.

If no 'fileName' is provided, use the title as before.

Ran tests and they still pass.

PR #28, opened 2015-08-30
Separated the application and executable name for Windows builds

AlicanC:
Let's say the name of our application is "Super Duper App" and the executable for it is "sda.exe".

Currently there is no way to keep that separation. You currently have two options:
- Change the application name to "sda"

This makes everything ugly. (Having "sda Setup.exe" instead of "Super Duper App Setup.exe", etc.) 

- Change the executable name to "Super Duper App.exe"

For this, I had to change a lot in my build process which was not that feasible.

### DIFF
--- a/lib/win.js
+++ b/lib/win.js
@@ -28,9 +28,10 @@ var WinBuilder = {
     options.log( '- Starting build for ´' + options.platform + '´ - ' );
 
     var configFilePath = writeConfigFile( 'installer.nsi', {
-      appPath : _windowsify( options.appPath ),
-      name    : options.config.win.title,
-      out     : _windowsify( options.out )
+      appPath    : _windowsify( options.appPath ),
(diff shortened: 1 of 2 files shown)


stefanjudis:
@AlicanC I think this is a duplicate of https://github.com/loopline-systems/electron-builder/pull/26 isn't it?

AlicanC:
I think #26 only changes the name of the output file. This also affects shortcuts and such.

stefanjudis:
Hmm, okay. maybe we want to discuss at #26. 

@jbavari PR is more complete and covers also the mac implementation. 

I've got no problems with including the shortcuts also in there.

stefanjudis:
I'm closing this one - as I'll prefer #26. 

But thanks. :)

AlicanC:
Yeah, sure.

PR #49, opened 2015-10-21
Clarify Linux install instructions

malept:
* Clarify that the Linux instructions are specifically for Ubuntu & its derived distributions
* Hint that `add-apt-repository` and `apt-get` usually require superuser privileges
* Other typographical changes in the vicinity


stefanjudis:
Thanks. :bowtie:

Issue #50, opened 2015-10-22
Add ability to specify your own installer.nsi.tpl


vxsx:
If i want my app on windows to run only as admin i would need to add 
```
WriteRegStr HKLM "Software\Microsoft\Windows NT\CurrentVersion\AppCompatFlags\Layers" \
                   "$INSTDIR\${APP_NAME}.exe" "RUNASADMIN"
```
to the installer.nsi, but at the moment it's only possible to do by modifying the template file inside of the module.
It would be great to have a possibility to specify a path to your own template to be able to make any customisations you want.

If you're interested in this functionality I'll be able to submit a PR around the weekend.

stefanjudis:
Sounds great - I'd love to take a pr solving this. As this is solving another issue into the list. :)

Thx.

PR #51, opened 2015-10-24
add ability to specify path to custom installer.nsi

vxsx:
#50

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,9 +24,9 @@ var Builder = {
    */
   build : function( options, callback ) {
 
-    options     = options || {};
-    options.log = options.log || console.log;
-    options.out = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
+    options             = options || {};
+    options.log         = options.log || console.log;
+    options.out         = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
     options.nsiTemplate = options.nsiTemplate || path.join( __dirname, 'templates/installer.nsi.tpl' );
 
     options.log(
--- a/index.js
+++ b/index.js
@@ -27,6 +27,7 @@ var Builder = {
     options     = options || {};
     options.log = options.log || console.log;
     options.out = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
+    options.nsiTemplate = options.nsiTemplate || path.join( __dirname, 'templates/installer.nsi.tpl' );
 
     options.log(
       '- Running electron-builder ' + require( './package' ).version
--- a/index.js
+++ b/index.js
@@ -24,10 +24,9 @@ var Builder = {
    */
   build : function( options, callback ) {
 
-    options             = options || {};
-    options.log         = options.log || console.log;
-    options.out         = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
-    options.nsiTemplate = options.nsiTemplate || path.join( __dirname, 'templates/installer.nsi.tpl' );
+    options     = options || {};
+    options.log = options.log || console.log;
+    options.out = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
 
     options.log(
       '- Running electron-builder ' + require( './package' ).version
--- a/usage.txt
+++ b/usage.txt
@@ -8,3 +8,4 @@ Usage
 
   Optional options:
     out:               path to output the installer (path must exist)
+    nsiTemplate:       path to override the installer.nsi template for win platform (path must exist)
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -27,6 +27,7 @@ var Builder = {
     options     = options || {};
     options.log = options.log || console.log;
     options.out = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
+    options.nsiTemplate = options.nsiTemplate || path.join( __dirname, 'templates/installer.nsi.tpl' );
```
stefanjudis: Can you please follow coding style here ( aligned `=` )?

stefanjudis:
@vxsx 

Except a little coding style issue  this looks good to me.

The only thing I'm asking myself is, if it is needed to pipe a custom NSI file trough the templating engine.

It feels a bit weird - doesn't it?

index.js
vxsx: oops, missed this one, fixed

vxsx:
it is a bit weird now that I think about it, but I think it's better to keep things consistent with the default template. and after all, enduser can easily hardcode things in it if he wants, templating is not that intrusive :)

stefanjudis:
@vxsx 

I just got my head around it a bit more. I think we can do it like that - but are there ways around setting a custom file?

I'm thinking of several config options or even areas where user can place command easily.
I'm just thinking out loud here - but maybe we can brainstorm a bit. :)

vxsx:
I think there are pros and cons to each approach. 

### 1. Options and customisable areas in config:
#### pros
some level of control over things
sounds simpler for the end user

#### cons
have to add every little bit as an option/area to the config (only a con if there are a lot more options, depends if the demand is high or low)
depending on demand (or perfectionism) - takes time to implement

### 2. Custom file:
#### pros
complete control
easy to implement (PR is right here :D)
not really a problem to add areas and options on top

#### cons
have to know what you're doing (harder for the end user)
maybe there would be issues about "how do i do this or that"
adding options on top may not be very good for the api
going away from this initial solution would be a backwards incompatible change

---
i probably missed lots of things

For me second approach is better, because digging in the source code and copying and adapting a template is fairly easy for me personally, and I kinda like the complete control thing. It also means that a problem is solved with minimum effort. Longterm i think combining approach 2 and 3 would probably be better in general, but it's gonna cost more time, and the answer is really "it depends".

Up to you to decide :)

Brankub:
+1 for @vxsx changes. Have to change .tpl file in electron-builder folder to add register settings.

FinalAngel:
:+1:

vxsx:
did you have a chance to think about the path you want to take with this @stefanjudis ?

index.js
```diff
@@ -24,9 +24,10 @@ var Builder = {
    */
   build : function( options, callback ) {
 
-    options     = options || {};
-    options.log = options.log || console.log;
-    options.out = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
+    options             = options || {};
+    options.log         = options.log || console.log;
+    options.out         = options.out ? path.resolve( process.cwd(), options.out ) : process.cwd();
+    options.nsiTemplate = options.nsiTemplate || path.join( __dirname, 'templates/installer.nsi.tpl' );
```
stefanjudis: As this windows specific this should be treated inside of the windows installer file, or?

usage.txt
```diff
@@ -8,3 +8,4 @@ Usage
 
   Optional options:
     out:               path to output the installer (path must exist)
+    nsiTemplate:       path to override the installer.nsi template for win platform (path must exist)
```
stefanjudis: This should also be handled in inside of the config file, or?
As it is really windows specific. :)

stefanjudis:
Okay, okay. :bowtie: 

I see a few people want this - so let's do it. :) 

Sorry for the delay on my side. :(

I made a few notes to the actual PR.
And I will have a further look, when these are done.

One additional note I have is, that we should try to engage people using a different nsi template to think of their nsi template and ask themselves if it might be useful for others. 

There might be some additions which we can add to the "base"-template.

Thanks a lot folks for your effort. :)

usage.txt
vxsx: yeah, makes sense. the path would be relative to the place you run the command / script from then, like with other asset paths, and not to the config (which may be a bit confusing for the user, but it's the way it is and still is fairly simple to understand)

stefanjudis: > the path would be relative to the place you run the command / script from then

It should handle relative and absolute paths. And currently it's supporting also both. So that shouldn't be a big deal :)

vxsx:
@stefanjudis updated the PR. not entirely sure about what to add to the readme beside the config option though

stefanjudis:
@vxsx Alright. Looks good. :)

Thanks a lot. ;)

vxsx:
@stefanjudis thank you :)

PR #57, opened 2015-11-06
Update desktop shortcut to use icon.ico

aaronfrost:
With these extra params, the Desktop shortcut that is created will have the icon of the icon provided in the config file. Without this change, I couldn't get the desktop icon to use the icon that I was providing in the config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Your application mount icon.
 Sizes of the icons included in `dmg` dialog.
 
 ### `osx.contents`
-Icons you wish to include in `dmg` dialog.
+This property contains the configuration for the OSX dmg window. This property is passed to `appdmg`, which builds the dmg package. For a deeper explanation of the different options that you can set in this property, visit [`appdmg`'s page](https://www.npmjs.com/package/appdmg).
 
 ### `win.title`
 Title of your application shown in generated windows installer.
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Your application mount icon.
 Sizes of the icons included in `dmg` dialog.
 
 ### `osx.contents`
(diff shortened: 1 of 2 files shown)


stefanjudis:
Looks good to me.

Which windows versions did you check?

aaronfrost:
Good question!!! Sorry for not mentioning that. I checked it on Windows 7. Once I added that, the icon of the link on the desktop changes from the default installer icon to the icon provided by my config.json file. Do you need me to check if it works on all versions of Windows? Or which ones should I check this on for you?

stefanjudis:
That's the funny part. 

I checked windows xp, 7 and 8 and it was working fine... 

> Do you need me to check if it works on all versions of Windows?

Well... if you could check a second version that would be great. Just as a safety check. 
But for me it looks really uncritical.

Thanks for taking the time. :)

aaronfrost:
I will DL another vm and check it out. Good points. I will let you know
what I find.

On Fri, Nov 6, 2015 at 1:59 AM Stefan Judis <notifications@github.com>
wrote:

> That's the funny part.
>
> I checked windows xp, 7 and 8 and it was working fine...
>
> Do you need me to check if it works on all versions of Windows?
>
> Well... if you could check a second version that would be great. Just as a
> safety check.
> But for me it looks really uncritical.
>
> Thanks for taking the time. :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/loopline-systems/electron-builder/pull/57#issuecomment-154348634>
> .
>
-- 
--
AF
@js_dev
aaron@frosty.io

aaronfrost:
Let me squash the commits and resubmit pull request.

stefanjudis:
This one is obselete - as is came with a different PR.

PR #59, opened 2015-11-06
Updated "How we used it so far" part of README.md

Frondor:
Fix: changed `rm -rf` (which doesn't work on win7 x86) for `rimraf`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,9 +180,9 @@ package.json
   "scripts": {
     "dev": "electron ./app",
 
-    "clean": "rm -rf ./dist",
-    "clean:osx": "rm -rf ./dist/osx",
-    "clean:win": "rm -rf ./dist/win",
+    "clean": "rimraf ./dist",
+    "clean:osx": "rimraf ./dist/osx",
+    "clean:win": "rimraf ./dist/win",
 
     "build": "npm run clean && npm run build:osx && npm run build:win",
     "build:osx": "npm run clean:osx && electron-packager ./app \"Loopline Systems\" --out=dist/osx --platform=darwin --arch=x64 --version=0.25.3 --icon=assets/osx/loopline.icns",


stefanjudis:
Thx. :bowtie: 

You might want to explain why rimraf is needed and point out how it should be installed?
There might be people not knowing why why it should be `rimraf` and not `rm -rf`...

Frondor:
Sure, I'll try to reproduce the error and google the SO question that helped me because I can't remember it right now

stefanjudis:
Nice, thanks a lot. :)

Frondor:
Well, this seems most likely to be a windows problem. so I don't think that `clean` should use `rimraf`. You can reject this PR and I can make a new one.
The problem was that my build process was failing at this [line](https://i.gyazo.com/6cfa00154430a9751961dfcc689f2143.png) so after a google search I found [this issue](https://github.com/react-bootstrap/react-bootstrap/issues/491) which is solved by switching from `rm -rf` to `rimraf` :neckbeard:

stefanjudis:
Ja I think a not somewhere for windows users is maybe enough, telling that they don't have the `rm -rf` command and should use the windows counterpart instead.

What do you think? :)

Frondor:
I'm the noob here :laughing: 
The only problem with changing `clean:win` only is that `build` task won't work because `clean` and `clean:osx` are still using `rm -rf`.
As long as `rimraf` works for both `osx` and `win`, I think there won't be any problem since both commands do the same job.
p.s.: Didn't understand that about *"I think a not somewhere for windows users is maybe enough" * :/

stefanjudis:
> I'm the noob here :laughing: 

no problem about that.

The problem is not the system you want to build for it's the system you want build on.

If you want to use `rm -rf` building anything it's not working.

Also `clean` is not part of `electron-builder` itself it's just a recommendation showing how we use it so far.

So a note like `If you want to clean you`re stuff and am on a windows machine please adjust rm -rf to use the windows command. Using libraries like rimraf may help to make it run on different environments`.

Frondor:
Alright, seems good to me... Maybe in the future there should be two different points (bullets) inside "How we used it so far", one `package.json` for OSX, and another for windows. Because the build processes vary a bit between both OS.
Shall I make a new PR or you are fine?

stefanjudis:
Jap go ahead then. :) Thx.

PR #60, opened 2015-11-06
Updated "How we used it so far" part of README.md

Frondor:
Added note for windows users as discussed here: #59


stefanjudis:
:+1:

PR #65, opened 2015-11-12
nsi.template update to fix missing desktop icon

aaronfrost:
Added some extras to the nsi template. When I added this, the desktop icon of the installed app was the icon that I provide, and not the default Electron icon. Without this, the app installs and the desktop icon is the Electron icon. So, this was a definite upgrade. I tested this on Win7 and Win10.


aaronfrost:
I tested this on both Win7 and Win10. So, I feel even more confident that it is a good thing. 

Without it, I get a desktop icon, but the icon image is that which is provided by default by Electron. When I add it, then the app installs and the desktop icon is one that matches the icon I provide.

I ran that test on both mentioned version of Windows. Thoughts?

stefanjudis:
No that's fine.

I think I'll merge it tomorrow. :bowtie: 

Thank you very much Sir. :+1:

PR #74, opened 2015-11-21
Add keywords to improve discoverability

etiktin:
It's difficult to find `electron-builder` on [npmjs.com](https://www.npmjs.com/). For example if you search for [electron installer](https://www.npmjs.com/search?q=electron+installer), it's not even listed.
I added a bunch of keywords that users might use when trying to find a solution for building/generating Electron installers. It should improve the discoverability of this project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,23 @@
     "url": "https://github.com/loopline-systems/electron-builder.git"
   },
   "keywords": [
-    "electron"
+    "electron",
+    "builder",
+    "build",
+    "installer",
+    "install",
+    "packager",
+    "pack",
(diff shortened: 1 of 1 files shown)


stefanjudis:
Cool thanks. :+1: 

Quick question: should `msi` be `nsi`?

etiktin:
I meant `msi`. It's a popular installer format for Windows. If someone's looking for it with some other keywords, it probably means he/she is looking to create an installer for an Electron app on Windows.

stefanjudis:
Okay fair enough. :)

Thanks. :bowtie:

PR #76, opened 2015-11-24
Added support for windows  Unistall/DisplayVersion

demetris-manikas:
Very small patch to allow for version property in the nsi templates.
Let me know what you think.
Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ config.json.sample:
   },
   "win" : {
     "title" : "Loopline Systems",
+    "version" : "x.x.x.x",
     "icon" : "assets/win/icon.ico",
     "nsiTemplate" : "path/to/custom/installer.nsi.tpl"
   }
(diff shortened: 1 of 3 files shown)


stefanjudis:
Thanks. :bowtie: 

Fine for me - but it's missing the readme mention about that?

demetris-manikas:
Updated the README.md but you probably have to rephrase it.
If version is not defined APP_VERSION is just an empty string

stefanjudis:
@demetris-manikas 
Now that I'm thinking of it defaulting to version in package.json would be nice maybe?

demetris-manikas:
I played with the idea but it needed more effort. Where is the package.json located? In my case it 's not even in the same dir as config.json. So things get complicated. Setting the path to the package.json would be a way. Should that be a global or a win param? These question made me go for the simple solution. Anyway it is very easy to produce a config.json containing the appropriate data.

stefanjudis:
Okay - I see. Fair enough. :bowtie:

PR #84, opened 2015-12-02
Stop process if already running

demetris-manikas:
If the application is already running the installation fails so just kill it. 
Maybe it would be a better solution to show some info and drop the installation but my knowledge about NSIS is still quite limited so as to implement this kind of behavior.

### DIFF
--- a/templates/installer.nsi.tpl
+++ b/templates/installer.nsi.tpl
@@ -29,31 +29,25 @@ InstallDir "$PROGRAMFILES\${APP_NAME}\"
 !insertmacro MUI_PAGE_FINISH
 !insertmacro MUI_LANGUAGE "English"
 
-!macro CheckAppRunning ABORT_MSG
-  test:
-
+!macro CheckAppRunning MODE
   ${nsProcess::FindProcess} "${APP_NAME}.exe" $R0
   ${If} $R0 == 0
-    MessageBox MB_ABORTRETRYIGNORE|MB_ICONEXCLAMATION "${APP_NAME} is already running." /SD IDABORT IDRETRY retry IDIGNORE doStopProcess
-        Abort "${ABORT_MSG}"
-    retry:
-      goto test
-    doStopProcess:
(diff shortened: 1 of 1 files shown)


stefanjudis:
> Is it possible to include a button on a dialog in case the user has something going before killing the running app?

That sounds great!!! :+1:

demetris-manikas:
Any comments on changed implementation?

stefanjudis:
Damn it. Sorry - I'll try to have a look at it tonight. :)

stefanjudis:
@demetris-manikas Just tried it and it looks good to me.

Only thing:

Can we make the dialog a bit more explanatory?

screenshot 2015-12-09 21 00 17

I'm thinking of something like. 

"App is already running. This may affect the installation process. Please close the application."

And I think the cta's could be better.

Also it'd be great if "ignore" could be "close & continue" showing, that this will shut down the application.

What do you think?

mkaz:
The solution I ended up going with was a  two button OK / CANCEL Message box.
Here's that code I used:

```
MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION "${APP_NAME} is running. $\r$\nClick OK to close it and continue with install" /SD IDCANCEL IDOK doStopProcess
  	    Abort
    doStopProcess:
         DetailPrint "Closing running ${APP_NAME} ..."
         ${nsProcess::KillProcess} "${APP_NAME}.exe" $R0
         DetailPrint "Waiting for ${APP_NAME} to close."
         Sleep 2000
```

The result looks like:

stop-dialog-2

stefanjudis:
@mkaz @demetris-manikas 

I have to say - I actually like @mkaz 's version more. 

What's do you two think?

demetris-manikas:
It 's fine by me. I only wanted the installer not to hang when the app is already running.
Your call.

demetris-manikas:
Incorporated changes proposed by mkaz.

stefanjudis:
:+1: Looks good to me thanks!

PR #87, opened 2015-12-10
Stop Process if Running

mkaz:
To support @demetris-manikas PR #84 
Switch to two button Ok - Cancel Message Box


stefanjudis:
@mkaz I took #84 as it covered also `uninstall`.

But anyways folks - @mkaz @demetris-manikas thanks for contributing. :)

PR #108, opened 2015-12-22
fix invalid json format in readme for config.json

baikon:
if users copy and paste the example config.json in the readme, they will run into a build error due to ``// optional`` comments:

```
Error: Could not load config file
```

This pull request just removes those comments so that it turns it back into valid json format

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,8 +112,8 @@ config.json.sample:
     "version" : "x.x.x.x",
     "icon" : "assets/win/icon.ico",
     "verbosity": 1,
-    "nsiTemplate" : "path/to/custom/installer.nsi.tpl", // optional
-    "fileAssociation": { // optional
+    "nsiTemplate" : "path/to/custom/installer.nsi.tpl",
+    "fileAssociation": {
       "extension": ".loop",
       "fileType": "Loopline Systems File"
     }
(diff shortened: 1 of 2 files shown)


stefanjudis:
Nice thanks. :+1: 

I was also thinking of extending this line here

https://github.com/loopline-systems/electron-builder/blob/master/index.js#L69.

Currently it's passing a new error object. Enriching it with the message of the thrown error would be more friendly. `JSON.parse` errors happen all the time and with a proper message you would have found the problem much quicker. :)

Possible to sneak it in here? :blush:

No problem if not. :)

baikon:
Not a problem, errors such as this are throwing now:

``ERROR---SyntaxError: /Users/username/public/config.json: Unexpected token /``

stefanjudis:
It shouldn't throw there...

According to node convention and the fact that we can expect an error there we should pass the error to the `callback` function rather then blowing up completely. Modules on top of `electron-builder` like `grunt-electron-builder` might wanna decided themselves how to treat the error. :)

Convention -> `callback( error, result )`

I was more thinking of a nicer error combining a message and the exception passed to the callback.

E.g. `callback( new Error( 'Could not load config file (' +  error.message + ')' ) );`

What do you think?

stefanjudis:
Closing here due to inactivity and solving it myself. :)

PR #121, opened 2016-01-24
Add option win.encoding for .nsi templates, dealing with Umlauts or chars with accent.

Eusebius1920:
As the `makensis` does not support full utf8 encoding, I added the possibility (via `iconv`) to specifiy a encoding which should be exported by electron-builder before invoking `makensis`.
In your `packager.json` just specify `win.encoding`. (e.g. `win.encoding = "ISO-8859-1"`)
This allows usage of Umlauts ö,ä,ü or accents without pain.

If there is any interest in this you can just merge with this request. I based it upon `0.2.6, if you need a version based on master (which is not running for me atm) then let me know.

**Edit:**
Sorry for the confusion. I deleted the first pull request, because I did the commits with the wrong user set up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,12 @@ Option to define a custom file association on Windows.
 Caution: when you use `win.nsiTemplate` option, `win.fileAssociation` option should only work
 if the custom nsi template is based on the original one.
 
+### `win.encoding` *( optional )*
+Option to define a custom encoding for output template.nsi on Windows.
+Default value is 'UTF-8'.
+For available values see [node-iconv's page](https://github.com/bnoordhuis/node-iconv).
+The packager.json config should always be UTF8 encoded.
+
 **Note:** You need to add something that might have value for others? Please consider a PR. ;)
 
 
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ if the custom nsi template is based on the original one.
 ### `win.encoding` *( optional )*
 Option to define a custom encoding for output template.nsi on Windows.
 Default value is 'UTF-8'.
-For available values see [node-iconv's page](https://github.com/bnoordhuis/node-iconv).
+For available values see [ashtuchkin/iconv-lite's page](https://github.com/ashtuchkin/iconv-lite/wiki/Supported-Encodings).
 The packager.json config should always be UTF8 encoded.
 
 **Note:** You need to add something that might have value for others? Please consider a PR. ;)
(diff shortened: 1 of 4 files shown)


stefanjudis:
@Eusebius1920. Can you please check why Travis is failing?

Eusebius1920:
The used image seems not capable of compile iconv, which needs native extensions. If this is a problem, I could try to switch to iconv-lite which apparently doesn't need them.

stefanjudis:
@Eusebius1920 Ja unfortunately I can not take it right it is... It should run easily on node 4 & 5.

If `iconv-lite` is solving this I can check again. :)

Eusebius1920:
I will change it later this day to iconv-lite and push again. 
Maybe you should mention in your `CONTRIBUTING.md` that one can not use modules with native extensions ;)

stefanjudis:
> Maybe you should mention in your CONTRIBUTING.md that one can not use modules with native extensions ;)

I think travis is pointing that out clearly and it should be the goal to run cross platform.
Don't want to add more points there, as it is detected anyways... A PR with failing builds will be merged rarely. :)

Eusebius1920:
I switched to `iconv-lite`, now travis CI should go through just fine and one can use characters (like ö,ä,ü,ß) apart from ascii in the nsis installers (e.g by using `ISO-8859-1` encoding for the template).

README.md
```diff
@@ -160,6 +160,12 @@ Option to define a custom file association on Windows.
 Caution: when you use `win.nsiTemplate` option, `win.fileAssociation` option should only work
 if the custom nsi template is based on the original one.
 
+### `win.encoding` *( optional )*
+Option to define a custom encoding for output template.nsi on Windows.
+Default value is 'UTF-8'.
+For available values see [ashtuchkin/iconv-lite's page](https://github.com/ashtuchkin/iconv-lite/wiki/Supported-Encodings).
+The packager.json config should always be UTF8 encoded.
```
etiktin: Does it have to contain the utf8 BOM (EF BB BF), or will it also work without it?
What ever is supported should probably be stated explicitly in the doc.

Eusebius1920: I changed that in README.md in the last commit, as I did not mean the packager.json but the used nsi template.

I think it is just as always with nodejs. As the BOM is not relevant for utf8 you can do either way If you supply a BOM, nodejs (At least this is what newer version of node do...) will just strip it in the decoded string. The same goes for writing an utf8 string back to disk with node: As the BOM is not relevant for utf8, node will not write it to the file.
(However I did not change anything with respect to this aspect in this pull request. Reading the template-file is just like before: A fs.readFile call of node's file module, reading an utf8 encoded file from disk. I just added the option to not dump it back as utf8 to the disk but encode it proir to this with a different encoding.)

develar:
@stefanjudis Could we close it as Obsolete?

stefanjudis:
Yeah Let's do it. :)
Am 11.03.2016 1:21 nachm. schrieb "Vladimir Krivosheev" <
notifications@github.com>:

PR #139, opened 2016-02-10
fix grammar and clarify electron-packager dependency

### LINKED ISSUE #137: Make dependency and flow clearer
People struggle with the instruction and the flow combination with `electron-packager`.
It should be made clearer.

cc. @citrusui

citrusui:
Fixes #137.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,47 +3,51 @@
 
 # electron-builder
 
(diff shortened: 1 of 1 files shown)


stefanjudis:
Thanks. @citrusui

PR #141, opened 2016-02-11
Linux deb: version and arch should be replaced in the out file name, …

develar:
Version and arch should be replaced in the out file name. Otherwise it is not possible to build package for several arch (because name will be the same).

Full name must be used to generate temp directory name, otherwise it is not possible to build in parallel (electron-complete-builder tests (AVA test runner) and build performed in parallel to speed up).

### DIFF
--- a/lib/linux.js
+++ b/lib/linux.js
@@ -41,18 +41,23 @@ var LinuxBuilder = {
       linux.archName = 'x86_64';
     }
 
-    var tmpFolder = path.join( os.tmpDir(), linux.executable );
+    var outFilename = linux.executable + '-' +
+                      linux.version + '-' + linux.archName + '.' +
+                      linux.target;
+    var tmpFolder = path.join( os.tmpDir(), outFilename );
     _createFolders( tmpFolder );
 
     _createDesktopEntry( tmpFolder, linux );
 
     var scripts = _createScripts( linux );
 
-    _copyApp( options, tmpFolder );
+    fs.copySync( options.appPath,
+      path.join( tmpFolder, 'opt', options.config.linux.executable ) );
 
-    _buildPackage( options, scripts, tmpFolder );
+    var destination = path.join( options.out, outFilename );
+    _buildPackage( options, scripts, tmpFolder, destination );
 
-    callback();
+    callback( null, destination );
   }
 };
 
@@ -62,12 +67,8 @@ var LinuxBuilder = {
  * @param {String} tmpFolder  folder where the packaged app is built
  */
 function _createFolders( tmpFolder ) {
-  fs.removeSync( tmpFolder );
-  fs.mkdirpSync( tmpFolder );
-  var usrShare = path.join( tmpFolder, 'usr', 'share' );
-  fs.mkdirpSync( path.join( usrShare, 'applications' ) );
-  fs.mkdirpSync( path.join( usrShare, 'icons', 'hicolors' ) );
-  fs.mkdirpSync( path.join( tmpFolder, 'opt' ) );
+  fs.emptyDirSync( tmpFolder );
+  fs.mkdirpSync( path.join( tmpFolder, 'usr', 'share', 'icons', 'hicolors' ) );
 }
 
 /**
@@ -90,7 +91,7 @@ function _createDesktopEntry( tmpFolder, linux ) {
   var tplFile = fs.readFileSync( desktopEntryTemplate, 'utf8' );
   var tpl = template( tplFile );
   var merge = tpl( linux );
-  fs.writeFileSync( dest, merge, 'utf8' );
+  fs.outputFileSync( dest, merge );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


lib/linux.js
```diff
@@ -90,7 +91,7 @@ function _createDesktopEntry( tmpFolder, linux ) {
   var tplFile = fs.readFileSync( desktopEntryTemplate, 'utf8' );
   var tpl = template( tplFile );
   var merge = tpl( linux );
-  fs.writeFileSync( dest, merge, 'utf8' );
+  fs.outputFileSync( dest, merge );
```
develar: Use power of fs-extra — avoid explicit mkdir.

lib/linux.js
```diff
@@ -62,12 +67,8 @@ var LinuxBuilder = {
  * @param {String} tmpFolder  folder where the packaged app is built
  */
 function _createFolders( tmpFolder ) {
-  fs.removeSync( tmpFolder );
-  fs.mkdirpSync( tmpFolder );
-  var usrShare = path.join( tmpFolder, 'usr', 'share' );
-  fs.mkdirpSync( path.join( usrShare, 'applications' ) );
-  fs.mkdirpSync( path.join( usrShare, 'icons', 'hicolors' ) );
-  fs.mkdirpSync( path.join( tmpFolder, 'opt' ) );
+  fs.emptyDirSync( tmpFolder );
```
develar: Use power of fs-extra — emptyDir does the job.

lib/linux.js
```diff
@@ -41,18 +41,23 @@ var LinuxBuilder = {
...
-    _copyApp( options, tmpFolder );
+    fs.copySync( options.appPath,
+      path.join( tmpFolder, 'opt', options.config.linux.executable ) );
 
-    _buildPackage( options, scripts, tmpFolder );
+    var destination = path.join( options.out, outFilename );
+    _buildPackage( options, scripts, tmpFolder, destination );
 
-    callback();
+    callback( null, destination );
```
develar: Callback returns out path (as electron-packager does). It is used in electron-complete-builder to publish artifact.

lib/linux.js
```diff
@@ -41,18 +41,23 @@ var LinuxBuilder = {
       linux.archName = 'x86_64';
     }
 
-    var tmpFolder = path.join( os.tmpDir(), linux.executable );
+    var outFilename = linux.executable + '-' +
```
develar: @stefanjudis Maybe we can set maximum line length to 160? It is a bit of annoying to use 80.

lib/linux.js
```diff
@@ -118,29 +119,15 @@ function _createScripts ( linux ) {
 }
 
 /**
- * Copy the app in the given folder
```
develar: Sorry, I don't see any reason to have one-line method here.

develar:
Also, we should not use `sync`. I am going to get rid of them. `electron-complete-builder` tests and build are performed in parallel, so, it is a showstopper. It will be part of another pull request, but don't you mind if I rewrite this file to use promises?  `electron-complete-builder` uses `async-await` and async code is clear and readable, but as a first step, I will just use promises without involving babel processing.

stefanjudis:
As a CLI utility there are no problems with `sync`, are they?

Or do you use it as a node module? 

I will have a look as soon as I reach the office. Thanks. :)

develar:
> As a CLI utility there are no problems with sync, are they?

No. Even more — to overcome such issues AVA run each test file in a separate Node.js process.

> Or do you use it as a node module?

Yes. Please see https://github.com/develar/electron-complete-builder/blob/c3d11d61caa531055415b5eb9bc5a754282307ae/src/deb.ts#L24

`electron-complete-builder` supports native application dependencies compilation (only if two-package.json project layout used), so, we cannot package in parallel. Before each "arch" we must rebuild node modules (well, there is a solution, but it is a minor issue due to some reasons (will be off topic here)).

But — [each task to build distributable format is performed asynchronously](https://github.com/develar/electron-complete-builder/blob/c3d11d61caa531055415b5eb9bc5a754282307ae/src/packager.ts#L124).

What does it mean? If you run `build --platform=all --dist` on OS X [here](https://github.com/develar/electron-complete-builder/blob/c3d11d61caa531055415b5eb9bc5a754282307ae/src/packager.ts#L133) you get:
* zip mac app (auto-update)
* build dmg
* build linux i386
* build linux amd64
* build win x64 installer&nugget
* build win ia32 installer&nugget

... and all this tasks are performed in parallel. Huge test and build speed up (fpm/zip/mono or other low-level tool consumes no more than 100% CPU, but 4 GHz Intel Core i7 supports 600% for example).

Also, if artifacts publishing is enabled, on dist task done, corresponding artifact will be uploaded in parallel.

lib/linux.js
stefanjudis: :+1:

lib/linux.js
stefanjudis: :+1:

lib/linux.js
stefanjudis: I'm not a big fan of these long lines... But we can surely decide together. :bowtie: 

As this file was mainly written by @xseignard. What do you think?

stefanjudis:
For me this looks fine...

Also switching to promises and mature the actual node module makes complete sense.

I'd say - we can give it a go when @xseignard agrees. :)

lib/linux.js
develar: It will be better just use standard, as electron-packager does. But, of course, I don't insist and don't want to start religious war. It is ok for me to run `npm test` before commit and fix all style issues.

stefanjudis: Not a fan of standard either... :) cool! Ja but we can surely tweak the rules if anything is really annoying you.

As long it s consistent I don't mind that much. :)

lib/linux.js
xseignard: The idea was to separate each steps of the building process, as it might involve more stuff over the time.

develar: In fact, we should not copy — fpm allows you to map any dir to any path. See https://github.com/jordansissel/fpm/wiki/Source:-dir

xseignard:
@stefanjudis I agree! 
Thanks for the smart catch @develar 

Regards!

Issue #138, opened 2016-02-10
Clarify collaboration between electron-builder and electron-builder-complete


stefanjudis:
> @stefanjudis To be clear — electron-complete-builder is not yet another tool to produce installers, but tool to combine all existing tools to "just publish my electron app".

> As we'll implement the official squirrel thingy
> In this case I will just switch back to use electron-builder again as before.

>Do you want to integrate https://github.com/develar/electron-complete-builder#code-signing, https://github.com/develar/electron-complete-builder#build-version-management and https://github.com/develar/electron-complete-builder/wiki/Publishing-artifacts into electron-builder?

> I ready to create pull requests and do so (e.g. #251, #255). But I think we should follow Unix Philosophy — Do one thing and do it well.

> Do you agree with me or I am wrong? Sorry, but I don't want to have gigantic package.json as you suggest — https://www.npmjs.com/package/electron-builder#how-we-use-it-so-far See my app — https://github.com/develar/onshape-desktop-shell/blob/master/package.json Only one line "release": "npm run compile && build" and app published to GitHub releases.

Taken from here: https://github.com/maxogden/electron-packager/issues/33#issuecomment-182311037

I will have a look later today. :) Thanks. @delevar

develar:
In general, I don't think that end users really need chain `electron-complete-builder` ->  `electron-builder` -> `electron-packager`.

It is not practical. Someone can blame that tool is bloated, but...

Each, each tool currently do to the same thing — read existing `package.json` (and other convenient places) to infer configuration. `electron-complete-builder` does it better than other tools, but it is a temporary situation, I guess (recently, "infer electron version" was implemented in `electron-packager` ). Because it is a vital idea. No one want to maintain gigantic package.json and build scripts. 

Currently, `electron-complete-builder` doesn't reinvent the wheel — only "deb" linux support is a my mistake — definitely, I will use `electron-builder` to build linux installer and get rid of my own prototype implementation (forgot to check it before writing my implementation).

Currently, it is not required to use `electron-builder` to build OS X (`appdmg` used directly, `electron-builder` doesn't add any value) and Windows— [windows-installer](https://github.com/electronjs/windows-installer) used directly to get auto-update ready app.

I cannot say yet how do I see our collaboration — wait your answer to my initial questions.

rishson:
From a user's point of view, both projects are great but I don't think the electron community needs more options.
I agree with @develar that there should just be 1 package that 'just works' :tm: 
[...]

PR #146, opened 2016-02-13
Merge electron-complete-builder project into electron-builder

develar:
First phase of #138.

Tests passed:
 * Linux & OS X: https://travis-ci.org/develar/electron-builder/builds/109034851
 * Windows: https://ci.appveyor.com/project/develar/electron-builder/build/1.0.2


stefanjudis:
@develar 

Ja looks okay for me... How do you plan to proceed then?

develar:
> How do you plan to proceed then

Phase 2: Update docs and discuss (as part of the `update readme` pull request) what we want to do with the existing `electron-builder` API. 

* Drop it — I don't think it is an option. We should provide easy migration.
* Mark as deprecated and don't describe in the docs. I prefer this variant.
* Continue to maintain. Well, I don't find this API user-friendly.

Please note — the main problem here that `electron-complete-builder` doesn't build windows installer as `electron-builder` does. I think, we cannot build only squirrel installer because it is not appropriate for all, so, we should somehow add this functionality to new API (currently, `platform` accepts only win32 — should be a way to specify what kind of installer we want to get).

Phase 3: migrate existing tests to AVA.

stefanjudis:
> Mark as deprecated and don't describe in the docs. I prefer this variant

Yeah sounds good but also including a nice migration guide.

> Please note — the main problem here that electron-complete-builder doesn't build windows installer as electron-builder does. I think, we cannot build only squirrel installer because it is not appropriate for all, so, we should somehow add this functionality to new API (currently, platform accepts only win32 — should be a way to specify what kind of installer we want to get)

Ja I'd keep `nsis` a short while and then simply drop it with a major release as Squirrel is the official recommended way to go.

Okay. :+1: 

I don't think I'm much help here as I don't have much time currently but yeah go ahead. :)

develar:
Thanks, could you please add project to [Appveyor](https://www.appveyor.com) (it is required to run windows test and my pull request contains required `appveyor.yml`)?

stefanjudis:
Done. 

https://ci.appveyor.com/project/stefanjudis/electron-builder

Issue #136, opened 2016-02-09
packages on linux don't show icon in desktop entry


citrusui:
Haven't seen an issue opened on this yet, so here you go.

xseignard:
Hello @citrusui 

It is a known issue I'm planning to tackle quick.

You're one of the first to try the `.deb` packaging and your feedback and patience is valuable, so thanks!

develar:
Docs: https://developer.gnome.org/integration-guide/stable/desktop-files.html.en

develar:
@xseignard  Will be fixed tomorrow in the electron-complete-builder (and our projects will be merged next week —https://github.com/loopline-systems/electron-builder/issues/138#issuecomment-183715299).

48x48 PNG icon is required, but we will automatically extract it from OS X icns file (if not explicitly provided).

PR #149, opened 2016-02-16
Fix packages on linux don't show icon in desktop entry #136

develar:
Docs (we use OS X `icns` to generate linux icon set) will be added as part of `update readme` pull request.


.gitignore
```diff
@@ -1,14 +1,13 @@
 node_modules/
-tmp/
```
develar: Otherwise `tmp.d.ts` is ignored. If it is still required, `/` maybe prepended to avoid such issues.

stefanjudis:
There is a lot of stuff going on in your PR's. We have to definitely lower the entry barrier at some point. I don't think a lot of people can simply jump in and contribute.

Honestly I'm not really aware of what all these changes in 20 files do and it feels like a huge overhead...
Not sure how I can help here and how "me clicking the button" has any benefit. So go ahead if you wish. :)

develar:
> can simply jump in and contribute.

Sorry. I don't see what is complex in my changes. 

`typings/main/ambient/gm/gm.d.ts` is type definitions (not written by me).

Maybe you mean precomputing of icons — promise in the package helper constructor and `await` in the `packageInDistributableFormat`. Yes, it adds complexity, but allows us:
* do not do it twice (number of linux architectures).
* do it in parallel (well, nodejs cannot to really perform parallel tasks, but at least IO).

Maybe you mean processing of OS X icons, but throwing error "Hi — you must double your work and repeat all icons in the yet another format for yet another OS" is not user friendly (also, Linux icons layout is complex (not all icons in one folder), so, in any case we have to somehow copy it in a smart way).

Also, OS X `icns`is not always contain 48x48 icon, so, we resize 128x128 icon.

Linux Icons "specification": http://stackoverflow.com/a/10981921

stefanjudis:
> Sorry. I don't see what is complex in my changes.

I rather easy thing with 20 changed files. The diff is huge. I don't know what is generated. I don't know what all these files are for. The project lost me already. Sorry.

That's all fine, the only thing I'm saying is that we have to make it as easy as possible to people to jump in. Meaning powerful docs and an easy workflow. TypeScript makes it already harder so there is a lot of guidance needed. ;)

PR #226, opened 2016-03-09
fix: ignore newline when parsing common name in .p12 certificates

pauliusuza:
The regex takes an extra newline when parsing the .p12 cert common name, in this example:

subject=/C=CY/L=Limassol/O=OrgName/CN=OrgName
issuer=/C=US/O=DigiCert

Common name ends up being: 
OrgName
issuer=

Expected:
OrgName

### DIFF
--- a/src/codeSign.ts
+++ b/src/codeSign.ts
@@ -60,7 +60,7 @@ async function importCerts(keychainName: string, authorityCertPath: string, deve
 function extractCommonName(password: string, certPath: string): BluebirdPromise<string> {
   return exec("openssl", ["pkcs12", "-nokeys", "-nodes", "-passin", "pass:" + password, "-nomacver", "-clcerts", "-in", certPath])
     .then(result => {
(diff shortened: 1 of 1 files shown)


pauliusuza:
@develar this should be the final fix for the custom .p12 to work properly

develar:
Could you please write a test? (see CodeSignTest.ts).

pauliusuza:
Sure, will open a separate PR for the test.

develar:
@pauliusuza You can add test as a part of this PR. Or do you want a new version with this fix ASAP :)?

pauliusuza:
I would appreciate if you could merge it ASAP. We are literally stuck here :)

The test you have in CodeSignTest.ts is valid, however the test.p12 is does not look like the .p12 cert we use in production. In your case there's no new line after common name. A proper approach here would be to test against 2 or 3 different certificates from different providers.

develar:
Ok, I will publish without test. Travis restarted (failed due yet another npm bug :().

pauliusuza:
Thanks

PR #227, opened 2016-03-10
feat: add console flag --ignore

TakT:
A pattern which specifies which files to ignore when copying files to
create the package(s). The out directory is ignored by default, along
with the electron-prebuilt and electron-packager Node modules, the .git
directory, and node_modules/.bin. Alternatively, this can be a predicate
function that, given the file path, returns true if the file should be
ignored or false if the file should be kept.


develar:
Sorry, but our position is: or use two-package json layout, or just specify required option in the `build` — "This object will be used as a source of electron-packager options. You can specify any other options here.".

PR #241, opened 2016-03-14
fix: Linux build fails at icon conversion #239

### LINKED ISSUE #239: Linux build fails at icon conversion
Building a `deb` with the latest release (2.9.5) I get the following error:

```
Unhandled rejection Error: Command failed: gm convert: Unable to open file (/tmp/electron-builder-30935NzzLT2KzUll7/icon_128x128x32.png) [No such file or directory].
```

I have provided a `build/icon.icns` file (generated using these steps: http://blog.macsales.com/28492-create-your-own-custom-icons-in-10-7-5-or-later). When I look what's in the `/tmp/electron-builder-30935NzzLT2KzUll7/` folder, I see that it created the following icons:

```
icon_16x16x32.png
icon_256x256x32.png
icon_32x32x32.png
icon_512x512x32.png
````

Looks like 64 and 128 is missing.

I am building the linux deb from an Ubuntu 15.10 machine.

Any ideas what's going on? Alternatively, could I just provide the icons "manually" (I already have all those pngs anyways)?

develar:
Closes #239

Well, it is not a real fix, because it seems that icns2png is completely broken on Linux (ironically, yes :()
Unlikely that it will be merged in this form.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "ava-tf": "^0.12.4-beta.6",
     "babel-plugin-array-includes": "^2.0.3",
     "babel-plugin-transform-es2015-parameters": "^6.7.0",
-    "electron-download": "^2.0.0",
+    "electron-download": "^2.1.0",
     "eslint": "^2.4.0",
     "eslint-plugin-ava": "sindresorhus/eslint-plugin-ava",
     "ghooks": "^1.0.3",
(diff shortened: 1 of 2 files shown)


develar:
Also, we must add special test icns to reproduce the issue in our tests — current is different than real production icns files.

PR #245, opened 2016-03-15
refactor: use windows-installer instead of fork.

havenchyk:
Replaced a fork with electron-winstaller. Let's wait for tests, because I have troubles with running them on my windows machine.


develar:
No, it is not possible. Please merge https://github.com/electronjs/windows-installer/pull/37 and then I open another PR, then another and finally it will be possible. But not now :)

PR #246, opened 2016-03-16
feat: ability to provide linux icon set

develar:
build/icons will be used as a source of Linux app icons if exists.
Or ICNS file will be converted – now we produce all missed required linux sizes — 24, 64, 96.


develar:
We use [sharp](https://github.com/lovell/sharp) instead of gm because "Most Windows (x64), Linux and ARMv6+ systems do not require the installation of any external runtime dependencies."

develar:
Blocked on https://github.com/lovell/sharp/issues/378

PR #280, opened 2016-03-30
Added basic badges.

havenchyk:
Missed one for appveyor, because it fails on master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,11 @@
+# electron-builder
+
+[![NPM version][npm-image]][npm-url]
+[![Build Status][travis-image]][travis-url]
+[![Dependency Status][depstat-image]][depstat-url]
+
+[![NPM][nodei-image]][nodei-url]
+
 Complete solution to build ready for distribution and "auto update" installers of your app for OS X, Windows and Linux.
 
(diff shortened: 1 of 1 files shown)


develar:
Why test and deps are useful? Tests in master should be green. Deps maintained by bot greenkeeper.

Npm version is useful, is it possible to fix command — `--save-dev` should be used.

havenchyk:
Since we use greenkeeper, deps will not be useful. As for npm version, it can be either small badge with version only or bigger badge from nodei.co. I think it will be enough to have version only.

As for tests, yes, they always should green on master, but the badge will indicate about meging issues, I guess.

havenchyk:
I'll provide an updated commit today, hopefully.

develar:
> the badge will indicate about merging issues, I guess.

About pull request? Why it is interesting for users?

havenchyk:
As a user i like to see that master is green and if it's not, the package is nit really well maintained. Anyway, if you don't see the reason to have it, I'll remove this badge.

develar:
CI runs tests on push. If no one push — no run. So it will be green after last push and, so, it is not indicator if activity.

I think, "small badge with version only" will be useful, other badges is a visual noise.

havenchyk:
Continued in #295

PR #344, opened 2016-04-20
feat: accept multiple default app dirs

demetris-manikas:
Have not written any tests since I am not yet familiar with the testing framework. 
Any comments are welcome


develar:
I will deprecated app in the CLI args (should be in the build.directories) and push your change.

Thanks! `www` definitely should be checked as app dir.

PR #348, opened 2016-04-21
test: recognize www as default app dir

demetris-manikas:


test/src/BuildTest.ts
```diff
@@ -61,6 +61,12 @@ test("version from electron-prebuilt dependency", () => assertPack("test-app-one
   }
 }))
 
+test("www as default dir", () => assertPack("test-app", allPlatformsAndCurrentArch(), {
```
develar: I think it is ok and better to test only current platform, not all platforms (so, our test suite will execute faster).

demetris-manikas:
Updated to run only on current platform

demetris-manikas:
However it snapped me that I had to fix the test-app/app/package.json since it his means that none of the current tests did test it and i cannot see why the "build in the app package.json" test does not show the errrors

PR #355, opened 2016-04-25
feat: check application package

demetris-manikas:


src/metadata.ts
```diff
@@ -24,6 +24,8 @@ export interface AppMetadata extends Metadata {
    */
   readonly description: string
 
+  readonly main?: string
```
develar: I think we should not add such option. User should use standard `package.json` main. And, so, our check should check such field in the package.json.

develar:
I think, we should perform this check before pack to asar due to performance reason.  i.e. — when app dir detected, we check it. Before pack, not after.

src/metadata.ts
demetris-manikas: If I understood correctly interface AppMetadata holds all "useful" properties of the application package.json for the sake ot typesafety. I thought I only gave access to main (an already existent option). Please correct me if I am wrong.

demetris-manikas:
Well, then your comment that I should be aware of asar what was it for?
That is was what made me do the check after the packaging.

src/metadata.ts
develar: Sorry, you are right.

demetris-manikas:
So should I apply the validation before or after packaging?

develar:
> Well, then your comment that I should be aware of asar what was it for?

Sorry, it was just a proposal. After your changes I see that we can just use app dir.

Also, it will fix OS X — currently, it is broken because on OS X path to `app.asar` is different — `Contents/Resources/app.asar

On other hand — I doubt that it will be a bottleneck (it seems, getting contents of asar file is very quickly). And, strictly speaking, it is better to check final result, not intermediate. 

So, :+1: to current ASAR check, just fix OS X.

src/platformPackager.ts
```diff
@@ -155,14 +157,36 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions>
 
     this.beforePack(options)
     await pack(options)
+    await this.sanityCheckPackage(appOutDir, options.asar)
+  }
+
+  private async statPackageFile(appOutDir: string, packageFile: string, isAsar: boolean): Promise<any> {
+    let fpath = path.resolve("/", packageFile)
```
develar: Please consider to use `const` here.

develar:
Also, please add `asar` dependency explicitly (currently, we inherit it from electron-packager).

demetris-manikas:
I believe all is done.

develar:
Thank you!

PR #370, opened 2016-04-29
Empathetic Docs

nickhudkins:
Perhaps include an example of building with the --dist flag so that the magic of `dist` and `dist:` prefix doesn't cause users headache.


develar:
@nickhudkins Thanks. Finally, did you name your script as `dist:` or not ;)?

PR #373, opened 2016-05-02
add Jenkins build number support

bontibon:
Jenkins stores the build number in the `BUILD_NUMBER` environment variable


hp011235:
I would really like to use this feature. Is there any documentation around it?

develar:
@hp011235 Please see https://github.com/electron-userland/electron-builder#build-version-management

hp011235:
Thanks @develar for the link. I tried that on my Jenkins server and it does set the CFBundleVersion, but does not set the version.build_number in CFBundleShortVersionString. Is it an issue or a known behavior?

Issue #397, opened 2016-05-13
custom build steps for Mac builds


ammmir:
on Mac builds for my app, i need to copy custom directory structures into the .app bundle, like XPC services, other .app bundles, etc. could you please add support for a custom build step like running a script or arbitrary command right before the deep codesign step?

develar:
Why [extraResources](https://github.com/electron-userland/electron-builder/wiki/Options#user-content-BuildMetadata-extraResources) is not suitable to copy  "custom directory structures"?

ammmir:
@develar because it'll copy them into the `Contents/Resources` dir if i'm reading the docs correctly

develar:
@ammmir If it will copy to root of the app (`yourApp/.`), will be it a solution for you?

ammmir:
@develar yes, that'll work, too :)

ammmir:
@develar actually, as part of my build process, i also need to modify `Contents/Info.plist` prior to codesigning, so a provision for running a custom command prior to the codesigning step would allow most flexibility imho.

develar:
Actually, there is an option `platformPackagerFactory` in the `PackagerOptions`, but it is very low-level. I will add custom function. BTW, why do you need to modify Info.plist, maybe we can add some options to avoid programmatic usage?

ammmir:
@develar in my case, i'm adding a privileged helper tool, which requires the main app bundle's Info.plist to contain a `SMPrivilegedExecutables` dict. there may be other uses too.. bat basically, i need to edit the Info.plist *and* add some files into various directories (MacOS, Library, etc.)

develar:
>  i'm adding a privileged helper tool, which requires the main app bundle's Info.plist to contain a SMPrivilegedExecutables dict

Please use https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#extend-info (in the `build`).

develar:
Implemented as:
```
      build: {
        afterPack: () => {
          called = true
          return Promise.resolve()
        }
      }
```

develar:
Do you copy files into `Contents/`?

develar:
@ammmir Do you copy files into `Contents/`?

hannanabdul55:
So does the step of copying `extraResources` files occur after codesigning is completed? 
I think it is so because if I specify an `extraResources` parameter, it is giving me an invalid DMG.

develar:
[...]

Issue #371, opened 2016-04-29
Application entry can't be found


caibirdme:
Hey,all
I followed the two package.json structure in the document, but when I ran `npm run build`, it reported an error that `Application entry file couldn't be found in package.Seems like a wrong configuration`
<img width="1140" alt="2016-04-29 10 46 43" src="https://cloud.githubusercontent.com/assets/8054803/14907200/e57d0a50-0df7-11e6-916a-c8bdf14f3ed5.png">

the package.json in the root of the project is:
```
  ...
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "postinstall": "install-app-deps",
    "clean": "rimraf dist",
    "build": "build"
  },
  "build": {
    "app-bundle-id": "com.smartstudy.toefl",
    "app-category-type": "public.app-category.education"
  },
  ...
```

and the structure of app directory is:
<img width="422" alt="2016-04-29 10 51 07" src="https://cloud.githubusercontent.com/assets/8054803/14907238/544d0f84-0df8-11e6-977b-df546dd83bd8.png">
the package.json in app directory is:
```
{
  "name": "toefl",
  "version": "0.0.1",
  "description": "this is toefl-pc",
  "main": "./index.js",
  "private": true,
  "author": "caibirdme",
  "license": "ISC",
  "dependencies": {
    "request": "^2.71.0",
    "sequelize": "^3.21.0",
    "sqlite3": "^3.1.3"
  }
}
```
version:
node 5.10.0
npm 3.8.3
electron 0.37.5
electron-builder 3.13.1

Vj3k0:
Try setting "main" property in your root package.json to be "app/index.js"

develar:
@Vj3k0 Fix will be tomorrow, for now I suggest just set version to `3.11.0`

develar:
@caibirdme Cannot reproduce. Test passed for me.

Notes:
1) ` "clean": "rimraf dist",` Please do not — electron-builder does it for you, — simplify your package.json and reduce build time.
2) Don't specify main explicitly — if file named as `index.js`, you don't need to specify it at all.

@demetris-manikas Maybe do you have any idea?

demetris-manikas:
@caibirdme is there an app.asar being built? If so could you provide the output of asar list app.asar?
@develar unfortunately I won't be able to do any tests before Monday.

demetris-manikas:
@caibirdme can you also please explain what is the "electron" listing in your app folder?

[...]

PR #401, opened 2016-05-13
feat: check asar existence and integrity

demetris-manikas:


demetris-manikas:
Forgot to mention that this patch was created in response to #371. 
Doesn't solve the problem but throws more descriptive errors.

develar:
If we want to report more clear message if asar file doesn't exists, maybe we should fix code 

```
try {
        return statFile(path.join(resourcesDir, "app.asar"), relativeFile) != null
      }
      catch (e) {
        // asar throws error on access to undefined object (info.link)
        return false
      }
```

? Here we just ignore error and returns false. I think, it is possible to throw error `throw new Error(`Package file ${asarFile} was not created.`)` here (checking error) and avoid yet another check.

src/platformPackager.ts
```diff
@@ -223,17 +226,38 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions>
...
+
+    if (outStat == null) {
+      throw new Error(`Package file ${asarFile} was not created.`)
+    }
+
+    try {
+      listPackage(asarFile)
+    }
+    catch (e) {
+      throw new Error(`Package file ${asarFile} is corrupted.`)
```
develar: Is it real case?

demetris-manikas: #371 suggests that it is real. Haven't been able to reproduce.

demetris-manikas:
Why not. It just feels more natural to me to first check for existence and then try to act on a file than trying to act on a file and on error check for its existence.

develar:
Makes sense since #397 I feel that there is area to optimise, but... let's do it later.

PR #405, opened 2016-05-14
fix(linux): use full path in .desktop file

yuya-oc:
In previous version, "Exec" parameter of .desktop file used only executable name.
This depends on user's $PATH, that might cause the problem when launching the app.
So this commit uses the full path for "Exec" parameter.


demetris-manikas:
Looks good to me.

PR #418, opened 2016-05-20
feat: Support finding electron version in package config

mcfedr:
Allows the user to avoid extra downloads of electron for current
platform when he might not need to. Also allows using versions that
are not released on electron-prebuilt

https://github.com/electron-userland/electron-packager/pull/374

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -109,6 +109,9 @@ export function spawn(command: string, args?: Array<string> | null, options?: Sp
 }
 
 export async function getElectronVersion(packageData: any, packageJsonPath: string): Promise<string> {
+  if (packageData.config['electron-version']) {
+    return packageData.config['electron-version']
+  }
   try {
     return (await readJson(path.join(path.dirname(packageJsonPath), "node_modules", "electron-prebuilt", "package.json"))).version
   }


src/util.ts
```diff
@@ -109,6 +109,9 @@ export function spawn(command: string, args?: Array<string> | null, options?: Sp
 }
 
 export async function getElectronVersion(packageData: any, packageJsonPath: string): Promise<string> {
+  if (packageData.config['electron-version']) {
+    return packageData.config['electron-version']
```
demetris-manikas: I believe there should be some kind of validity checking and error reporting of the user defined version.

develar:
Thanks, in the next release you can specify `electronVersion` in the `build`.

src/util.ts
develar: In any case user will get clear error on electron-download.

Electron version is not documented at all, yeah... We should document it some day.

mcfedr:
Thanks, thats really helpful

develar:
> will that work without the changes to electron-packager

Yes, we pass all data to electron-packager and electron-packager doesn't infer. In any case now we use our own fork.

develar:
3.27.0 is released, but it is a `next` release as always.

Issue #416, opened 2016-05-19
Code Signing Issue on Server 2008 R2


dennyferra:
node version 5.9.0
npm version 3.7.3
electron 0.38.9
electron-builder 3.23.0

I'm having trouble code-signing my app with electron-builder. This issue is happening on our Team City build server which is running Server 2008 R2.

The error that is being generated is:

```
Done Adding Additional Store
SignTool Error: An unexpected internal error has occurred.
Error information: "Error: SignerSign() failed." (-2147024846/0x80070032)
```

I'm not able to find much information on the error code given above. I extracted the complete SignTool command that is being used (and pointed to the certificate locally), this gives the error mentioned above:

`signtool sign /tr http://timestamp.comodoca.com/rfc3161 /f C:\certificate.p12 /fd sha256 /td sha256 /d Downloader /as /p ***** D:\TeamCity\buildAgent\work\3cb96f5abb9e4582\release\downloader-win32-x64\Downloader.exe`

Interestingly code signing works fine on my development machine which is running Windows 10.

develar:
We must sign using dual-code sign — sha1 for old windows versions and sha256 for new windows 10 (required by Microsoft). So, bundled signtool requires windows 10 to be run. I assume, Server 2008 R2 is too old. Is it possible to upgrade to Windows Server 2012 & 2012 R2? In any case I am not yet sure that Server 2008 R2 is not possible to support but... I doesn't have it to test.

develar:
@dennyferra BTW, please update to 3.25 because 3.23 has critical regression if you build win on win.

PR #434, opened 2016-05-24
feature: specification of signing algorithms

heinzbeinz:
added options for windows signing: signingHashAlgorithms (sha1, sha256) and signcodePath (path to signtool.exe)

**note**: sha256 only signing works only in conjunction with pull request for signcode-tf


develar:
> signcodePath

I don't like it. It is ok for low-level tool `signcode`, but not for electron-builder. Maybe env option is suitable for you? Like `SIGNTOOL_PATH`? Or do you add this param to use osslsigncode on Linux? In this case `USE_SYSTEM_SIGNCODE` will be added.

heinzbeinz:
Thanks for your feedback. I used it in Windows Server 2008, The bundled signtool seems to new for this windows edition (looks like from Windows 10 SDK). An SIGNTOOL_PATH would be perfectly fine.

develar:
@heinzbeinz Send me your version of signtool and I will test it — as solution of https://github.com/electron-userland/electron-builder/issues/416 we can just downgrade signtool if it passes tests for win10.

heinzbeinz:
Here you go. (exe could not be directly uploaded, I wrapped it in a zip)
[signtool.zip](https://github.com/electron-userland/electron-builder/files/279718/signtool.zip)

heinzbeinz:
Closing because of #435

PR #447, opened 2016-05-30
doc: added signingHashAlgorithms to winbuild

heinzbeinz:
Updated Options-documentation


develar:
Options.md should not be edited, `signingHashAlgorithms` is not documented in the `WinBuildOptions`, so, `Options.md` doesn't contain about it. I will add your comment to `metadata.ts`.

heinzbeinz:
Ah now I see, thanks for clarifying and sorry for the inconvenience

PR #478, opened 2016-06-08
add TypeScript Badge

ellerbrock:


develar:
TS is not a reason be proud. TS is ugly, outdated, not modern language. I don't say that MS cannot develop modern language — TS is great compared to JS and MS did it right — where is dart or kotlin?

But TS is not a [kotlin](http://kotlinlang.org) and we cannot use kotlin because for most of the web developers it will be a reason to not contribute (or even to not use).

develar:
Thanks for your contribution, but 👎

PR #496, opened 2016-06-14
feat: adding dot asar option

markgardner:
This will allow users to have the ability to include dot folders into their asar file. This should maintain existing default functionality.

### DIFF
--- a/src/asarUtil.ts
+++ b/src/asarUtil.ts
@@ -14,6 +14,7 @@ export async function createAsarArchive(src: string, resourcesPath: string, opti
   const files = (await new BluebirdPromise<Array<string>>((resolve, reject) => {
     glob = new Glob("**/*", {
       cwd: src,
+      dot: options.dot
     }, (error, matches) => {
       if (error == null) {
         resolve(matches)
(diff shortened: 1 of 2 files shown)


develar:
Could you please specify the reason why you have to do it? Name of required files? My plan is to get feedback and add explicit inclusion for such files (name starts with dot).

markgardner:
We just keep our built files in a .build folder. It helps us keep the app directory clean when building locally

develar:
If I understand correctly — you build source js using Babel/TS to dir `.build`? Hmm.... You also have to modify `files` option, because dot files are ignored as well by default.

Ok, it seems yor setup is not common and it cannot be automatically discovered. Or maybe is there some way — field in the package.json, Babel config?

markgardner:
We have a homegrown asset management tool that is like webpack. We use that to create all of the minified release files.

We are doing a few custom things to build all of our assets that is definitely an edge case. We did modify the files array to include our build folder (that seemed pretty straight forward to me once I released how glob ignores dot folders by default).

Side Note, I started from scratch upgrading our complex build process for 0.36.0 that would call electron-packager directly. Using electron-builder is sooo much easier. Thank you for all you have done with this.

develar:
Please be aware that you should not just add pattern `**/.*`, you should add pattern `.build`. To ignore `node_modules/.bin`, `. DS_Store` and so on.

PR #532, opened 2016-06-22
Remove reference to obsolete target parameter. Fixes #531

### LINKED ISSUE #531: `build --target dir` doesn't work
https://github.com/electron-userland/electron-builder/blob/master/README.md

The README says to install a `pack` script like this:

```javascript
"scripts": {
  "postinstall": "install-app-deps",
  "pack": "build --target dir",
  "dist": "build"
}
```

> And then you can run `npm run dist` (to package in a distributable format (e.g. dmg, windows installer, deb package)) or `npm run pack`.

But the `build` command doesn't accept a `target` parameter, so if you actually try to `npm run pack`, you just get an error message, `Unknown argument: target`

dfabulich:
I figure that since `--target` is removed, this `pack` script must no longer be useful. (What was it originally for?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,11 +67,10 @@ For a production app you need to sign your application, see [Where to buy code s
     ```json
     "scripts": {
       "postinstall": "install-app-deps",
-      "pack": "build --target dir",
       "dist": "build"
     }
     ```
-    And then you can run `npm run dist` (to package in a distributable format (e.g. dmg, windows installer, deb package)) or `npm run pack`.
+    And then you can run `npm run dist` (to package in a distributable format (e.g. dmg, windows installer, deb package)).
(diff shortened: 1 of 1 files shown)


dfabulich:
AppVeyor has a failed test that just can't _possibly_ be due to my documentation update, right?! 😬 

`httpRequestTest » download to nonexistent dir`

develar:
Thanks. Hmmm... Interesting, should we fix docs or implementation... Our `--osx` and other platform flags accepts target list — including dir.

develar:
I think, we should add `--dir` to build dir target for current platform.

develar:
Thanks for PR, I decided to fix implementation :)

PR #562, opened 2016-07-03
Document how to disable DMG icons.

jwheare:
My first instinct was to set this to null, but the deepAssign in dmg.ts doesn't copy null values so it wasn't passed on to appdmg.


develar:
> My first instinct was to set this to null

And you are right. I fixed implementation, rather than document bug :) 

Thank you for pull request.

Issue #95, opened 2015-12-17
`Question`: Why appdmg is an optional dependency?


havenchyk:
Hi @stefanjudis!

I founded that `appdmg` is in a list of optional dependencies, could you please clarify, why?

stefanjudis:
To not break windows builds as it's not working on windows. :)

havenchyk:
But why should it break windows build? appdmg can be or cannot be required, that's it.

stefanjudis:
Can you try to install appdmg on a windows build and check if it's succeeding? 

This was my latest state. maybe it's not valid anymore.

havenchyk:
Sure, I'll check. I just haven't found on my linux machine appdmg in the node_modules and it's was very strange :)

ntrpy:
This is my log output when trying to install it under Windows 7:

```javascript
C:\Users\sven\Documents\Projects\LoopLine\frontend>npm i appdmg
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "i" "appdmg"
npm ERR! node v4.2.3
npm ERR! npm  v2.14.7
npm ERR! code EBADPLATFORM

npm ERR! notsup Unsupported
npm ERR! notsup Not compatible with your operating system or architecture: appdmg@0.3.5
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   win32
npm ERR! notsup Actual Arch: x64

npm ERR! Please include the following file with any support request:
npm ERR!     C:\Users\sven\Documents\Projects\LoopLine\frontend\npm-debug.log

C:\Users\sven\Documents\Projects\LoopLine\frontend>```

stefanjudis:
Thanks @ntrpy.

@havenchyk I'd say the same happens on your linux machine. ;)

havenchyk:
Yep, same here. Thanks!

fyockm:
Would you mind adding a note to the README or wiki to call out the fact that `appdmg` is an `optionalDependency`? I can submit a PR if you think it belongs in the README.

I had npm configured to _not_ install optional dependencies by default. It took me a while to figure out why I had to manually install `appdmg` when the README declares it "used under the hood". I actually had the same problem with `7zip-bin-osx`.

develar:
[...]

PR #576, opened 2016-07-11
README note about optional deps

fyockm:
Add note to README about `appdmg` and `7zip` packages being optional dependencies.

Ref: https://github.com/electron-userland/electron-builder/issues/95#issuecomment-226819644


develar:
Thanks!

PR #603, opened 2016-07-19
docs: Add Squirrel Updates Server to auto update section

djpereira:
Include Squirrel Updates Server as an alternative to serve releases.


develar:
Thanks!

PR #676, opened 2016-08-17
disable dual-signing for .msi installer

tony2n2:
MSI file format does not support dual signing.

see: http://social.technet.microsoft.com/wiki/contents/articles/32288.windows-enforcement-of-authenticode-code-signing-and-timestamping.aspx#How_to_handle_MSI_files


develar:
Thanks! A little bit more correct fix pushed.

PR #677, opened 2016-08-17
skip rcedit for .msi installer

tony2n2:
rcedit can only edit .exe resources


develar:
But it not fails, right? Or you get some error?

tony2n2:
it fails with this error if msi is enabled:
`Fatal error: Unable to load file`

develar:
Thanks!

Issue #683, opened 2016-08-20
OS X: Error: Module version mismatch. Expected 48, got 49


meltedspark:
* **^5.30.0**: 

* **1.3.3**: 

When trying to build for OS X getting the following error:
`tsc && node-gyp rebuild && build --mac`

> Error: Module version mismatch. Expected 48, got 49.
>     at Error (native)
>     at Object.Module._extensions..node (module.js:583:18)
>     at Module.load (module.js:473:32)
>     at tryModuleLoad (module.js:432:12)
>     at Function.Module._load (module.js:424:3)
>     at Module.require (module.js:483:17)
>     at require (internal/module.js:20:19)
>     at Object.<anonymous> (/Users/Jenia/Desktop/git/electron-angular2-native/node_modules/macos-alias/lib/create.js:7:13)
>     at Module._compile (module.js:556:32)
>     at Object.Module._extensions..js (module.js:565:10)
>     at Module.load (module.js:473:32)
>     at tryModuleLoad (module.js:432:12)
>     at Function.Module._load (module.js:424:3)
>     at Module.require (module.js:483:17)
>     at require (internal/module.js:20:19)
>     at Object.<anonymous> (/Users/Jenia/Desktop/git/electron-angular2-native/node_modules/macos-alias/index.js:1:18)
>     at Module._compile (module.js:556:32)
>     at Object.Module._extensions..js (module.js:565:10)
>     at Module.load (module.js:473:32)
>     at tryModuleLoad (module.js:432:12)
>     at Function.Module._load (module.js:424:3)
>     at Module.require (module.js:483:17)
>     at require (internal/module.js:20:19)
>     at Object.<anonymous> (/Users/Jenia/Desktop/git/electron-angular2-native/node_modules/ds-store/index.js:2:13)
>     at Module._compile (module.js:556:32)
>     at Object.Module._extensions..js (module.js:565:10)
>     at Module.load (module.js:473:32)
>     at tryModuleLoad (module.js:432:12)
> From previous event:
>     at tsAwaiter (/Users/Jenia/Desktop/git/electron-angular2-native/node_modules/electron-builder/src/util/awaiter.ts:10:47)
>     at Object.build (/Users/Jenia/Desktop/git/electron-angular2-native/node_modules/electron-builder/src/builder.ts:195:52)
>     at Object.<anonymous> (/Users/Jenia/Desktop/git/electron-angular2-native/node_modules/electron-builder/src/build-cli.ts:27:28)
[...]

PR #691, opened 2016-08-22
docs: clarify native addons build process

meltedspark:
Add clarification regarding how native addons that are not dependencies but a part of the project should be treated

https://github.com/electron-userland/electron-builder/issues/683


develar:
Awesome addition, nothing to correct, thanks a lot.

PR #703, opened 2016-08-25
Improved the section about the postinstall helper

Toflar:


develar:
Thanks a lot for improving documentation.

Issue #738, opened 2016-09-09
NSIS - No app icon in Add/Remove program


mbrainiac:
* **Version**: 6.5.2

* **Target**: Windows 10, 64

![no-icon](https://cloud.githubusercontent.com/assets/1453038/18380393/53171a9e-76a1-11e6-82a6-360788ecc676.jpg)

Problem:
 * multiUser.nsh: $appExe is empty.

PR #739, opened 2016-09-09
fix(nsis): no app icon in Add/Remove program #738

mbrainiac:


develar:
Yes, we should add tests for it, work is started, but after some update wine cannot execute nsis installer. I will add tests later.

develar:
@stefanjudis Don't you mind if I will disable mention-bot for some time?

stefanjudis:
@develar na, go for it. I think it's handy - but as I'm not involved anymore, sure thing! 👍

PR #759, opened 2016-09-19
fix: support asar in package.json's `main` property even when not kaging to asar during the build

wojtkowiak:
This pull request fixes the case when the app is packaged to `.asar` earlier than in the process of building with electron-builder. 
In my case I have the app split into 3 smaller asar archives for some modularity reasons. I have tried with `asar: false` but it failed even though Electron normally supports it. You can provide any path in the `main` of `package.json` e.g. it could be `somePath/some.asar/evenMore/index.js`.
I did not create an issue because I went to fixing it immediately. If an issue needs to be filed first, let me know.
In the tests I have added `(custom asar)` to distinguish it from using `asar: true` - not sure if this naming is ok.


develar:
Binary files (like `asar`) should be stored using GIT LFS (https://git-lfs.github.com). Committed for you. 

Thanks for pull request.

wojtkowiak:
@develar thanks for correcting :smile: and quick response.

PR #773, opened 2016-09-23
feat(windows): use appId to identify package instead of name

### LINKED ISSUE #753: Separate test and production builds on Squirrel.Windows
* **6.3.0**: 

* **Mac/Windows**: 

I am trying to be able to create separate test and production builds for my application. On mac this works fine by changing the appId from `my.domain.appid` to `my.domain.appid-test` for my test version. Both the mac prod build and the mac test build are able to run at the same time with no problems. 

When I try to do the same thing with the windows build, the more recently installed build always overwrites the least recently installed build causing the shortcut for the older build to do nothing. I'm assuming this is because appId is only supported for `nsis` distributions on windows and that lack of support is not allowing the applications to have unique ids and thus overwriting each other. 

Is there any way support for appId can be added to the `Squirrel.Windows` target? If not, is it possible to get this behavior another way?

symbyte:
Change unique identifier to appId instead of name for Squirrel.Windows
builds. This alligns the behavior of packaging a Squirrel.Windows app with
packaging a Squirrel.Mac app (as mac app uses appId for uniqueness).
This makes creating separate builds for one app (i.e. one for test and
one for production) more intuitive.

Closes #753


develar:
> This alligns the behavior of packaging a Squirrel.Windows app with
packaging a Squirrel.Mac app (as mac app uses appId for uniqueness).

In general, it is not possible. Squirrel.Windows and NSIS don't have option to specify different installation path and even if you set different `appId` — app will be installed using your `productName` (and of `productName` is not defined — by name).

> I tried with NSIS and everything worked fine.

How? Ah, because `productName` was explicitly set and changed to different value, right?

This PR cannot be accepted as is because it will break all existing applications. And since Squirrel.Windows will be deprecated soon, I don't want change default.

In 30 minutes I will push you PR with one change — `build.squirrelWindows.useAppIdAsId` will be required to enable this behaviour.

develar:
Thank you.

symbyte:
> How? Ah, because productName was explicitly set and changed to different value, right?

Yes, that is correct.

> And since Squirrel.Windows will be deprecated soon, I don't want change default.

Oh really? What will be the default target for auto updating applications on windows? Will Squirrel.Mac be deprecated too?

symbyte:
That is great news. He have had a lot of quirky issues with Squirrel.Windows behavior (auto updates taking a really long time to start being the biggest). Thanks for taking steps for a better solution! I'm looking forward to it.

Issue #548, opened 2016-06-28
windows installer not signed when building on mac 


symbyte:
- **Version**: 5.7.0

- **Target**: mac and win (both default)

Expected behavior:

When building on local mac dev machine with a valid code signing identity using the build -ow command to build for both windows and mac, generate signed mac application and signed Squirrel.Windows installer.

Actual behavior:

Only the mac application is signed using valid code signing identity, windows installer gives warning of unknown publisher on installation.

Have I misunderstood what code signing is supported on mac development machine?

develar:
How do you specify win cert, specify used options?

symbyte:
I wasn't specifying the win cert because I thought that it would be picked up automatically like for mac signing, maybe that is my problem?

My build log (removed dep install because it is very long):

```
Signing app (identity: Oseberg)

Creating DMG
Creating MacOS zip
Skip app dependencies rebuild because platform is different
Packaging for platform win32 x64 using electron 1.2.5 to build/download/win-unpacked

Building Squirrel.Windows installer
```

develar:
> would be picked up automatically like for mac signing, maybe that is my problem

You cannot use Apple cert to sign windows app. So, we don't use discovered cert on Mac to sign windows app. Or do you have non-Apple cert in your MacOS keychain and expect that it will be used?

symbyte:
I see, I thought that code signing certificates are universal, my mistake. 

If I obtain a windows specific code signing cert and keep it on my mac dev machine am I able to specify it to be used for the windows build while still having the mac cert in my keychain be auto detected and used to sign the mac build?

develar:
Yes. Using cert file/password. I will support cert from keychain to sign win app soon.

Mercieral:
@develar 

Sorry for bringing up an old issue, but can you elaborate further on "Using cert file/password"? By this do you mean setting the CSC_LINK and CSC_KEY_PASSWORD variables? And if we do set these to the windows code signing cert, it will still pick up the Apple cert in keychain automatically?

develar:
@Mercieral If you build macOS and windows on one machine, you cannot specify different `CSC_LINK`/`CSC_KEY_PASSWORD` — but you need to (because certs are different). And what to do? Solution — set `build.win.certificateFile`/`build.win.certificatePassword` (or reuse `CSC_KEY_PASSWORD` to avoid password in the `package.json`  — you can use the same password for both certs).

[...]

PR #774, opened 2016-09-23
docs: add missing certificateFile and certificatePassword options description

wojtkowiak:
I was just trying to code sign on Windows and spent a moment to find whether it is already supported and how to use it so I allowed myself to update the docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ dist/
 /.idea/encodings.xml
 /.idea/copyright/profiles_settings.xml
 /.idea/workspace.xml
+/.idea/deployment.xml
 /typings/browser/
 /typings/browser.d.ts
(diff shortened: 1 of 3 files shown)


develar:
Please reflect my comment — https://github.com/electron-userland/electron-builder/issues/548#issuecomment-234536292

`certificateFile` is required only and only if you build on mac, otherwise you should use `CSC_LINK`

`certificatePassword` is required only and only if you build on mac AND you win cert password is not equals to mac cert password.

wojtkowiak:
Good to know :) Updated, let me know if this way it reflects better the proper way of using it.

develar:
Thanks. Please edit macOptions.ts/winOptions.ts files in the future.

Issue #647, opened 2016-08-04
Native compilation doesn't happen if the package is using prebuilt (eg leveldown)


cosmosgenius:
electron-builder@5.22.2

OSX el capitan x64

My app directory package.json has a dependency to leveldown 1.4.0. Since leveldown is natively compiled, I was expecting install-app-deps to compile with electron-prebuilt deps on dev package.json. While installing leveldown, it uses prebuilt to download native binary instead of compiling.

Passing `--build-from-source` as args in the `doSpawn` function fixes the issues and compile with the correct version.

cosmosgenius:
Is there any update on this? or is there some option which can force a rebuild in such cases?

develar:
Is it possible to attach sample project?

cosmosgenius:
sample project https://github.com/cosmosgenius/electron-leveldb

cosmosgenius:
What i am doing currently is modifying the `npmExecArgs ` variable in `spawnNpmProduction` by adding `--build-from-source` to args list. Is there anyway to passing this args to install-app-deps? That would solve this issue.

develar:
Thanks for clear explanation and sample project. `--build-from-source` added by default.

PR #790, opened 2016-10-02
Make a setting for --build-from-option flag

### LINKED ISSUE #787: Install app deps always trigger rebuild
* **Version**: 7.10.0

* **Target**: Mac and Linux

Hello, I'm using electron-builder with native module (https://github.com/lyssdod/node-libtorrent). The module itself uses `node-pre-gyp` and has precompiled binaries. It installs flawlessly if I'm not using `electron-builder` directly (running `npm install` from `app` directory works). But if I do, it fails with

```
Error output:
module.js:457
    throw err;
    ^

Error: Cannot find module 'adm-zip'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
```

The fun fact is, this module (`adm-zip`) is specified in `node-libtorrent-ng`s `devDependencies` and is handled by regular `npm install` just fine. 

So my questions are:
1. Why `electron-builder` ignores default action of `node-gyp-rebuild` despite binaries are available?
2. Is there a way to debug 'install-app-deps' step?

Thanks!

lyssdod:
This seems to fix #787 

The only place I'm unsure of is (haven't touched yet):
https://github.com/electron-userland/electron-builder/blob/a4eeaded7a7e0da77112a1c96786d80f90e47083/src/install-app-deps.ts#L33

### DIFF
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -80,6 +80,7 @@ Don't customize paths to background and icon, — just follow conventions.
 | compression | <a name="BuildMetadata-compression"></a>The compression level, one of `store`, `normal`, `maximum` (default: `normal`). If you want to rapidly test build, `store` can reduce build time significantly.
 | afterPack | <a name="BuildMetadata-afterPack"></a>*programmatic API only* The function to be run after pack (but before pack into distributable format and sign). Promise must be returned.
 | npmRebuild | <a name="BuildMetadata-npmRebuild"></a>*two package.json structure only* Whether to [rebuild](https://docs.npmjs.com/cli/rebuild) native dependencies (`npm rebuild`) before starting to package the app. Defaults to `true`.
+| npmSkipBuildFromSource | <a name="BuildMetadata-npmSkipBuildFromSource"></a>*two package.json structure only* Whether to omit using [--build-from-source](https://github.com/mapbox/node-pre-gyp#options) flag when installing app native deps. Defaults to `false`.
 | nodeGypRebuild | <a name="BuildMetadata-nodeGypRebuild"></a>Whether to execute `node-gyp rebuild` before starting to package the app. Defaults to `false`.
 | electronDist | <a name="BuildMetadata-electronDist"></a>The path to custom Electron build (e.g. `~/electron/out/R`). Only macOS supported, file issue if need for Linux or Windows.
 
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -202,6 +202,11 @@ export interface BuildMetadata {
   readonly npmRebuild?: boolean
 
   /*
+   *two package.json structure only* Whether to omit using [--build-from-source](https://github.com/mapbox/node-pre-gyp#options) flag when installing app native deps. Defaults to `false`.
+   */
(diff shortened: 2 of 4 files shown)


lyssdod:
My main concern for this — I have to create installers for all platforms, but I have only one (Mac) available. Since I won't be able to rebuild my native modules with MSVC on Mac, I would like to use already present precompiled binaries for bundling

jordansteeves:
Hi! Any word on a review here? @develar

develar:
Did you investigate why https://github.com/electron-userland/electron-builder/issues/647 doesn't work? I mean — why " it uses prebuilt to download native binary instead of compiling."? Why incorrect  binaries was used? 

I don't want to introduce yet another flag, it is better to investigate — is it really bug in the precompiled binaries.

lyssdod:
@develar I`ll try to investigate 5.22.2 release

develar:
@lyssdod Issues is not in the electron-builder, should be reproducible using latest version also.

lyssdod:
The trick is in the `--build-from-source` flag which's being handled by `node-pre-gyp` and forces compilation instead of checking for binaries. In my particular case compilation fails because I've moved all compilation deps to `devDependencies` to minimize final package size (for fast `npm install` which will install only needed stuff). Moving these deps back to `dependencies` eliminates the benefits of using  `node-pre-gyp` completely, because this way it will compile the code for one particular platform (host platform `electron-builder` is running at), but not for others.

develar:
@lyssdod I understand you completely. But I want to understand — does `node-pre-gyp` works correctly or not. Bug #647 was reproduced by me — incorrect precompiled binaries were used. It is very, very critical — because you cannot test your app on all platforms and archs every release.

So, I ask you to check — is it `node-pre-gyp` bug or not. Who is guilty :) To ensure, that your approach really works and correct binaries will be used in ALL cases.

If you don't want to do this work, I can accept PR but this flag will be not documented and marked as experimental. And not by default.

lyssdod:
I'm all up for solving this issue on the `node-pre-gyp`'s side, it just seems to me that they're not very active and responsive to issues (in contrast to you :+1: ): https://github.com/mapbox/node-pre-gyp/issues/239

I`ll look into bug #647 more thoroughly now.

lyssdod:
Hi again @develar,

I've looked into #647 but didn't see any useful info there (release used there already has this flag enabled). Instead I've prepared a test repo which uses prebuilt modules after patching `electron-builder`:

https://github.com/lyssdod/electron-abi-test


> I don't want to introduce yet another flag

http://electron.atom.io/docs/tutorial/using-native-node-modules/ says there's another method of specifying it: 
```shell
# Tell node-pre-gyp to build module from source code.
export npm_config_build_from_source=true
```

I can update PR to handle this option instead of handling the flag if it makes sense.

develar:
@lyssdod Ok, thanks. 

> So in general it is recommended to always build native modules from source code.

For this flag will be not enabled by default, according to this note.

lyssdod:
Thank you!

PR #797, opened 2016-10-06
make install-app-deps also handle skip build flag

lyssdod:
Hi,
I am very sorry but that file I'm mentioned in my original PR and not touched is actually required. I forgot to add it to original PR. Can you please merge this also?

I am very sorry for this.


lyssdod:
Thank you!

Issue #529, opened 2016-06-20
NSIS Auto-Update


kunkinkan:
Creating this issue just to keep track of status, or when it is planned.

black-snow:
@develar @kunkinkan How's this gonna work? Will it be analogue to Squirrel?

develar:
@black-snow yes, only feed url will be changed in the user code. Auto-update API impl will be bundled.

black-snow:
Great. How are things going? Any help needed?

atypicalprogrammer:
Will this be better than Squirrel auto-update? I don't like with them that I cannot control the process; I should be able to allow the user to choose whether or not to update to the newest version, instead of forcing it. (And also; better security. Allow me access to the downloaded auto-update so I can for example compare md5 hashes, etc). So a method after downloading the update to verify it.

black-snow:
@atypicalprogrammer afaik *Squirrel* [does verify checksums](https://github.com/Squirrel/Squirrel.Windows/search?utf8=%E2%9C%93&q=checksum).

atypicalprogrammer:
@black-snow how do you figure? I'm talking about me providing a txt file with a checksum to compare with a downloaded file. It's a custom function that should be possible to provide better security. Something that is ran before the update itself is installed, simply a function that can cancel the further progression of install. The checksum validation I can add myself.

develar:
Started to work on this.

First version will support only Bintray:
* no rate limits (opposite to GitHub) to access latest version and version files. Rate limit free, no need to parse html or http headers, just use Rest API (opposite to GitHub — token requirement is not user-friendly).
* better security — Bintray reports file sha256/sha1 on version files request. We use it to verify downloaded file (because windows just sucks opposite to macOS — we cannot rely on code sign completely).
* `nsis-auto-updater` will be published to NPM, but only to provide flexibility — will be just prebundled on pack.

zh99998:
any good news now?

mkazlauskas:
It seems like work is being done. Please write a small readme and I'll be glad to do some testing.

develar:
@mkazlauskas Yes. It is nearly finished, but depends on a lot of small, but important things. I hope it  will be ready to test this week if no other critical NSIS issues (like #735) will be discovered.

virror:
Are there any docs for this yet so i can start developing this or are they coming when all is done?

develar:
> are they coming when all is done?

Yes.

develar:
[...]

PR #828, opened 2016-10-16
fix: enable GH_TOKEN env variable for releases

jamesmacfie:
A bug was introduced in the previous release where the token was set to env.GH_TOKEN but then an error thrown immediately. This change ensures that the token is set in the right place and an error is only thrown if no token exists

[Link to commit where this was introduced](https://github.com/electron-userland/electron-builder/commit/b1ae7d5b3c00de2305d6f8b2cb1981ceb574b4d0#diff-942adfda4f3ed355d1699273799dba23R47)

### DIFF
--- a/src/publish/gitHubPublisher.ts
+++ b/src/publish/gitHubPublisher.ts
@@ -42,9 +42,8 @@ export class GitHubPublisher implements Publisher {
   }
 
   constructor(private info: GithubOptions, private version: string, options?: PublishOptions, private isPublishOptionGuessed: boolean = false, config?: GithubOptions | null) {
-    let token = info.token
+    let token = info.token || process.env.GH_TOKEN
     if (isEmptyOrSpaces(token)) {
-      token = process.env.GH_TOKEN
       throw new Error(`GitHub Personal Access Token is not set, neither programmatically, nor using env "GH_TOKEN"`)
     }
 


willyb321:
I actually just reported this exact issue [here](https://github.com/electron-userland/electron-builder/issues/529#issuecomment-254075836). Good luck on your PR! 👍

jamesmacfie:
If someone could help me out with those failing builds then that would be appreciated. I'm not sure whether the build is being flakey or my change caused the errors

willyb321:
@jamesmacfie I'll have a bit of a look through to see whats going on!

src/publish/gitHubPublisher.ts
```diff
@@ -42,9 +42,8 @@ export class GitHubPublisher implements Publisher {
   }
 
   constructor(private info: GithubOptions, private version: string, options?: PublishOptions, private isPublishOptionGuessed: boolean = false, config?: GithubOptions | null) {
-    let token = info.token
+    let token = info.token || process.env.GH_TOKEN
```
willyb321: I'd try this bit without info.token for a start.

willyb321:
Actually, I've just noticed this..  
```
{
  "owner": "develar",
  "repo": "test",
  "token": "__test__",
  "version": "2.0.0",
  "tag": "2.0.0"
}

is not equal to

{
  "version": "2.0.0",
  "token": "[redacted]",
  "tag": "2.0.0"
}```

develar:
Sorry, it was stupid error and for some reasons our tests didn't catch it. I will fix tests.

PR #842, opened 2016-10-22
Make packager handle available native deps when building on non-native platforms

lyssdod:
Hi,

Given the skip build flag handling you've merged previously, I was able to successfully use precompiled binaries when creating package for native platform. However, I've stumbled upon an issue when I was trying to package app with native deps for Windows on Mac. It just skipped deps installation and wrapped stuff I've got after initial `npm install` (Mac-related) into Windows build, which of course didn't run. This change should allow the making of a truly crossplatform build on the same machine.


lyssdod:
Commit message is kinda confusing, I meant `platforms`, not `archs`

Issue #874, opened 2016-11-03
Create Mac Package Installer for non-MAS distribution


eriedl:
* **Version**: 17.19.1

* **Target**: Mac

According to Apple's code signing documentation you need to sign your app and the package installer with `Developer ID Application` respectively `Developer ID Installer` certificates. `3rd Party Mac Developer Application` and `3rd Party Mac Developer Installer` are only valid for distribution through the Mac Store.
We plan on distributing our app outside of the Mac Store as a package installer and not as a disk image or archive. However, that configuration is right not supported in electron-builder.

I propose an additional target `pkg` that will sign the app and installer with the `Developer ID *` certificates and create a package installer vs a disk image for distribution

develar:
Duplicates #52 Coud you please explain why do you want package installer instead of DMG?

eriedl:
A couple of reasons. For one Apple is pushing for real installers over distribution of disk images, which just mounts like an external disk on your computer. I also talked to our internal Mac OS developers and they also suggested that distribution should be in form of a package installer. You can read through their documentation and you will find they talk only about installers and not about disk images anymore: [Distributing Your Mac App](https://developer.apple.com/macos/distribution/) at the very bottom, [App Distribution Guide](https://developer.apple.com/library/content/documentation/IDEs/Conceptual/AppDistributionGuide/DistributingApplicationsOutside/DistributingApplicationsOutside.html#//apple_ref/doc/uid/TP40012582-CH12-SW2)

With a disk image, a lot of users don't get it that you have to drag the app to your local disk somewhere, ideally to the Applications folder, to run it. They just run it from the disk image directly, which can lead to problems if the image gets unmounted at some point. If the user is part of an enterprise that uses something like Casper to distribute software internally, they also like package installers over disk images as they can be better scripted. You may have also run into an situation where companies have had their installers packaged into a disk image.
It might seem that having an installer is too much overhead, but I disagree. One could argue also that the installer for Windows or the Debian package support is overhead, too. I find installers very handy and confident inspiring for the end user as they know exactly what's going to happen. Especially, if the installer is also signed. Keep in mind, not all end-users are tech-savvy enough to understand the difference between a DMG and a PKG. They just like to install that software on their system and get going.
Having a PKG would also allow us to add future components that may need to be installed in the system outside of the app folder.

[...]

PR #875, opened 2016-11-03
feat: Support for creation and signing of package installers and sign…

eriedl:
…ing for distribution of Mac apps outside of the Mac App Store

Issue #874


develar:
In general, +1. I will review and merge soon. Coud you please explain why do you want package installer instead of DMG?

develar:
Thanks! Only because of your PR and answer `pkg` target is supported.

PR #881, opened 2016-11-05
feat: add ability to specify additional npm rebuild args

mmckegg:
I am building an electron app that uses native dependencies targeting multiple platforms. [Prebuild](https://github.com/mafintosh/prebuild) binaries are available for all of them. 

However when `npmSkipBuildFromSource` is enabled, `prebuild` downloads the binary for the installed node version instead of the electron version. This can be easily handled by specifying a `--abi=ELECTRON_ABI_VERSION` argument, but electron-builder does not allow this.

**This PR adds support for specifying any number of additional `npm rebuild` arguments to allow working around issues outlined above.**

### DIFF
--- a/src/packager.ts
+++ b/src/packager.ts
@@ -237,12 +237,11 @@ export class Packager implements BuildInfo {
       }
       else {
         const forceBuildFromSource = this.devMetadata.build.npmSkipBuildFromSource !== true
-        const additionalArgs = this.devMetadata.build.npmArgs
         if (platform.nodeName !== process.platform && forceBuildFromSource) {
           log("Skip app dependencies rebuild because platform is different")
         }
         else {
-          await installDependencies(this.appDir, this.electronVersion, Arch[arch], forceBuildFromSource, (await statOrNull(path.join(this.appDir, "node_modules"))) == null ? "install" : "rebuild", additionalArgs)
+          await installDependencies(this.appDir, this.electronVersion, Arch[arch], forceBuildFromSource, (await statOrNull(path.join(this.appDir, "node_modules"))) == null ? "install" : "rebuild", this.devMetadata.build.npmArgs)
         }
       }
     }
--- a/src/packager.ts
+++ b/src/packager.ts
@@ -237,11 +237,12 @@ export class Packager implements BuildInfo {
       }
       else {
         const forceBuildFromSource = this.devMetadata.build.npmSkipBuildFromSource !== true
+        const additionalArgs = this.devMetadata.build.npmArgs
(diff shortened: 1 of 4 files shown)


src/util/util.ts
```diff
@@ -54,6 +54,14 @@ export function spawnNpmProduction(command: string, appDir: string, forceBuildFr
     npmExecArgs.push("--build-from-source")
   }
 
+  if (additionalArgs) {
+    if (Array.isArray(additionalArgs)) {
+      additionalArgs.forEach(arg => npmExecArgs.push(arg))
+    } else {
```
develar: Please put else on next line.

src/util/util.ts
```diff
@@ -54,6 +54,14 @@ export function spawnNpmProduction(command: string, appDir: string, forceBuildFr
     npmExecArgs.push("--build-from-source")
   }
 
+  if (additionalArgs) {
+    if (Array.isArray(additionalArgs)) {
+      additionalArgs.forEach(arg => npmExecArgs.push(arg))
```
develar: Please use spread operator here: `npmExecArgs.push(...additionalArgs)`

src/packager.ts
```diff
@@ -237,11 +237,12 @@ export class Packager implements BuildInfo {
       }
       else {
         const forceBuildFromSource = this.devMetadata.build.npmSkipBuildFromSource !== true
+        const additionalArgs = this.devMetadata.build.npmArgs
```
develar: Var is used only in the one else branch, please inline.

develar:
Probably we should always pass `--abi=ELECTRON_ABI_VERSION`?

mmckegg:
@develar yes, that would work too. But `ELECTRON_ABI_VERSION` needs to be computed, that text was just a placeholder. This can be obtained by executing `electron --abi`.

develar:
>  This can be obtained by executing electron --abi

I suppose it will be more good solution — to avoid explicit configuration. To simplify developer life.

src/util/util.ts
mmckegg: Fixed!

src/util/util.ts
mmckegg: Fixed!

mmckegg:
I've addressed those review notes. I think that electron-builder should handle setting the `--abi` for the user, but still think this is worth merging to handle any other strange cases like this.

@develar Shall I create an issue?

develar:
PR about abi version is welcome :) Thanks for spotting it.

PR #921, opened 2016-11-18
Add electron download options

FTAndy:
```javascript
function createDownloadOpts(opts: any, platform: string, arch: string, electronVersion: string) {
  const downloadOpts = Object.assign({
    cache: opts.cache,
    strictSSL: opts["strict-ssl"]
  }, opts.download)
```

the code has this option, but the doc does not declare it.


develar:
Thanks. In the future please edit `metadata.ts`.

PR #938, opened 2016-11-25
remove unnecesary comma

danielnieto:
text editors "complain" when using this snippet because this extra comma


develar:
Thanks for improving docs.

PR #955, opened 2016-11-30
fix: wine version processing

ml1nk:
If wine is installed on Fedora 24/25 over the repository, the version string is rather uncommon:
$ wine --version
wine-1.9.23 (Staging)

The version comparison fails with "TypeError: Invalid Version: 1.9.23 (Staging)" if you try to create a windows build. This pull should fix the problem by cutting the version string on the first whitespace.


develar:
Thanks!

Avien:
This is still happens to me on latest version 10.5.0 when trying to build on ubuntu ci (gitlab)

PR #962, opened 2016-12-02
Support calling from renderer process

dustintownsend:
I am handling the auto update code from the renderer process (using redux-saga). I added an additional OR condition: `require("electron").remote.app`.

Do you think this is useful? And is this the correct way to handle this?

### DIFF
--- a/nsis-auto-updater/src/NsisUpdater.ts
+++ b/nsis-auto-updater/src/NsisUpdater.ts
@@ -28,7 +28,7 @@ export class NsisUpdater extends EventEmitter {
   constructor(options?: PublishConfiguration | BintrayOptions | GithubOptions) {
     super()
 
-    this.app = (<any>global).__test_app || require("electron").app
+    this.app = (<any>global).__test_app || require("electron").app || require("electron").remote.app
 
     if (options == null) {
       this.clientPromise = this.loadUpdateConfig()


develar:
Sorry for late response. I don't think that it is useful because you should not use auto updater from renderer process — spawn/download should be executed in the main process.

dustintownsend:
I am calling this from the renderer to check if an update is available. This is disappointing.

develar:
I suggest to call main process to check updates. https://github.com/electron/electron/blob/master/docs/tutorial/quick-start.md#differences-between-main-process-and-renderer-process

dustintownsend:
@develar but isn't calling remote.app from the renderer basically the same thing? Since under the hood remote is just using IPC to send the calls from the renderer to the main. Or am I miss understanding?

I may open a new issue to discuss. Maybe there is something I can do to avoid needing to use remote for my use case. It's clear I need to do some more research on this first.

I will say that I currently have this running with beta users and it is working great. The nsis auto updater is great!

develar:
@dustintownsend Update is downloaded in the renderer process anyway. I don't think that it is good.

spawn is from renderer process, not main. I am not sure about security. 

Ok, I am not going to teach users.

nsis-auto-updater/src/NsisUpdater.ts
```diff
@@ -28,7 +28,7 @@ export class NsisUpdater extends EventEmitter {
   constructor(options?: PublishConfiguration | BintrayOptions | GithubOptions) {
     super()
 
-    this.app = (<any>global).__test_app || require("electron").app
+    this.app = (<any>global).__test_app || require("electron").app || require("electron").remote.app
```
develar: Please save `require("electron")` to temp var to avoid  several calls to require("electron")

dustintownsend:
@develar I got my use case working using IPC, so no need for this! Sorry to waste your time with this.

In case anyone needs to know how to do this, it's actually very simple.
Use `webContents.send` from the Main and `ipcRenderer.on` in the Renderer to forward the events from autoUpdater (like update-downloaded). And use `ipcRenderer.send` and `ipcMain.on` for `checkForUpdates` and `quitAndInstall`.

Issue #783, opened 2016-09-27
Description displayed rather than name?


robwilkerson:
- **Version**: 7.3.0

- **Target**: Windows

It looks like the `description` value from our `app/package.json` file might be appearing where it should be showing the `productName` value. Or maybe I'm just doing something wrong. Specifically, we see the description in the Task Manager.

Screenshots below, but here's a snippet from our `app/package.json` file:

``` javascript
{
  "name": "hughesnet-status-meter",
  "author": "Hughes Network Systems, LLC",
  "productName": "HughesNet Status Meter",
  "description": "Status Meter is like a fuel gauge for your HughesNet Data Allowance.",
  "main": "app.js",
  "version": "6.5.1-rc.4",
  ...
}
```

![name_taskmanager](https://cloud.githubusercontent.com/assets/7735/18876150/92118920-8496-11e6-88e5-90d950521026.PNG)
![name_start-up](https://cloud.githubusercontent.com/assets/7735/18876152/921420b8-8496-11e6-917a-e37b3a1eb1b4.PNG)

develar:
Please try NSIS target.

PR #965, opened 2016-12-02
Use productName for FileDescription on Windows

unindented:
`FileDescription` is used when right-clicking on a program in the task bar for example:

screen shot 2016-12-02 at 18 52 27

So I believe we should be displaying `appInfo.productName` there, as `appInfo.description` is more like a one-sentence description of your app, which could be used in an *About* dialog.

Edge, Internet Explorer, etc. use the same value for `ProductName` and `FileDescription`:

screen shot 2016-12-02 at 18 49 17

### DIFF
--- a/src/winPackager.ts
+++ b/src/winPackager.ts
@@ -149,7 +149,7 @@ export class WinPackager extends PlatformPackager<WinBuildOptions> {
     const args = [
       file,
       "--set-version-string", "CompanyName", appInfo.companyName,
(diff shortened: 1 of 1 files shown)


unindented:
My app was displaying the wrong string in some places:

screen shot 2016-12-02 at 18 47 27

screen shot 2016-12-02 at 18 48 50

develar:
Thanks a lot. Closes #783.

PR #969, opened 2016-12-03
Allow customization of NSIS installer icon

unindented:


develar:
Thanks! I don't ask you to add tests because I am going to simplify test assertions soon.

PR #979, opened 2016-12-06
feat: First start after update will start with a flag

maczikasz:
After updating with the auto updater, the first start will start with a "--updated" flag

### DIFF
--- a/templates/nsis/boringInstaller.nsh
+++ b/templates/nsis/boringInstaller.nsh
@@ -7,7 +7,13 @@
 
 !ifndef BUILD_UNINSTALLER
   Function StartApp
-    ${StdUtils.ExecShellAsUser} $0 "$SMPROGRAMS\${PRODUCT_FILENAME}.lnk" "open" ""
+    ${GetParameters} $R0
+    ${GetOptions} $R0 "--update" $R1
+    ${IfNot} ${Errors}
+      ${StdUtils.ExecShellAsUser} $0 "$SMPROGRAMS\${PRODUCT_FILENAME}.lnk" "open" "--updated"
+    ${Else}
+      ${StdUtils.ExecShellAsUser} $0 "$SMPROGRAMS\${PRODUCT_FILENAME}.lnk" "open" ""
+    ${endif}
   FunctionEnd
 
   !define MUI_FINISHPAGE_RUN
--- a/templates/nsis/boringInstaller.nsh
+++ b/templates/nsis/boringInstaller.nsh
@@ -16,10 +16,19 @@
     ${endif}
   FunctionEnd
 
+  Function licensePre
+      ${GetParameters} $R0
+      ${GetOptions} $R0 "--update" $R1
+      ${IfNot} ${Errors}
+        Abort
+      ${endif}
+  FunctionEnd
+
   !define MUI_FINISHPAGE_RUN
   !define MUI_FINISHPAGE_RUN_FUNCTION "StartApp"
 
   !ifdef LICENSE_FILE
+    !define MUI_PAGE_CUSTOMFUNCTION_PRE licensePre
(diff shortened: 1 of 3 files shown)


maczikasz:
I didn't add tests yet because I wasn't sure where it would make most sense, and how much should I modify in the original files or how much stuff can I add for this very small change

develar:
Please modify https://github.com/electron-userland/electron-builder/blob/master/templates/nsis/boringInstaller.nsh#L10 also (sorry, I didn't mentioned this place).

develar:
> I didn't add tests yet

Yes. it is a little bit complicated and not enabled by default. Please see https://github.com/electron-userland/electron-builder/blob/master/test/src/windows/nsisTest.ts#L86

Writing test will be much more complex than implementation.

Have you tested it manually?

maczikasz:
I did test it manually and it worked

I will add to the boring installer, I thought it was on purpose, because if you are running an update you wouldn't want the boring updater, but yeah also possible.

One other thing I was thinking about, but that's probably not part of the PR just an idea and I am interested in your opinion, if you do an update, do you really want to show the licence again?

develar:
> do you really want to show the licence again

Nice catch. Well... What if license was changed? But in general, 👍  In most cases it makes sense to not show license again. The same as in case of macOS — if dmg has license, it is not included into the zip package update.

maczikasz:
I changed the boring installer (still only manual tests)

Would you want me to implement the lincence thing or would you'd like to think about a smarter solution (changing licence or smth)

develar:
> Would you want me to implement the lincence thing

Yes, simple check is enough. http://nsis.sourceforge.net/Skipping_Pages

maczikasz:
I added the skipping and tested manually (both boring and oneClick)

templates/nsis/boringInstaller.nsh
```diff
@@ -7,13 +7,28 @@
...
+    ${GetParameters} $R0
+    ${GetOptions} $R0 "--update" $R1
+    ${IfNot} ${Errors}
+      ${StdUtils.ExecShellAsUser} $0 "$SMPROGRAMS\${PRODUCT_FILENAME}.lnk" "open" "--updated"
+    ${Else}
+      ${StdUtils.ExecShellAsUser} $0 "$SMPROGRAMS\${PRODUCT_FILENAME}.lnk" "open" ""
+    ${endif}
+  FunctionEnd
+
+  Function licensePre
```
develar: Could you please move `licensePre` function to [common.nsh](https://github.com/electron-userland/electron-builder/blob/master/templates/nsis/common.nsh) (after line `!ifndef BUILD_UNINSTALLER`)?

maczikasz: sure

maczikasz: Error: resolving install function "licensePre" in function "mui.LicensePre_22.5.19"
 it doesn't work any more, do I need to do something like export or change visiblity?

develar: Strange. Ok, leave it as is.

develar:
> warning: install function "licensePre" not referenced - zeroing code (306-315) out
> https://circleci.com/gh/electron-userland/electron-builder/352?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Please wrap function into `!ifdef LICENSE_FILE` to fix this warning.

maczikasz:
Done I moved it to the already existing ifdef

develar:
Thanks a lot! Nice feature.

maczikasz:
Could you pls release a version of "electron-auto-updater" that contains change?

develar:
electron-auto-updater@0.7.0 published.

maczikasz:
awesom thanks vm

PR #986, opened 2016-12-07
fix: improve error message for case of wrongly set electron* dependencies

seeekr:


seeekr:
I cloned the project to my local machine and tried to commit, but always got an error from the `npm run precommit` hook that my commit message is empty, which it was clearly not.
Did a little bit of digging and in the `validate-commit-msg` the line `var commitMsgFile = process.argv[2] || getGitFolder() + '/COMMIT_EDITMSG';` is causing the issue -- there's no commit message file argument, and neither exists a file `.git/COMMIT_EDITMSG`. In case anyone wants to look further.

seeekr:
@develar Holy crap, you're too fast for me ;) I was still fiddling around with the commit error issue. :)

develar:
> I cloned the project to my local machine and tried to commit,

Do you use IDEA or terminal?

seeekr:
Tried both. First WebStorm, then via terminal.
```
✗ git --version
git version 2.11.0
```

develar:
Ok, I have experienced this error as well. Well, I am going in this case remove this validation since it it doesn't matter for contributors – in any case I can change commit message on squash.

seeekr:
Yeah, probably a good call. Stuff like this must be rock-solid, otherwise the loss of potential contribution is just too high -- needs to be really low-friction. :)

Issue #956, opened 2016-12-01
Building Electron app symlink error


Jflinchum:
* **Version**:  10.3.0

* **Target**: macOS

When trying to run `build --mac` for my Electron app, I get this error:
```
yarn run v0.17.9
$ build --mac
Skip app dependencies rebuild because npmRebuild is set to false
⚠️  Packaging using asar archive is disabled — it is strongly not recommended.
Please enable asar and use asarUnpack to unpack files that must be externally available.
Packaging for darwin x64 using electron 1.4.7 to dist/mac
Error: ENOENT: no such file or directory, symlink 'Versions/Current/Electron Framework' -> '/Users/jflinchum/Documents/myApp/dist/mac/Electron.app/Contents/Resources/app/node_modules/electron-prebuilt/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Electron Framework'
    at Error (native)
From previous event:
    at /Users/jflinchum/Documents/myApp/node_modules/electron-builder/src/util/fs.ts:178:13
    at next (native)
From previous event:
    at walk (/Users/jflinchum/Documents/myApp/node_modules/electron-builder/out/util/fs.js:236:26)
    at /Users/jflinchum/Documents/myApp/node_modules/electron-builder/src/util/fs.ts:74:44
    at /Users/jflinchum/Documents/myApp/node_modules/graceful-fs/polyfills.js:279:18
    at FSReqWrap.oncomplete (fs.js:123:15)
From previous event:
    at /Users/jflinchum/Documents/myApp/node_modules/electron-builder/src/util/fs.ts:62:10
    at runCallback (timers.js:574:20)
    at tryOnImmediate (timers.js:554:5)
    at processImmediate [as _immediateCallback] (timers.js:533:5)
From previous event:
    at /Users/jflinchum/Documents/myApp/node_modules/electron-builder/src/util/fs.ts:59:27
    at next (native)
    at go$readdir$cb (/Users/jflinchum/Documents/myApp/node_modules/graceful-fs/graceful-fs.js:149:14)
    at FSReqWrap.oncomplete (fs.js:123:15)
From previous event:
    at walk (/Users/jflinchum/Documents/myApp/node_modules/electron-builder/out/util/fs.js:90:22)
    at copyDir (/Users/jflinchum/Documents/myApp/node_modules/electron-builder/src/util/fs.ts:176:3)
[...]

PR #998, opened 2016-12-12
Fixes bug affecting OSX in which copyDir walks to a symlink referenci…

f3z0:
…ng a file that has not yet been copied by postponing the linking step until after the full walk is complete. Fixes issue #956


develar:
Thank, I added test and merged.

PR #1016, opened 2016-12-15
appx, use arch from parameters instead of process.arch

black-snow:
Use the architecture (ia32 / x64) that's been passed as a parameter instead of process.arch when calling makeappx.exe. Allows for 32-bit builds on 64-bit machines.

### DIFF
--- a/src/targets/appx.ts
+++ b/src/targets/appx.ts
@@ -63,7 +63,7 @@ export default class AppXTarget extends Target {
     use(this.options.makeappxArgs, (it: Array<string>) => args.push(...it))
     // wine supports only ia32 binary in any case makeappx crashed on wine
     // await execWine(path.join(await getSignVendorPath(), "windows-10", process.platform === "win32" ? process.arch : "ia32", "makeappx.exe"), args)
-    await spawn(path.join(await getSignVendorPath(), "windows-10", arch == Arch.ia32 ? "ia32" : "x64", "makeappx.exe"), args)
+    await spawn(path.join(await getSignVendorPath(), "windows-10", arch === Arch.ia32 ? "ia32" : "x64", "makeappx.exe"), args)
 
     await packager.sign(destination)
     packager.dispatchArtifactCreated(destination, packager.generateName("appx", arch, true))
@@ -102,7 +102,7 @@ export default class AppXTarget extends Target {
             return safeName
             
           case "arch":
-            return arch == Arch.ia32 ? "x86" : "x64"
+            return arch === Arch.ia32 ? "x86" : "x64"
 
           default:
             throw new Error(`Macro ${p1} is not defined`)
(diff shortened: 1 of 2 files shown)


black-snow:
I wonder why there's a change on line 110 ...

black-snow:
Manifest will now have the correct value for `ProcessorArchitecture` and the correct `makeappx.exe` will be used.

black-snow:
Well, seems like it's not :D

src/targets/appx.ts
```diff
@@ -100,11 +100,14 @@ export default class AppXTarget extends Target {
 
           case "safeName":
             return safeName
+            
+          case "arch":
+            return arch == Arch.ia32 ? "x86" : "x64"
```
develar: Please use `===` here.

black-snow: @develar CI failed when using triple =, see #34bfea6

src/targets/appx.ts
```diff
@@ -55,21 +55,21 @@ export default class AppXTarget extends Target {
...
+      this.writeManifest(templatePath, preAppx, safeName, arch)
     ])
 
     const destination = path.join(this.outDir, packager.generateName("appx", arch, false))
     const args = ["pack", "/o", "/d", preAppx, "/p", destination]
     use(this.options.makeappxArgs, (it: Array<string>) => args.push(...it))
     // wine supports only ia32 binary in any case makeappx crashed on wine
     // await execWine(path.join(await getSignVendorPath(), "windows-10", process.platform === "win32" ? process.arch : "ia32", "makeappx.exe"), args)
-    await spawn(path.join(await getSignVendorPath(), "windows-10", process.arch, "makeappx.exe"), args)
+    await spawn(path.join(await getSignVendorPath(), "windows-10", arch == Arch.ia32 ? "ia32" : "x64", "makeappx.exe"), args)
```
develar: Please use `===` here.

src/targets/appx.ts
black-snow: Or did it fail because of the Arch/string thing?

develar:
Are you sure that 64bit makeappx cannot produce 32bit appx package? Could you please check it?

black-snow:
@develar I think it can, that's why we're doing this at all. Did you mean "can" instead of "cannot"?

develar:
> seems to be the way to go - I can ask Microsoft if you wish, though

You can check generated appx package — does it work or no. Ok — in any case for now it is better to use 32 bit to produce appx for 32.

black-snow:
Yes it works. But that has nothing to say for I'm on a 64bit machine. If the generated appx was still x64 I wouldn't feel any difference (unless Windows minds that it's x86 in the manifest but a 64 bit exe or something).
Anyhow, I think we did it right. Can probably confirm on wednesday. The Windows Store accepted both my packages and correctly listed them as x86/x64.

Issue #959, opened 2016-12-01
NSIS auto-update doesn't use system HTTP proxy


dbkr:
* **Version**: 10.4.1
* **Target**: Windows (NSIS)

The NSIS auto-updater appears not to use a system HTTP proxy if Windows is set up with one and instead tries to connect to the update URL directly. On machines where the only web access is via the proxy, this means the app will just not update (even though the embedded chromium appears to use the proxy).

develar:
Yes. We must not use node http API, but instead use chromium electron API.

maczikasz:
So I spend some time looking into this feature and assuming you haven't started any work on this, I would propose and do the following:

- Consolidate all HTTP related requests behind a common interface (stuff from publish/restApiRequest and util/httpRequest
- and create a HTTP based implementation for this (for publishing and stuff done by the builder)
- and create an Electron NET api based implementation for the stuff done in AutoUpdater

So then this IF will have to be used by all Providers and Publishers

develar:
Yes. Electron Chrome Network API mimics node, so should be easy to migrate existing helpers publish/restApiRequest and util/httpRequest.

maczikasz:
Yes, my point was, that it should be done in a manner where it's possible to use normal HTTP, as net api requires app ready event, also when the app is building or the tests are running there is no electron app, I will create the PR either during the weekend or latest on monday

maurovitale:
Hi to all,
my electron proxy setting doesn't work with AutoUpdater.  On the main process I have this configuration : 
      
```
session.defaultSession.setProxy({
    proxyRules : proxyString
 }, function () {
    console.log("proxy ok")
 });

```
This works in the entire electron app. I can see the logs on squid. AutoUpdater should use Electron Chrome Network API but in my case this not works, the method 'checkForUpdates' of AutoUpdater bypass the proxy. Is there something that I miss ?

akioyamamoto1977:
I have same issue with maurovitale.
Is there any conclusion if there is a way to indicate a proxy server by program code ?

akioyamamoto1977:
> I have same issue with maurovitale.
> Is there any conclusion if there is a way to indicate a proxy server by program code ?

Excuse me putting a comment on this closed issue but I found a  solution to indicate a default proxy (for Electron) from program code as below.

```
app.commandLine.appendSwitch('proxy-server', '127.0.0.1:2222');
```

elonzh:
> > I have same issue with maurovitale.
[...]

PR #1024, opened 2016-12-17
NSIS auto-update doesn't use system HTTP proxy (#959)

maczikasz:
feat: Use Electron.Net to send http requests for auto updater

There was a refactoring around httpRequests and util/restApiRequests, they both forward the calls to an instance of the HttpExecutor interface, which can be Node or Electron. Node is used by default and electron is only used in case it's an electron app running the auto updater


maczikasz:
I haven't tested it yet on the corporate proxy, but electron net works so I assume it will work also, but I can verify on monday

develar:
I have merged PR with some changes:

* no need to pass httpExecutor — we can just use executorHolder
* no need to add electron to dependencies.
* no need to add electron typings to nsis-auto-updater — parent typings is enough. Probably later nsis-auto-updater will be standalone package, but for now it anyway uses sources of electron-builder.

I didn't publish nsis-auto-updater because i didn't check it manually and our current test suite doesn't test auto update anymore — we must run nsisUpdaterTest against electron, but not node (probably using https://github.com/segmentio/nightmare). It will be cool if you will implement it, because no sense to test auto updater using node if it differs from the real app.

develar:
@maczikasz Thank you, without your PR this important task was postponed for a unknown time.

maczikasz:
Thank you for merging the PR, I will do a manual test today on our proxy ( I did manual test w/o proxy just proving regression). Thanks for cleaning it up a bit, as you see I am not a JS/TS dev so I had some fears that I had written in a very stupid/weird manner :)

I completely agree with your statement about the tests, I will try to implement them with the lib you suggested, let's see if I can make it work.

maczikasz:
Fyi it works behind our corporate proxy manually, but I will still try to do the tests (possibly with Spectron rather than nightmare)

maczikasz:
Hi @develar do I see right you released a new version of the auto-updater after all? I see a release in npm, but I thought it will not get released.

Anyway I will finish the tests in beginning on Jan, I have a FW that would work but I have an issue with Spectron (https://github.com/electron/spectron/issues/153) which stops me from being able to actually do it :(

Will update you if there is any progress there

develar:
I trust you and as you reported that it works I decided to publish because feature that you implemented is awesome.

Issue #1056, opened 2016-12-28
pack-updater script is broken


badams:
* **Version**: 10.13.1
* **Target**:  N/A

```bash
npm run pack-updater
> electron-builder@0.0.0-semantic-release pack-updater /Users/byron/dev/electron-builder
> cd nsis-auto-updater && node ./test/vendor/yarn.js --production && cd ..

module.js:471
    throw err;
    ^

Error: Cannot find module '/Users/byron/dev/electron-builder/nsis-auto-updater/test/vendor/yarn.js'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
```

As far as I can tell this has been broken since [eff85c38](https://github.com/electron-userland/electron-builder/commit/eff85c38e99a7e11c47c1e40464258dfcae069bd#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)

badams:
https://github.com/electron-userland/electron-builder/pull/1057#issuecomment-269434192

PR #1057, opened 2016-12-28
fix(electron-auto-update): fix pack-updater script

badams:
Fix for #1056


develar:
This is script is not used. But not yet deleted :)

badams:
Ah I was unsure what its purpose was, but ran it anyway when packing my fork for nsis-auto-updater

develar:
Sorry for confusion, I will remove it.

PR #1060, opened 2016-12-29
fix(linux): executable breakes if productName given

stanfieldr:
Change the path to the executable so the symlink in /usr/local/bin does not break


src/targets/fpm.ts
```diff
@@ -50,6 +50,7 @@ export default class FpmTarget extends Target {
     const templateOptions = Object.assign({
       // old API compatibility
       executable: this.packager.executableName,
+      productName: this.packager.info.metadata.productName
```
develar: Please change to `this.packager.appInfo.productFilename`

develar:
Thanks, merged.

PR #1061, opened 2016-12-29
NsisUpdater:the protocol of server can be configured

kazusa0106:


develar:
Please fix compilation error. Have you tested your solution?

mattlyons0:
I was assuming http wasn't supported as a design decision as there isn't really any reason not to have a https server because of LetsEncrypt.

PR #1066, opened 2016-12-30
Install chrome & xvfb to allow for e2e testing

Superjo149:
Just added the xvfb package and chrome stable browser from google's repo. Installing this will enable you to also use this image for headless testing. 

This will allow you to do something like this:
`xvfb-run npm run test-e2e`


develar:
Docker images wil be rebuild in 10-60 minutes. Thanks for contribution.

PR #1067, opened 2016-12-30
fix(electron-auto-updater): use Github public API on Github provider

s-coimbra21:
It's possible to get the latest release using the github api, see [reference](https://developer.github.com/v3/repos/releases/#get-the-latest-release)

I believe this method to be more robust and efficient than reading the redirect location of a request to the /:owner/:repo/releases/latest endpoint.

EDIT: I'm going to open an issue instead because I just saw that you had already decided to not use the API due to the rate limit, sorry for the rushed PR.


s-coimbra21:
I should add that I was getting a ` 406 Not Acceptable` response from Github when trying to auto-update with the current release of `electron-auto-update`.

develar:
> due to the rate limit, sorry for the rushed PR.

exactly. Github has very funny small limit. Not suitable for updates. Especially for Russia / China / Belarus / etc.

Issue #1058, opened 2016-12-28
How to auto-update in windows, completely lost


KYSSelectron:
I'm trying to understand how auto-update works with electron-builder when using windows (10) and packing for windows. I'm however lost in the docs (mainly https://github.com/electron-userland/electron-builder/wiki/Auto-Update).

What do I have to write in my dev package.json?

```
{
  "name": "myapp",
  "license": "GPL-3.0",
  "productName": "myapp",
  "author": "me",
  "description": "yes",
  "homepage": "https://myapp.github.io/",
  "repository": {
    "type": "git",
    "url": "https://github.com/myapp/myapp.github.io"
  },
  "devDependencies": {
    "electron": "^1.4.12",
    "electron-builder": "^10.7.0"
  },
  "dependencies": {
  },
  "scripts": {
    "start": "./node_modules/.bin/electron app",
    "build": "build --win --x64",
    "postinstall": "install-app-deps"
  },
  "build": {
    "appId": "my.app",
    "productName": "myapp",
    "win": {
      "icon": "./build/icon.ico"
    },
    "publish": {
      "provider": "github",
      "repo": "https://github.com/myapp/myapp.github.io"
    }
  }
}
```

Then, in my main.js page I have this

```
var autoUpdater = require('electron-auto-updater').autoUpdater;
// import {autoUpdater} from "electron-auto-updater" -> unexpected token import

app.on('ready', function() {
    autoUpdater.checkForUpdates();
});

autoUpdater.addListener("update-available", function(event) {
});

autoUpdater.addListener("update-downloaded", function(event, releaseNotes, releaseName, releaseDate, updateURL) {
    autoUpdater.quitAndInstall();
    return true;
});

autoUpdater.addListener("update-not-available", function() {
});
```

And that's it. What am I supposed to upload to the github repository? How can I check what's going on? The build works well in the sense that I get the installer .exe that installs and creates a shortcut of the application and this works. But other than that, auto-update is not there and I don't know what can I try.

develar:
Step 1: Install electron-auto-updater as app dependency. Did you install it?

> What am I supposed to upload to the github repository

It will be automatically upload to GitHub repository on CI server. Define `GH_TOKEN`. On dev machine you can force publish using `--publish always `

kazusa0106:
Is the server for auto-update must be https?I got "ERR_SSL_PROTOCOL_ERROR" after cheking-for-update.

KYSSelectron:
[...]

Issue #1038, opened 2016-12-22
auto-update error: 406 not acceptable


cawa-93:
* **Version**: 10.9.3

* **Target**: Windows --ia32

After `checking-for-update` event fire `error` event with error: [debug.log](https://github.com/electron-userland/electron-builder/files/669861/debug.2.txt)

**My app**:
* [releases](https://github.com/cawa-93/ad_maker/releases/)
* [main.js](https://github.com/cawa-93/ad_maker/blob/master/app/main.js#L44)
* [package.json](https://github.com/cawa-93/ad_maker/blob/master/package.json)
* [app package.json](https://github.com/cawa-93/ad_maker/blob/master/app/package.json)

s-coimbra21:
* Version: 10.15.1
* Target: Windows (NSIS)

Happening to me consistently. I tried to debug this for a bit and saw that the request containing the Location header (for the redirect) wasn't being picked up.

badams:
This seems to be caused by a change on Github's end, am looking into it

s-coimbra21:
Does electron's `net.request` let you handle 302 redirects yourself?

badams:
@s-coimbra21 I believe so, but I've opted for a different approach when retrieving the latest version tag in my PR (#1068)

develar:
@badams did a great job, electron-auto-updater@0.8.6 (tag: next) published. Tag `next` because I don't have time to make all tests and we don't yet have special electron tests. Please try it.

PR #1068, opened 2016-12-31
fix(electron-auto-updater) Checking for updates from github was failing

badams:
Appears that github are now more strict about the `Accept` header, and invalid requests are resulting in a [406 response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406)

I've adjusted the httpExecutor.request method to allow for custom header objects to be passed, so user can specify header (eg. `Accept: application/json`)

I've also updated the way we retrieve the latest version tag; 
when sending `Accept: application/json` to `/:user/:repo/releases/latest` github will respond with a JSON payload like below

```json
{
  "tag_name": "v1.0.0",
  ...
}
```

So it can be easily extracted without relying on redirect headers or the official API

#1038 #1058

### DIFF
--- a/nsis-auto-updater/src/GitHubProvider.ts
+++ b/nsis-auto-updater/src/GitHubProvider.ts
@@ -11,10 +11,16 @@ export class GitHubProvider implements Provider<VersionInfo> {
   async getLatestVersion(): Promise<UpdateInfo> {
     // do not use API to avoid limit
     const basePath = this.getBasePath()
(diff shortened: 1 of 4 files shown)


develar:
Amazing! Thanks a lot. For some technical (unrelated) reasons PR cannot be accepted right now, but I hope to merge it today.

develar:
Don't worry about merge conflicts :)

develar:
electron-auto-updater@0.8.6 (tag: next) published with your fix. Tag `next` because I don't have time to make all tests and we don't yet have special electron tests.

Issue #1045, opened 2016-12-24
Checksum mismatch on Windows NSIS


Daymannovaes:
* **Version**: 10.8.1

* **Target**: NSIS
(building from windows too)

* **Node version**: 6.9.1
* **npm version**: 4.0.3

I'm getting checksum mismatch, like following in the screenshot

![screenshot 18](https://cloud.githubusercontent.com/assets/3648665/21464447/0b89b716-c964-11e6-9988-cd59709bfed1.png)

I uploaded the executables with `--publish=always` at github and didn't change anything.
The fact that I'm checking for updates a lot of times even ater the download starts, can affect this?

Because just for test purpose, I put the check interval to 30 seconds, so I get `update-available` a bunch of times.

Plus, I'm getting everytime a different checksum. What's going on?

My autoupdater module is very simple:

```
autoUpdater.addListener('update-available', () => {
  logger.info('update-available');
});
autoUpdater.addListener('update-downloaded', (event, releaseNotes, releaseName, releaseDate, updateURL) => {
  autoUpdater.quitAndInstall();
});
autoUpdater.addListener('error', (error) => {
  logger.error('error when trying to check for update');
  logger.error(error);
});
autoUpdater.addListener('checking-for-update', () => {
  logger.info('checking-for-update');
});
autoUpdater.addListener('update-not-available', () => {
  logger.info('update-not-available');
});

async function setup() {
  autoUpdater.checkForUpdates();

  await sleep(30 * 1000);
  setup();
}
setup();
```

And but package config is simple too:

```
"build": {
    "appId": "some-id",
    "files": [
      "node_modules/**/*",
      ".env",
      "app/**/*"
    ],
    "publish": [
      {
        "provider": "github",
        "repo": "some-repo"
      }
    ],
    "win": {
      "certificateFile": "./certs/my_signing_key.pfx",
      "certificatePassword": ""
    }
  }
```

develar:
Probably. Confirmed that we don't queue check updates requests. Should be fixed.

tvanro:
**electron-builder**: 10.16.0
**electron-auto-updater**: 0.9.0

**Target**: NSIS
(building from docker on Linux)

I'm receiving the exact same error so there's still something going on I'm afraid.
Like with @Daymannovaes, the second checksum is different on each check from electron-builder. My guess is that it's doing the check on a whole file-download request instead of only the file.

develar:
[...]

PR #1077, opened 2017-01-03
fix: add request.pause() to notification about progress

maczikasz:
See issue #1045


develar:
cc @badams I think proper fix is to not listen `data`, but use `pipe` in the same way as we already use `progress-stream` for download.

develar:
To be clear — I don't to use `progress-stream` here to reduce external deps. So, probably approach like our DigestTransform should be used here.

maczikasz:
ok will take a look how it's done

maczikasz:
@develar I pushed a new commit, did you mean like that?

develar:
@maczikasz Thank you! 0.9.1 published with your fix.

badams:
@develar @maczikasz Nice work, this is a much cleaner solution 👍

PR #1103, opened 2017-01-11
fix(codeSign): signing with 'Mac Developer' failed

ericzhangjx:
Removed codes of throwing exceptions inside `findIdentity()`, and let the `sign()` in `macPackager.ts` itself handle the returned value of `findIdentity()` (which should be string|null) and then try to sign with certType==='Mac Developer' before exporting unsigned application;

### DIFF
--- a/packages/electron-builder/src/codeSign.ts
+++ b/packages/electron-builder/src/codeSign.ts
@@ -10,7 +10,7 @@ import { homedir } from "os"
 import { statOrNull } from "electron-builder-util/out/fs"
 import isCi from "is-ci"
 
-export const appleCertificatePrefixes = ["Developer ID Application:", "Developer ID Installer:", "3rd Party Mac Developer Application:", "3rd Party Mac Developer Installer:"]
+export const appleCertificatePrefixes = ["Developer ID Application:", "Developer ID Installer:", "3rd Party Mac Developer Application:", "3rd Party Mac Developer Installer:", "Mac Developer:"]
 
 export type CertType = "Developer ID Application" | "Developer ID Installer" | "3rd Party Mac Developer Application" | "3rd Party Mac Developer Installer" | "Mac Developer"
 
@@ -221,11 +221,7 @@ export async function findIdentity(certType: CertType, qualifier?: string | null
     for (const prefix of appleCertificatePrefixes) {
       checkPrefix(identity, prefix)
     }
-    const result = await _findIdentity(certType, identity, keychain)
-    if (result == null) {
-      throw new Error(`Identity name "${identity}" is specified, but no valid identity with this name in the keychain`)
-    }
-    return result
+    return await _findIdentity(certType, identity, keychain)
   }
 }
 


packages/electron-builder/src/codeSign.ts
```diff
@@ -10,7 +10,7 @@ import { homedir } from "os"
 import { statOrNull } from "electron-builder-util/out/fs"
 import isCi from "is-ci"
 
-export const appleCertificatePrefixes = ["Developer ID Application:", "Developer ID Installer:", "3rd Party Mac Developer Application:", "3rd Party Mac Developer Installer:"]
+export const appleCertificatePrefixes = ["Developer ID Application:", "Developer ID Installer:", "3rd Party Mac Developer Application:", "3rd Party Mac Developer Installer:", "Mac Developer:"]
```
develar: Not necessary and can lead to major issues.

packages/electron-builder/src/codeSign.ts
```diff
@@ -221,16 +221,12 @@ export async function findIdentity(certType: CertType, qualifier?: string | null
     for (const prefix of appleCertificatePrefixes) {
       checkPrefix(identity, prefix)
     }
-    const result = await _findIdentity(certType, identity, keychain)
-    if (result == null) {
-      throw new Error(`Identity name "${identity}" is specified, but no valid identity with this name in the keychain`)
-    }
-    return result
+    return await _findIdentity(certType, identity, keychain)
```
develar: If check removed here, it should be added to macPackager.ts

ericzhangjx: [macPackager.ts](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/src/macPackager.ts#L149)
```
    let name = await findIdentity(isMas ? "3rd Party Mac Developer Application" : "Developer ID Application", isMas ? masQualifier : this.platformSpecificBuildOptions.identity, keychainName)
    if (name == null) {
      if (!isMas) {
        name = await findIdentity("Mac Developer", this.platformSpecificBuildOptions.identity, keychainName)
        if (name != null) {
          warn("Mac Developer is used to sign app — it is only for development and testing, not for production")
        }
      }

      if (name == null) {
        const message = process.env.CSC_IDENTITY_AUTO_DISCOVERY === "false" ?
          `App is not signed: env CSC_IDENTITY_AUTO_DISCOVERY is set to false` :
          `App is not signed: cannot find valid ${isMas ? '"3rd Party Mac Developer Application" identity' : `"Developer ID Application" identity or custom non-Apple code signing certificate`}, see https://github.com/electron-userland/electron-builder/wiki/Code-Signing`
        if (isMas || this.platformSpecificBuildOptions.forceCodeSigning) {
          throw new Error(message)
        }
        else {
          warn(message)
          return
        }
      }
    }
```

As for the current logic in `macPackager.ts`, when provided with an identity string, it first try to get it as a "3rd Party Mac Developer Installer"/"Developer ID Installer"; If returns null, it will do an additional identity query for Darwin as 'Mac Developer'; If returns null again, it will decide to either throw an error or display a warning message depending on the build options(`isMas || this.platformSpecificBuildOptions.forceCodeSigning`);

So here in `findIdentity()` I don't see why it should throw an error when failed to get identity. It can just return `string|null` denoting whether it find the requested identity or not, and let `macPackager.ts` do the result checking part. `findIdentity()` should not throw error before every options are tried in `macPackager.ts`.

develar: if `this.platformSpecificBuildOptions.identity` is specified explicitly, we must throw error. Currently, we will throw error in the macPackager only for `mas`. If check removed from `findIdentity`, here, in the macPackager, we must add corresponding check.

ericzhangjx: Get it, thanks

develar:
@ericzhangjx Thank you for investigation.

Issue #1112, opened 2017-01-13
GitHub auto-update on Windows returning error


JustinBeckwith:
* **Version**: 10.12.0
* **Target**:  Windows

I'm having some trouble getting auto-updates for Windows to work.  I've installed the `electron-auto-updater` npm module, and mostly got everything set up. My builds are producing app-latest.yml files.  When I try running my installed app, I'm getting this error:

```
[2017-01-13 07:53:13:0250] [error] SyntaxError: Unexpected end of JSON input
SyntaxError: Unexpected end of JSON input
    at IncomingMessage.response.on.e (C:\Users\beckwith\AppData\Local\Programs\app-engine-trebuchet\resources\app.asar\node_modules\electron-auto-updater\src\electronHttpExecutor.ts:132:55)
    at emitNone (events.js:86:13)
    at IncomingMessage.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:974:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
From previous event:
    at ElectronHttpExecutor.doApiRequest (C:\Users\beckwith\AppData\Local\Programs\app-engine-trebuchet\resources\app.asar\node_modules\electron-auto-updater\src\electronHttpExecutor.ts:89:5)
    at ElectronHttpExecutor.request (C:\Users\beckwith\AppData\Local\Programs\app-engine-trebuchet\resources\app.asar\node_modules\electron-builder-http\src\httpExecutor.ts:52:17)
    at request (C:\Users\beckwith\AppData\Local\Programs\app-engine-trebuchet\resources\app.asar\node_modules\electron-builder-http\src\httpExecutor.ts:147:38)
    at C:\Users\beckwith\AppData\Local\Programs\app-engine-trebuchet\resources\app.asar\node_modules\electron-auto-updater\src\GitHubProvider.ts:14:26
    at next (<anonymous>)
From previous event:
    at GitHubProvider.getLatestVersion (C:\Users\beckwith\AppData\Local\Programs\app-engine-trebuchet\resources\app.asar\node_modules\electron-auto-updater\out\GitHubProvider.js:60:11)
    at C:\Users\beckwith\AppData\Local\Programs\app-engine-trebuchet\resources\app.asar\node_modules\electron-auto-updater\src\NsisUpdater.ts:110:38
From previous event:
[...]

PR #1114, opened 2017-01-13
fix(electron-auto-updater-github): Add better error handling for gith…

badams:
…ub releases #1112


badams:
haven't had a chance to test this solution entirely, will confirm

PR #1116, opened 2017-01-14
fix(docker): add missing libcurl3 for osslsigncode

salomvary:
Windows build *without* code signing works fine but when code signing is
enabled `osslsigncode` fails. It's apparently due to libcurl shared
dependency mismatch.

    /root/.cache/electron-builder/winCodeSign/winCodeSign-1.6.0/linux/osslsigncode:
    error while loading shared libraries: libcurl.so.4: cannot open
    shared object file: No such file or directory

Installing libcurl3 fixes the issue.


develar:
Thanks. Build started. Will be ready in 1 hour.

salomvary:
Works on the latest Docker image 👍

PR #1148, opened 2017-01-19
fix: Arch & pacman use i686 instead of i386 and pkg.tar.xz

hleitzell:
Digging through the code, this is probably not the ideal way to do this but it follows what was done for the 'deb' packaging.

Arch Linux (pacman) does not have packages available for i386 and only builds for i686 architecture or better.

Also added in the dependencies of Electron on Arch and changed the filename from ".pacman" to the correct "pkg.tar.xz".


develar:
Cool. Thanks a lot.

PR #1173, opened 2017-01-26
Show welcome and finish page duing uninstall for boring uninstall

### LINKED ISSUE #885: option to delete app data on manual uninstall
https://github.com/electron-userland/electron-builder/issues/528#issuecomment-258673100

alexbrazier:
This change will show a welcome and finish page during the uninstall process if using the boring installer.
The welcome page removes the need to have the message box asking if you are sure you want to uninstall.
The finish page makes it more clear that it was successfully uninstalled. It also makes it possible to customise the finish page as seen in the screenshot.

image

image

To show the remove user data box you can add an include script that contains something like the following, which closes #885 :

```
!ifdef BUILD_UNINSTALLER
  Function un.AddAppData
    RMDir /r "$APPDATA\Rocket.Chat+"
  FunctionEnd

  ; Using the read me setting to add option to remove app data
  !define MUI_FINISHPAGE_SHOWREADME
  !define MUI_FINISHPAGE_SHOWREADME_TEXT "Remove user data"
  !define MUI_FINISHPAGE_SHOWREADME_NOTCHECKED
  !define MUI_FINISHPAGE_SHOWREADME_FUNCTION un.AddAppData

!endif
```

### DIFF
--- a/packages/electron-builder/templates/nsis/boringInstaller.nsh
+++ b/packages/electron-builder/templates/nsis/boringInstaller.nsh
@@ -47,10 +47,12 @@
   !insertmacro MUI_PAGE_INSTFILES
   !insertmacro MUI_PAGE_FINISH
 !else
+  !insertmacro MUI_UNPAGE_WELCOME
   !ifndef INSTALL_MODE_PER_ALL_USERS
     !insertmacro PAGE_INSTALL_MODE
   !endif
   !insertmacro MUI_UNPAGE_INSTFILES
+  !insertmacro MUI_UNPAGE_FINISH
 !endif
 
 !macro initMultiUser
--- a/packages/electron-builder/templates/nsis/uninstaller.nsh
+++ b/packages/electron-builder/templates/nsis/uninstaller.nsh
@@ -2,10 +2,10 @@ Function un.onInit
   !insertmacro check64BitAndSetRegView
 
   ${IfNot} ${Silent}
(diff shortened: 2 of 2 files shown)


develar:
Cool.

> To show the remove user data box you can add

Could you please file PR to add this functionally by default?

develar:
Published as 11.7.0

moulinierf:
Hi Alex, based on this feature, it seems there is an issue to interact with the unwelcome page.
by default when we use electron builder with nsis, we have now during the uninstallation , the unwelcome page and the unfinish page like you described bellow. And it's cool
but we didn't interact with them like this

!define MUI_PAGE_CUSTOMFUNCTION_PRE un.custom_pre_page
!define MUI_PAGE_CUSTOMFUNCTION_LEAVE un.custom_leave_page
!define MUI_WELCOMEPAGE_TITLE_3LINES
!define MUI_WELCOMEPAGE_TITLE 'unwelcome_title'
!define MUI_WELCOMEPAGE_TEXT 'unwelcome_text'
!insertmacro MUI_UNPAGE_WELCOME

so when we have are in a language where the default text is too long we don't have the 3 lines.
and we can't change the text. see the screenshot.

I've test many ways to solve it, manage anothe unpage welcome and it works for the othe page but the default page that you added is never removed.
I tried to manipulate the window item to hide it, try to click on the next button with a command to navigate 
 but it doesn't work. And I verified that handles are correct.
 
 so what it the solution in this context to interact with your unwelcome page or to remove it?
 thanks in advance

image

PR #1204, opened 2017-02-02
AutoUpdater docs Fix

cawa-93:


docs/Auto Update.md
```diff
@@ -71,7 +71,7 @@ Emitted when checking if an update has started.
 
 * `info` [UpdateInfo](#UpdateInfo) for generic and github providers. [VersionInfo](#VersionInfo) for Bintray provider.
 
-Emitted when there is an available update. The update is downloaded automatically if `autoDownload` is not set to `false`.
+Emitted when there is an available update. The update is downloaded automatically if `autoDownload` is `true`.
```
develar: So, it will be better to mention here that default value is `true`.

PR #1209, opened 2017-02-02
[WIP] S3 AWS Publishing

Anmo:
Enable publishing to a S3 AWS bucket.
Remove `isAuthTokenSet` and let each Publisher check the auth tokens.
For `auto-update` just use the `GenericProvider`.

Missing the tests ;)

Any comment is welcome.

### DIFF
--- a/packages/electron-builder/package.json
+++ b/packages/electron-builder/package.json
@@ -66,6 +66,7 @@
     "parse-color": "^1.0.0",
     "plist": "^2.0.1",
     "progress": "^1.1.8",
+    "s3": "^4.4.0",
     "sanitize-filename": "^1.6.1",
     "semver": "^5.3.0",
     "tunnel-agent": "^0.4.3",
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -5,7 +5,7 @@ import { RequestHeaders, executorHolder } from "electron-builder-http"
 import { Provider, UpdateCheckResult, FileInfo, UpdaterSignal } from "./api"
 import { BintrayProvider } from "./BintrayProvider"
 import BluebirdPromise from "bluebird-lst-c"
-import { BintrayOptions, PublishConfiguration, GithubOptions, GenericServerOptions, VersionInfo } from "electron-builder-http/out/publishOptions"
+import { BintrayOptions, PublishConfiguration, GithubOptions, S3Options, GenericServerOptions, VersionInfo } from "electron-builder-http/out/publishOptions"
 import { readFile } from "fs-extra-p"
 import { safeLoad } from "js-yaml"
 import { GenericProvider } from "./GenericProvider"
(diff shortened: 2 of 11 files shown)


packages/electron-updater/src/AppUpdater.ts
```diff
@@ -248,11 +248,19 @@ function createClient(data: string | PublishConfiguration | BintrayOptions | Git
   switch (provider) {
     case "github":
       return new GitHubProvider(<GithubOptions>data)
+    case "s3":
     case "generic":
+      if (provider === "s3") {
+        let s3Data = {
```
Anmo: Is there a way to do this in the previous `case`?

Was giving me a `fall through`, is there a way to say that was deliberate?

develar: I don't see any error — tslint or IDEA are happy.

But in any case consider to do it in a separate branch: `new GenericProvider` is not a duplicated code :)

develar:
LGTM. I will do more detailed review and merge in 2 days. Thanks for implementing long-awaited feature.

packages/electron-builder/package.json
```diff
@@ -66,6 +66,7 @@
     "parse-color": "^1.0.0",
     "plist": "^2.0.1",
     "progress": "^1.1.8",
+    "s3": "^4.4.0",
```
develar: Dependency should be added to the root package.json as well (for tests). But I am going to any case move S3 publisher to a separate package (to reduce dependencies if user doesn't use S3 publisher), so, I will do it.

packages/electron-updater/src/AppUpdater.ts
Anmo: Are you suggest this:
```typescript
    case "s3":
      let s3Data = {
        url: `https://s3.amazonaws.com/${(<S3Options>data).bucket || ""}`,
        channel: (<S3Options>data).channel || ""
      }
      return new GenericProvider(<GenericServerOptions>s3Data)
    case "generic":
      return new GenericProvider(<GenericServerOptions>data)
```

develar: Yes. To make it clear. `s3Data` can be inlined here.

Anmo: Done 👍

packages/electron-builder/package.json
Anmo: Done 👍

develar:
I see that s3 library is outdated (published 2 years ago) https://github.com/andrewrk/node-s3-client/issues/129 Yeach...

develar:
Merged. Will be pushed tomorrow.

develar:
https://github.com/electron-userland/electron-builder/commit/fd4fc0f6af0032d28799357001192997aa59cd44

13.1.0 will be published today or tomorrow.

develar:
aws-sdk used directly as all current s3 npm modules are outdated. And now aws-sdk is good enough to upload (I hope).

develar:
S3 publisher requires  explicit installation of `electron-publisher-s3` module. I will later update docs to reflect it. As separate module because AWS SDK is not a lightweight dependency. And electron-builder must be lightweight.

develar:
Github and Bintray moved to `electron-builder-publisher`. `electron-buidler` explicitly depends on this module.

PR #1220, opened 2017-02-07
Update doc for download-progress event support

heldr:
Just a note why electron-updater `download-progress` event is not supported on Mac yet.


heldr:
Nice! Thanks! 🍵 

When possible, could you please `yarn run update-wiki`?

develar:
Yes, as part of releasing 13.1.2 (s3 publisher should be documented). Thanks for fixing docs.

PR #1240, opened 2017-02-10
feat: log GitHub request paths on error

iffy:
I have no idea if my commit format is right and neither `npm i` nor `npm test` pass.  But this PR expresses some debugging info that I'd *really* like to have.


develar:
Applied, thanks.

PR #1253, opened 2017-02-13
Added nsis-web in windows target options

akashnimare:


develar:
Thanks!

develar:
In the future please do not modify Options.md directly — it is generated from .ts files (in this case — winOptions).

Issue #1298, opened 2017-02-24
Add option to specify what files to uninstall


mastergberry:
* **Version**: 13.5

* **Target**: NSIS

So some of my users apparently think it's a good idea to install their application onto their Desktop or into their Documents folder.....Terrible idea. To be clear instead of installing it into `C:\Users\<username>\Desktop\<Application Name>` they are doing `C:\Users\<username>\Desktop`

https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/templates/nsis/uninstaller.nsh#L42

This line is nuking their other files if they install it into a dumb location.

Is there a way to remove just the files we installed by default and possibly extra files also such as logs or something? I would even just be content with uninstalling the original files we installed and then use the customUnInstall macro to clean up the other ones.

mastergberry:
I submitted a basic PR that will be a smart "temporary" solution assuming the programmer knows what is being bundled in the installation.

In theory all original installed files should be accounted for somewhere in the uninstaller. The current solution gives more power to the programmer to make the decision how they want to uninstall the program.

develar:
It is why we don't allow to choose installation directory by default.

This feature was rejected to be implemented because by default user cannot choose installation directory.

> So some of my users apparently think it's a good idea to install their application onto their Desktop or into their Documents folder.....Terrible idea. 

Proper solution will be to automatically detect such folders and create subdir for app. Could you please file PR?

mastergberry:
Sure I can attempt it. Will try to submit something in a couple of hours or so.

mastergberry:
@develar So I found where to shove the code and the way I am thinking of doing this at the moment would check the `$APP_FILENAME` and see if it still in `$INSTDIR` after the `!insertmacro MUI_PAGE_DIRECTORY`

I have found this function to help with this http://nsis.sourceforge.net/StrContains

Is there some easier way to do it without having to include this extra function into the installer? If not where should this function be put?

develar:
> where should this function be put?

`include` dir in the `nsis`.

mastergberry:
I created a possible fix although I have not yet had time to test it. I can test it later this weekend or I can submit a PR for you to test/merge it in @develar

develar:
@mastergberry Looks good for me (but not tested). You can file a PR.

PR #1300, opened 2017-02-24
feat(nsis): Add option to define custom files to be removed

mastergberry:
Instead of just nuking installation directory there is an option for the programmer to specify which files should be uninstalled when uninstalling the application. Temporary solution for #1298


mastergberry:
Tagging @develar for this instead of recommended reviewer.

PR #1301, opened 2017-02-24
fix(nsis): Enforce a proper application location is made when user ov…

### LINKED ISSUE #1298: Add option to specify what files to uninstall

mastergberry:
…errides default location, fixes #1298

WARNING: Not tested yet.

### DIFF
--- a/packages/electron-builder/templates/nsis/boringInstaller.nsh
+++ b/packages/electron-builder/templates/nsis/boringInstaller.nsh
@@ -1,5 +1,6 @@
 !include UAC.nsh
 !include StdUtils.nsh
+!include StrContains.nsh
 
 !ifndef INSTALL_MODE_PER_ALL_USERS
   !include multiUserUi.nsh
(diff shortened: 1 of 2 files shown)


mastergberry:
Looks like the build failed because the condition didn't get hit and the `StrContains` wasn't used.

packages/electron-builder/templates/nsis/boringInstaller.nsh
```diff
@@ -1,5 +1,6 @@
 !include UAC.nsh
 !include StdUtils.nsh
+!include StrContains.nsh
```
develar: Move this line under `!ifndef BUILD_UNINSTALLER`

Issue #1303, opened 2017-02-25
Publishing always fails on circleci


arekkas:
* **Version**: 13.11.1

* **Target**: OSX

![grafik](https://cloud.githubusercontent.com/assets/3372410/23329906/ab8119e8-fb42-11e6-87ae-6374c407e048.png)

![grafik](https://cloud.githubusercontent.com/assets/3372410/23329909/bbe5a790-fb42-11e6-89ad-a2ef29590e28.png)

PR #1304, opened 2017-02-25
Publishing always fails on circleci #1303

arekkas:
I'm pretty sure that value is either `undefined` or empty (`''`) instead of `null` if the env var is not set.


arekkas:
@develar (no pressure) what's the timeline to the next release?

develar:
@arekkas Thanks for fix. 14.4.1 published.

arekkas:
Thanks, confirmed working

PR #1321, opened 2017-03-02
Update Publishing Artifacts.md (path syntax for S3 publishing)

Sukram21:
Give an example how the syntax of a non-default path should look like (is not completely intuitive as the slash prefix has to be omitted).


develar:
I will fix mentioned bug.

PR #1326, opened 2017-03-02
Grammar correction

jacobq:
Integrate fragment into previous sentence.


develar:
English is not my native lang. Here 2 sentences to emphasize that additional configuration it is bad. It is not correct?

jacobq:
Before the edit it was written as:
> But it requires additional configuration. For each CI (since AppVeyor can build only Windows and Travis only macOS / Linux).
The first sentence would be OK, but the second one is actually a [fragment](https://owl.english.purdue.edu/owl/resource/620/1/) since it is just a [prepositional phrase](https://owl.english.purdue.edu/owl/resource/594/01/). There needs to be a subject and verb, but if you make it part of the previous sentence it works.

develar:
Wiki will be synced on Saturday.

PR #1343, opened 2017-03-08
Create private GitHub provider, resolves #1266

### LINKED ISSUE #1266: electron-updater Support for private Github repository
electron-builder version: **13.7.0**: 

electron-updater version: **1.6.4**

Target **Windows**: 

Electron updater seems unable to locate my private Github repository even when the GH_TOKEN environment variable is set. I've tried setting the env variable through the system settings, programatically by altering `process.env.GH_TOKEN`, and even passing the `token` parameter in app-update.yml (which seems it might be respected after looking into electron-updater's code). I've verified via logging that `process.env.GH_TOKEN` is properly set to my token which has all permission scopes granted.

Is there something I am missing or is there a known issue with private Github repositories? If I make the repo public then updater finds it and updates perfectly, so the integration is good.

AlienHoboken:
Creates a PrivateGitHubProvider for auto updater. Private GitHub Provider is used if the `GH_TOKEN`
environment variable is set. Necessarily uses the GitHub API with the provided `GH_TOKEN`. Rate limiting can become less of a worry via [conditional requests](https://developer.github.com/v3/#conditional-requests) (will update with example code for this).

Also adds the ability to all providers to pass custom headers to the download for the update file.

### DIFF
--- a/docs/Auto Update.md
+++ b/docs/Auto Update.md
@@ -26,6 +26,12 @@ Simplified auto-update is not supported for Squirrel.Windows.
 
 `latest.yml` (or `latest-mac.json` for macOS) will be generated and uploaded for all providers except `bintray` (because not required, `bintray` doesn't use `latest.yml`).
 
+## Private Update Repo
+
+You can use a private repository for updates with electron-updater by setting the `GH_TOKEN` environment variable. If `GH_TOKEN` is set, electron-updater will use the GitHub API for updates allowing private repositories to work.
+
+**Note:** The GitHub API currently has a rate limit of 5000 requests per user per hour. An update check uses up to 3 requests per check. If you are worried about hitting your rate limit, consider using [conditional requests](https://developer.github.com/v3/#conditional-requests) before checking for updates to reduce rate limit usage.
+
 ## Debugging
 
 You don't need to listen all events to understand what's wrong. Just set `logger`.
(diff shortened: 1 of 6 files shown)


AlienHoboken:
Forgot to lint, embarrassing. 😳

Tests aren't running completely on my machine, lacking Wine. But should be good.

develar:
Will be merged soon, thanks!

AlienHoboken:
Awesome! I'll update the documentation on the PR

develar:
Thanks, merged with some small changes.

docs/Auto Update.md
```diff
@@ -26,6 +26,12 @@ Simplified auto-update is not supported for Squirrel.Windows.
 
 `latest.yml` (or `latest-mac.json` for macOS) will be generated and uploaded for all providers except `bintray` (because not required, `bintray` doesn't use `latest.yml`).
 
+## Private Update Repo
+
+You can use a private repository for updates with electron-updater by setting the `GH_TOKEN` environment variable. If `GH_TOKEN` is set, electron-updater will use the GitHub API for updates allowing private repositories to work.
+
+**Note:** The GitHub API currently has a rate limit of 5000 requests per user per hour. An update check uses up to 3 requests per check. If you are worried about hitting your rate limit, consider using [conditional requests](https://developer.github.com/v3/#conditional-requests) before checking for updates to reduce rate limit usage.
```
develar: Not clear for what "conditional requests" means for end user.

AlienHoboken: The link explains the concept of a conditional request. Would it be best to give some explanation here as well?

develar: Not clear — where end users can use conditional requests. Have you checked — maybe latest URL trick works if token specified?

PR #1349, opened 2017-03-10
feat(deployment): expand macros in S3 options

piotrdubiel:
S3 options can now contain macros to allow something like this:
```
"publish": {
    "provider": "s3",
    "bucket": "some-bucket",
    "path": "${os}/${env.SOME_ENV}"
}
```
Now `bucket`, `path` and `channel` can contain macros. Previously these options couldn't be configured with macros.

### DIFF
--- a/packages/electron-builder/src/publish/PublishManager.ts
+++ b/packages/electron-builder/src/publish/PublishManager.ts
@@ -81,11 +81,6 @@ export class PublishManager implements PublishContext {
       }
 
       let publishConfig = publishConfigs[0]
-      if ((<GenericServerOptions>publishConfig).url != null) {
-        publishConfig = Object.assign({}, publishConfig, {
-          url: packager.expandMacro((<GenericServerOptions>publishConfig).url, null)
-        })
-      }
 
       if (packager.platform === Platform.WINDOWS) {
         const publisherName = await (<WinPackager>packager).computedPublisherName.value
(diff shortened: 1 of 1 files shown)


develar:
Resolving in this place is ok, because in any case result is not cached — to be able to specify publish config per target. Anyway, arch is not used currently.

packages/electron-builder/src/publish/PublishManager.ts
```diff
@@ -380,7 +380,33 @@ export async function getPublishConfigs(packager: PlatformPackager<any>, targetS
...
+  return await (<Promise<Array<PublishConfiguration>>>BluebirdPromise.map(asArray(publishers), it => getResolvedPublishConfig(packager.info, typeof it === "string" ? {provider: it} : it)))
+    .then(publishConfigs => expandPublishConfigs(packager, publishConfigs))
+}
+
+function expandPublishConfigs(packager: PlatformPackager<any>, publishConfigs: Array<PublishConfiguration>) {
+  return publishConfigs.map(publishConfig => expandPublishConfig(packager, publishConfig))
+}
+
+function expandPublishConfig(packager: PlatformPackager<any>, publishConfig: PublishConfiguration) {
+  if ((<S3Options>publishConfig).bucket != null) {
```
develar: I think we can just expand macro for all keys in the object. It also will allow us to remove explicit expand for generic URL (see usages in this file). Could you please do it?

piotrdubiel: Done. One question: is it ok not to add arch in `computeDownloadUrl` to expand?

piotrdubiel: Too soon, need to fix.

develar: Only snap test should be failed (unrelated).

piotrdubiel: I think it's fixed.

Issue #1373, opened 2017-03-15
URLs generated by s3Url result in certificate error on request -> breaks S3 autoUpdate


Sukram21:
* **Version**: 15.4.2

* **Target**: mac, win

In https://github.com/electron-userland/electron-builder/commit/7041b5d3e2b17983973d5cea41215b5c371609d7#diff-36ba9eb20bed1b6c45f54cf4faedfcabL81, the `s3Url` method behavior was changed from `https://s3.amazonaws.com/${options.bucket}` to `https://${options.bucket}.s3.amazonaws.com`. 

**Resulting Problem:** If the bucket name contains any dots, requesting new URL will result in a SSL certificate warning (AWS S3 wildcard certificate is `*.s3.amazonaws.com, s3.amazonaws.com` -> doesn't match). This is already mentioned here: https://github.com/electron-userland/electron-builder/commit/7041b5d3e2b17983973d5cea41215b5c371609d7#diff-b8b7b963245d02b44186a38315e09df6L85 

> format `https://s3.amazonaws.com/bucket_name` [must be used](http://stackoverflow.com/a/11203685/1910191)

I guess this just got lost when committing.

**Result:** S3 auto-update checks don't work as they get stuck in a SSL certificate warning.

**Solution:** Revert the `s3Url` method behavior. 

I will create a pull request soon. Maybe I'm missing something, but was there a reason for this change?

develar:
Because

```
develar-imac:~ develar$ curl -v -L https://s3.amazonaws.com/develar/test/latest.yml
*   Trying 54.231.40.2...
* TCP_NODELAY set
* Connected to s3.amazonaws.com (54.231.40.2) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate: s3.amazonaws.com
* Server certificate: DigiCert Baltimore CA-2 G2
* Server certificate: Baltimore CyberTrust Root
> GET /develar/test/latest.yml HTTP/1.1
> Host: s3.amazonaws.com
> User-Agent: curl/7.51.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< x-amz-bucket-region: eu-central-1
< x-amz-request-id: 1D7811AE111228D4
[...]

PR #1374, opened 2017-03-15
[WIP] fix: S3 URLs (#1373)

Sukram21:
Generate correct S3 urls that won't result in SSL certificate warnings anymore. Outlined in #1373.


packages/electron-builder-http/src/publishOptions.ts
```diff
@@ -115,7 +120,8 @@ export interface S3Options extends PublishConfiguration {
 }
 
 export function s3Url(options: S3Options) {
-  let url = `https://${options.bucket}.s3.amazonaws.com`
+  let url = `https://s3.${options.region}.amazonaws.com/${options.bucket}`
+
   if (options.path != null) {
```
develar: As this URL format can leads to issues, let's use it only bucket name contains dots.

packages/electron-builder/src/publish/PublishManager.ts
```diff
@@ -89,6 +90,12 @@ export class PublishManager implements PublishContext {
         }
       }
 
+      if (publishConfig.provider === "s3") {
+        const s3 = new S3({signatureVersion: "v4"})
```
develar: As commented above, we should do it only if bucket name contains dot — avoid net calls.

develar:
@Sukram21 please see other my comments.

I have fixed the issue related to API doc generation, diff should be small and reasonable.

Sukram21:
Pushed a new commit dealing with your comments 🙂

develar:
Will be published tomorrow. I don't want to add aws-sdk dep to main module, but it requires some publisher API changes. It is the reason why s3 publisher is not part of electron-builder module.

develar:
Thanks a lot for PR, it makes s3 publisher more user-friendly.

Sukram21:
You're welcome, I'm glad to contribute to this awesome project :)

Issue #1367, opened 2017-03-14
Installed release app cannot found app-update.yml


changkun:
* **Version**: 15.4.2

* **electron-updater**: 1.9.0

* **Target**: macOS & Windows

Followed the document, here is my updater code:

```js
const electron = require('electron')
const { autoUpdater } = require('electron-updater')
const dialog = electron.dialog

let updater

function setupUpdateEvents () {
  autoUpdater.autoDownload = false
  autoUpdater.on('error', (event, error) => {
    dialog.showErrorBox('error', error)
  })
  autoUpdater.on('checking-for-update', () => {
  })
  autoUpdater.on('update-available', () => {
    dialog.showMessageBox({
      type: 'info',
      title: 'Found Updates',
      message: 'Found Updates, do you want update?',
      buttons: ['Sure', 'No']
    }, (buttonIndex) => {
      if (buttonIndex === 0) {
        autoUpdater.downloadUpdate()
      } else {
        updater.enabled = true
        updater = null
      }
    })
  })
  autoUpdater.on('update-not-available', () => {
    dialog.showMessageBox({title: 'Install Updates', message: 'No Updates Available.'})
    updater.enabled = true
    updater = null
  })
  autoUpdater.on('update-downloaded', () => {
    dialog.showMessageBox({title: 'Install Updates', message: 'Will start update, app will quit...'}, () => {
      autoUpdater.quitAndInstall()
    })
  })
}

function checkForUpdates (menuItem, focusedWindow, event) {
  updater = menuItem
  updater.enabled = false
  setupUpdateEvents()
  autoUpdater.checkForUpdates()
}

module.exports.checkForUpdates = checkForUpdates
```

Basically I was following the example given from the AutoUpdate example. The export function `checkForUpdates` is used for menuItem click callback. I signed my app and configured the pusher in package.json and successfully published to GitHub repo. Then I download the app from Github and installed into Application folder on macOS. However, it still give me an error for check the `app-update.yml`, why? Is there anything did I miss for implementation?

Here is the error:

```
Cannot check for updates: Error: ENOENT: no such file or directory, open '/Applications/MyApp.app/Contents/Resources/app-update.yml'
    at /Applications/MyApp.app/Contents/Resources/app.asar/node_modules/electron-updater/src/AppUpdater.ts:246:27
    at Generator.next (<anonymous>)
From previous event:
[...]

PR #1377, opened 2017-03-15
docs(updater): update Auto Update quick setup guide

changkun:
1. Add zip target attention note for macOS
2. Add `electron-updater` quick setup code snippet example

See: https://github.com/electron-userland/electron-builder/issues/1367

### DIFF
--- a/docs/Auto Update.md
+++ b/docs/Auto Update.md
@@ -32,7 +32,7 @@ A [complete example](https://github.com/iffy/electron-updater-example) showing h
 
 **Manual Update**
 
-The following code snippet gives another example, which illustrate an encapsulated menual update via menu.
+The following code snippet gives another example, which illustrate an encapsulated manual update via menu.
 
 ```js
 // updater.js
--- a/docs/Auto Update.md
+++ b/docs/Auto Update.md
@@ -1,7 +1,5 @@
 See the [Publishing Artifacts](https://github.com/electron-userland/electron-builder/wiki/Publishing-Artifacts) section of the [Wiki](https://github.com/electron-userland/electron-builder/wiki) for more information on how to configure your CI environment for automated deployments.
 
-A [complete example](https://github.com/iffy/electron-updater-example) showing how to use.
-
 Simplified auto-update is not supported for Squirrel.Windows.
 
 ## Quick Setup Guide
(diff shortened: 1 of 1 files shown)


develar:
Maybe it is better to use notifications instead of dialog box (https://github.com/develar/onshape-desktop-shell/blob/master/src/AppUpdater.ts)?

changkun:
@develar It is arguable to use messaging or dialog box.

According to my practice, one of the drawbacks for notification is to set up an ipcRenderer somewhere else if menu creates on ipcMain, which may lead coupling for afterward maintains. I would rarely use it. In this snippet, the complete update function was encapsulated, and the `updater` is not related to anyone else.

Please let me know more benefits for notification usage scenario if I am wrong.

develar:
@changkun I don't like message box because it interrupts user work. Notifications are much more user-friendly. I don't want to promote message box.

Yeach, electron supports notifications only for renderer process and as result, example will be complicated.

Also, `autoUpdater.autoDownload = false` is promoted, but I think that for most users auto download is better.

Maybe we should add note about it like "this example for manual update via menu".

docs/Auto Update.md
```diff
@@ -26,9 +26,13 @@ Simplified auto-update is not supported for Squirrel.Windows.
...
 ### Examples
 
+**Auto Update**
+
 A [complete example](https://github.com/iffy/electron-updater-example) showing how to use.
 
-The following code snippet give another example of how to implement few `electron-updater` events, which gives a optional update function.
+**Manual Update**
+
+The following code snippet gives another example, which illustrate an encapsulated menual update via menu.
```
develar: menual -> `manual`

changkun:
@develar You are definitely right about the interruption. I also hate it. For manual update, user should aware that they are updating software and will be interrupt in the furture, does it reasonable a little?

develar:
> You are definitely right about the interruption. I also hate it. For manual update, user should aware that they are updating software and will be interrupt in the furture, does it reasonable a little?

For manual auto update message box is ok. But in general I don't want to promote such implementation. I am afraid that most users will stupidly and blindly copy example.

develar:
Thanks! I am not happy to promote this example, but for now ok.

changkun:
You're welcome. Perhaps you can remove the example from your subsequent commit :-)

Issue #1370, opened 2017-03-14
Downloading update from private GitHub repository — mac support


derek-duncan:
electron-updater
* **Version**: 1.9.0

* **Target**: dmg

I've been debugging `Update download failed. The server sent an invalid response. Try again later.` after `electron-updater` attempts to download from a private GitHub repo. I wanted to confirm the authentication token wasn't the problem, so I replicated the request in Postman. Here's the `GET` request I'm using: 

```
Url: "https://api.github.com/repos/__OWNER__/__REPO__/releases/assets/__ASSET_ID__"
Headers: {
  Accept: "application/octet-stream"
  Authorization: __TOKEN__
}
```

This request fails, because Github requires query-token authentication over header authentication for octet stream requests. See the error returned:

```
<?xml version="1.0" encoding="UTF-8"?>
<Error>
    <Code>InvalidArgument</Code>
    <Message>Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified</Message>
    <ArgumentName>Authorization</ArgumentName>
    <ArgumentValue>token __TOKEN__</ArgumentValue>
    <RequestId>88A9E737F583B6C9</RequestId>
    <HostId>TqzUjnu1qvKvN8WfyGNt7d0sNXkhF58kH3KsKZu1WCqZALaP4UADqQXl42v6+8NLbkDVueuYLhI=</HostId>
</Error>
```

Anyways, the solution to my postman error was changing the authentication to a query string:

```
Url: "https://api.github.com/repos/__OWNER__/__REPO__/releases/assets/__ASSET_ID__?access_token=__TOKEN__"
Headers: {
  Accept: "application/octet-stream"
}
```

The file from the private GitHub repo is then successfully downloaded.

Since this request in Postman worked, I manually made the change in the npm module to see if it fixed the `Update download failed. The server sent an invalid response. Try again later.` error. However, the same error message is still returned, and the download fails. I'm stuck for now and need some input on how to proceed. 🤔 

Thank you 👍 

(btw: I am using placeholder `__VARS__` throughout my example code. Their actual values do appear in the error messages)

derek-duncan:
I'll also note that trying to use both authentication types will throw the same error. So anytime the header `Authorization token __TOKEN__` is present, the request fails.
```
<?xml version="1.0" encoding="UTF-8"?>
<Error>
    <Code>InvalidArgument</Code>
[...]

PR #1382, opened 2017-03-17
fix(updater): GitHub private repo for windows, partial solution for #1370

AlienHoboken:
This is a partial solution for #1370. This fixes private github provider for auto updates on Windows.

However it is know that Mac still does not work, this is due to how squirrel.mac uses feed URL and cannot have its headers controlled on redirect. A solution for this is being investigated.


develar:
In the future please do `yarn lint` ;)

Issue #1415, opened 2017-03-28
Prebuild binaries not being downloaded


neurosnap:
* **Version**: 16.4.2
* **Target**: mac x64

It doesn't seem like electron-builder is attempting to use prebuild-install at all.

```
DEBUG=electron-builder ./node_modules/.bin/build -m --x64
  electron-builder Two package.json structure is used (dev: /Users/erock/notion/desktop/package.json, app: /Users/erock/notion/desktop/app/package.json) +0ms
  electron-builder Effective config: {
  electron-builder   "npmSkipBuildFromSource": true,
  electron-builder   "appId": "ai.notion.desktop",
  electron-builder   "asarUnpack": [
  electron-builder     "**/*.node"
  electron-builder   ],
  electron-builder   "publish": [
  electron-builder     {
  electron-builder       "provider": "s3",
  electron-builder       "bucket": "notion-desktop"
  electron-builder     },
  electron-builder     {
  electron-builder       "provider": "github"
  electron-builder     }
  electron-builder   ],
  electron-builder   "dmg": {
  electron-builder     "contents": [
  electron-builder       {
  electron-builder         "x": 410,
  electron-builder         "y": 150,
  electron-builder         "type": "link",
  electron-builder         "path": "/Applications"
  electron-builder       },
  electron-builder       {
  electron-builder         "x": 130,
  electron-builder         "y": 150,
  electron-builder         "type": "file"
  electron-builder       }
  electron-builder     ]
  electron-builder   },
  electron-builder   "mac": {
  electron-builder     "category": "public.app-category.productivity"
  electron-builder   },
  electron-builder   "linux": {
  electron-builder     "target": [
  electron-builder       "deb",
  electron-builder       "AppImage"
  electron-builder     ]
  electron-builder   }
  electron-builder } +3ms
Rebuilding native production dependencies for darwin:x64
  electron-builder Spawning npm rebuild notion-spellchecker +368ms

> notion-spellchecker@3.3.2 install /Users/erock/notion/desktop/app/node_modules/notion-spellchecker
> prebuild-install || node-gyp rebuild

  CXX(target) Release/obj.target/hunspell/vendor/hunspell/src/hunspell/affentry.o
  CXX(target) Release/obj.target/hunspell/vendor/hunspell/src/hunspell/affixmgr.o
  CXX(target) Release/obj.target/hunspell/vendor/hunspell/src/hunspell/csutil.o
[...]

PR #1450, opened 2017-04-09
fix(rebuild): Log stdout and stderr

lgeiger:
This logs the output of the rebuild process to help debug build failures like #1415


develar:
Because we do log stderr/stdout but only if error occurred. But process (npm rebuild) exited ok due to bash script `prebuild || node-gyp`.

PR #1453, opened 2017-04-10
Fix appindicator icons with snap build

### LINKED ISSUE #1452: Indicator icons are broken in snaps of Electron applications.
Which version of electron-builder are you using?
  * **Version**: 16.8.2

What target are you building for?
* **Target**: Linux. snap

When building snaps for Linux the Indicator icons (if supported by the applications) have broken image icons. This can be resolved by coercing the `TMPDIR` path for Chromium Framework/Electron to ensure `libappindicator` has readable resources. For example:

```
apps:
  example:
    # Correct the TMPDIR path for Chromium Framework/Electron to
    # ensure libappindicator has readable resources.
    command: env TMPDIR=$XDG_RUNTIME_DIR ${SNAP}/bin/desktop-launch $SNAP/usr/bin/example
```

Coercing TMPDIR is a safe default for all snapped Electron applications.

geekgonecrazy:
Closes #1452

Credit to: @flexiondotorg for the exact fix :smile:

### DIFF
--- a/packages/electron-builder/src/targets/snap.ts
+++ b/packages/electron-builder/src/targets/snap.ts
@@ -61,7 +61,7 @@ export default class SnapTarget extends Target {
 
     snap.apps = {
       [snap.name]: {
-        command: `desktop-launch $SNAP/${packager.executableName}`,
+        command: `env TMPDIR=$XDG_RUNTIME_DIR desktop-launch $SNAP/${packager.executableName}`,
         plugs: replaceDefault(options.plugs, ["home", "x11", "unity7", "browser-support", "network", "gsettings", "pulseaudio", "opengl"])
       }
     }


develar:
Are your sure that it is safe to change `TMPDIR` for the whole app? Why default `TMPDIR` is not readable?

geekgonecrazy:
My understanding from the snap team is that it's safe for any graphical application like electron to always have this.  

I'll see if I can figure out the snapTest command.

develar:
Why TMPDIR is not readable? Could you please explain?

flexiondotorg:
XDG_RUNTIME_DIR is set by `snapd` and points to `/run/user/<UID>/snap.<snapname>/`

  * https://github.com/snapcore/snapd/blob/7d46f0d32c3db67543ccc14c580a643097727944/snap/snapenv/snapenv.go#L94

develar:
So... it is more safe to change `TMPDIR`. Ok. Could you please update snapshot to fix CI? Or I can do it.

geekgonecrazy:
@develar thanks for merging! :+1:

develar:
Release will be tomorrow morning CET.

PR #1461, opened 2017-04-12
chore: remove exec permission from non-exec files

dchest:


develar:
Strange :)

dchest:
Yeah, I sometimes get similar mysterious permission screw ups on a Mac, not sure if it's a text editor's fault or git's.

PR #1467, opened 2017-04-13
Add support for additional certificate file

akorchev:
I had the need to add an additional certificate file via the `/ac|-ac` option (similar to [this SO question](http://stackoverflow.com/questions/16571978/signed-using-pfx-file-but-uac-still-prompts-for-unknown-publisher)). This PR adds a new `additionalCertificateFile` option to `WinBuildOptions`.


develar:
I think env will be better. Or not?

akorchev:
I guess yes however I don't know how to support env as well. Using a file was easier to implement.

akorchev:
Any chance of getting this merged?

develar:
Today I retuned from vacation. I don't want to merge it as is, but I am going to add support for env and then merge. In 2 days.

develar:
On other hand, you can use `--em` and, so, PR is valid in this light.

develar:
Will be releases today evening CET. Thanks!

akorchev:
No, thank you!

PR #1510, opened 2017-04-27
Fix broken link re: bundleVersion

liliakai:
Bad markdown was causing this link to point to #see instead of the relevant github comment.


develar:
Thanks, in the future please modify files in the `*Options.ts` files.

PR #1530, opened 2017-05-05
add proxy authentication support to electron-updater

### LINKED ISSUE #1528: [electron-updater] proxy authentication
* **Version**: latest

Hi,
currently it seems like the `login` events of the net.requests created in [`electronHttpExecutor`](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/electronHttpExecutor.ts) cannot be intercepted by the user which is obviously problematic if the app is used behind a proxy that requires authentication.
Would a PR to resolve this be accepted? If yes, do you have any design guidelines/ preferences?

BurningEnlightenment:
With this approach we simply relay the login event from the electron requests to the `autoUpdater` instance through a callback provided by an (optional) constructor parameter of `ElectronHttpExecutor`.
The test suite runs; but I haven't had time to test the change specifically, but uploaded it in order to allow for early feedback on the general approach.

resolves #1528

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -94,7 +94,10 @@ export abstract class AppUpdater extends EventEmitter {
     }
     else {
       this.app = require("electron").app
-      executorHolder.httpExecutor = new ElectronHttpExecutor()
+      executorHolder.httpExecutor = new ElectronHttpExecutor((authInfo: Electron.LoginAuthInfo,
+        callback: (username: string, password: string) => void) => {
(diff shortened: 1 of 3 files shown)


BurningEnlightenment:
@develar Why does `doApiRequest()` use the default session instead of the electron updater session?

develar:
I think it is bug.

BurningEnlightenment:
> I think it is bug.

Shall I add a commit fixing it to this PR?

develar:
Yes, please.

develar:
Danke!

develar:
electron-updater 1.15.0

PR #1534, opened 2017-05-08
feat: local path to custom electron build (windows support)

### LINKED ISSUE #1342: Local path to custom electron build
* **Version**: 15.1.1

* **Target**: Windows, OSX

re: https://github.com/electron-userland/electron-builder/issues/760

The commit states, "The path to custom Electron build (e.g. `~/electron/out/R`). Only macOS supported, file issue if need for Linux or Windows."

We need this to work on Windows, as it's necessary for us to supply our own, custom build of Electron. We are aware of the env. variable solution for specifying a mirror but a local directory would be a better solution for our automated build system.

MariaDima:
closes #1342


develar:
Thanks :)

develar:
17.5.0 published.

PR #1536, opened 2017-05-08
Fix LoginHandler definition

BurningEnlightenment:
Fixes a small oversight from #1530, I'm sorry for the inconvenience.
I also re-exported the NET_SESSION_NAME symbol from the api header file so it can be used without relying on implementation details (the existence of electronHttpExecutor)

### DIFF
--- a/packages/electron-updater/src/api.ts
+++ b/packages/electron-updater/src/api.ts
@@ -57,7 +57,7 @@ export interface UpdateCheckResult {
 
 export const DOWNLOAD_PROGRESS = "download-progress"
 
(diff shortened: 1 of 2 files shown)


develar:
> I also re-exported the NET_SESSION_NAME symbol from the api header file so it can be used without relying on implementation details (the existence of electronHttpExecutor)

I guess it should be exported in the `main.ts` to be able to use it from `electron-updater` (not `electron-updater/out/api`), right?

BurningEnlightenment:
> I think it should be exported in the main.ts to be able to use it from electron-updater (not electron-updater/out/api), right?

Ì'm fine with either option. Will update.

BurningEnlightenment:
@develar fyi I applied the change you suggested.

PR #1565, opened 2017-05-22
Snap for extra after parts and build packages support

hychen:
For support unity global menubar integration, it needs to specify which [launcher](https://developer.ubuntu.com/en/blog/2016/07/06/announcing-new-snap-desktop-launchers/)  the app use. 

In our case, we need to specify `desktop-gtk2` in `after` in snapcraft.yaml.

This patch extends SnapBuildOption for specify after parts and necessary packages in building step.


develar:
Please fix CI failures. (please run `yarn test` before PR).

In general, 👍 Thanks.

PR #1574, opened 2017-05-23
feat: allow passing through binaries and requirements options

feens:
In order to properly sign helpers, such as keytar (used for keychain access), the specific binary needs to be signed.
electron-osx-sign allows this but there was no way to pass these through, so allow the option(s) in package.json.
Additionally, the mac (not app store) apps need requirements sent to ensure helper acts are recognized properly.

Also: the _mas-dev_ target wasn't being allowed, so I fixed that.

### DIFF
--- a/packages/electron-builder/src/macPackager.ts
+++ b/packages/electron-builder/src/macPackager.ts
@@ -174,7 +174,16 @@ export default class MacPackager extends PlatformPackager<MacOptions> {
         }
       }
     }
-
+    let requirements
+    if (!isMas) {
+      requirements = this.platformSpecificBuildOptions.requirements
+    }
+    let binaries
+    if (isMas && masOptions) {
+      binaries = masOptions.binaries
+    } else if (!isMas && this.platformSpecificBuildOptions) {
+      binaries = this.platformSpecificBuildOptions.binaries
+    }
     const signOptions: any = {
       skipIdentityValidation: true,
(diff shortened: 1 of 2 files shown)


feens:
Back story: we were having issues getting our desktop app to properly sign the for MAS.  Our app uses keytar to handle keychain storage, but it (keytar) needs to be properly signed otherwise it will show a warning when accessing keychain items that were created with the app.

develar:
> Paths of any extra binaries that need to be signed

But electron-osx-sign must automatically detect such binaries. User should not explicitly specify it.

feens:
It doesn't necessarily though.  We found that out the hard way when we were using electron-packager and electron-osx-sign to build the non-MAS version.  It seems that keytar seems to be a bit of a special case and we're still figuring out some issues with it. This at least allows users to pass in the option, otherwise it will detect as it normally would.

develar:
> This at least allows users to pass in the option, otherwise it will detect as it normally would.

👍 In any case I cannot promise that I will investigate this case in a reasonable time. But user should be able to pass it explicitly.

feens:
That's fine.  I think being as it's likely a one-off there's not much sense in you spending a ton of time trying to figure out why.  I think it makes sense to just pass through as many options as possible along with sensible defaults...that way other developers can just override when they need specific behaviour.

PR #1583, opened 2017-05-25
Add note about node_modules inclusion

rmccue:
Was wondering why my production dependencies (`electron-updater` in particular) weren't working. As it turns out, I wasn't including `node_modules` in the build.

This PR includes the docs update for this, but an alternative could be to add `node_modules/**/*` similar to how `package.json` is always included. Happy to close this and open a different PR if you'd prefer that.


develar:
Thanks!

PR #1592, opened 2017-05-28
fix: add missing Exec variable for passing URLs as arguments

kylehuff:
Add Exec variable '%U' to desktop file generation to match AppImage generated desktop file.


develar:
Thanks!

PR #1614, opened 2017-06-04
fix(tar): fix invalid sym-/hardlink targets in archive

mrodem:
The tar --transform expression did not work as intended, leading to broken sym-/hardlinks inside the archive. The intention is to match `^\.`, but the escape character is evaluated in JS before the expression is passed to tar. As a result, the first character of all paths is transformed (`^.`).

The expression still works for regular files, as all paths begin with a `.`. However, tar also transforms sym-/hardlink targets. This e.g. means that a `foo -> bar` symlink is transformed to `foo -> appname-1.0.0ar`.

Adding an extra escape character, so that the correct transform expression is used.


develar:
Thanks!

develar:
Included into the 18.4.0 release.

PR #1621, opened 2017-06-05
fix: Do not pack "elevate.exe" when not needed

### LINKED ISSUE #1620: Do not pack "elevate.exe" when not needed
* **Version**: 18.5.0 (behavior changed in 13.6.0)

* **Target**: Windows

When an electron application is installed in a user-controlled location, then elevation functionality seems to not be needed.

Since there is an option to allow or not elevation (allowElevation), is it possible to use it in order to not pack at all the "elevate" executable?

MariaDima:
Close #1620


develar:
Discussion in the issue. PR cannot be accepted in this form for now.

develar:
I am sorry for misunderstanding — `elevate.exe` is not required for NSIS. It is required only for updater. You can simply add option like `doNotPackElevateHelper`. Your second commit is not required. First (7e9cdb9d9c8c38f919da184142c8bb04e9bf88cf) is enough — you only need to add and respect `doNotPackElevateHelper`.

MariaDima:
Yes! The updater, on error, uses the executable to rerun the installer!

I'll revert the second commit and add the new "doNotPackElevateHelper" option.

One question:
    Can we do the same for the signtool?
   On my second commit I generalized it the removed all files that do not relate to the application and are not needed after installation.
   If this is ok, maybe I can call the new option "doNotPackHelperFiles"

develar:
> Can we do the same for the signtool?

For some reasons I decided that powershell is not suitable to verify code signature. But after your PR it seems it is not required anymore. We can simply do not pack and remove this line.

I prefer separate options.

MariaDima:
Ok, thanks!
I'll updated as following:

* Revert to the first commit
* Define "doNotPackElevateHelper" under Nsis options
* Use the new option instead of the existing "allowElevation"
* Do not pack the signTool.exe at all

Please correct me, if I misunderstood.

develar:
Yes. 👍

MariaDima:
@develar are you aware of the test failures?
I see it's a general issue, not related to the changes of this patchset.
Are you working on this? Can I help here?

develar:
@MariaDima Do not worry, all is ok and PR will be merged in minutes.

develar:
Thanks for PR!

MariaDima:
Thanks!

MariaDima:
@develar we forgot to mention "packElevateHelper" in the wiki. Should I add it or you prefer to not expose this option?

develar:
@MariaDima Wiki will be synced a little bit later, because 18.7.0 not yet latest release, but `pre-release`.

MariaDima:
Perfect, thanks!

Issue #987, opened 2016-12-07
appx - wrong icon / how to set assets


black-snow:
* **Version**: 10.6.1

* **Target**: appx (32/64)

When building for windows store the `icon.ico` isn't used in a couple places. E. g. the installed software list shows a default electron icon. Start menu as well.
I found those default icons in `assets`. I haven't yet figured out how to use my own icon for those assets.

develar:
If you can and want — please file PR.

black-snow:
Will try, guess I'll have to start here https://github.com/electron-userland/electron-builder/blob/master/src/targets/appx.ts#L49

develar:
@black-snow Ouch, sorry, it is already implemented :) Just put `44x44.png` and other such files into your build resources directory.

black-snow:
@develar You sure? I put `myappname.44x44.png` and the others (50x50, ...) into my `build` folder but it's still using the default ones.

develar:
> ah wait, I must leave out the app name, right?

Yes. Only size.

black-snow:
Alright, that works. Some quirks, though.

* Install Screen (dbl click appx) - image is blurry (upscaled)
* Start Menu & Task Bar - icon just looks weird ![Image of Taskbar](http://www.black-snow.net/tmp/weird_taskbar.png)

Sources look okay, though. The images in `pre-appx--ia32` are just fine and the manifest seems okay ...

develar:
> Fo AppX sha256 is always used.

Thanks, fixed.

> Where does the Square150x150Logo etc. in the manifest come from? I

I don't know. To investigate.

@black-snow  Would you like to help and fix this issue?

black-snow:
@develar of course. At first I'll unpack the appx and stuff all the assets in there that a VS plugin generated for me. Will report back.

black-snow:
Also there's some manual steps that have to be done in order to publish the appx in the Windows Store. Maybe we can automate these, too. There's four fields in the manifest that need to be set:

* Package.Identity.Name
* Package.Identitiy.Publisher
* Package.Properties.DisplayName
* Package.Properties.PublisherDisplayName

[...]

PR #1657, opened 2017-06-12
Add support for custom AppX assets and asset resolutions.

surajreddy:
This PR includes custom the ability to add a custom `assets` folder for `AppX` builds, along with support to specify the names for the `AppX` assets. This is based on https://github.com/electron-userland/electron-builder/issues/987#issuecomment-267406485

It leaves the default behavior untouched.

**Should I add option notes for the builds in this issue, or should I be updating the wiki directly?**

**Note:** this doesn't add support for the `BadgeLogo` visual asset option.

**Related:** https://github.com/electron-userland/electron-builder-binaries/pull/3

**Some Reference Docs:**
https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/how-to-create-a-basic-package-manifest
https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/what-s-changed-in-windows-10

cc @feens

### DIFF
--- a/packages/electron-builder/src/targets/appx.ts
+++ b/packages/electron-builder/src/targets/appx.ts
@@ -2,9 +2,8 @@ import BluebirdPromise from "bluebird-lst"
 import { Arch, getArchSuffix, Target } from "electron-builder-core"
 import { exec, spawn, use } from "electron-builder-util"
 import { copyDir, copyFile } from "electron-builder-util/out/fs"
-import { emptyDir, readFile, writeFile } from "fs-extra-p"
+import { emptyDir, readdir, readFile, writeFile } from "fs-extra-p"
 import * as path from "path"
-import sanitizeFileName from "sanitize-filename"
 import { AppXOptions } from "../options/winOptions"
 import { getSignVendorPath, isOldWin6 } from "../windowsCodeSign"
 import { WinPackager } from "../winPackager"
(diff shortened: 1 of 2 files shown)


develar:
> Should I add option notes for the builds in this issue, or should I be updating the wiki directly?

Just run `yarn api`.

packages/electron-builder/src/targets/appx.ts
```diff
@@ -44,38 +44,60 @@ export default class AppXTarget extends Target {
...
-        }
-        return copyFile(path.join(vendorPath, "appxAssets", `SampleAppx.${size}.png`), target)
-      }),
-      copyDir(appOutDir, path.join(preAppx, "app")),
-      this.writeManifest(templatePath, preAppx, safeName, arch, publisher)
-    ])
+    
+    const customAssetsFolder = this.options.assetsFolder
+    if (customAssetsFolder) {
+      const customAssetsPath = path.resolve(this.packager.projectDir, customAssetsFolder)
```
develar: Custom dir must be relative to build resources directory by default. If doesn't exist, relative to project dir. Hint — use method statOrNull.

develar:
yarn lint ;)

packages/electron-builder/src/targets/appx.ts
```diff
@@ -44,38 +44,60 @@ export default class AppXTarget extends Target {
...
+          return copyFile(path.join(vendorPath, "appxAssets", `SampleAppx.${size}.png`), target)
+        }),
+        copyDir(appOutDir, path.join(preAppx, "app")),
+      ])
+    }
 
-    const destination = path.join(this.outDir, packager.expandArtifactNamePattern(this.options, "appx", arch))
-    const args = ["pack", "/o", "/d", preAppx, "/p", destination]
-    use(this.options.makeappxArgs, (it: Array<string>) => args.push(...it))
+    this.writeManifest(templatePath, preAppx, safeName, arch, publisher, this.options.assetNames)
```
develar: If custom dir specified, we should populate missed values in the asset names. Right? Otherwise strange — why we require explicit configuration of names.

packages/electron-builder/src/options/winOptions.ts
```diff
@@ -347,4 +357,19 @@ export interface AppXOptions {
...
+
+  /**
+   * Runs makepri on the folder before running makeappx.
+   */
+  readonly makePri?: boolean | null
+
+  /**
+   * Specify a custom assets folder to use for the assets.
+   */
+  readonly assetsFolder?: string | null
```
develar: This dir should be equals by default to some convenient location, e.g. appxAssets. So, user can simply put files without explicit conf.

surajreddy:
@develar I've updated the PR based on our conversations and linted the changes. Let me know if there's any other improvements you would like.

# README Changes
## AppX Assets Support
AppX assets need to be placed in the `appxAssets` folder in the buildResources folder (by default this is `build`).

The assets should follow these naming conventions:

- Logo: `StoreLogo.png`
- Square150x150Logo: `Square150x150Logo.png`
- Square44x44Logo: `Square44x44Logo.png`
- [Optional] BadgeLogo: `BadgeLogo.png`
- Wide310x150Logo: `Wide310x150Logo.png`
- [Optional] Square310x310Logo: `LargeTile.png`
- [Optional] Square71x71Logo: `SmallTile.png`
- [Optional] SplashScreen: `SplashScreen.png`

All official AppX asset types are supported by the build process. These assets can include scaled assets by using `target size` and `scale` in the name. See https://docs.microsoft.com/en-us/windows/uwp/controls-and-patterns/tiles-and-notifications-app-assets for more info

Default assets will be used for `Logo`, `Square150x150Logo`, `Square44x44Logo` and `Wide310x150Logo` if not provided. For assets marked `[Optional]`, these assets will not be listed in the manifest file if not provided.

> **Note:** Support for assets being provided in the default `build` directory with the names `44x44.png`, `50x50.png`, `150x150.png`, `310x150.png` is dropped. Please move your assets to the `build/appxAssets` folder and name them according to the convention above.

black-snow:
+500 interwebs for you, looking forward to having this in `electron-builder`

develar:
Merge in process. I will change "appx_assets" to "appx" since this folder in build resources and it is clear that it is appx assets.

>  is dropped

In general it violates our rules — should be at least one major release before final drop, but
* it is non-default windows target.
* your changes are awesome and wanted.
* usages if Appx target is not very popular.

develar:
@black-snow Hi, I need your help. To avoid intermediate file copying, now [file mappings](https://docs.microsoft.com/en-us/windows/uwp/packaging/create-app-package-with-makeappx-tool#mapping-files) is used.

Problem is that not clear for me what should be used as `/ProjectRoot` arg to `makepri new`. And no test for it. For now I changed it to `userAssetDir`, but could you please check that my changes doesn't break this functionality? And add missed test? I pushed changes to master.

PR #1658, opened 2017-06-12
fix: macOS failing when there is no old icon

djpereira:
This avoids throwing an ENOENT exception, if for any reason the old icon is not found to be deleted during the mac installer creation process.
```
Error: ENOENT: no such file or directory, unlink '/mypath/dist/canary/mac/Electron.app/Contents/Resources/myapp.icns'
    at Object.<anonymous> (/mypath/node_modules/electron-builder/src/packager/mac.ts:154:19)
From previous event:
    at Object.createApp (/mypath/node_modules/electron-builder/out/packager/mac.js:138:21)
```

### DIFF
--- a/packages/electron-builder/src/packager/mac.ts
+++ b/packages/electron-builder/src/packager/mac.ts
@@ -3,7 +3,7 @@ import BluebirdPromise from "bluebird-lst"
 import { asArray, getPlatformIconFileName, use } from "electron-builder-util"
 import { copyFile, copyOrLinkFile, unlinkIfExists } from "electron-builder-util/out/fs"
 import { warn } from "electron-builder-util/out/log"
-import { readFile, rename, unlink, utimes, writeFile } from "fs-extra-p"
+import { readFile, rename, deleteFile, utimes, writeFile } from "fs-extra-p"
 import * as path from "path"
 import { build as buildPlist, parse as parsePlist } from "plist"
 import { normalizeExt, PlatformPackager } from "../platformPackager"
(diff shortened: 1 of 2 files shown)


develar:
Do you use extendInfo?

djpereira:
@develar, I am not using extendInfo.

develar:
So, how error is possible if electron always has default icon?

djpereira:
@develar, I am not sure, but I am running `build` more than once in different output directories and it fails the second time:
```
build --em.build.directories.output=dist/release
build --em.build.directories.output=dist/canary
build --em.build.directories.output=dist/dev
```

develar:
Merged, but I will take a look why it is happened.

djpereira:
@develar, thanks for doing that.
Do you mind publishing 19.2.2 to npm even as pre-release?

djpereira:
@develar, I have another finding about this issue.
Since I am running multiple builds consecutively, the second build is trying to "unlink" the an icon with the name of the previous build. Obviously, that one is not there, instead electron.icns is there.

I am setting a different icon for each build:
```
build --em.build.directories.output=dist/release --em.build.mac.icon=build/icon1.icns
build --em.build.directories.output=dist/canary --em.build.mac.icon=build/icon2.icns
build --em.build.directories.output=dist/dev --em.build.mac.icon=build/icon3.icns
```

However as I said, it seems the script is trying to remove the wrong "old icon"...

Error: ENOENT: no such file or directory, unlink '/mydirectory/dist/release/mac/Electron.app/Contents/Resources/release.icns'

djpereira:
@develar, I think the problem here is that when we use electron-builder with the "electronDist" option, the custom dist gets copied over using "hard-links" instead of actually copying the content.

The problem with this, is that as the content of the electron dist "copy" gets modified (icons, name, etc.), the original electronDist source folder "mirrors" the changes.

This is not an issue for each use case, but you definitely get into trouble very often.
I think that ideally we want to make an actual copy of the files each time, not a hard link.

Issue #1545, opened 2017-05-15
Allow to start app after silent install


maczikasz:
* **Version**:  latest

 latest

* **Target**:  nsis

Hi, we would like to do silent updates on our app, but using our one click installer that's at the moment not possible. We are calling quitAndInstall on the update-downloaded trigger, which does not start in silent mode.

The other thing is that we would like to do it seemlessly so the app should start after a silent install.

There are two things there then, to allow to specify on quit and install to do a silent one , or an other method, and the other thing is to be able to specify for the installer a flag at runtime or compile time that overrides the if in : https://github.com/electron-userland/electron-builder/blob/002a7149ca71d931e061b42f5520458ac60262f0/packages/electron-builder/templates/nsis/installSection.nsh#L192

develar:
If you call `quitAndInstall` explicitly (not auto install after user update), why installation should be silent?

maczikasz:
what we want to achieve is to install as soon as the update is downloaded then start the app again, and this to be completely silent

Our app normally runs in the background (on the tray, this way we can achieve silent updates without the user being notified) if there is a better way to achieve this then pls tell me

maczikasz:
My understanding how this works at the moment:

Install on app quit -> silent mode -> app wont start automatically
Install by shutting down the app -> non silent (now silent possible) -> app starts automatical

develar:
App will be not started automatically on silent install. As you pointed (https://github.com/electron-userland/electron-builder/blob/002a7149ca71d931e061b42f5520458ac60262f0/packages/electron-builder/templates/nsis/installSection.nsh#L192). So, I think, this issue should be opened until it is not fixed.

maczikasz:
I closed by accident, probably clicked on the wrong button, sry

tiangolo:
I just had an issue using TypeScript, it seems like I have to call `autoUpdater.quitAndInstall(isSilent: boolean)` with an explicit `false` for the `isSilent` parameter, as in `autoUpdater.quitAndInstall(false)`. 

Otherwise I see a `Supplied parameters do not match any signature of call target.`.

---

But as I understand, that flag should have a default to `false` and be needed only for Windows / NSIS, right? That's what I infer from this comment: https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/AppUpdater.ts#L265

[...]

PR #1712, opened 2017-06-20
feat(nsis): Add flag to force start on silent install

maczikasz:
If --force-run flag is specified on installer then will start the app on silent install
Exposed forceRun flag on NSIS updater quit and install, default value is false

https://github.com/electron-userland/electron-builder/issues/1545


develar:
Thanks!

PR #1734, opened 2017-06-22
Auto update does not work on machines with Powershell version < 3

### LINKED ISSUE #1732: Auto update does not work on machines with Powershell version < 3
* **Version**:  LATEST 

* **Target**:  NSIS

The auto updater was changed to use "Convert-from-json" function which was introduced in powershell 3, so on windows 7 machines without powershell 3 it will not work. If that's intented we will need to adjust our system requirements, please tell me if there is a way to make it backward compatible.

Thanks

MariaDima:
Close #1732

### DIFF
--- a/packages/electron-updater/src/NsisUpdater.ts
+++ b/packages/electron-updater/src/NsisUpdater.ts
@@ -1,5 +1,5 @@
 import BluebirdPromise from "bluebird-lst"
-import { execFile, spawn } from "child_process"
+import { execFile, execFileSync, spawn } from "child_process"
 import { CancellationError, CancellationToken, DownloadOptions } from "electron-builder-http"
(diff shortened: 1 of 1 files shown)


MariaDima:
I was experimenting with ConvertTo-CSV command that is supported by powershell < 3.

It could be used to support Windows 7 users.
However, the output is not good at all: It's more like a string, from which we need to extract data using either a regular expression or a substring logic (easier).

What's your thoughts on this?
The output could depend on the way the certificate was created (which options were provided during creation), but it could be an attempt to support more users before bypassing verification.

develar:
Why not just check windows version?

develar:
Thanks!

develar:
Until https://github.com/electron/electron/pull/9648 is not accepted, code signing doesn't matter at all and users in any case not protected.

PR #1771, opened 2017-06-30
feat: throw error when token contains invalid characters

meowtec:
I was building my app on Windows, but it failed with `The header content contains invalid characters`. 
After debugging, I found that the env `GH_TOKEN` on my WIndows was wrong, it ends with `\n` (`xxxxxxxx\n`), not a valid HTTP header value.

So I am adding characters checking before request, making the case clear.

### DIFF
--- a/packages/electron-builder-util/src/util.ts
+++ b/packages/electron-builder-util/src/util.ts
@@ -163,8 +163,12 @@ export function getTempName(prefix?: string | null | undefined): string {
   return `${prefix == null ? "" : `${prefix}-`}${tempDirPrefix}-${(tmpDirCounter++).toString(16)}`
 }
 
-export function isEmptyOrSpaces(s: string | null | undefined) {
+export function isEmptyOrSpaces(s: string | null | undefined): s is "" | null | undefined {
   return s == null || s.trim().length === 0
+}
+
+export function isTokenCharValid(token: string) {
+  return /^[\w\/=+-]+$/.test(token)
 }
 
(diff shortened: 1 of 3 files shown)


packages/electron-builder-util/src/util.ts
```diff
@@ -163,10 +163,14 @@ export function getTempName(prefix?: string | null | undefined): string {
   return `${prefix == null ? "" : `${prefix}-`}${tempDirPrefix}-${(tmpDirCounter++).toString(16)}`
 }
 
-export function isEmptyOrSpaces(s: string | null | undefined) {
+export function isEmptyOrSpaces(s: string | null | undefined): s is "" | null | undefined {
   return s == null || s.trim().length === 0
```
develar: return type here is boolean, isn't?

develar:
I think we should trim token.

packages/electron-builder-util/src/util.ts
meowtec: `:arg is type` does not mean the return type. It is type guards (https://www.typescriptlang.org/docs/handbook/advanced-types.html)

It mean that if isEmptyOrSpaces return true, then `s` should be `''` or `null` or `undefined`.
Otherwise, `s` should be: 

`string | null | undefined` - `'' | null | undefined` = `string`

```
let str: string | null
if (!isEmptyOrSpaces(str)) { 
  str.slice // str is string only 
}
```

develar:
Thanks a lot!

Issue #1772, opened 2017-06-30
BadDigest: The Content-MD5 you specified did not match what we received.


johnf:
* **Version**: 19.11.0

* **Target**:  Windows, Mac and Linux

Since upgrading to 19.11.0 I'm getting BadDigest issues when uploading to S3. Rolling back to 19.8.0 and the problem goes away. I'm assuming it's related to b57dc8a9e47a79eaaefb083cf176e4ff365837a0

Please let me know if there is anything I can do to help debug.

```
hubber-setup-0.0.5.exe to GitHubBadDigest: The Content-MD5 you specified did not match what we received.
    at Request.extractError (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/services/s3.js:577:35)
    at Request.callListeners (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
    at Request.emit (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/request.js:682:14)
    at Request.transition (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/request.js:684:12)
    at Request.callListeners (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
    at Request.emit (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/request.js:682:14)
    at Request.transition (/Users/travis/build/echohubio/hubber/node_modules/aws-sdk/lib/request.js:22:10)
[...]

PR #1775, opened 2017-07-01
fix: prevent latest-${os}.{yml,json} files from being rewritten

johnf:
Prevent multiple updates of latest-${os}.{yml,json}, when multiple
publishers are specified. This was causing BadDigest MD5 errors when
uploading to S3.

Closes: #1772


develar:
Thanks!

PR #1782, opened 2017-07-03
Update installer.nsi

sante85:
added preInstall macro

### DIFF
--- a/packages/electron-builder/templates/nsis/installer.nsi
+++ b/packages/electron-builder/templates/nsis/installer.nsi
@@ -6,6 +6,10 @@ Var desktopLink
 !include "multiUser.nsh"
 !include "allowOnlyOneInstallerInstance.nsh"
 
+!ifmacrodef preInstall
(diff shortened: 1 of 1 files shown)


develar:
Could you please explain for what? Maybe `customHeader` is enough?

develar:
Or `preInit`?

sante85:
i want prevent installation, if a condition is satisfied, and do custom action.

i have tested **preInit**, that is called at build installer, and **customInstall** is called after setup installer.
**customHeader** is after executing setup installer, right?!

develar:
As you can see, we call `check64BitAndSetRegView` and `ALLOW_ONLY_ONE_INSTALLER_INSTANCE ` in the `.onInit`. And it is the right and correct place to add some custom check before install. So, I don't understand why `preInit` macro is not suitable for you to add custom check.

sante85:
Yes, you have right mode, i set Preinit again and do correctely.

Best regards

Issue #1816, opened 2017-07-11
19.15.1 reports wrong version during build


HitomiTenshi:
* **Version**: 19.15.1

* **Target**: Windows ia32 NSIS

Output:

```
> build --ia32 --dir

electron-builder 0.0.0-semantic-release
Using electron-builder.yml configuration file
Skip app dependencies rebuild because npmRebuild is set to false
Packaging for win32 ia32 using electron 1.7.4 to dist\win-ia32-unpacked
```

It should say electron-builder `19.15.1`.

black-snow:
This is the spot: https://github.com/electron-userland/electron-builder/blob/452085b7ba9034e0a7a9f6664de1b749ea172bc6/packages/electron-builder/package.json#L4

black-snow:
develar has addressed this - guess it will be fixed in the next release

PR #1818, opened 2017-07-11
Fixed wrong version (0.0.0 to 19.5.1)

black-snow:


black-snow:
Fixes https://github.com/electron-userland/electron-builder/issues/1816

black-snow:
Missing dep `electron-webpack` --- also seing this in latest `master` when `yarn test`.

develar:
Thanks, but problem is different — this version was released not correctly — `set-versions` was not called before `compile`.

black-snow:
Alright. I already thought that you don't manually set them each time ;)

PR #1848, opened 2017-07-21
Correct link to electron-builder-http doc

buffcode:
Document was moved


develar:
Hi, thanks, but please use wiki (https://github.com/electron-userland/electron-builder/wiki) instead of reading files in the `docs` directly ;)

PR #1860, opened 2017-07-25
feat(squirrel.windows): set version-string, product-version and icon to Update.exe

yuya-oc:
Show application's information instead of GitHub's one in the task manager.

For example, that's meaningful when Update.exe is registered to startup via `node-auto-launch`.
(I used `mattermost/desktop` as an example)

v19.16.3:
before

This PR:
after


develar:
Thanks. Fix is one line — change `sign` to `signAndEditResources. But after that app will be not started. It is Squirrel.Windows known issue. We are not going to investigate and fix it, because — use reliable and simple NSIS instead.

PR #1864, opened 2017-07-26
Added path normalization for file filter

Metaur:
The problem occurs on Windows, when the project files are extracted in local drive root. 
That way, the _src_ argument may become 'Z:\' (drive path), which will lead to filtering out all node modules, because of taking incorrect start index in substring (_it.substring(src.length + 1)_).


develar:
Thanks!

develar:
19.17.0

PR #1872, opened 2017-07-28
Ensure folder paths to have trailing path separator

Metaur:
After my recent pull request, have noticed, that windows file system is doing some tricky magic with double backslashes (path separator), and sometimes does not. so here is the fix for filter.js, and for file copier as well.


develar:
I have fixed tests, will be merged soon.

Issue #1702, opened 2017-06-18
Unable to upload files: Debian distribution, component and architecture were not found on repo and were not specified when deploying to Bintray


jviotti:
* **Version**: 19.5.0

* **Target**: debian

I'm testing automatic publishing to Bintray with a dummy application called "digital-clock". This is how my `package.json` file looks like:

```json
{
  "name": "digital-clock",
  "version": "1.0.0",
  "description": "Digital Clock is a dummy Electron application that displays the current time",
  "author": "John Doe <johndoe@example.com>",
  "license": "BSD",
  "main": "main.js",
  "homepage": "digitalclock.com",
  "build": {
    "appId": "com.example.digitalclock",
    "productName": "Digital Clock",
    "publish": [
      {
        "provider": "github",
        "repo": "digital-clock",
        "owner": "jviotti"
      }
    ],
    "linux": {
      "category": "Utils",
      "icon": "assets/icons",
      "packageCategory": "utils",
      "maintainer": "Jane Doe <janedoe@example.com>",
      "target": [ "deb", "rpm", "AppImage", "snap" ],
      "synopsis": "A minimalistic digital clock rendered in monospace font"
    },
    "deb": {
      "publish": [
        {
          "package": "digital-clock",
          "repo": "digital-clock-debian",
          "user": "jviotti",
          "provider": "bintray"
        }
      ],
      "depends": [
        "gconf2",
        "gconf-service",
        "libnotify4",
        "libappindicator1",
        "libxtst6",
        "libnss3"
      ]
    },
    "rpm": {
      "publish": [
        {
          "package": "digital-clock",
          "repo": "digital-clock-rpm",
          "user": "jviotti",
          "provider": "bintray"
        }
      ],
      "depends": [
        "GConf2",
        "libnotify",
        "nss",
        "libappindicator",
        "libXtst"
      ]
    },
    "snap": {
      "confinement": "strict",
      "summary": "A minimalistic digital clock rendered in monospace font",
      "grade": "stable",
      "buildPackages": [],
      "stagePackages": [
        "libnotify4",
        "libappindicator1",
        "libxtst6",
        "libnss3",
        "libxss1",
        "fontconfig-config",
        "gconf2",
        "libasound2",
        "pulseaudio"
      ],
      "plugs": [
        "default",
        "unity8-calendar"
      ]
    }
  },
  "scripts": {
    "pack": "build --dir",
    "dist": "build",
    "release": "build",
[...]

PR #1884, opened 2017-08-01
feat: propagate arch to publishers upload functions

jviotti:
This is the first towards adding proper Bintray Debian support.

See: https://github.com/electron-userland/electron-builder/issues/1702

### DIFF
--- a/packages/electron-publish/src/BintrayPublisher.ts
+++ b/packages/electron-publish/src/BintrayPublisher.ts
@@ -54,7 +54,7 @@ export class BintrayPublisher extends HttpPublisher {
     }
   }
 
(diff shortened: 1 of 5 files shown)


packages/electron-publish/src/BintrayPublisher.ts
```diff
@@ -72,6 +72,7 @@ export class BintrayPublisher extends HttpPublisher {
             "Content-Length": dataLength,
             "X-Bintray-Override": "1",
             "X-Bintray-Publish": "1",
+            "X-Bintray-Debian-Architecture": arch
```
jviotti: I'll add the other Bintray headers later. I'm passing this one here just to make use of the `arch` variable we're propagating.

jviotti: See https://github.com/electron-userland/electron-builder/pull/1885 for a WIP of the other headers.

PR #1916, opened 2017-08-09
call rebuild with version

akaJes:
the solution for this https://github.com/EmergingTechnologyAdvisors/node-serialport/issues/1180#issuecomment-320975958 issue


develar:
Спасибо.

develar:
Fixed in 19.20.1.

Issue #1917, opened 2017-08-09
NSIS: Remove the runAfterFinish dependency on Start Menu Shortcut


MariaDima:
* **Electron-Builder Version**: 19.19.1

* **Electron-Updater Version**: 2.4.4

* **Target**: Windows

If the NSIS option "runAfterFinish" is set to true (default behavior), then the electron application is being launched after an update.
This auto-launch mechanism is based on the existence of a start menu shortcut.
Some users don't want to have menu shortcuts, or select to move/rename them.
This action causes the auto-launch to fail.

It will be beneficial for more users if we could depend on the actual executable's location for starting the application after an update.

Issue #1767, opened 2017-06-30
Application with createDesktopShortcut=false doesn't launch automatically


clusterberries:
* **Version**:  19.10.0 or 19.11.1

* **Target**: nsis

According to the [documentation ](https://github.com/develar/electron-builder/blob/350e24106aba4882bb63e184ce9a964092728b79/docs/Options.md) if new option **createDesktopShortcut** is set to false, no desktop shortcut is created after installation. But in addition the application doesn't launch automatically and there is no application in the start menu. In this case users don't have ability to start the application after it is installed.
Is it expected behavior?

develar:
No :)

clusterberries:
Ok, thanks! Waiting for the fix then

develar:
@clusterberries Simply because feature was shipped to hear feedback and now we have one :)

PR #1918, opened 2017-08-09
feat(nsis): Define custom shortcuts name, Remove runAfterFinish dependency on shortcuts

### LINKED ISSUE #1827: how to set desktop shortcut name by windows?

MariaDima:
Close #1827, #1917, #1767

### DIFF
--- a/packages/electron-builder/templates/nsis/boringInstaller.nsh
+++ b/packages/electron-builder/templates/nsis/boringInstaller.nsh
@@ -8,9 +8,9 @@
 !ifndef BUILD_UNINSTALLER
   Function StartApp
     ${if} ${Updated}
-      ${StdUtils.ExecShellAsUser} $0 "$INSTDIR\${APP_EXECUTABLE_FILENAME}" "open" "--updated"
+      ${StdUtils.ExecShellAsUser} $0 "$launchLink" "open" "--updated"
     ${else}
-      ${StdUtils.ExecShellAsUser} $0 "$INSTDIR\${APP_EXECUTABLE_FILENAME}" "open" ""
+      ${StdUtils.ExecShellAsUser} $0 "$launchLink" "open" ""
     ${endif}
   FunctionEnd
 
--- a/packages/electron-builder/templates/nsis/boringInstaller.nsh
+++ b/packages/electron-builder/templates/nsis/boringInstaller.nsh
@@ -8,9 +8,9 @@
 !ifndef BUILD_UNINSTALLER
   Function StartApp
     ${if} ${Updated}
-      ${StdUtils.ExecShellAsUser} $0 "$startMenuLink" "open" "--updated"
+      ${StdUtils.ExecShellAsUser} $0 "$INSTDIR\${APP_EXECUTABLE_FILENAME}" "open" "--updated"
     ${else}
-      ${StdUtils.ExecShellAsUser} $0 "$startMenuLink" "open" ""
+      ${StdUtils.ExecShellAsUser} $0 "$INSTDIR\${APP_EXECUTABLE_FILENAME}" "open" ""
     ${endif}
   FunctionEnd
 
@@ -28,7 +28,7 @@
 
   !ifdef allowToChangeInstallationDirectory
     !include StrContains.nsh
-          
+
     !insertmacro MUI_PAGE_DIRECTORY
 
     # pageDirectory leave doesn't work (it seems because $INSTDIR is set after custom leave function)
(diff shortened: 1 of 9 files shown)


MariaDima:
I'm sorry for mixing issues.
I started working on providing a custom shortcuts name, but I kept failing on verifying various scenarios (related to the rest of the nsis available options).
I ended up modifying more parts of the existing business logic, in order to make sure that different update scenarios work.

If that's a problem, I can re-work the proposal based on your feedback.

packages/electron-builder/templates/nsis/boringInstaller.nsh
```diff
@@ -8,9 +8,9 @@
 !ifndef BUILD_UNINSTALLER
   Function StartApp
     ${if} ${Updated}
-      ${StdUtils.ExecShellAsUser} $0 "$startMenuLink" "open" "--updated"
+      ${StdUtils.ExecShellAsUser} $0 "$INSTDIR\${APP_EXECUTABLE_FILENAME}" "open" "--updated"
     ${else}
-      ${StdUtils.ExecShellAsUser} $0 "$startMenuLink" "open" ""
+      ${StdUtils.ExecShellAsUser} $0 "$INSTDIR\${APP_EXECUTABLE_FILENAME}" "open" ""
```
develar: @MariaDima We cannot do it. We must open app usng start menu link because only in this case Application User Model ID (https://msdn.microsoft.com/en-us/library/windows/desktop/dd378459(v=vs.85).aspx) will be applied correctly. We set this *** MS id only for link.

Not clear for me — can we set it for executable or not. To investigate.

MariaDima: I understand the concern, but I'm not sure whether we will face issues.
Normally, the electron application developers use app.setAppUserModelId(id) to set the AUMID.
Maybe I'm overconfident since I tested the patchset with applications that take care of setting the AUMID, but I think that this related to the owner's responsibilities.

I couldn't find a way to set the AUMID using the installer.
Unless I'm mistaken, this call System::Call 'Shell32::SetCurrentProcessExplicitAppUserModelID(w "${APP_ID}")' is meant to be used by the running process (similarly to the electron API to set the AUMID).

What do you think of the following:
 * Check whether the menu link exists
 * If yes, launch the app as it is already being done (using the shortcut).
 * If not, use the executable itself (instead of doing nothing).
Or we could add an option to let the application owner control how to launch its application (by default, start menu would be used).

It would be nice if we could also support somehow the people that make sure AUMID is set correctly for their apps.

develar: I think, we should start using link if it is exists and start using executable otherwise.

MariaDima: Thanks.
I'll update the patchset (no option will be added. I'll just fallback to the executable in case the shortcut does not exist)

develar:
Merge in process, please do not commit new changes :)

packages/electron-builder/templates/nsis/uninstaller.nsh
```diff
@@ -33,16 +33,29 @@ Section "un.install"
...
+    ${If} $R1 != ""
+      RMDir /r /REBOOTOK "$SMPROGRAMS\$R1"
+    ${Else}
+      Delete "$oldStartMenuLink"
+    ${EndIf}
+    Delete "$oldDesktopLink"
+  ${else}
+    ClearErrors
+    ${GetParameters} $R0
+    ${GetOptions} $R0 "--no-desktop-shortcut" $R1
```
develar: Am I right — if new installer "do not create desktop shortcut", then old uninstaller should delete desktop shortcut?

develar: I thinks it is not a good behavior. If `DO_NOT_CREATE_DESKTOP_SHORTCUT`, we just keep old shortcuts.

develar:
19.21.0

packages/electron-builder/templates/nsis/uninstaller.nsh
MariaDima: @develar , you are right.
I wasn't sure what was best to do here.

We would either enhance the installer to remove any existing oldStartMenuShortcut or let the application owner plan his update (meaning he would have to do 2 subsequent updates).
I selected the second option to not mess the installer with an uninstaller business logic.
But I see that this path leaves garbage: If the owner at the same time changes the shortcut path (shortcut name or menu directory), the desktop icon will never be removed!!!!
The owner would have to control carefully his updates ... That can be frustrating for a lot of people :-(

develar: @MariaDima Thanks for clear explanation. 👍

PR #1941, opened 2017-08-12
feat(appx): attempt to use identityName for manifest name

jviotti:
It is very useful to set a custom `identityName` when the name of the
app in package.json contains hyphens (which are not permitted in
`appxmanifest.yml`).

Even if you set a non-hyphenated name as `identityName`,
electron-builder will still use the package.json's `name` property for
the `name` appxmanifest.yml value.

Thus, we attempt to re-use `identityName` for that case as well.

### DIFF
--- a/packages/electron-builder/src/targets/appx.ts
+++ b/packages/electron-builder/src/targets/appx.ts
@@ -125,7 +125,7 @@ export default class AppXTarget extends Target {
             return appInfo.versionInWeirdWindowsForm
 
           case "name":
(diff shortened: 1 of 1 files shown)


develar:
I see that you are right. Thanks.

develar:
Released as 19.22.1 Decided to not release as major release because appx is not default target and identity name is not required option.

jviotti:
Sounds great to me, thanks a lot!

PR #1944, opened 2017-08-14
feat: custom Windows sign tool timeout via SIGNTOOL_TIMEOUT

dchest:
Hello,

Since we use EV cert for singing Windows binaries, we have to use a USB token for it, so we wrote a custom tool (used via `SIGNTOOL_PATH`), which copies the file to sign into Parallels, which then asks the developer to enter the PIN code, signs, and then copies the signed file back to macOS. Unfortunately, 2 minutes is sometimes not enough to finish the signing, so I'd like to make it customizable. Would this work for you?

Thank you!

### DIFF
--- a/packages/electron-builder/src/windowsCodeSign.ts
+++ b/packages/electron-builder/src/windowsCodeSign.ts
@@ -139,7 +139,7 @@ async function spawnSign(options: SignOptions, inputPath: string, outputPath: st
 
   const toolInfo = await getToolPath()
   return await exec(toolInfo.path, args, {
-    timeout: +(process.env.SIGNTOOL_TIMEOUT || 120 * 1000),
+    timeout: parseInt(process.env.SIGNTOOL_TIMEOUT as any, 10) || 120 * 1000,
     env: toolInfo.env || process.env
   })
 }
--- a/packages/electron-builder/src/windowsCodeSign.ts
+++ b/packages/electron-builder/src/windowsCodeSign.ts
@@ -139,7 +139,7 @@ async function spawnSign(options: SignOptions, inputPath: string, outputPath: st
 
   const toolInfo = await getToolPath()
   return await exec(toolInfo.path, args, {
-    timeout: 120 * 1000,
+    timeout: +(process.env.SIGNTOOL_TIMEOUT || 120 * 1000),
     env: toolInfo.env || process.env
   })
 }


develar:
We want to provide ability to sign file on remote servers. Despite that fact, your approach is also fully valid.

Yes, what I want to see — it is PR that adds ability to delegate signing to some API (and by default uses  current code). But it requires extra work.

I don't think that we need `SIGNTOOL_TIMEOUT` to be passed always. We can omit timeout if `SIGNTOOL_PATH` is set. But... it will be not logically for other users. So, 👍 to your fix.

packages/electron-builder/src/windowsCodeSign.ts
```diff
@@ -139,7 +139,7 @@ async function spawnSign(options: SignOptions, inputPath: string, outputPath: st
 
   const toolInfo = await getToolPath()
   return await exec(toolInfo.path, args, {
-    timeout: 120 * 1000,
+    timeout: +(process.env.SIGNTOOL_TIMEOUT || 120 * 1000),
```
develar: For what we need leading `+` here?

dchest:
Thanks! If I have time, I'll work on adding signing delegation.

Sorry for the errors with this PR, I hope I got them fixed now :)

packages/electron-builder/src/windowsCodeSign.ts
dchest: It converts process.env.SIGNTOOL_TIMEOUT from string to number

develar: Not quite evident. I will change to `parseInt(process.env.SIGNTOOL_TIMEOUT as any, 10) || 120 * 1000`

dchest:
thanks 👍

develar:
19.22.2 released.

develar:
@dchest Is it possible to share your tool, to bundle as part of electron-builder?

dchest:
It's here: https://github.com/PeerioTechnologies/desktop-release-builder (specifically, osslsigncode replacement is https://github.com/PeerioTechnologies/desktop-release-builder/blob/master/osslsigncode.js and the part that must run on Windows is  https://github.com/PeerioTechnologies/desktop-release-builder/blob/master/winsigner.js). As you can see it's not very generic, so I'm not sure you'd want to bundle it without changes. It requires some setup as described here https://github.com/PeerioTechnologies/desktop-release-builder#installation

Some notes:

* I'd probably use a network service instead of shared folders for passing files back and forth if I were to rewrite it.
* Directory watcher is polling a directory instead of using fswatch because Parallels breaks FS events.
* The repo has no license I'll fix this today.

develar:
@dchest Cool. Have you tried to use `prlctl` (http://download.parallels.com/desktop/v11/docs/en_US/Parallels%20Desktop%20Pro%20Edition%20Command-Line%20Reference.pdf)?

dchest:
@develar it would be very nice to launch the machine automatically, indeed. Since the operator needs to see PIN entry GUI for EV token to sign, it wasn't a big deal to additionally ask them to launch this tool manually, but yeah, it would be nice to have everything completely automated with prlctl (or its VirtualBox alternative if going full open source). We haven't used it yet, but maybe will do in the future.

develar:
Default SIGNTOOL_TIMEOUT changed from 2 minutes to 10 minutes.

PR #1952, opened 2017-08-16
fix(builder): use the same artifact name for AppImage when publishing

dchest:
AppImages had a different name when published on GitHub than on the disk.

(Note: I haven't tested these changes yet, but I believe they will solve the problem.)

### DIFF
--- a/packages/electron-builder/src/targets/appImage.ts
+++ b/packages/electron-builder/src/targets/appImage.ts
@@ -36,7 +36,8 @@ export default class AppImageTarget extends Target {
 
     // https://github.com/electron-userland/electron-builder/issues/775
     // https://github.com/electron-userland/electron-builder/issues/1726
-    const resultFile = path.join(this.outDir, this.options.artifactName == null ? packager.computeSafeArtifactName("AppImage", arch, false) : packager.expandArtifactNamePattern(this.options, "AppImage", arch))
+    const safeArtifactName = this.options.artifactName == null ? packager.computeSafeArtifactName("AppImage", arch, false) : packager.expandArtifactNamePattern(this.options, "AppImage", arch)
+    const resultFile = path.join(this.outDir, safeArtifactName)
     await unlinkIfExists(resultFile)
 
     const appImagePath = await appImagePathPromise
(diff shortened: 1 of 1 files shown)


develar:
Cannot be accepted. GitHub has strange rules about artefact name — no space, only ascii.

develar:
What problem do you have? GitHub publisher uploads file using not file name, but specified safe name. On disk you have correctly named (without strange GitHub restrictions) file.

dchest:
Ah, so `expandArtifactNamePattern` doesn't sanitize the name for GitHub? The problem I have is that the published name is not `artifactName` template expanded (it doesn't have spaces). Is there a function to sanitize the name for GitHub that I can apply after calling `expandArtifactNamePattern`? I think I saw something like this.

dchest:
That is, I have this template:

```
${name}-${os}-${arch}.${ext}
```

and it expands correctly on disk: 

```
appname-linux-x86_64.AppImage
```

but on GitHub I still have the version number in it and no OS (default template):

```
appname-v1.2.3-x86_64.AppImage
```

develar:
We don't want to force users to use GitHub restricted file name. It is not acceptable. So, our solution is — update info and artifactCreated event have special field to specify safe name according to *** Github rules.

The only issue — if you want to publish manually without our publisher.

> but on GitHub I still have the version number in it and no OS (default template):

Ok. So, proper solution — if unsafe normal artifact name conforms to Github rules (I cannot find any official documentation, so, `no spaces, only ascii.`), use it. If not — use hardcoded safe name pattern.

Currently, there is no such function. Yes — 👍 it will be amazing fix (and applicable not only for AppImage).

develar:
> Is there a function to sanitize the name for GitHub that I can apply after calling expandArtifactNamePattern

You simply cannot do it. I have tried. Problem — if name contains non-ascii and another not-acceptable form, resulting safe name can be ugly/not unique. So, I decided to use hardcoded safe name pattern.

dchest:
> Ok. So, proper solution — if unsafe normal artifact name conforms to Github rules (I cannot find any official documentation, so, no spaces, only ascii.), use it. If not — use hardcoded safe name pattern.

OK, will send a PR for that.

Issue #1951, opened 2017-08-16
Publish to multiple Bintray configurations from a single app


jviotti:
* **Version**: 19.22.1

* **Target**: deb, rpm, AppImage

Take the following configuration object as an example:

```json
{
  "appId": "com.example.digitalclock",
  "productName": "Digital Clock",
  "publish": [
    {
      "provider": "github",
      "repo": "digital-clock",
      "owner": "jviotti"
    }
  ],
  "linux": {
    "category": "Utils",
    "icon": "assets/icons",
    "packageCategory": "utils",
    "maintainer": "Jane Doe <janedoe@example.com>",
    "target": [ "deb", "rpm", "AppImage" ],
    "synopsis": "A minimalistic digital clock rendered in monospace font"
  },
  "deb": {
    "publish": [
      {
        "package": "digital-clock",
        "repo": "digital-clock-debian",
        "user": "jviotti",
        "component": "digital-clock",
        "provider": "bintray"
      }
    ],
    "depends": [
      "gconf2",
      "gconf-service",
      "libnotify4",
      "libappindicator1",
      "libxtst6",
      "libnss3"
    ]
  },
  "rpm": {
    "publish": [
      {
        "package": "digital-clock",
        "repo": "digital-clock-rpm",
        "user": "jviotti",
        "provider": "bintray"
      }
    ],
    "depends": [
      "GConf2",
      "libnotify",
      "nss",
      "libappindicator",
      "libXtst"
    ]
  }
}
```

There is a global `publish` GitHub object, and both `deb` and `rpm` define their own `publish` object, both to Bintray, but pointing to different repositories.

When building both `rpm` and `deb` targets at the same time, `electron-builder` publishes both deb and rpm files to the Debian Bintray repository, basically ignoring the `rpm` specific `publish` configuration.

See https://travis-ci.org/jviotti/digital-clock/builds/265001992 as an example:

```
Building AppImage for arch ia32
Building deb
Publishing to Github (owner: jviotti, project: digital-clock, version: 1.0.0)
Uploading digital-clock-1.0.0-x86_64.AppImage to GitHub
Building deb
Uploading digital-clock-1.0.0-i386.AppImage to GitHub
Building rpm
Publishing to Bintray (user: jviotti, owner: jviotti,  package: digital-clock, repository: digital-clock-debian, version: 1.0.0)
Uploading digital-clock_1.0.0_amd64.deb to Bintray
Building rpm
[...]

PR #1958, opened 2017-08-17
fix(publish): cache publishers based on the target name

jviotti:
Publishers are currently cached by the publisher provider (e.g.
"bintray") and by the platform name (e.g. "linux"). This means that if
the user has different publish configurations for targets that belong to
the same OS (e.g. deb and rpm), then one will override the other on the
cache.

Fixes: https://github.com/electron-userland/electron-builder/issues/1951

### DIFF
--- a/packages/electron-builder/src/publish/PublishManager.ts
+++ b/packages/electron-builder/src/publish/PublishManager.ts
@@ -135,7 +135,7 @@ export class PublishManager implements PublishContext {
           break
         }
 
-        const publisher = this.getOrCreatePublisher(publishConfig, packager)
+        const publisher = this.getOrCreatePublisher(publishConfig, packager, event.target)
         if (publisher == null) {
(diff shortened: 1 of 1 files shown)


packages/electron-builder/src/publish/PublishManager.ts
```diff
@@ -158,8 +158,12 @@ export class PublishManager implements PublishContext {
     }
   }
 
-  private getOrCreatePublisher(publishConfig: PublishConfiguration, platformPackager: PlatformPackager<any>): Publisher | null {
-    const providerCacheKey = `${publishConfig.provider}-${platformPackager.platform.name}`
+  private getOrCreatePublisher(publishConfig: PublishConfiguration, platformPackager: PlatformPackager<any>, target: Target | null): Publisher | null {
+    let providerCacheKey = `${publishConfig.provider}-${platformPackager.platform.name}`
+    if (target) {
```
jviotti: The conditional is only needed because `target` might be `null` (not sure under what circumstances that could happen)

PR #2025, opened 2017-09-01
Fix parsing of CLI '-c' arguments when combining values with a config file path

### LINKED ISSUE #2016: unable to override config options with CLI arguments when using custom config file
* **Version**: 19.27.2

* **Target**: macOS, Windows

**TL;DR:** I'm trying to specify both `--config.compression=maximum` and `--config ./electron.config.json` as CLI arguments, and I'm seeing an error/no effect, depending on the order of the arguments.

I currently have two scripts in my package.json:

```json
{
  "scripts": {
    "electron-release": "electron-builder --config ./electron.config.json --mac --win --publish always",
    "electron-test": "electron-builder --config ./electron.config.json --mac"
  }
}
```

I'm using a separate electron.config.json file to separate my app's dependencies from the build instructions. I'd like to override the `compression` option from the electron.config.json in my `electron-test` script so that it will run a little faster. I tried changing the `electron-test` script to the following:
`"DEBUG=electron-builder electron-builder --config ./electron.config.json --mac --config.compression=store`

However, I see this effective config, using the `config.compression` option specified in the electron.config.json file instead of the one I specified using the CLI argument.

```
  electron-builder Effective config:
  ...
  electron-builder compression: maximum
  ...
```

I also tried changing the `electron-test` script to this command:
`DEBUG=electron-builder electron-builder --config.compression=store --config ./electron.config.json --mac`

But then I get the following effective config in the debug messages.

```
  electron-builder Effective config:
  electron-builder '0':
  electron-builder   compression: store
  electron-builder '1': ./electron.config.json
  electron-builder length: 2
```

And the following error:

```
Error: Configuration is invalid.
 - configuration has an unknown property 'length'. These properties are valid:
...
```

Would it be possible to make both of these arguments work at the same time? If you point me in the right direction in the code, I can submit a PR.

elsbree:
Fixes #2016

### DIFF
--- a/packages/electron-builder/src/builder.ts
+++ b/packages/electron-builder/src/builder.ts
@@ -4,8 +4,10 @@ import { CancellationToken } from "electron-builder-http"
 import { addValue, Arch, archFromString, isEmptyOrSpaces, warn } from "builder-util"
 import { executeFinally } from "builder-util/out/promise"
 import { PublishOptions } from "electron-publish"
+import { deepAssign } from "read-config-file/out/deepAssign"
 import { DIR_TARGET, Platform } from "./core"
 import { normalizePlatforms, Packager } from "./packager"
+import { Configuration } from "./configuration"
 import { PackagerOptions } from "./packagerApi"
 import { PublishManager } from "./publish/PublishManager"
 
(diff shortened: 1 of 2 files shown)


develar:
Thanks!

Yes, Yargs creates array.

develar:
Released as 19.27.7 (`next` release).

PR #2031, opened 2017-09-04
Update configuration.md - improve docs

Sibiraj-S:
few doc improvements

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -30,12 +30,12 @@ Env file `electron-builder.env` in the current dir ([example](https://github.com
 ## Configuration
 
 <!-- do not edit. start of generated block -->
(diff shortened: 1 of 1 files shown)


docs/configuration/configuration.md
```diff
@@ -128,7 +128,7 @@ Env file `electron-builder.env` in the current dir ([example](https://github.com
   
   If `react-scripts` in the app dev dependencies, `react-cra` will be set automatically. Set to `null` to disable automatic detection.
 * <code id="Configuration-extraMetadata">extraMetadata</code> any - Inject properties to `package.json`.
-* <code id="Configuration-forceCodeSigning">forceCodeSigning</code> = `false` Boolean - Whether to fail if application will be not signed (to prevent unsigned app if code signing configuration is not correct).
```
develar: What's wrong with `will be`?

Sibiraj-S: actually, the statement mentions about an action that is `to fail`.

Signing should happen before the process which is going to fail happens

Grammatically, this will be more correct.

Sibiraj-S:
@develar can close this PR if this is GTG.

Issue #748, opened 2016-09-12
AppImage launcher icon defaults to Electron default icon


eriedl-kiban:
Which version of electron-builder are you using?
* 6.7.3

What target are you building for?
* Linux (AppImage target on Ubuntu 16.04)

I have the image sizes 16, 24, 48, 64, 128, 256 and 512 in the icons sub-directory as PNGs. I had to put them there as the Mac OS X ICNS file is missing the 24, 48 and 64 sizes for now.
Anyway, the AppImage is successfully built and shows the correct icon. However, when launching the app the icon shown on the launcher bar defaults to Electron's default icon.
I'm not sure what I'm missing here to display the app icon also on the launcher. Mac OS X and Windows builds show the correct icons when running.

develar:
> launcher bar defaults to Electron's default icon

Please use launch menu to start app, not app image.

Please see https://github.com/probonopd/AppImageKit/issues/174

@probonopd First user issue :( It is really not user-friendly.

probonopd:
This is also a bit an issue with how icons are handlet by e.g., the Unity launcher. I cannot change this, but there is a workaround: In Qt, one can set an icon to the main window using `setWindowIcon(QIcon("icon.png"));` and Unity (and other launchers) will then use this icon also for the dock. I don't know how to do the same without Qt yet but it should definitely be possible.

This way, the correct icon would show up regardless of how the user launches the app.

develar:
http://stackoverflow.com/questions/31529772/setting-app-icon-for-electron-atom-shell-app

`new BrowserWindow({width: 800, height: 600,icon: __dirname + '/Bluetooth.ico'});`

It is not convenient, but can be a temporary solution. I will try to fix https://github.com/probonopd/AppImageKit/issues/174 after #529

probonopd:
Could it use `.DirIcon`?

eriedl-kiban:
> Please use launch menu to start app, not app image.

I can confirm that searching for the app and then starting it from the Launcher search results shows the correct icon. The correct icon is also shown in the launcher when installed from a Debian package and then run.

> new BrowserWindow({width: 800, height: 600,icon: __dirname + '/Bluetooth.ico'}); (keep in mind, that currently icons are not copied to packaged app, so, you have to include it somehow).

[...]

PR #2041, opened 2017-09-06
Escape .desktop to fix RESOURCE_NAME detection

crazyserver:
Not sure if this is the only place where this change should be done.
Also don't know about the last line change.

Related to https://github.com/electron-userland/electron-builder/issues/748


develar:
Thanks, because we use ejs now, I will get rid of these bloody shell vars.

PR #2076, opened 2017-09-13
fix(deployment): properly check options.draft value in gitHubPublisher

githubbob42:


develar:
Will be published 22 September.

PR #2089, opened 2017-09-15
License btns config

etherElric:
Create a licensebuttons_[lang_code].json to overwrite the license buttons labels when creating a dmg.
File structure:
{
  "lang": "English",
  "agree": "Agree",
  "disagree": "Disagree",
  "print": "Print",
  "save": "Save",
  "description": "Here is my own description"
}


develar:
Please fix CI failure — add missing deps to dmg-builder package.json.

Could you please use iconv-lite to encode instead of codepage? Because codepage has 3 deps, and iconv-lite zero.

packages/dmg-builder/src/licenseButtons.ts
```diff
@@ -0,0 +1,189 @@
...
+import { parseJson } from "builder-util-runtime"
+import _debug from "debug"
+export const debug = _debug("electron-builder")
+const iconv = require("iconv-lite")
+
+export async function getLicenseButtonsFile(packager: PackageBuilder): Promise<Array<LicenseButtonsFile>> {
+  const files = (await packager.resourceList)
+    .filter(it => {
+      const name = it.toLowerCase()
+      return name.startsWith("licensebuttons_") && name.endsWith(".json")
```
develar: Please name.toLowerCase() here to allow any letter case.

develar:
We should also allow to use yaml and json5. But do not waste your time — I will made final cosmetic changes on release. Will be released 22 September.

develar:
Thanks a lot for a cool and useful feature.

Issue #1992, opened 2017-08-25
Choose Installation Options page - wrong internationalization - Portuguese Brazil language


hemavidal:
* **Version**: 19.24.1

* **Target**: NSIS

I've selected the option:
package.json
```json
{
  "nsis": {
      "oneClick": false,
      "allowElevation": true,
      "allowToChangeInstallationDirectory": true,
      "createDesktopShortcut": true,
      "displayLanguageSelector": true
  }
}
```
And the first window **Choose Installation Options** appears in english:
![image](https://user-images.githubusercontent.com/1553814/29720311-f277c58c-898f-11e7-8bb2-85d4fb028a2d.png)

- The window title is correct "Instalação de myapp"
- The buttons are correct "Seguinte >" and "Cancelar"
- But the other texts don't change.

**I saw that in Deutsch is correct.**
![image](https://user-images.githubusercontent.com/1553814/29720395-3e290f4a-8990-11e7-973e-9cf4222da55e.png)

Is there a way/option to change it or it is a bug?

develar:
Please translate https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/templates/nsis/messages.yml and https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/templates/nsis/assistedMessages.yml

You to add `pt`.

hemavidal:
With **pt** it worked, but with pt_BR doesn't work.
The electron-builder/NSIS make distinction of **pt** and **pt_BR**?

develar:
What do you mean "worked"?

> The electron-builder/NSIS make distinction of pt and pt_BR?

In the translations of electron-builder you can use both variants. `pt_BR` can be also used.

hemavidal:
The key **pt** have to receive pt_PORTUGAL translation not pt_BRAZILIAN translations.
I've put the **pt_BR** in assistedMessages.yml but nothing hapens...
So i put the pt_BRAZILIAN message in pt_PORTUGAL key.(there is not the correct way)
I must to enable pt_BR in other place to installer accept both of them?
```yml
# assisted installer messages, see messages.yml for one-click installer messages
chooseInstallationOptions:
  en: Choose Installation Options
  de: Installationsoption wählen
  ru: Выберите опции установки
  sk: Vyberte možnosti inštalácie
  pt: Escolha as opções de instalação
  pt_BR: Doesn't work :/
```

develar:
[...]

PR #2095, opened 2017-09-18
feat(nsis): brazilian portuguese language

marceloavf:
**Feature:** Brazilian portuguese language for `nsis` installation

Refs to #1992


develar:
Will be released 22 September.

hemavidal:
Thanks Marcelo!

develar:
19.29.1 marked as latest and released for all. Please verify that it works.

Note — official NSIS supports `pr_BR` but this locale was not bundled (as we bundle only proven/requested languages). Now bundled by default. It means that by default if your OS locale is pr_BR, this translation will be used.

PR #2121, opened 2017-09-27
Make BUILD_NUMBER higher priority than CI-defined build numbers

joaocc:
This change allows having custom BUILD_NUMBERs even when running in a CI environment, without having to modify packages.json.
For instance, to have BUILD_NUMBER based on UTC timestamp instead of sequential ID, a build script only needs to do the following:
```
BUILD_NUMBER=$(date -u +"%Y%m%d%H%M%S")
```

### DIFF
--- a/packages/electron-builder/src/appInfo.ts
+++ b/packages/electron-builder/src/appInfo.ts
@@ -20,7 +20,7 @@ export class AppInfo {
       buildVersion = info.config.buildVersion
     }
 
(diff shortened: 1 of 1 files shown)


develar:
Thanks! Please try 19.31.3

joaocc:
Thx!
Is there any way to test this before the release appears on npm? Installing via direct git dependency in packages.json doesn't seem to work (prob because the packages are within folders).

develar:
@joaocc Just set version to ` 19.31.3` (tagged as `next`, not as `latest`).

joaocc:
Tested. Works great. Thx!

PR #2161, opened 2017-10-05
Use BUILD_BUILDNUMBER env var as buildNumber for VSTS CI

Sirikon:
In VSTS Builds, the build variable containing the build number is `BUILD_BUILDNUMBER`

https://docs.microsoft.com/en-us/vsts/build-release/concepts/definitions/build/variables


develar:
19.34.2 wil be tomorrow morning CET.

Issue #1547, opened 2017-05-15
Adding a license file for Mac .pkg target


gniezen:
* **Version**: 17.0.3

* **Target**: mac, pkg

It looks like it's possible to add a license file for Windows NSIS as well as Mac .dmg, but not for Mac .pkg target?

genbandbrent:
any forecast for this feature?

develar:
Only if will be donated.

genbandbrent:
Thanks. How long would it take if a donation were made?

develar:
@genbandbrent https://github.com/electron-userland/electron-builder/issues/1639#issuecomment-307354319 You can expect fix this week if.

PR #2212, opened 2017-10-19
feat: #1547 Added pkg license

Egorikhin:


packages/electron-builder/src/targets/pkg.ts
```diff
@@ -70,6 +71,11 @@ export class PkgTarget extends Target {
     let distInfo = await readFile(distInfoFile, "utf-8")
     const insertIndex = distInfo.lastIndexOf("</installer-gui-script>")
     distInfo = distInfo.substring(0, insertIndex) + `    <domains enable_anywhere="${options.allowAnywhere}" enable_currentUserHome="${options.allowCurrentUserHome}" enable_localSystem="${options.allowRootDirectory}" />\n` + distInfo.substring(insertIndex)
+    if (options.license) {
```
develar: We should not require explicit option.

1. use getLicenseFiles util method to autodetect (see other implementations).
2. disable autodetect of option explicitly set to null — triple equals

Egorikhin: Would you like me to do something like this? 
http://www.cocoabuilder.com/archive/xcode/319021-add-localized-readme-license-etc-in-distribution-package-using-productbuild.html

>in your resource folder, add the localized folders, but named English.proj, French.proj, German.proj (they will be renamed as en.proj, fr.proj, de.proj by productbuild) containing your localized Welcome.rtf (for example), and use the --resources option.

develar: At first I think we should add basic implementation, and add localized later. If you want to add localization righ now — great. If not — it is ok too, you are not forced to prepare polished full featured PR.

Thanks for info. Yes, we can automatically creates such files for productbuild since localized licenses supported for DMG/nsis.

Egorikhin: In order not to make unnecessary pushes, could you please tell me if my approach is right or not:
```ts
    if (options.license !== null) {
      const licenseFiles = await getLicenseFiles(this.packager)
      const license = await this.packager.getResource(options.license, ...licenseFiles.map(f => f.file))
      if (license != null) {
        const licensePath = path.join(this.packager.info.projectDir, license)
        distInfo = distInfo.substring(0, insertIndex) + `    <license file="${licensePath}"/>\n` + distInfo.substring(insertIndex)
      }
    }
```

Egorikhin:
Any comments on my new changes, please?

develar:
Released as 19.43.0, thanks.

Issue #1167, opened 2017-01-25
Electron Builder AutoUpdater download-progress event is not called on MAC.


pdabhiargus:
Electron Version: 1.4.3
Electron Builder Version: 11.4.4
Target: macOS 10.12.2
Node Version: 7.2.1
NPM Version: 4.1.1

I am trying to use the electron-auto-updater module on MAC. Everything works all OK expect that the 
download-progress event of auto updater never gets called.

I have the same solution for Windows and it works perfectly on it.

develar:
FYI: electron-auto-updater renamed to electron-updater.

Yes, because under the hood we use native Squirrel.Mac updater and update downloaded by Squirrel.Mac.

tborychowski:
@develar,
is there a feature-request (or a bug) reported to Squirrel.Mac?
Do you know if that's in progress?
Currently the UX of the autoupdate is crap: use will see that download has started and somewhere between **10 seconds - 5 hours** they might see "Download complete" (if they didn't close the app in the mean time).

develar:
> is there a feature-request (or a bug) reported to Squirrel.Mac?

I am not aware about it. You can search / file issue.

If Squirrel.Mac will reject this idea — probably I can make a PR to Squirrel.Mac.

tborychowski:
Thanks. Issue added: https://github.com/Squirrel/Squirrel.Mac/issues/200

develar:
`electron-updater` 2.0.0 released.

No more any difference compared to windows version.

* Secure sha512 integrity check.
* Full support of any existing electron-updater features on macOS.
    * Progress is supported.
    * Any dynamic or relative URL is supported.
    * Any provider as you want without any restrictions. Yes — private github repo is supported (#1370). Yes — Bintray is supported now (#1172). 

Wants more? Stay tuned — lzma2 instead of zip or even direct support of DMG as update file is coming.

biroplane:
what kind of event does the progress fires?
i've tried with both `progress` and `download-progress`, but nothing happen.
any chance to have some snippet?!
Thanks!

danieldbird:
@biroplane did you find out how to get the download-progress?

biroplane:
Nope! still trying

pdabhiargus:
@danieldbird  @biroplane 

I have the following code working all OK on Windows.  I have used mainWindow.setProgressBar to show the progress at the taskbar icon. On MAC auto-update works but without progress.

Here is the package.json

```
{
[...]

PR #2242, opened 2017-10-30
Updating docs: `download-progress` works on MacOS

kraenhansen:
Related to https://github.com/electron-userland/electron-builder/issues/1167#issuecomment-340384415 - I can verify that I am receiving observing these events on MacOS now.


develar:
Thanks for updating docs :)

PR #2245, opened 2017-10-30
Correct typo in section title

jfix:
'How do Publish' should be 'How to Publish'

### DIFF
--- a/docs/configuration/publish.md
+++ b/docs/configuration/publish.md
@@ -16,7 +16,7 @@ You can publish to multiple providers. For example, to publish Windows artifacts
 }
 ```
 
-## How do Publish
+## How to Publish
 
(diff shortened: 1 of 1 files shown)


develar:
Are you sure? https://www.google.de/search?client=safari&rls=en&q=%22how+do+publish%22&ie=UTF-8&oe=UTF-8&gfe_rd=cr&dcr=0&ei=bVX3WfqhE4nZ8Afs76q4CQ

jfix:
Yep, quite sure. :-)   Do the same search with the 'to' and you'll get 65
millions results, vs 21000 for the original phrase. ;-)

The difference is 'How to do this or that' which is not so much a question
but an instruction or a tutorial, an explanation, if you want.
'How do (these work)?' is a question, and I think this section explains how
to do something.

Hope this makes it clearer.


cheers,
Jakob.

develar:
Thanks from russian speaking guy :)

PR #2262, opened 2017-11-02
Include application name in update notification

iffy:
Also made "Quit" a lowercase "exit"


develar:
On macOS — `Quit`, on Linux/Windows — ?

iffy:
@develar I'm not sure.  Even on macOS, the English reads strange how it is.  Some other options could be:

> `Application` version `version` is downloaded and will automatically be installed after it is closed

> `Application` will automatically update to version `version` after it is closed

> `Application` will automatically update to version `version` after quitting

develar:
I am agree that on macOS we should use suggested `Quit`. But not sure that `Quit` applicable for Windows.

PR #2278, opened 2017-11-08
Fixed --force-run

### LINKED ISSUE #2179: autoUpdater.quitAndInstall does not force run with oneClick=false installation and silent install
* **Version**: 
19.27.7

* **Target**: 
2.10.2

I have an app with an installation which is not oneClick and allows to change installation directory, and when auto-updating the app the silent installation finishes but the app doesn't run. When I change the installation to oneClick the app runs after the silent installation.
This is its nsis:
```
"nsis": {
      "oneClick": "false",
      "allowToChangeInstallationDirectory": "true"
    }
```

I'm using setImmediate as it was given as a solution in #1604.
```
setImmediate(function() {
                app.removeAllListeners("window-all-closed");
                updateWindow.close();
                autoUpdater.quitAndInstall(true, true);
            });
```

JohnyCilohokla:
Moved StartApp macro to common.nsh
Got --force-run (isForceRun) working in all configurations
Setup isOneClickRun for ONE_CLICK + RUN_AFTER_FINISH
When isForceRun is defined or when isOneClickRun is true the installer will start the app after the installation is complete

Note: this PR fixes #2179

### DIFF
--- a/packages/electron-builder/templates/nsis/common.nsh
+++ b/packages/electron-builder/templates/nsis/common.nsh
@@ -72,3 +72,19 @@ Name "${PRODUCT_NAME}"
 
   !define MUI_PAGE_CUSTOMFUNCTION_PRE licensePre
 !macroend
+
+
+!macro StartApp
+  Var /GLOBAL startAppArgs
+  ${if} ${isUpdated}
+    StrCpy $startAppArgs "--updated"
+  ${else}
+    StrCpy $startAppArgs ""
+  ${endif}
+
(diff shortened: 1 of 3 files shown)


GhostfromTexas:
According to the docs, force-run only works with OneClick enabled. This commit changes that. ForceRun will work in either OneClick mode or not. Tested across my system multiple times and it works fine in both cases now.

Docs will need to be adjusted if this change is accepted.

develar:
Merge in process, please do not resolve conflicts :)

develar:
Thanks a lot!

PR #2285, opened 2017-11-08
fix(nsis): NSIS installer is crashing on "Exit" (Admins on Win7)

### LINKED ISSUE #2284: NSIS installer is crashing on "Exit" (Admins on Win7)
* **electron-builder**: 19.45.2

* **electron-updater**: 2.16.1

* **Target**: Windows 7 (x64)

We are experiencing the same behavior reported by tylerlong (https://github.com/electron-userland/electron-builder/issues/2134):
_I tried the NSIS installer on Windows 7, it crashed at the end of the installation. But the app had been installed successfully and I could use the installed app without problem._

I decided to open this ticket to track the issue, in case we can do something.
At the moment it mostly affects user experience.
The crash is occurring "right before" the installer would exit gracefully.

The issue affects only windows 7 users and specifically Admins.
I tested windows 10 and non admin users and there is no problem with the installer.

MariaDima:
Close #2284

### DIFF
--- a/packages/electron-builder/templates/nsis/oneClick.nsh
+++ b/packages/electron-builder/templates/nsis/oneClick.nsh
@@ -8,11 +8,7 @@
         StrCpy $startAppArgs ""
       ${endif}
 
-      !ifdef INSTALL_MODE_PER_ALL_USERS
-        ${StdUtils.ExecShellAsUser} $0 "$launchLink" "open" "$startAppArgs"
-      !else
-        ExecShell "" "$launchLink" "$startAppArgs"
-      !endif
+      ${StdUtils.ExecShellAsUser} $0 "$launchLink" "open" "$startAppArgs"
     !macroend
   !endif
 


MariaDima:
I cannot tell that I fully understand why this issue appeared now.

I did an investigation on this and realized that the installer crashes when all following conditions meet:
 * The installer is executed by an admin user (I didn't select to install for all available users, just me)
 * ExecShell is used to run the application after installation is complete
 * The application is started using the start menu link.

It seems that something is problematic in regards to the location of the start menu link.
Even though the application starts fine, the Exec command seems to be crashing the installer (I'm not sure whether is relates to an exit code we are not waiting for).

If we start the application either by using the ExecShellAsUser command, or by using the ExecShell and the actual location of the exe file (not the start menu), everything works fine.

So I was thinking of proposing the use of the actual exe location.
But then I thought that there is no reason to use the ExecShell command (at least not visible to me at the moment).
If we are simple users, we are fine.
If not, we are still fine, since an admin is actually a user.
In general, no matter how the installer is being executed, I think that the application is ok to be started with the users actual privileges.

What do you think @develar ?
Is there a reason for needing the ExecShell command?

develar:
> Is there a reason for needing the ExecShell command?

It is used only due to historical reasons. Originally was only per-user and scenario, when installer is opened by admin user, was not checked. Looks good.

MariaDima:
Thanks for the merge!
You are working with the speed of light!

PR #2300, opened 2017-11-14
fix(AppImage): AppRun.sh sometimes fails to install icons

Doccrazy:
If the ~/.local/share/applications folder contains files owned by root (caused by broken packages or sudo usage), the chmod on $DESTINATION_DIR_DESKTOP/* will fail with "permission denied", causing the script to exit early. Also, I would consider modifying files not owned by this package bad practice.

This fixes the behavior by using the desktop-file-install --mode parameter instead of the chmod.


develar:
Thanks a lot!

PR #2306, opened 2017-11-16
Mention how to disable code-signing on macos

jszaszvari:
It's previously implied but people seem to struggle with figuring this out.


eranimo:
How do you disable code signing in Windows?

Issue #2330, opened 2017-11-24
Check dependencies is run on both devPackageFile and appPackageFile


XuluWarrior:
electron-builder: 19.45.5
Target: --dir

I have a project where the root folder has a package.json with a dependency on "electron" and a "app" directory with package.json without this dependency.

When I run electron-builder it gives the error 
`Package "electron" is only allowed in "devDependencies". Please remove it from the "dependencies" section in your package.json.`

If I have provided an app package file I don't see why it should check also check the dev package file for the electron dependency.  There is no problem as the app one is used for installing node_modules before packaging.

diegoflorez:
move `electron` from `dependencies` to `devDependencies` in **package.json** and append next lines to the **eslintrc** file:
 ```
 // https://github.com/benmosher/eslint-plugin-import#importcore-modules
  settings:{
    'import/core-modules': ['electron']
  }

```

XuluWarrior:
@diegoflorez did you mean to comment on #2330?  I'm not sure how it's related.

PR #2331, opened 2017-11-24
#2330

XuluWarrior:
If devPackageFile and appPackageFile are both defined, only check electron dependencies in appPackageFile.


develar:
I don't fully agree, because you should specify dependencies even in the `devPackageFile` as `devDependencies` but I understand, that some users can find it illogical. And so, electron builder cannot force you and require some inconvenient for you changes.

PR #2343, opened 2017-11-28
Grammar changes

emilymalecbrown:
Hi there! Just a small PR to change grammar. Got tripped up here because `whereas` usually means `in contrast to` rather than `consequently`. Thanks for your work on this!


develar:
Thanks for native English :)

PR #2373, opened 2017-12-11
[fix] Add the listener of redirect event

nicholaslee119:
Problem:
There is no event listener for redirect(302) event, and the redirect event won't fire the response event listeners.


develar:
Do you use private GitHub provider?

develar:
Redirect is handled in the parent class, please describe how to reproduce the issue.

nicholaslee119:
oops, sorry for the PR a wrong branch

Issue #2374, opened 2017-12-11
Private GitHub Provider: neither event "update-available" nor event "update-not-available" fired after checkForUpdates()


nicholaslee119:
* **Version**: 19.43.3

* **electron-updater Version**: 2.16.1

* **Target**:  Mac

I use autoUpdater for my private github repo like this:

```js
autoUpdater.setFeedURL({
  provider: 'github',
  token: 'MY_TOKEN',
  owner: 'nicholaslee119',
  private: true,
  repo: 'electron-updater-example'
});

console.log("App start....")

autoUpdater.on('checking-for-update', () => {
  console.log('Checking for update...');
})

autoUpdater.on('update-available', (info) => {
  console.log('Update available.');
})
autoUpdater.on('update-not-available', (info) => {
  console.log('Update not available.');
})

autoUpdater.on('error', (err) => {
  console.log('Error in auto-updater. ' + err);
})

app.on('ready', function()  {
  console.log("ready!!!");
  autoUpdater.checkForUpdates();
});

```

and here is output

```
App start....
ready!!!
Checking for update...

```

It seems like the neither update-available nor update-not-available were fired after checkForUpdates()

develar:
Reopened because @jpeizer reported that it is still broken. https://github.com/electron-userland/electron-builder/issues/2379#issuecomment-351420471

jpeizer:
Logs:
```
[11:11:59.539] [info] Checking for update
[11:12:03.157] [info] Found version <version> (url: <package>.exe)
[11:12:03.158] [info] Downloading update from <package>.exe
```
This is where it appears to stop.  Let me know if you want any more details.

nicholaslee119:
@jpeizer Thanks for your report, I will check it in a couple of days

nicholaslee119:
@develar 
I am handling the problem of downloading update from private Github,
and I have seen there is a local Proxy Server for downloading assets,
Could you tell me why are you putting this local Proxy Server in that way?

develar:
@nicholaslee119 Because Squirrel.Mac doesn't support arbitrary URL, but we cannot use predefined URLs for GitHub releases.

PR #2375, opened 2017-12-11
fix the problem of redirect

nicholaslee119:
https://github.com/electron-userland/electron-builder/issues/2374

fix the problem on this issues

according to electron document
[clientRequest](https://electronjs.org/docs/api/client-request#requestfollowredirect)

the event "response" and "redirect" are two separate events. 

So I split them up

### DIFF
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -132,6 +132,19 @@ Please double check that your authentication token is correct. Due to security r
       return
     }
 
+    const redirectUrl = safeGetHeader(response, "location")
+    if (redirectUrl != null) {
+      if (redirectCount > 10) {
+        reject(new Error("Too many redirects (> 10)"))
+        return
+      }
+
+      this.doApiRequest(prepareRedirectUrlOptions(redirectUrl, options), cancellationToken, requestProcessor, redirectCount)
+        .then(resolve)
+        .catch(reject)
+      return
+    }
+
     let data = ""
     response.setEncoding("utf8")
     response.on("data", (chunk: string) => {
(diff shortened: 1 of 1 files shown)


develar:
Thanks for spotting this recent regression.

Problem is that this http executor is used not only for Electron, but also for NodeJS, where no `redirect` event. Could you please do not remove old logic and just add a new redirect event?

nicholaslee119:
Maybe it would be better if I put that logic into ElectronHttpExecutor?

develar:
👍 Yes. 

FYI: we must not use electron auto redirect because we must remove Authorization header (see prepareRedirectUrlOptions method impl).

develar:
Have you tested it locally?

nicholaslee119:
I tried to put the logic to ElectronHttpExecutor.js, 
but It seems that I must refactor the abstract function `doRequest` in HttpExecutor, So it would an impact on others subclass  which inherited HttpExecutor, 
and I think It does not deserve to refactor all of them just because of the little "odd behavior" of electron.
anyway, maybe there will be the same "odd behavior" of another sub httpExecutor in the future
So I just leave the redirecthandler in HttpExecutor.

nicholaslee119:
I have tried to `npm run test-all` even on the origin master without changing anything
maybe it seems like there is something wrong with test cases?

develar:
@nicholaslee119 Thanks! Please verify electron-updater 2.17.5.

jpeizer:
**Failed**

**Setup:**
yarn _v1.3.2_
node: _v8.1.2_
Electron: _v1.8.2-beta.3_
Electron-Builder: _v19.49.2_
Electron-Updater: _v2.17.6_
OS: _Windows 10 64 bit_
Github: _Private Repo_
**main.js**
`autoUpdater.setFeedURL({
    provider: 'github',
    repo: '<repo>',
    owner: '<owner>',
    vPrefixedTagName: true,
    token: '<token>',
    private: true,
  });` < for testing without building

The last trigger seen was downloading.  'update-downloaded' has failed to trigger.

Built result: same as testing environment.

Last known good configuration:
Electron-Builder: v19.44.0
Electron-Updater: v2.16.0

PR #2389, opened 2017-12-15
Update README.md

purefan:
Minor grammar improvement


develar:
is, because typings it is one object. But now I am not sure...

PR #2397, opened 2017-12-18
I'd like to make computeSignToolArgs usable in custom signing.

spicykoala:
computeSignToolArgs is available on the configuration object passed into a custom signing method. This was really helpful to me, but in order to make it usable I had to remove the custom sign check from the winPackager.

I'm not sure if it's the best way to solve the issue, but in my view we can just leave out certificateSubjectName and certificateSha1 and certificateFile for the same effect. Either way, I wanted to create this pull request so that it could be discussed.

Thanks for all of your fantastic work - this tool has been awesome.


develar:
CI passed, but I am not sure that solution is correct.

> in order to make it usable

Could you please clarify why it is not usable now? If because your custom signer still want to get code singing info, please explain why.

spicykoala:
Right now if you call it, for example:
```js
configuration.computeSignToolArgs(true);
```

It blows up inside windowsCodeSign.ts on 
```ts
const certificateFile = (options.cscInfo as FileCodeSigningInfo).file
```

Because it's undefined, undefined.file throws an error.

I thought the args were handy to have, because I'm using Parallels to run the signing, but I'm manually making the calls for other reasons.

Issue #2379, opened 2017-12-12
Auto-update fails on Windows (and maybe Linux)


jayhasyee:
- electron-builder: 19.48.3
- electron-updater: 2.17.4
- electron: 1.7.9
- target: win ia32/x64 nsis

I use the `autoUpdater.checkForUpdatesAndNotify()` method.

The logs are:
```
[2017-12-12 11:48:51.200] [info] App starting...
[2017-12-12 11:48:51.271] [info] Checking for update
[2017-12-12 11:48:55.964] [info] Found version 0.0.2 (url: astroffers-au-setup-0.0.2.exe)
[2017-12-12 11:48:55.967] [info] Downloading update from astroffers-au-setup-0.0.2.exe
```

The downloading has been started, but it never resolves, however I see it in the network monitoring: the download starts end ends, but I never got any notification or log.
Before #2376 has been fixed (with `2.17.4`), Linux auto-update fell back to full download. There was the same issue. The download has been started, I saw it in the network monitoring, but it never resolved.

I think it relates to #2377 

This issue was broken after `electron-updater@2.16.3`. This version still works. I use it now for production releases without Linux distribution.

jpeizer:
Same result:

Setup:
yarn v1.3.2
node: v8.1.2
Electron: v1.8.2-beta.3
Electron-Builder: v19.49.2
Electron-Updater: v2.17.6
OS: Windows 10 64 bit
Github: Private Repo
main.js
autoUpdater.setFeedURL({ provider: 'github', repo: '<repo>', owner: '<owner>', vPrefixedTagName: true, token: '<token>', private: true, }); < for testing without building

The last trigger seen was downloading. 'update-downloaded' has failed to trigger.

Built result: same as testing environment.

Last known good configuration:
Electron-Builder: v19.44.0
Electron-Updater: v2.16.0

jpeizer:
Additional Information:

%localappdata%/Temp <up directory> created
No file created

develar:
@jasonhinkle Do you also use private GitHub repo?

jayhasyee:
Notice, if it's helpful, that I use public github repo: https://github.com/jayhasyee/astroffers
You can also find there my `build` preferences in `package.json`.

develar:
@jayhasyee If you use public repo, than @jpeizer case is not related to this issue, but to #2374

develar:
@jayhasyee You use nsis, not nsis-web, right?

develar:
[...]

PR #2413, opened 2017-12-26
fix the bug of redirect event in electron.net

nicholaslee119:
fix the bug which was mentioned in 
https://github.com/electron-userland/electron-builder/issues/2374
and
https://github.com/electron-userland/electron-builder/issues/2379


develar:
Thanks a lot, new release with this fix will be released tomorrow.

PR #2431, opened 2018-01-03
Change the default desktop helper to desktop-gtk2 for snaps

flexiondotorg:
I work on the [Snapcraft](https://snapcraft.io) team. This pull request changes the default [desktop helper](https://github.com/Ubuntu/snapcraft-desktop-helpers) for snaps to `desktop-gtk2` which is the preferred default for Electron applications for practically all use cases.

PR #2432, opened 2018-01-03
Update the default stage-packages for snaps

flexiondotorg:
I work on the [Snapcraft](https://snapcraft.io) team. This pull request updates the `stage-packages:` for snaps that suit practically all Electron applications.

  * `libappindicator1` is dropped because the the [desktop helpers](https://github.com/Ubuntu/snapcraft-desktop-helpers) provide the toolkit appropriate library. See pull request #2431 
  * `libgconf2-4` provide everything needed for `gconf2` support.
  * Likewise `libpulse0` provides everything needed to talk to the PulseAudio server.
  * `libnspr4` and `libpcre3` have been added as they are frequently required.


develar:
Hi @flexiondotorg Thanks a lot for your PR. Finally I got some reaction (https://forum.snapcraft.io/t/where-is-complete-full-up-to-date-docs-about-creating-desktop-apps/2843) :)

It is not clear from my post, but idea of using `gnome-platform` was rejected because actually size of resulting snap is bigger.
In this PR I see that you suggest to use `desktop-gtk2`. Interesting, I will investigate how does it impact final size.

develar:
Before: 72M
After: 108M.

36MB increase. Need to think about it....

Ideally, shared snap should be used, but for now it is not possible, because shared snap installed automatically only if from the same developer. And we need to ask Canonical to whitelist such shared snap if we want to avoid "fist, install shared snap".

But... I am not sure that "ideally" is really ideal solution. What does 36MB mean in our world? Nothing. What does missed dependency or increased maintenance cost mean? Malfunctioned app and a lot of wasted money.

PR will be merged and new version released tomorrow CET after some thinking. Thanks a lot.

flexiondotorg:
@develar Happy to help 😀

Using `desktop-gtk2` does increase the size of the snap a bit, but this is mitigated by snap delta updates.

We are also working on improving desktop theme integration and when than lands the size of snaps using `desktop-gtk2`, `desktop-gtk3`, etc will come down a good deal.

With regard to your post of the Snapcraft forum, you explained that you wanted the Snap experience using `electron-builder`  to be the best is can be. We do too. The PRs from me today mean that simply adding the `snap` target will produce a snap that integrates well in the vast majority of cases. We snapped over 20 assorted Electron apps using this in the last couple of days with great results.

I'll also be filing pull requests here as recommended defaults evolve for Electron snaps. If you ever have questions please feel free to contact me, @popey or @evandandrea directly.

develar:
A little bit increased build time also doesn't matter — as part of https://github.com/electron-userland/electron-build-service, prebuilt default snap template will be used to make build time constant (9 seconds).

develar:
Docker images will be updated in a hour (since it is Automated Build on Docker Cloud for security reasons).

Issue #2421, opened 2017-12-29
AutoUpdate takes 60 seconds to fail validating signature on Windows 7 due to PowerShell version


jlgreen2012:
* Electron-builder version **19.47.1**: 

* Electron-updater version 2.16.3

* **Target**: NSIS 32bit and 64bit

On Windows 7, the ConvertTo-Json command is not available when attempting to validate the signature of the update. I see a check in the code here to handle this:
https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
```
export function verifySignature(publisherNames: Array<string>, tempUpdateFile: string, logger: Logger): Promise<string | null> {
  return new BluebirdPromise<string | null>((resolve, reject) => {
    execFile("powershell.exe", [`Get-AuthenticodeSignature '${tempUpdateFile}' | ConvertTo-Json -Compress`], {
      timeout: 60 * 1000
    }, (error, stdout, stderr) => {
      if (error != null || stderr) {
        try {
          execFileSync("powershell.exe", ["ConvertTo-Json test"], {timeout: 10 * 1000})
        }
        catch (testError) {
          logger.warn(`Cannot execute ConvertTo-Json: ${testError.message}. Ignoring signature validation due to unsupported powershell version. Please upgrade to powershell 3 or higher.`)
          resolve(null)
          return
        }
...
```

Reading the code, I would expect the try statement to fail after 10 seconds at most, fall into the catch and move on. However, according to my logs, it's taking an entire 60 seconds (the full timeout value of the first execFile call) to get to the `logger.warn` statement in the catch block.

This is leading to a very poor user experience, and I'm hoping the issue might be clearer to you than to me. Is there a way to make this fail faster? Thanks!

develar:
We don't check Windows version because user can update PS. But in case of error we can just return null if `winVersion.startsWith("6.") && !winVersion.startsWith("6.3")`.

develar:
Timeout reduced from 60 to 30 seconds.

develar:
2.18.2

PR #2446, opened 2018-01-05
Add parameters to PowerShell invocation

### LINKED ISSUE #2385: electron-updater: Signature validation fails if powershell profile results in console warning
* **Version**: 19.29.2, `electron-updater` 2.16.1

* **Target**: Windows, Linux, MacOS

Signal Desktop update fails on WIndows in some cases, with the error "Unexpected token W in JSON."

This is caused by the user's powershell profile. If its configuration results in a warning on the console, you'll get output like this, which is invalid JSON:

```
> powershell.exe "Get-AuthenticodeSignature Signal.exe | ConvertTo-Json"
WARNING: Could not load PSReadline.
{
...
}
```

More information: https://github.com/WhisperSystems/Signal-Desktop/issues/1879

Recommendation: add `-NoProfile` to the powershell command here: https://github.com/electron-userland/electron-builder/blob/1ec4ba287664a54faa5d6d03aeafefeed9a204f3/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts#L11

briantist:
Adds the following arguments:

* `-NoProfile`: fixes #2385 (and is best practice)
* `-NonInteractive`: tells PowerShell to die instead of prompt in certain cases (this may also prevent the timeout seen in #2421, but not certain)
* `-Command`: this was implicit, but it's better explicit now that there are additional parameters

See also: https://stackoverflow.com/a/25172234/3905079

### DIFF
--- a/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
+++ b/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
@@ -10,7 +10,7 @@ import { Logger } from "./main"
 export function verifySignature(publisherNames: Array<string>, tempUpdateFile: string, logger: Logger): Promise<string | null> {
   return new BluebirdPromise<string | null>((resolve, reject) => {
     // https://github.com/electron-userland/electron-builder/issues/2421
-    execFile("powershell.exe", [`Get-AuthenticodeSignature '${tempUpdateFile}' | ConvertTo-Json -Compress`], {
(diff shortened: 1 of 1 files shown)


briantist:
The failed tests in circleci seem to be related to hitting a usage limit on some API during the tests (unrelated to the code)? Seems other PRs have that problem too.

I can't make heads or tails out of the AppVeyor output; if it's something I need to fix I'd appreciate some help on figuring out what that is.

develar:
Thanks a lot! Will be released in 2 days.

develar:
I will change another usages of PS in our code, thanks for spotting this issues.

PR #2452, opened 2018-01-08
afterSign hook

jeremco:
Hi

In documentation there is 2 hooks available : afterPack and beforeBuild.
I propose to add another hook : afterSign.
If defined, the callback function is called after code-signing ( but before pack into distributable format).

Regards
Jeremie


develar:
Could you please describe use case?

jeremco:
Hi

Thanks for merging.
For my project, we use widevine with offline DRM content (for the moment this is a fork of electron to manage offline licences), but to use electron with widevine offline DRM content, the exe needs another signature file. The VMP sig file must be computed after code signing. 

Look at https://github.com/castlabs/electron-releases

The idea here is to manage the VMP signature in afterSign hook

Regards
Jeremie

PR #2474, opened 2018-01-15
feat(electron-updated, nsis): Custom location for downloading & installing updates

### LINKED ISSUE #2403: Nsis: Update is not installed if user restarts OS after 'update-downloaded' event.
* **Version**: 19.45.5

* **Electron-updater**: 2.16.1

* **Target**: nsis

When update is ready for installation and user just shutdowns his PC or restarts it, update is not installed. It is a very common case, as a result many our users doesn't have up-to-date application. 
Also if we check for updates after system restart, it is downloaded one more time.

One additional question: how can we get name of temp folder where update is downloaded?

MariaDima:
Close #2403


develar:
Previously downloaded file can be used only if checksum matches (specified in the update info).

MariaDima:
The rule will still apply for the cases the application hasn't terminated before the update has been installed.
The attempt is to avoid downloading again the same file and instruct the updater which file to install.
There is a risk that the file to have been modified by a malicious user and it won't pass validation again, but isn't it still possible to happen (due to the time difference between the download action and the actual installation one).

MariaDima:
@develar is there a security concern regarding the proposal?

develar:
Yes — we cannot use file as is without validation. If previously downloaded file found — checksum (actual and required specified in the update info) must be checked.

develar:
> isn't it still possible to happen (due to the time difference between the download action and the actual installation one)?

Well, strictly speaking you are right, but this time gap is minimal and issue is not so critical as in case of previously downloaded file.

MariaDima:
So if I understand correctly, the concern is about the set of the installer path for quitAndInstall, right?
If the computer gets restarted without closing the application gracefully, then this information is lost.
I could save a file containing the updateInfo at the user computer, along with the installer, but I guess it would have the same security concern.

What do you think about the option to set our own download path?
The application owner could install the file on his own if download by electron builder was successful, without worrying about the file being deleted.

develar:
1. I believe that proposed solution should be by default. By default we should download file to deterministic location. Using atomic move — download file to temporary location and then atomic move to constant location.
2. On `quitAndInstall` we should verify checksum.
3. You don't need to "save a file containing the updateInfo at the user computer". Instead, we check new version and get new update info.
4. If there is file in the cache (file name includes version), we should just use it without downloading.
5. And... for upcoming differential update we **have to store full installer somewhere**, so, upcoming electron-builder **will and must** put used installer to some location on a user machine. And updater **will and must** put downloaded installer to this location.

MariaDima:
This sounds good. The traffic for checking a new version shouldn't cause much more traffic to the servers. The devs won't have the capability of falling back to an older downloaded version if needed, but this is not supported today either.

Thanks, I'll work on the proposed changes.

develar:
>  you electron builder should set a non temp location for the copy, even if the developer hasn't set a custom download folder

Yes. And it is requirement for differential update (we need to have old full installer file).

MariaDima:
Ok, Any preference for the location?

I'm thinking that then the custom set of the location by the user might be needed after all.
The whole point was setting a path that would survive a possible windows cleanup action.

develar:
> Any preference for the location?

New electron-builder version copies installer to `"$APPDATA\${APP_INSTALLER_STORE_FILE}"` where  `APP_INSTALLER_STORE_FILE` it is ``${appInfo.productFilename}\\installer.exe``

So, updater should download file to `path.join(this.app.getPath("userData"), "new-installer.exe")` and code 

```
CopyFiles /SILENT "$EXEPATH" "$APPDATA\${APP_INSTALLER_STORE_FILE}"
```

should be modified, `${if} ${isUpdated}` then delete `installer.exe` and rename new-installer.exe to installer.exe But... because Windows sucks, you cannot rename executing file, so, maybe no need to rename and updater will check both files `installer.exe` and `new-installer.exe` Or we can add version to this file (I guess, it will be a better solution).

MariaDima:
I got a bit confused:
  Why do we need the builder to copy the installer?

I was under the impression that the updated could work isolated:
  1. Check for updates/Download an installer (first action after a system reboot, nothing have been downloaded by the active process):
    * Check if there is already a downloaded installer in the predefined folder.
    * If there is, check the checksum (SHA512 and/or SHA256).
    * If the download seems to be the one we need, don't download it again (at the moment an downloaded event is being sent, even if no download occurred).
      Otherwise proceed with the download, rename the existing installer to old-installer and move the new one to the required location (APPDATA/installer.exe).
      We could keep versions, but this would be another feature request and we might need to think about cleanups. So currently only

  2. QuitAndInstall
    * Check for an installer existence.
    * If the current process hasn't downloaded an installer, check whether the application developer has set the feed by himself to get the latest.yml data.
    * If we have the latest.yml data check the checksum (SHA512 and/or SHA256).
    * If the file is ok, then proceed with the installation.
      Otherwise return error.

What do you think?

MariaDima:
Hmmm. this wouldn't work for differential updates if there are more downloads before an installer has actually been executed.

develar:
> Why do we need the builder to copy the installer?

Because for differential download we need to have full old file.

MariaDima:
What if we save installed version in registry and name the last executed installer using this version? Then we would download new installers as many times as we want but make sure that the one that has been used, is in place.

For full downloads we will have no problem,
For differential updates, I see that versioning is already checked (I'll look closely on this). If there is incompatibility between the installer and the user's version, a full download will be required.
We have a lot of users that have older versions than the one we provide (they miss updates).

MariaDima:
Hi @develar.
What do you think about the updated commits?
Are we in sync regarding high-level approach?

I tested the builder with one application and it seems ok:
    * Installer 1 (installer.exe) is run.
    * Installer copies itself inside the app data corresponding directory.
    * Installer 2 (new-installer.exe) is downloaded under the same app data directory.
    * Application executes quitAndUpdate.
    * Installer 2 is being run.
    * Installer 2 copies itself under "installer.exe"
    * The directory contains installer.exe and new-installer.exe which the same Installer 2 version.
    * If Installer 3 gets available, it while be downloaded under "new-installer.exe" name, replacing the old downloaded file.

In general, the "new-installer.exe" after execution copies itself into "installer.exe" and remains as is (no deletion).
For a new version update, the checksum will be checked against the latest available download, before any other action.

develar:
* builder-util cannot be added  as dependency, because it is heavy. I suggest to copy-paste `copyFile` function.

MariaDima:
I added the dependency for not copying the hashfile related functionality.
I have already tested without adding the dependency, but I felt like copying too much staff.
Do you prefer I copy the related methods or maybe transfer it to another package (like builder-util-runtime that is already a dependency)?

develar:
Ok, I will do this boring thing. Thanks for PR — I will review in 2 days.

MariaDima:
Ok, I'll wait for your feedback.

I also have a concern about SHA2.
Do we have to check it or I can remove the relevant check (it will cause problems with non window machines if we leave it like this)?

MariaDima:
Hi @develar,
Any feedback on this PR? Is there anything I can work on to accelerate the procedure?

rpastro:
@develar What is the status on this PR? This is a major issue for our product as well and we are waiting anxiously for this enhancement. Thanks for your support.

develar:
Merging in progress.

> installerInfo.sha2 is deprecated. Should we even check this? sha512 is taking over

No. Sha512 is enough.

develar:
> // todo: installerInfo.sha512 is deprecated.

I fixed it. Simply pass ResolvedUpdateFileInfo here.

develar:
To make clear — we do not validate package path for Web Installer because Web Installer in any case verifies checksums.

develar:
Pushed to my fork as https://github.com/develar/electron-builder/commit/f98ae83dc796e28e446738942e59491b3481e198 Will be tested manually today and after that pushed to master and released. Please review final changes. I have removed ability to set custom cache location — I think not required for now (better to keep updater API as simple as possible).

develar:
@MariaDima Thanks a lot for your amazing commits and ideas. It makes our updater and builder more production ready and more reliable. Thank you.

MariaDima:
@develar, thanks a lot for looking into this issue!

I checked the updated code and also tested it.
I'm wondering about the following:
  If the download for some reason fails (I tested signature validation failure), then an attempt is done to remove the problematic installer. The tests on my computer (win 7 x64) had the following problem:
    The file could not be removed, but it was "locked".
    Further attempts to download an installer were failing.
    When I terminated the application the "lock" disappeared and the file was deleted.
    After restarting the application I could successfully download a new installer.
    It could be my environment, so I'll double check with another machine.

I remember that you suggested this, so just for confirmation reasons:
  If the installer haven't been downloaded by the active process the UpdateInfo, and as a consequence the downloadedUpdateHelper file, does not exist.
  Even though we have downloaded the installer (previous day for example), it cannot be installed.
  In case we know that an installer have already been downloaded, should we execute the checkForUpdates() first to feed the UpdateInfo and then execute the quitAndInstall()?

develar:
> signature validation failure

sha512, right?

develar:
Possible it is because some file is not closed (but I am sure that we close all files correctly), possibly due to bloody Windows bugs. Anyway — now we download to temp file and rename on success.

develar:
2.21.0 released as `next` version.

MariaDima:
Thanks a lot for the whole handling! This feature will resolve many headaches!

youyinnn:
any updates here?
this feature is very important

Issue #2319, opened 2017-11-21
Windows build failure with: __shim.js not under the source directory 


alexstrat:
* **Version**:  electron-builder@19.46.3 ([latest](https://yarnpkg.com/en/package/electron-builder) is 19.45.5 🤔 )

* **Target**: squirrel

On build, stack-trace:
```
Error: File "C:\Users\Alex\Documents\GitHub\browserX/__shim.js" not under the source directory "C:\Users\Alex\Documents\
GitHub\browserX"
    at getDestinationPath (C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\src\util\appFileCop
ier.ts:24:15)
    at C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\src\asar\asarUtil.ts:81:71
From previous event:
    at AsarPackager.createPackageFromFiles (C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\ou
t\asar\asarUtil.js:225:11)
    at C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\src\asar\asarUtil.ts:34:52
    at Generator.next (<anonymous>)
From previous event:
    at AsarPackager.pack (C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\out\asar\asarUtil.js
:134:11)
    at taskManager.addTask._computeFileSets.then.fileSets (C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron
-builder-lib\src\platformPackager.ts:244:136)
From previous event:
    at WinPackager.copyAppFiles (C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\src\platformP
ackager.ts:244:10)
    at C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\src\platformPackager.ts:186:10
    at Generator.next (<anonymous>)
    at runCallback (timers.js:781:20)
    at tryOnImmediate (timers.js:743:5)
    at processImmediate [as _immediateCallback] (timers.js:714:5)
From previous event:
    at WinPackager.doPack (C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\out\platformPackage
r.js:267:11)
    at C:\Users\Alex\Documents\GitHub\browserX\node_modules\electron-builder-lib\src\platformPackager.ts:115:16
[...]

PR #2476, opened 2018-01-15
Added "path.resolve()" to "const shimPath"

Megajin:
This will resolve the "__shim.js" not found issue. #2319

### DIFF
--- a/packages/electron-builder-lib/src/util/AppFileCopierHelper.ts
+++ b/packages/electron-builder-lib/src/util/AppFileCopierHelper.ts
@@ -152,7 +152,7 @@ async function compileUsingElectronCompile(mainFileSet: ResolvedFileSet, package
   })
 
   // add shim
-  const shimPath = path.resolve(`${mainFileSet.src}/${ELECTRON_COMPILE_SHIM_FILENAME}`)
+  const shimPath = `${mainFileSet.src}${path.sep}${ELECTRON_COMPILE_SHIM_FILENAME}`
   mainFileSet.files.push(shimPath)
   mainFileSet.metadata.set(shimPath, {isFile: () => true, isDirectory: () => false, isSymbolicLink: () => false} as any)
   if (mainFileSet.transformedFiles == null) {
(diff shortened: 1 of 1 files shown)


develar:
Here should be projectDir, right?

Megajin:
> Here should be projectDir, right?

Yes, the problem was that windows did not resolve the shim.js path. 
For example before the changes:
`C:\ProjectDir\App/__shim.js`

after the change:
`C:\ProjectDir\App\__shim.js`

develar:
Thanks for spotting it! 

So, better fix will be using `path.sep` instead of using `path.resolve` just to normalise file path, right?

Megajin:
> Thanks for spotting it!
> So, better fix will be using path.sep instead of using path.resolve just to normalise file path, right?

I'm not sure. Both can work obviously. However `path.sep` will only normalize the seperators, whereas `path.resolve`:

> If after processing all given path segments an absolute path has not yet been generated, the current working directory is used. The resulting path is normalized and trailing slashes are removed unless the path is resolved to the root directory. Zero-length path segments are ignored. If no path segments are passed, path.resolve() will return the absolute path of the current working directory.

Based on the docs I'd say it is safer to use `path.resolve`.

develar:
`mainFileSet.src` should be absolute path. In any case it is not correct to assume that current working dir will be equal to project dir, so, `projectDir` should be used.

Megajin:
> mainFileSet.src should be absolute path. In any case it is not correct to assume that current working dir will be equal to project dir, so, projectDir should be used.

True... so I will change this:

`  const shimPath = path.resolve(`${mainFileSet.src}/${ELECTRON_COMPILE_SHIM_FILENAME}`)`

to that:

  const shimPath = `${mainFileSet.src}${path.sep}${ELECTRON_COMPILE_SHIM_FILENAME}`

right?

develar:
Yes. It is not about "do not call path.resovle if not need", but to make code clear and avoid question "why path.resolve is used if mainFileSet.src guarantied to be absolute?"

Megajin:
> Yes. It is not about "do not call path.resovle if not need", but to make code clear and avoid question "why path.resolve is used if mainFileSet.src guarantied to be absolute?"

Done =).

develar:
Thanks!

PR #2519, opened 2018-01-24
add sha2 in updateInfo

ochukai:


develar:
For what? sha2 is deprecated and not used. sha512 is used instead.

PR #2546, opened 2018-02-02
Avoid delay due to powershell.exe waiting for input

mrodem:
On Windows 7 we are currently seeing a 30 second delay between the 100% `download-progress` event and `update-downloaded` event. We are not seeing the delay on Windows 10. Investigation reveals that the 30 seconds are spent during the `verifySignature` phase when calling powershell.exe. The process appears to be stuck waiting for input after invoking the command.

Adding `-InputFormat None` fixes the issue on Windows 7 (Powershell v2). Also tested that Windows 10 (Powershell v5) still works. The fix was found on https://serverfault.com/a/639007.


develar:
Thanks a lot!

PR #2558, opened 2018-02-06
Remove encodeUri from squirrel-windows packaging

### LINKED ISSUE #2557: Issues with spaces in exe name with Squirrel.Windows
Version: `19.49.0` (`electron-builder` and `electron-builder-squirrel-windows`).
Target: `squirrelWindows`

### Summary of impact

If the productName contains spaces, it will be replaced with `%20` in the final exe. Apart from the cosmetic issue, it also causes `electron.autoUpdater.quitAndInstall()` to break.

### Technical details

Since the update to Squirrel.Windows 1.7.x, there seems to be a discrepancy in how spaces (and probably other characters) are escaped in the `.nupkg` file. Previously Squirrel.Windows escaped the spaces with `%20`, and unescaped when extracting. With the newer version, this escaping/unescaping has been removed. However, `electron-builder-squirrel-windows` still escapes the spaces when building, which means the filenames are still escaped after extracting.

For most files I can simply choose to not use spaces. However, for the exe this is more difficult:
 * The exe name can only be controlled by setting `"productName": "My Application"`.
 * `productName` affects a whole bunch of other things, most significantly the desktop shortcut, and `Roaming/...` folder name.

Since I need spaces in my shortcut name (having "MyApplication" there is just weird), I can't feasibly remove them from the exe name because of the above.

So because of this, my application ends up as "My%20Application.exe".

If this was just a cosmetic change, it would still be ok. However, this also breaks the `electron.autoUpdater.quitAndInstall()` process (restarting the app after installing the update). Internally, this calls Squirrel's `Update.exe --processStartAndWait "My%20Application.exe"`. Squirrel tries to be smart, and replaces the `%20` with a space when it looks for the exe. Unable to find it, it crashes and shows a nasty error dialog on some machines.

My proposed fix would be to simply remove the `encodeURI` step from https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder-squirrel-windows/src/squirrelPack.ts#L245. However, I don't know enough about this to know what else this would affect.

Logs from `Update.exe` crashing:

```
2018-02-06 03:54:51> Program: Starting Squirrel Updater: --processStartAndWait My%20Application.exe
2018-02-06 03:54:51> Program: Want to launch '...\app-4.31.2\My Application.exe'
2018-02-06 03:54:51> Program: File ...\app-4.31.2\My Application.exe doesn't exist in current release
2018-02-06 03:54:51> Unhandled exception: System.ArgumentException: Value does not fall within the expected range.
   at Squirrel.Update.Program.ProcessStart(String exeName, String arguments, Boolean shouldWait)
   at Squirrel.Update.Program.executeCommandLine(String[] args)
   at Squirrel.Update.Program.main(String[] args)
```

rkistner:
Since Squirrel.Windows 1.7.x, Squirrel doesn't do URI encoding or decoding of paths anymore. Having `encodeUri` in here leaves the filenames in the escaped form after installation, which causes some issues.

Fixes #2557.

### DIFF
--- a/packages/electron-builder-squirrel-windows/src/squirrelPack.ts
+++ b/packages/electron-builder-squirrel-windows/src/squirrelPack.ts
@@ -165,7 +165,7 @@ async function pack(options: SquirrelOptions, directory: string, updateFile: str
   </metadata>
 </package>`
   debug(`Created NuSpec file:\n${nuspecContent}`)
-  archive.append(nuspecContent.replace(/\n/, "\r\n"), {name: `${encodeURI(options.name).replace(/%5B/g, "[").replace(/%5D/g, "]")}.nuspec`})
+  archive.append(nuspecContent.replace(/\n/, "\r\n"), {name: `${options.name}.nuspec`})
 
   //noinspection SpellCheckingInspection
   archive.append(`<?xml version="1.0" encoding="utf-8"?>
@@ -242,7 +242,7 @@ async function encodedZip(archive: any, dir: string, prefix: string, vendorPath:
       }
 
       // GBK file name encoding (or Non-English file name) caused a problem
-      const relativeSafeFilePath = encodeURI(file.substring(dir.length + 1).replace(/\\/g, "/")).replace(/%5B/g, "[").replace(/%5D/g, "]")
+      const relativeSafeFilePath = file.substring(dir.length + 1).replace(/\\/g, "/"))
       archive._append(file, {
         name: relativeSafeFilePath,
         prefix,
--- a/packages/electron-builder-squirrel-windows/src/squirrelPack.ts
+++ b/packages/electron-builder-squirrel-windows/src/squirrelPack.ts
@@ -242,7 +242,7 @@ async function encodedZip(archive: any, dir: string, prefix: string, vendorPath:
       }
 
       // GBK file name encoding (or Non-English file name) caused a problem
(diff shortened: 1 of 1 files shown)


develar:
Thanks a lot. Have you build your app with this change?

rkistner:
For my app I made this change directly to the built JS file (it's a lot more effort to use a forked repo).

Initial testing looks good, but I still have to test what happens when upgrading from a very old version (where Squirrel still does the unescaping) to the new build.

develar:
Ignore CI failure — I will snapshots.

rkistner:
I've confirmed that this works for my case for:
 * A fresh install.
 * Upgrading from an electron-builder 19.49.0 build.
 * Upgrading from an electron-builder 5.23.2 build.

In all those cases the exe is correctly installed as `My Application.exe` (not `My%20Application.exe`), and `electron.autoUpdater.quitAndInstall()` worked as expected.

However, I only tested the case with a space in the name - I did not test any non-ASCII characters (I'm not that familiar with how the encoding is supposed to work in those cases).

develar:
Could you please set product name to `真实光影V4低配` and test setup?

rkistner:
That does seem to work, both for clean install and on update:

image

The MSI installer does fail to build with a name like that, but I believe that's an unrelated limitation (same happens without this change).

develar:
Thanks. Fix will be applied and released tomorrow evening CET.

rkistner:
Thanks!

PR #2564, opened 2018-02-08
deleteAppDataOnUninstall is ignored if perMachine = true

riverspirit:
Noted that deleteAppDataOnUninstall doesn't delete app data if `perMachine` is set to `true`.


develar:
It should delete. If it is still doesn't work, please file issue.

Issue #1182, opened 2017-01-27
Building and Releasing using Channels


consense:
* **Version**:  11.2.4

* **Building on**:  win64

* **Target**:  win32/64

* **Aim**: Allow users to receive alpha, beta or release builds via autoupdate. The type of channel would be set at runtime (depending on user-settings returned from a remote REST api). Generic server-side provider prefered. User set to channel alpha would receive alpha, beta and release updates, beta users beta and release, release users only release.

* **Problem**: The process to accomplish the above may already be possible, but some information is spread across different ticket comments and at least for me not comprehensively understandable. In particular I am unsure whether my understanding describes the recommended workflow of electron-builder, seeing the project is (loving it) going through a lot of iterations in the last months. If you allow me to summarize my understanding of the necessary steps and accompanying questions:

1. Build 

    1. new version is pushed with `app/package.json` containing a version such as `0.12.1-alpha.1`

    2.  CI server could (in some external script) parse the version (e.g. `0.12.1-alpha.1`) from app/package.json and execute the electron-builder build cli with channel-injecting parameter `--em.build.publish.channel=alpha` . This would result in alpha.yml file being created alongside of `AppName 0.12.1-alpha.1.exe` on the CI server.
**Question: Is electron-builder itself capable of recognizing from `app/package.json` that the version contains an alpha/beta prerelease tag and determine the correct `publish.channel` parameter itself without relying on a custom script for this?**

    3. CI server uploads the yml and exe file to generic http host (e.g. using s3-deploy npm package)

    4. So S3 contains eventually a growing number of distribution versions (exe) and 3 yml files (`alpha.yml`, `beta.yml` and `release.yml`)

2. Auto-update

    1. Electron app initializes autoupdater passing in the desired release channel at runtime. e.g. `autoUpdater.setFeedURL({url: 'https://mys3bucketurl.com', channel: 'alpha'});`
 
    2. Autoupdater fetches the yml corresponding to channel property passed into `setFeedUrl` from the generic http server. 
Autoupdater then checks, whether the version from the downloaded yml file is greater than the installed version (using `semver.gt` as in https://github.com/electron-userland/electron-builder/blob/7c2973dda46434827fc43a1c330bbf9da3923053/packages/electron-updater/src/AppUpdater.ts#L156)
[...]

PR #2595, opened 2018-02-16
Doc - Release using channels

popod:
todo:
- [ ] improve documentation
- [ ] review and fix english typos

### DIFF
--- a/docs/tutorials/release-using-channels.md
+++ b/docs/tutorials/release-using-channels.md
@@ -1,28 +1,25 @@
 > !!! This documentation is in "beta" and needed to be tested !!!
 
-## Auto-updates with channels / Releasing with channels
+# Release using channels / Auto-updates with channels
 
-### Description
+## Description
+Channels are useful to distribute "beta" or "alpha" releases of your application to a choosen set of users. This allow to test an application before release it as "stable".
 
-Channels are usfull to distribute "beta" or "alpha" releases version of your application to a choosen set of your users. This allow your to test your application before release it as a "stable" version.
-
-Users which receivs "beta" version of your app will get "stable" versions too when it will be released. Otherwhise users who doesn't wanna beta version will only get "stable" releases.
+Users which receive "beta" version will get "stable" versions too. Otherwise users who doesn't wanna "beta" will only get "stable" releases.
 
 They are tree channels level ordered by stability :
-1. "stable", your application is stable and this is the default one (example version 1.3.2)
-2. "beta" which mean your application works, but should have some bugs (example version 1.3.2-beta)
-3. "alpha" which mean your application is not stable and in active development (example version 1.3.2-alpha)
-
-See more about that in xxxxx -> links with more channels informations.
+1. "stable", your application is stable and this is the default one (example: 1.3.2)
+2. "beta" which mean your application works, but should have some bugs (example: 1.3.2-beta)
+3. "alpha" which mean your application is not stable and in active development (example: 1.3.2-alpha)
 
 
-### Configuration
+## Configuration
+To release using channels, you should config electron-builder and define the channels to use in client side.
 
-#### Electron-Builder
+### Electron-Builder
+By default (without using channels), all application releases use the "stable" channel.
 
-By default (without using channels), all your application releases use the "stable" channel.
-
-If you want to use channels, you should add this to your package.json
+If you want to use channels, you should add this to your package.json:
 
 ```
 "version": "x.x.x-beta",
@@ -33,31 +30,34 @@ If you want to use channels, you should add this to your package.json
 }
 ```
 
-> Note: `allowDowngrade` is automatically set to `true` when `generateUpdatesFilesForAllChannels = true`, so you doesn't need to add it.
+> Note: `allowDowngrade` is automatically set to `true` when `generateUpdatesFilesForAllChannels = true`, so you doesn't need to set it.
 
-All you have to do to release version in a channel is to set the `package.json` version. Add "-beta" or "-alpha" (nothing for stable) to the version number to automatically build for the related channel.
+All you have to do to release using channels is to define the channel in the version tag of the `package.json`. Add "-beta" or "-alpha" (nothing for stable) to automatically build for the related channel.
 
 
-#### Your application
-
+### Your application
 All you need to do here is to define which channel the user will receive.
 
 `autoUpdater.setFeedURL({url: 'https://www.yoursite.com', channel: 'beta'});`
 
 > note: remove the need to define url !
 
-The folower related versions will be distributed to the users according to channel you set in `setReedURL()` :
+The following versions will be distributed to the users according to the channel you set in `setFeedURL()` :
 - "stable": users will only get "stable" versions
 - "beta": users will get "beta" and "stable" version
 - "alpha": users will get "alpha", "beta" and "stable" version
 
 
-### How to use it / Example
+### Advanced (optional)
+todo:
+- allow downgrade
+- 
 
-Imagine your application is in version 1.0.1 (stable).
+## How to use it / Example
+Imagine your application version is 1.0.1 (stable).
 
-If you want to release a beta for your 1.1.0 version, you only need to define your package.json "version" to "1.1.0-beta".
+If you want to release a beta for the 1.1.0 version, you only need to update the package.json "version" with "1.1.0-beta".
 
-And when your application is stable enouth and you want to release the version for all your users, only remove the "-beta" tag from your package.json "version".
+And when your application is stable enouth and you want to release it to all users, only remove the "-beta" tag from the package.json "version" tag.
 
 
--- a/docs/tutorials/release-using-channels.md
+++ b/docs/tutorials/release-using-channels.md
@@ -1,6 +1,6 @@
 > !!! This documentation is in "beta" and needed to be tested !!!
 
-# Release using channels / Auto-updates with channels
+# Release Using Channels / Auto-Updates With Channels
 
 ## Description
 Channels are useful to distribute "beta" or "alpha" releases of your application to a choosen set of users. This allow to test an application before release it as "stable".
@@ -8,9 +8,9 @@ Channels are useful to distribute "beta" or "alpha" releases of your application
 Users which receive "beta" version will get "stable" versions too. Otherwise users who doesn't wanna "beta" will only get "stable" releases.
 
 They are tree channels level ordered by stability :
-1. "stable", your application is stable and this is the default one (example: 1.3.2)
-2. "beta" which mean your application works, but should have some bugs (example: 1.3.2-beta)
-3. "alpha" which mean your application is not stable and in active development (example: 1.3.2-alpha)
+1. "stable", your application is stable and this is the default one (example: "1.3.2")
+2. "beta" which mean your application works, but should have some bugs (example: "1.3.2-beta")
+3. "alpha" which mean your application is not stable and in active development (example: "1.3.2-alpha")
 
 
 ## Configuration
@@ -35,25 +35,23 @@ If you want to use channels, you should add this to your package.json:
 All you have to do to release using channels is to define the channel in the version tag of the `package.json`. Add "-beta" or "-alpha" (nothing for stable) to automatically build for the related channel.
 
 
-### Your application
+### Your Application
 All you need to do here is to define which channel the user will receive.
 
-`autoUpdater.setFeedURL({url: 'https://www.yoursite.com', channel: 'beta'});`
+`autoUpdater.channel('beta')`
 
-> note: remove the need to define url !
-
-The following versions will be distributed to the users according to the channel you set in `setFeedURL()` :
-- "stable": users will only get "stable" versions
+The following versions will be distributed to users depending on the channel defined:
+- "stable" or nothing: users will only get "stable" versions
 - "beta": users will get "beta" and "stable" version
 - "alpha": users will get "alpha", "beta" and "stable" version
 
 
 ### Advanced (optional)
 todo:
-- allow downgrade
+- `allowDowngrade`
 - 
 
-## How to use it / Example
+## How To Use It / Example
 Imagine your application version is 1.0.1 (stable).
 
 If you want to release a beta for the 1.1.0 version, you only need to update the package.json "version" with "1.1.0-beta".
--- a/docs/tutorials/release-using-channels.md
+++ b/docs/tutorials/release-using-channels.md
@@ -46,10 +46,8 @@ The following versions will be distributed to users depending on the channel def
 
 
 ## How To Use It / Example
-Imagine your application version is 1.0.1 (stable).
+Imagine that your application is stable and in version 1.0.1.
 
-If you want to release a beta for the 1.1.0 version, you only need to update the package.json "version" with "1.1.0-beta".
+If you want to release a beta for the new 1.1.0 version, you only need to update the package.json "version" with "1.1.0-beta".
 
-And when your application is stable enough and you want to release it to all users, only remove the "-beta" tag from the package.json "version" tag.
-
-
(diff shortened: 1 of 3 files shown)


docs/tutorials/release-using-channels.md
```diff
@@ -0,0 +1,63 @@
+> !!! This documentation is in "beta" and needed to be tested !!!
+
+# Release using channels / Auto-updates with channels
```
develar: Please in the future use form  "Header Must Use Capital Case".

docs/tutorials/release-using-channels.md
```diff
@@ -0,0 +1,63 @@
...
+
+> Note: `allowDowngrade` is automatically set to `true` when `generateUpdatesFilesForAllChannels = true`, so you doesn't need to set it.
+
+All you have to do to release using channels is to define the channel in the version tag of the `package.json`. Add "-beta" or "-alpha" (nothing for stable) to automatically build for the related channel.
+
+
+### Your application
+All you need to do here is to define which channel the user will receive.
+
+`autoUpdater.setFeedURL({url: 'https://www.yoursite.com', channel: 'beta'});`
```
develar: Now user can set `channel`, `setFeedURL` not required, right?

kwent:
One scary [section](https://github.com/electron-userland/electron-builder/blob/e409c0e9a4542204ef3d16048700470edb4b02df/docs/auto-update.md#quick-setup-guide) to me was:

```
Do not call setFeedURL. electron-builder automatically creates app-update.yml file for you on build in the resources (this file is internal, you don't need to be aware of it).
```

Might want to modify this as well :)

docs/tutorials/release-using-channels.md
kwent: I'm confused. Where do we set `channel` ?

kwent: Oh this: https://github.com/electron-userland/electron-builder/blob/ba4809aeb3bd3b27f66ee684c3e6e8ab0cf55186/packages/electron-updater/src/AppUpdater.ts#L73. Needs to be documented

docs/tutorials/release-using-channels.md
popod: Thanks for the comment, I've change this.

docs/tutorials/release-using-channels.md
popod: @kwent and @develar Thanks for feedback. I've remove the "old" `setFeedURL` and replace it with `autoUpdater.channel()`. I will add the doc for `channel()` too.

docs/tutorials/release-using-channels.md
```diff
@@ -36,9 +37,9 @@ All you have to do to release using channels is to define the channel in the ver
 
 
 ### Your Application
-All you need to do here is to define which channel the user will receive.
+All you need to do here is to define which channel the user will receive with:
 
-`autoUpdater.channel('beta')`
+`autoUpdater.channel('beta')` (see [the documentation here](../auto-update.md#module_electron-updater.AppUpdater+channel))
```
kwent: I have been doing `autoUpdater.channel = 'beta'`. Is this valid as well ?

popod: @kwent my bad.. channel('beta') is false. I will correct this.

popod:
@develar and @kwent what do you think about this current "small tutorial" ? Have I miss something ?

docs/tutorials/release-using-channels.md
```diff
@@ -0,0 +1,55 @@
+<a name="release_using_channels"></a>
+# Release Using Channels / Auto-Updates With Channels
+
+## Description
+Channels are useful to distribute "beta" or "alpha" releases of your application to a chosen set of users. This allows to test an application before release it as "stable".
+
+Users which receive "beta" version will get "stable" versions too. Otherwise, users who don't want "beta" will only get "stable" releases.
+
+They are tree channels level ordered by stability:
```
rahilsondhi: typo. i think you meant to say "There are three channels, ordered by stability"

popod: Thanks ! Yes this is exactly what I'm trying to say ;)

kwent:
LGTM.

One thing i saw is if i set channel to `stable` the library was looking for `stable-mac.yml`. (I imagine same thing for windows and linux) but the `packager` is creating `latest-mac.yml`. Setting the channel to `latest` instead `stable` fixed it in my case. Not sure if the `autoupdater` or the `packager` is wrong but something is not in phase.

Might not be the best place and might be worst opening an issue instead tho. cc @develar

popod:
@kwent thanks: this is an error! The channels to use are `latest`, `beta` and `alpha`, not `stable`. I think the easy way is that I use `latest` in place of `stable` in this doc. But `stable` seems to be more intuitive..

@develar do you see an other possibility ? Like as example set the channel to `latest` both when the user set it to `stable` and `latest` ?

>(curiously you can set the channel to any string value as well...). Not sure if the autoupdater or the packager is wrong but something is not in phase.

And do you think that this is useful to limit the channel to `latest`, `beta` and `alpha` ?

kwent:
Maybe we can alias stable to latest ?

develar:
> And do you think that this is useful to limit the channel to latest, beta and alpha ?

No, user should be able to use whatever he want. But! Where do you set? updater.channel or electron-builder publish config channel?

develar:
Obviously, that you cannot set updater.channel to `stable` if you use `generateUpdatesFilesForAllChannels`. 

`stable` in our case it is `latest`. But tutorial has a lot of usages of word `stable` :) And actually, to set  "stable" channel (as tutorial suggests), you need to set it to `null` or `latest`.

popod:
@develar as I wrote in my latest comment, do you think that it is possible to do this:

When doing `autoUpdater.channel = 'stable'`, in the setter function, change `stable` by `latest` ? As is, the tutorial will be correct. I think that the word "stable" is more logic than "latest" for a "user point of view" and according to the tutorial description.

develar:
Yeach... you are right. `latest` is not clear. Because actually, `beta` will point to stable release if for now stable release is the latest.

> all application releases use the "stable" channel.

Well... no channel means that channels are not used and default name of update file is `latest` :)

Need to think about it. We cannot map `stable` to `latest` in the electron-updater, because electron-updater is not aware of `generateUpdatesFilesForAllChannels`. And electron-builder (if generateUpdatesFilesForAllChannels is set to default false) will generate `stable` update file instead of `latest`. I do not want to add this magic transformation.

Possible solution — if `generateUpdatesFilesForAllChannels` is set, `stable` will be generated instead of `latest` (and logic `getDefaultChannelName` of electron-updater will be not used because we will set channel by default explicitly to `stable`).

popod:
Thanks for your reply @develar ! I understand and agree with the fact that we should no do "magic transformation" :)

I think that the auto-updater should be "compatible" if we use `generateUpdatesFilesForAllChannels` or not. So we need to keep the `latest` label. I think the easy way to solve this problem is that I update the tutorial with `latest` in place of `stable`.

I will do it tomorrow. If you have an other solution, let me know ;)

jake:
Been loosely following this thread, and have actually got this multi-channel support working in my app. So, thanks for these WIP docs!

Isn't the feature already complete? Are we discussing improving how it works, or just documenting the way it currently works?

At the risk of getting off topic... I was definitely annoyed and disappointed that I had to use these magic variable names ("alpha" -> "beta" -> "latest"). For my workflow, I want "canary" -> "preview" -> "release". 

I think hardcoding these release channel names is a very weird idea. Not only does it limit functionality, it's less clear how they relate. If you let the user define the flow of channels themselves, they can customize whatever (and however many) make sense for their application, and it also will likely make more sense overall since they set up the relationships themselves.

There should be a new concept introduced in this release channels feature, one of channel order. I think in the build settings, defining an ordered array would be great.

Maybe, if you wanted to keep a sane default around, you could make `channels: true` expand to:

```
channels:
  - alpha
  - beta
  - latest
```

Or, if you want something else as I'd like to you could define your own array:

```
channels:
  - canary
  - preview
  - release
```

If `generateUpdatesFilesForAllChannels` is set, the same logic that exists today is followed, where anything _before_ the currently-being-built-to-channel in that array gets pushed the release as well. (e.g., if I was building "preview" it would also push to "canary").

Also, FWIW, `generateUpdatesFilesForAllChannels` is also a very confusingly named parameter. It doesn't generate updates for _all_ channels. It controls whether or not to publish the update to any channel "before" the currently built to channel.

popod:
@jake Thanks for your feedback !

>Isn't the feature already complete? Are we discussing improving how it works, or just documenting the way it currently works?

Here, I think we should only "documenting the way it currently works". But we think about improving it too :)

Auto update with channel is available since a few month and no documentation existe about it until now.. As we discuss here: https://github.com/electron-userland/electron-builder/issues/1182#issuecomment-324944370, I already have suggest to let the user define it's own "channel name". But we kept things simple for the first working implementation.

I think that we should keep the "stable, beta, alpha" channels by default to provide an "easy to use" functionality, but allow advanced users to change this..

I don't know what think @develar about that, but I think we could open a new issue for let the users chose their "channels names" and the related order. I really think this is a pretty feature !

develar:
> Isn't the feature already complete? 

Complete.

> generateUpdatesFilesForAllChannels is also a very confusingly named parameter. It doesn't generate updates for all channels

Well, this option does exactly what name means — generates _update info files_ for all channels :) As update info files is generated and uploaded, your users get updates.

@jake Do you realise and understand, that `canary` update info file will contain info about ** release**  channel update when do you release a new version of app? e.g. after canary 20.3.0 you release 20.40.30 and so, all users of `canary` or `preview ` channels must get updates from `release` channel.

> For my workflow, I want "canary" -> "preview" -> "release".

@jake Do you use semver or not? I mean — `canary` Please provide example of `canary` version number.

develar:
> I think hardcoding these release channel names is a very weird idea. Not only does it limit functionality, it's less clear how they relate. 

BTW, right now nothing stops you :) But it is your responsibility to copy/create/upload such update info files. And thanks a lot to @popod, it is not good — electron-builder should help you to avoid manual uploading/copying.

jake:
Sorry, I don't quite follow what you're asking. I also just wanted to reiterate I'm just trying to be helpful by providing some real world feedback :)

I don't use semver, as I think it's very non-human-friendly. I try to build software for humans, not robots. I understand its value in internal library world, but Electron is for end users. I'm not sure how relevant it is to this discussion, but here's our versioning strategy. Ideally the names would be what I described earlier. FWIW, Framer.js also uses a similar [whole-version-number](https://framer.com/updates/) strategy.

<details>
<summary>my non-standard versioning strategy</summary>
Per the release channel docs ([issue](https://github.com/electron-userland/electron-builder/issues/1182), [pull](https://github.com/electron-userland/electron-builder/pull/2595)), there are three release channels: Alpha, Beta, Stable.

Build artifacts have their version number (and thus their release channel) in the filename.

Once you download a release from a channel, you are subscribed to that channel. You can easily switch by downloading a release from a different channel.

Conceptually, we are working on `3.0.0` within the Alpha and Beta channels, and finally releasing it through the Stable channel. Once `3.0.0` is released on Stable we start work on `4.0.0` within the Alpha channel and repeat the workflow.

### Alpha
Rolling time-based snapshots. Version is next potential for "Stable" channel. Internal only, bleeding edge. Commonly broken. Possibly CI-driven every commit.

- `3.0.0-alpha.20180218.093433`
- `3.0.0-alpha.20180218.094454`
- `3.0.0-alpha.20180218.112235`

Note: Subscribers of Alpha also receives updates published to Beta and Stable.

### Beta
Latest snapshot promoted to Beta channel for consideration. Limited pool of external testers. *Supposed* to work and be ready for public consumption.

- `3.0.0-beta`

In the event a beta release has an issue, we issue hotfixes to the Beta channel.

- `3.0.0-beta.1`
- `3.0.0-beta.2`
- `3.0.0-beta.3`

Note: we'd also be issuing additional releases on the Snapshot channel, e.g. `3.0.0-snapshot.TIMESTAMP`, and once we solved the issue issuing new Beta channel releases.

Note: Subscribers of Beta also receives updates published to Stable.

Consider: TIMESTAMP on hotfix releases?

- `3.0.0-beta.20180218.112424`
- `3.0.0-beta.20180218.125310`
- `3.0.0-beta.20180219.152239`

### Stable

The publicly consumed versions. Guaranteed to work, as they passed through Snapshot and Beta channels. Only ever whole major versions with zeroed minor and patch versions.

- `1.0.0`
- `2.0.0`
- `3.0.0`

Note: Subscribers of Stable only receive updates to Stable.
</details>

---

In your example, is it not still pushing to both the currently selected channel and any channels it supersedes?

- If I'm building `latest` it pushes to `latest`, `beta`, and `alpha`
- If I'm building `beta`, it pushes to `beta` and `alpha`
- If I'm building `alpha`, it pushes to `alpha`

Isn't this always the case? Release channels cascade down. You never push a channel release "upstream"-- you wouldn't build an `alpha` release and push it to `latest`. However, you'd want people subscribed to a "downstream" channel to get anything upstream.

---

> Well, this option does exactly what name means — generates update info files for all channels :) As update info files is generated and uploaded, your users get updates.

How so? I have this option enabled, and when I publish an "alpha" build my "latest" release file is not regenerated, or even re-uploaded as far as I can tell. Doesn't this option control the feature that pushes releases to downstream channels in addition to the selected channel?

> right now nothing stops you

Are you proposing I fork `electron-builder` and change those hardcoded values?

develar:
> Are you proposing I fork electron-builder and change those hardcoded values?

These values are not harcoded — you can use any channel name as you want (but in this case electron-builder cannot help you to upload/regenerate corresponding update info files).

> when I publish an "alpha" build my "latest" release file is not regenerated, or even re-uploaded as far as I can tell.

Yes. Because `latest` it is stable, and as you build `alpha`, only `alpha` is regenerated and uploaded.

Ok... so, as far I see from your example versions, you use pre-release version tags `alpha` and `beta`. But... you name (want to think in these terms, want to name it so) channels as `canary` and `preview, right?

jake:
> These values are not harcoded — you can use any channel name as you want (but in this case electron-builder cannot help you to upload/regenerate corresponding update info files).

Oh, I see. I guess I'm somewhat conflating arbitrary release channels with the "sorted" release channels as I've been describing (e.g. canary -> preview -> stable).

The `alpha` -> `beta` -> `latest` release hierarchy _is_ hardcoded though, right?

> you use pre-release version tags alpha and beta. But... you name (want to think in these terms, want to name it so) channels as canary and preview, right?

Pretty much! I'm fighting against `npm`'s  annoyingly-hardcoded semver requirements. If I was just setting up a system without any limitations, I'd just use e.g. version `v3-canary` for the "canary" channel, `v3-preview` for "preview", and `v3` for "release".

Anyway, if I can be of any help testing the future fully-customizable release channel hierarchy I'm game!

jake:
Oh, just to clarify:

Have "latest" **always** mean "ready for production" _unless you've defined your own release channel list_. In which case, the first entry in your list represents "ready for production", and it's whatever name you've chosen for it.

develar:
> Then it's easy to do channel[0] to find the most-stable

Hmm... currently, we detect it using prerelease component of current app version. If no pre release component — only Most-stable channel file will be generated.

kwent:
@popod @develar I think we need to fix the `stable` / `latest` ambiguous and this should be good to merge.

popod:
@develar @kwent I've change the documentation by using "latest" in place of "stable". I think this is clear enough to deploy.

kwent:
LGTM

develar:
@popod Thanks a lot. The only reason why this awesome addition is not yet merged — I want to do some changes to address @jake suggestions.

JsseL:
For those looking for a **[link to the docs](https://www.electron.build/tutorials/release-using-channels)**.

PR #2622, opened 2018-02-21
Fix 2381

### LINKED ISSUE #2381: nsis installer removes existing app links in start menu subdirectory
`electron-builder@19.45.5`
`platform: Windows 10`

nsis config:
```yml
nsis:
  guid: XXXXX-XXXXX-XXXXX-XXXXX
  multiLanguageInstaller: true
  oneClick: false
  perMachine: true
  allowToChangeInstallationDirectory: true
  menuCategory: true
```

We have multiple applications all added to the start menu under a `Company Name` directory. When installing our Electron app, the contents of this directory are getting removed and after we are only left with a link to our Electron app.

I know very little about nsis scripts so I'm guessing it has something to do with [cleanupOldMenuDirectory or keepShortcuts](https://github.com/electron-userland/electron-builder/blob/541605ff2038c1b9485d8f0a3c88167f61e80b8d/packages/electron-builder-lib/templates/nsis/include/installer.nsh#L181) but I haven't managed to get my head around the logic yet.

timfish:
This should fix #2381

### DIFF
--- a/packages/electron-builder-lib/templates/nsis/uninstaller.nsh
+++ b/packages/electron-builder-lib/templates/nsis/uninstaller.nsh
@@ -42,7 +42,7 @@ Section "un.install"
       ${if} $R1 == ""
         Delete "$oldStartMenuLink"
       ${else}
-        RMDir /r "$SMPROGRAMS\$R1"
+        RMDir "$SMPROGRAMS\$R1"
       ${endIf}
     !endif
   ${endIf}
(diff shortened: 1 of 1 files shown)


timfish:
Actually, that change only stops the start menu directory from being deleted. 

It doesn't remove the old shortcut link. That will take some more work.

develar:
20.0.9 released.

Issue #2677, opened 2018-03-09
Bintray publishing stopped working


johannesjo:
* **Version**: 20.4.0
* **Target**: Linux (deb)

A bintray configuration that worked fine before now throws a 404 when trying to connect to the api. The build doesn't fail however. 

Full log: https://api.travis-ci.org/v3/job/351081787/log.txt

johannesjo:
Looks to me as the url used by electron builder is wrong. According to the info on the bintray package page the url should be something like this `https://api.bintray.com/content/...` electron builder tries to access `https://api.bintray.com/packages/...` instead.

develar:
Our Bintray test passed. 

`content` is used only for uploading content. https://github.com/electron-userland/electron-builder/blob/master/packages/electron-publish/src/BintrayPublisher.ts#L64

Please ensure that your API token is correct, and username / repo name is correct.

johannesjo:
Seems like this was an issue on bintrays end. Without changing anything the deployment started to work again for my latest build.

develar:
Yes, Bintray is not reliable.

PR #2684, opened 2018-03-11
fix(deployment): correct bintray api path

johannesjo:
As discussed in #2677 attempts to fix the publishing issue with bintray.


develar:
Thanks, but please see https://github.com/electron-userland/electron-builder/issues/2677#issuecomment-372921551

Issue #2685, opened 2018-03-12
Path must be a string. Received true.


jbreckmckye:
* **Version**: 20.4.1 (also 20.2.0)

* **Target**: Mac OS X

## Details

When I try to run my Electron build, I get a fatal exception, and no executable is outputted.

This is the exception:

```
  • electron-builder version=20.4.1
  • loaded configuration file=package.json ("build" field)
  • writing effective config file=dist/electron-builder.yaml
  • Unresolved node modules: @tweenjs/tween.js, browserify-css, deep-assign, envify, load-bmfont, object-assign, present, promise-polyfill, style-attr, three-bmfont-text, webvr-polyfill
  • no native production dependencies
  • packaging       platform=darwin arch=x64 electron=1.8.2 appOutDir=dist/mac

TypeError: Path must be a string. Received true
    at assertPath (path.js:28:11)
    at Object.relative (path.js:1267:5)
    at AppFileWalker.handleSymlink (/Users/jimmybreck-mckye/Projects/navigation-wrangler/node_modules/electron-builder-lib/src/util/NodeModuleCopyHelper.ts:34:23)
    at /Users/jimmybreck-mckye/Projects/navigation-wrangler/node_modules/electron-builder-lib/src/util/NodeModuleCopyHelper.ts:63:24
    at Generator.next (<anonymous>)
From previous event:
    at AppFileWalker.collectNodeModules (/Users/jimmybreck-mckye/Projects/navigation-wrangler/node_modules/electron-builder-lib/out/util/NodeModuleCopyHelper.js:161:11)
    at then.it (/Users/jimmybreck-mckye/Projects/navigation-wrangler/node_modules/electron-builder-lib/src/util/AppFileWalker.ts:54:21)
    at runCallback (timers.js:785:20)
    at tryOnImmediate (timers.js:747:5)
    at processImmediate [as _immediateCallback] (timers.js:718:5)
From previous event:
    at AppFileWalker.handleNodeModulesDir (/Users/jimmybreck-mckye/Projects/navigation-wrangler/node_modules/electron-builder-lib/src/util/AppFileWalker.ts:50:8)
    at AppFileWalker.consume (/Users/jimmybreck-mckye/Projects/navigation-wrangler/node_modules/electron-builder-lib/src/util/AppFileWalker.ts:32:21)
    at /Users/jimmybreck-mckye/Projects/navigation-wrangler/node_modules/builder-util/src/fs.ts:77:69
From previous event:
[...]

PR #2690, opened 2018-03-13
Fix: handle dependencies with package 'link' fields

jbreckmckye:
Linked bug: #2685 

Some projects, like Mozilla Aframe, set a 'link' property at the root of their package.json file.
This causes problems for our packageDependencies job, which extends that object and adds 'link'
to signify the dependency dir is a symbolic link.

This PR makes sure any existing `link` in package.json is always overwritten with the appropriate value.


develar:
Thanks! 20.5.1 released.

Issue #2701, opened 2018-03-15
 Signing of windows executable fails if proxy is required on osx


johvo:
* **Version**: "electron-builder": "^19.47.1",

* **Target**: Build for Windows (on OSX)

* **Issue**: I'm behind a proxy and it seems the proxy settings are not used while signing with windows executable (when build on a osx machine).
The used osslsigncode tool has an option "-p my_proxy_url" which seems to be unused by electron-builder.

* **Error-message**:

 `>> gulp dist
[10:19:19] Using gulpfile ~/dev/__ext/bagga.plugin.collab/gulpfile.js
[10:19:19] Starting 'clean'...
[10:19:19] Starting 'compile-menubar'...
[10:19:21] Finished 'clean' after 2.51 s
[10:19:22] Version: webpack 3.8.1
     Asset     Size  Chunks                    Chunk Names
menubar.js   919 kB       0  [emitted]  [big]  main
 style.css  3.46 kB       0  [emitted]         main
[10:19:22] Finished 'compile-menubar' after 2.94 s
[10:19:22] Starting 'compile'...
[10:19:22] Finished 'compile' after 5.08 ms
[10:19:22] Starting 'dist'...
electron-builder 19.47.1
Rebuilding native production dependencies for win32:x64
Packaging for win32 x64 using electron 1.7.9 to dist/win-unpacked
Signing Plugin.exe (certificate file: "/Users/tiv/dev/Codesinging-Cert_full.pfx")
[10:20:48] 'dist' errored after 1.43 min
[10:20:48] Error: Exit code: 255. Command failed: /Users/tiv/Library/Caches/electron-builder/winCodeSign/winCodeSign-1.9.0/darwin/10.12/osslsigncode -in /Users/tiv/dev/__ext/bagga.plugin.collab/dist/win-unpacked/Plugin.exe -out /Users/tiv/dev/__ext/bagga.plugin.collab/dist/win-unpacked/Plugin-signed-sha1.exe -t http://timestamp.verisign.com/scripts/timstamp.dll -pkcs12 /Users/tiv/dev/Codesinging-Cert_full.pfx -h sha1 -n Plugin -pass ******************* (sha256 hash) failure: Couldn't connect to server
authenticode timestamping failed

Failed

CURL failure: Couldn't connect to server
authenticode timestamping failed

Failed

    at /Users/tiv/dev/__ext/bagga.plugin.collab/node_modules/builder-util/src/util.ts:109:16
[...]

PR #2702, opened 2018-03-15
fix(osx): Code signing might need proxy to reach internet (#2701)

johvo:


develar:
Thanks, makes sense. One note — `-p` not available for signtool.

PR #2720, opened 2018-03-20
fix(docker): add missing libgconf-2-4 for chromedriver

yuya-oc:
chromedriver couldn't start because `libgconf-2-4` was missing.
So Spectron also couldn't start Electron apps for testing.

```
# ./node_modules/.bin/chromedriver
/project/node_modules/electron-chromedriver/bin/chromedriver: error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory
```


probonopd:
Maybe related to https://github.com/electron-userland/electron-builder/issues/2732

Issue #2640, opened 2018-02-25
Auto updater fails to launch UAC update on full download fallback


jonzlin95:
* **Version**: 19.55.2

* Electron Updater Version: 2.20.1

* **Target**: Windows (32 & 64bit)

Hi friends,

We're using the auto-update feature of the electron-builder, and 90%+ of the time it works great. However for some users, they run into

```
[2018-02-24 20:47:35.078] [error] Cannot download differentially, fallback to full download: Error: EPERM: operation not permitted,
```

which triggers a full download. This works fine (leads to)

```
[2018-02-24 21:24:38.746] [info] New version 0.5.0 has been downloaded to C:\Users\Username\AppData\Local\Temp\up-7jJaOf\AppSetup.exe
```

But upon an attempt to quitAndInstall

```
[2018-02-24 21:24:39.747] [info] Install on explicit quitAndInstall
[2018-02-24 21:24:39.747] [info] Install: isSilent: false, isRunAfter: true
[2018-02-24 21:24:39.761] [info] Access denied or UNKNOWN error code on spawn, will be executed again using elevate
```

The final line fails, and so the user is unable to update. Upon restarting the app they are stuck in an update loop, rendering the app unusable until a fresh install of the latest version. 

Our current fix is to tell users to run as admin, but ideally the UAC would prompt and allow users to opt into admin.

Here is our relevant package.json lines

```
    "nsis": {
      "oneClick": true,
      "include": "build/installer.nsh", // Simply checks if VC++ is installed, and attempts an install if not.
      "perMachine": true
    },
```

Let me know if any more information would be useful for debugging.

Thanks,
Jonathan

afroewis:
One single client has the same problem. He has the latest version installed, yet it will always download the latest version, so he is in an update loop. I just told him to run as an administrator and will get the results soon. However, interestingly, the setup *does* replace the files (The "Created date" of the files changes).

jonzlin95:
After a few more days of debugging this issue, we've made some new/further? discoveries

1. Running the application as Administrator *successfully* auto-updates to the next version
2. On a failed installation what happens is 
   1. We see black cmd window flash (what I assume is the elevate.exe call)
   2. Old app is uninstalled successfully
   3. Everything else fails and user is left without the application.

We tried adding some custom nsis, but it didn't solve the problem.
```
!macro customInit
    ${IfNot} ${UAC_IsAdmin}
        ShowWindow $HWNDPARENT ${SW_HIDE}
[...]

Issue #2363, opened 2017-12-06
NSIS all users installation auto-update


Egorikhin:
* **electron-builder version**: 
19.47.1
* **electron-updater version**
2.16.3

* **Target**: 
win

I'm facing the issue with updating application installed to the Program Files folder. 
I have `oneClick: false` in config and i'm installing updates with:
```
        setImmediate(() => {
            app.removeAllListeners("window-all-closed");
            autoUpdater.quitAndInstall(true, true);
        });
``` 
So when I'm trying to update application running under not admin account, it show UAC notificaton to enable old-uninstaller.exe execution. After I click on Yes it successfully deletes everything from application folder in Program Files but not installs new one. Altough, when I'm trying to update it under admin account it will be updated successfully.

Maybe it's relevant to this [issue](https://github.com/electron-userland/electron-builder/issues/1106#issuecomment-274787783).
Could you please take a look on it or maybe suggest where should i look?

Egorikhin:
Ok, figured out myself.
You can use this customInit script in order to make it work as expected:
```
!macro customInit
    ${if} $installMode == "all"
        ${IfNot} ${UAC_IsAdmin}
            ShowWindow $HWNDPARENT ${SW_HIDE}
            !insertmacro UAC_RunElevated
            Quit
        ${endif}
    ${endif}
!macroend
```

develar:
Better to prepare PR. Anyway, thanks for solution.

Egorikhin:
I don't know if it will be the correct code to fix in the PR, so i'll try to ask here

In the `multiUser.nsh` on line 45 there is the following code:
```
  !macro setInstallModePerAllUsers
    StrCpy $installMode all
    SetShellVarContext all

    **!ifdef BUILD_UNINSTALLER**
      ${IfNot} ${UAC_IsAdmin}
        ShowWindow $HWNDPARENT ${SW_HIDE}
        !insertmacro UAC_RunElevated
        Quit
      ${endif}
    **!endif**
```
If we delete the bold !ifdef, it will work as expected, but I'm not sure if it's correct to delete this.
Could you please assist me on this?

develar:
Cannot answer right now, UAC is quite complicated, need to deep reinvestigate. Could you please provide your full nsis config?

Egorikhin:
config:
```
    "nsis": {
        "oneClick": false,
        "allowToChangeInstallationDirectory": true,
        "include": "./configuration/zzz/include.nsh",
[...]

PR #2739, opened 2018-03-25
Allow --package-file arg to escape spaces in filenames

jonzlin95:
Right now package file doesn't escape spaces in the filename, which causes a bug where the package file passed to the nsis installer to be incorrect. This means that when the installer runs, it will fail on folders such as 

`C:\Users\first last\appdata\roaming\package-version.7z`

This issue only seems to occur upon a failed Differential update (due to perMachine) followed by a full download to the users' appdata folder.

This fixes it by forcing the whole file path to be a single arg. I believe this also fixes the following issue

#2640 

I'm unsure if #2363 is the same issue, but I see similar symptoms

### DIFF
--- a/packages/electron-updater/src/NsisUpdater.ts
+++ b/packages/electron-updater/src/NsisUpdater.ts
@@ -106,7 +106,7 @@ export class NsisUpdater extends BaseUpdater {
     const packagePath = this.downloadedUpdateHelper.packageFile
     if (packagePath != null) {
       // only = form is supported
-      args.push(`--package-file=${packagePath}`)
+      args.push(`--package-file="${packagePath}"`)
     }
 
     const spawnOptions = {


develar:
I doubt that it fixed mentioned issues — fix doesn't related to permissions. Anyway — thanks a lot, it may be a part of existing issues.

jonzlin95:
It did fix the first one, because it was a misdiagnose on my end. I can look into fixing other ones with similar issues since I’ve gone through all the Nsis code and also through most of the update code

jonzlin95:
@develar -> I was investigating this some more, and I noticed the permissions issue is related somehow to the oneClick setting. Haven't had lot of time to kind of dig through outside of the day job, but the fix above only worked on assistedInstaller, but not on oneClick. Will keep updates on this thread.

develar:
Probably you are right — we pass args to elevate and so.. yeach. Thanks a lot.

PR #2750, opened 2018-03-27
feat(electron-updater): extensible sources

benjamincburns:
With this change, users can register their own provider factory
functions. This allows users to write their own auto update providers,
or to extend the ones provided by this library. To do this, a user calls
`providerFactory.registerProviderType` passing a name for the provider
type as a string, and a function which accepts a `data` argument of type
`PublishConfiguration | AllPublishOptions` and an instance of
`AppUpdater`, and returns an instance of `Provider<any>`.

### DIFF
--- a/packages/electron-updater/src/providerFactory.ts
+++ b/packages/electron-updater/src/providerFactory.ts
@@ -1,45 +1,77 @@
 import { AllPublishOptions, BaseS3Options, BintrayOptions, GenericServerOptions, getS3LikeProviderBaseUrl, GithubOptions, newError, PublishConfiguration } from "builder-util-runtime"
+import { Provider } from "./main"
 import { AppUpdater } from "./AppUpdater"
 import { BintrayProvider } from "./BintrayProvider"
 import { GenericProvider } from "./GenericProvider"
 import { GitHubProvider } from "./GitHubProvider"
 import { PrivateGitHubProvider } from "./PrivateGitHubProvider"
 
-export function createClient(data: PublishConfiguration | AllPublishOptions, updater: AppUpdater) {
+export type ProviderFactoryFunction = (data: PublishConfiguration | AllPublishOptions, updater: AppUpdater) => Provider<any>
+
+interface ProviderFactoryList {
+  [key: string]: ProviderFactoryFunction
+}
+
+const providerList: ProviderFactoryList = {
+  github: createGitHubProvider,
+  spaces: createSpacesOrS3Provider,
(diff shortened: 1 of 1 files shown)


benjamincburns:
@develar I'm offering up this change as hopefully a less contentious alternative to #2743, though I'd still hope we can find a way for both PRs to be acceptable.

develar:
That PR is not closed, so, let's continue discussion. So, I will not review this PR in this case until that PR is not resolved.

develar:
Closed as  #2743 is merged now. Feel free to reopen if need.

benjamincburns:
@develar I suggest reopening this (I have no button available to me to do so myself). This is a separate feature which lets others implement their own update providers.

packages/electron-updater/src/providerFactory.ts
```diff
@@ -1,45 +1,77 @@
...
-    default:
-      throw newError(`Unsupported provider: ${provider}`, "ERR_UPDATER_UNSUPPORTED_PROVIDER")
+  const providerFactoryFunction = providerList[data.provider]
+  if (!providerFactoryFunction) {
+    throw newError(`Unsupported provider: ${data.provider}`, "ERR_UPDATER_UNSUPPORTED_PROVIDER")
+  }
+  return providerFactoryFunction(data, updater)
+}
+
+export function registerProviderType(name: string, factoryMethod: ProviderFactoryFunction) {
```
benjamincburns: The extension point is here. See type `ProviderFactoryFunction` for the required function signature.

develar:
Do you need this functionality to be merged?

samuelmaddock:
@develar I was considering looking into auto-update support for [Dat protocol](https://www.datprotocol.com/). Would you consider merging this functionality in? It would be helpful for my use case.

develar:
@samuelmaddock For support Dat protocol you don't need this PR — you can change electron-updater directly and so, your code will be maintained as part of this project. Simply implement your provider as any other existing providers  and send PR :)

samuelmaddock:
@develar One thing I'd be concerned about is adding a lot of extra npm dependencies. Dat protocol isn't http so it would require a few extra modules.

develar:
Ok, in this case it makes sense.

benjamincburns:
This PR has been open for quite some time (over 2.5 years) and the maintainer hasn't accepted it. I have little motivation to update this given that it's been so long and I've moved on to other projects, so I'm closing it.

PR #2754, opened 2018-03-27
Updater: note about dev-app-update.yml

gryzzly:
Adds a note about `dev-app-update.yml` file that is required for working against user triggered update flow without packaging.


develar:
Thanks!

PR #2777, opened 2018-04-06
Fix docker link

stefansl:


develar:
It is docs sources, actual docs on https://www.electron.build

Issue #2166, opened 2017-10-07
Add a configuration option to specify the location of a provisioning profile


jwheare:
Following on from #2142:

When building for the `mas-dev` target, by default, a provisioning profile in the working directory is automatically used for code signing by `electron-osx-sign`. It would be good to be able to pass a CLI arg or set a config file parameter to specify a location for the file and pass it on with e.g.

`electron-osx-sign path/to/my.app --provisioning-profile=path/to/my.provisionprofile`

or using the `provisioning-profile` option in calls to `require('electron-osx-sign').sign/signAsync`

holgersindbaek:
Is anyone working on this? Would it be possible to specify a provisioning profile in the settings?

Something like this:

```
"build": {
    "appId": "com.copany.someapp",
    "productName": "someapp",
    "buildVersion": "1.0.11",
    "mac": {
      "target": [
        "mas-dev"
      ],
      "category": "public.app-category.games",
      "icon": "src/electron/assets/store.icns",
      "entitlements": "src/electron/entitlements.mac.plist",
      "type": "development"
    },
    "mas-dev": {
      "category": "public.app-category.games",
      "entitlements": "src/electron/entitlements.mas.plist",
      "type": "distribution",
      "provisioning-profile": "src/electron/mas-dev.provisioningProfile"
    },
    "files": [
      "src/electron.js",
      "src/electron/**/*",
      "package.json"
    ],
  },
```

AllSeeingEye:
I've sent a PR for this feature, please test 

`provisioningProfile` field can be assigned for `dmg`, `mas` and `mas-dev` target options.

mcfedr:
As I understand this is needed not only for `mas` for but for mac builds that are signed as well.

I'm currently using a bit of code in `afterPack` that copies the correct provisionprofile according to mac/mas.

holgersindbaek:
@mcfedr Can you share that piece og code?

mcfedr:
Something like this..

```javascript
let profile = {
    'mas': 'signing/mac_dist.provisionprofile',
    'darwin': 'signing/dev_id_dist.provisionprofile',
  }

  electronBuilder.build({
    ...,
    afterPack: ({ appOutDir, packager, electronPlatformName }) => {
          console.log('appOutDir', appOutDir, electronPlatformName)
          return Promise.resolve()
            .then(() => {
              if (packager.platform.name !== 'mac') {
                return
              }
              return new Promise((resolve, reject) => {
[...]

PR #2782, opened 2018-04-10
feat(mac): pass the provisioning profile path to electron-osx-sign

AllSeeingEye:
Implements #2166.

`provisioningProfile` field can now be passed on the `dmg`, `mas` and `mas-dev` target options;
field should contain a path - absolute or relative to the app root - to the respective provisioning profile.

### DIFF
--- a/packages/electron-builder-lib/src/macPackager.ts
+++ b/packages/electron-builder-lib/src/macPackager.ts
@@ -84,9 +84,10 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     const prepackaged = this.packagerOptions.prepackaged
 
     if (!hasMas || targets.length > 1) {
+      const hasDmg = targets.length !== 0 && targets.some(it => it.name === "dmg")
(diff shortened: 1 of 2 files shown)


AllSeeingEye:
@develar We need the `provisioningProfile` option for `dmg` because developer may want to sign a stand-alone app - and stand-alone apps have different profiles than App Store ones.

(I have this very situation, and tested both dmg and mac options)

develar:
> stand-alone app - and stand-alone apps have different profiles than App Store ones.

Well, for AppStore you define in the `mas`, for non store in the `mac`. For what do you need in the DMG?

AllSeeingEye:
@develar ok, this would make sense. I'll amend my PR.

AllSeeingEye:
@develar This feature has been reworked; `provisioningProfile` option now works the same way as `entitlements` option does:

> Well, for AppStore you define in the `mas`, for non store in the `mac`.

develar:
Thanks! 20.9.0

PR #2850, opened 2018-04-30
Prevent download notification queueing

caseywebdev:
The current implementation using `this.signals.updateDownloaded` is creating an event listener that is never removed. The problem with this is that if you call `checkForUpdatesAndNotify`, no update is found, then call `checkForUpdatesAndNotify` again an hour later and an update *is* found, the notification in this method is actually created twice. This PR updates the method to notify the user when an update is available and the download is successful instead of binding to the `updateDownloaded` event.


develar:
Thanks! Please try 2.21.9

PR #2858, opened 2018-05-02
Check for package file existence before updating

jonzlin95:
Fixes a bug where a user can have their app uninstall on an update.

Right now it's possible for the package path to not exist, but be passed in anyway. One way to set up application as "assistedInstaller: true, perMachine: true". Then upon the elevate.exe being called on the first attempt at updating, to press "No". Then restart the application and press "Yes" on the next elevate when updating. This only happens in production - I was unable to reproduce this bug in development

If you check the path of the package, there isn't a package, so the installer tries to run on an empty file, but fails causing an uninstall for the user (since that part of the NSIS code will run).

There may be a more elegant solution, and since there are steps to reproduce it isn't too bad.

### DIFF
--- a/packages/electron-updater/src/NsisUpdater.ts
+++ b/packages/electron-updater/src/NsisUpdater.ts
@@ -9,6 +9,7 @@ import { GenericDifferentialDownloader } from "./differentialDownloader/GenericD
 import { newUrlFromBase, ResolvedUpdateFileInfo } from "./main"
 import { findFile, Provider } from "./Provider"
 import { unlink } from "fs-extra-p"
+import { pathExistsSync } from "fs-extra"
 import { verifySignature } from "./windowsExecutableCodeSignatureVerifier"
 
 export class NsisUpdater extends BaseUpdater {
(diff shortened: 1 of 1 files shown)


develar:
Please try electron-updater 2.21.10

I guess real fix — https://github.com/electron-userland/electron-builder/commit/7fe72da

jonzlin95:
Thanks for the quick look @develar. I'll try this fix and see if it fixes the bug I'm facing and report back!

develar:
Closed for now.

jonzlin95:
@develar sorry been a bit swamped recently, but got some time to get back to this bug. Your fix w/ _packageFile = null in clear() doesn't work, will explain below.

Right now the (auto)-update flow is as follows (my understanding of the code)

Check latest.yml on server -> If new version true -> Download packageFile + Installer -> On completion trigger update.

On update trigger -> Check for "installer" existence. -> If true -> Run the installer, with --packageFile packageFile. 

Issue with this flow is that if for some reason, the packageFile failed to download (computer restarted during the downloads, some sort of connection error etc.), but the installer didn't, it will cause a failure.  (This is easy to happen since installer is ~1MB, packageFile is 10-100's of MB's). Just changing packageFile to null on clear() unfortunately doesn't fix this, because I've seen done() be called while packageFile is empty. 

I think the only way to know is to check for existence of packageFile as well as installer, but I'm not sure where the code for this is best to be put though. 

Thanks for all the hard work you put into the library! Let me know if there's anything else I can tell you that might help w/ this bug.

Issue #2860, opened 2018-05-02
Add support for current user file associations on Windows


nikku:
* **Version**: 20.11.1

* **Target**: Windows

Per-user file associations on Windows are possible without inferring with UAC. This is something I've implemented a while ago as part of [an electron apps behavior](https://github.com/camunda/camunda-modeler/blob/master/app/lib/platform/windows/file-associations.js). 

The solution is to alter [`FileAssociation.nsh`](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder-lib/templates/nsis/include/FileAssociation.nsh) to write the relevant registry keys to either `HKEY_LOCAL_MACHINE\Software\Classes` or to `HKEY_CURRENT_USER\Software\Classes` depending on whether the installer operates in `all` or `current` user mode.
 
`HKEY_CLASSES_ROOT` is simply a unified view on both parts of the registry tree (cf. [this post](https://www.itninja.com/question/hkey-local-machine-software-classes-and-hkey-classes-root)).

To ease the implementation NSIS comes with a built in switch by using `SHELL_CONTEXT` to either point to the per-user or per-machine registry root (cf. [`WriteRegExpandStr` documentation](http://nsis.sourceforge.net/Docs/Chapter4.html#writeregexpandstr)).

Issue #772, opened 2016-09-22
Must be error if file association is set, but perMachine not


virror:
- **Version**: 7.3.0

- **Target**: win32

After updating to latest electron-builder, file association including icon stopped working. Worked perfectly before updating, i think last version was the last 6.x, but cant remember exactly.
Using NSIS
"fileAssociations": {
      "ext": ".aepwx",
      "name": "myName",
      "icon": "build/icon.ico"
    }

develar:
Works for me. Nothing changed. Please provide sample project or set debug env `DEBUG=electron-builder` and attach logs.

> "icon": "build/icon.ico"

Do not specify app icon — it is default value in any case.

virror:
Sadly i cant send you the project, but im willing to help in other ways. We have tested on several machines now and non of them works. How do you set the debug variable?

develar:
> We have tested on several machines now and non of them works

Please try to not set icon at all. Since it is equal to app icon, no need to set it explicitly.

> How do you set the debug variable?

It is an environment variable. Set `DEBUG=electron-builder`

develar:
> non of them works.

No icon? Could you please attach screenshot of file in the File Explorer?

virror:
Yes, app does not open on double click and icon is just a white standard "no specific icon" icon.
When i check properties of the file it says that no program is associated to this file type.
I will try with the debug mode and see if i can get some data.
Were do the log file end up?

develar:
Do you use `perMachine: true`?

virror:
No.
"nsis": {
      "oneClick": true,
      "perMachine": false,
      "installerHeaderIcon": "build/icon.ico"
    },

develar:
Because Windows is *** Windows, per-user installed program cannot register file associations.

Yes — it is not documented. Yes — we must throw error if file associations set, but `perMachine` (or boring installer) not.

virror:
But you just said that "per-user installed program cannot register file associations", doesn't that mean that you should set error if file associations set and perMachine is TRUE? For my case its set to false?

develar:
> doesn't that mean that you should set error if file associations set and perMachine is TRUE?

`perMachine: true` means that app will be installer for all users. Ugly *** UAC in action — user cannot install app only per-user — admin rights is required to install.

> For my case its set to false?

`false` is the default value.

virror:
Ahh, i thought perMachine true was that it should install for one user only, i will try change this then : )

[...]

PR #2861, opened 2018-05-02
feat(nsis): implement file associations per user

nikku:
By writing the relevant entries either to

    HKEY_LOCAL_MACHINE\Software\Classes

or to

    HKEY_CURRENT_USER\Software\Classes

we're able to support file associations on user installations, too.

Closes https://github.com/electron-userland/electron-builder/issues/2860.

### DIFF
--- a/packages/electron-builder-lib/src/targets/nsis/NsisTarget.ts
+++ b/packages/electron-builder-lib/src/targets/nsis/NsisTarget.ts
@@ -554,10 +554,6 @@ export class NsisTarget extends Target {
 
     const fileAssociations = packager.fileAssociations
     if (fileAssociations.length !== 0) {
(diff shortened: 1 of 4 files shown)


nikku:
Could you give me a few pointers where to write tests for that feature?

nikku:
I've removed the assertion testing for `perMachine`, as this effectively fixes https://github.com/electron-userland/electron-builder/issues/772.

develar:
@clusterberries Hi, it seems per-user is possible :)

develar:
Ouch, it really works. @nikku You are hero.

develar:
> where to write tests for that feature?

existing test is enough, for now not possible to automate this check (checked manually that association works).

Issue #2872, opened 2018-05-07
Registering fileAssociation and custom mime type on AppImage install


AdrianoFerrari:
* **Version**: 20.11.1

* **Target**: AppImage

I'm trying to register a custom mime-type along with my `fileAssociation`. This seems to be the only way to actually get Linux to recognize a new custom file type to associate with an application.

It *seems* like adding `--add-mime-type=application/x.mycustommime` flag to `desktop-file-install` [here](https://github.com/electron-userland/electron-builder/blob/9dced9ceaf245a86f0ef33df373f3fd44871e80a/packages/electron-builder-lib/templates/linux/AppRun.sh#L170) is one of the two necessary steps.

The second is to add an XML file that looks like this and somehow pass that to `xdg-mime install` [here](https://github.com/electron-userland/electron-builder/blob/9dced9ceaf245a86f0ef33df373f3fd44871e80a/packages/electron-builder-lib/templates/linux/AppRun.sh#L194):
```
<?xml version="1.0"?>
<mime-info xmlns='http://www.freedesktop.org/standards/shared-mime-info'>
  <mime-type type="application/x.gingko">
    <comment>Gingko Tree</comment>
    <glob pattern="*.gko"/>
    <generic-icon name="x-office-document"/>
  </mime-type>
</mime-info>
```

But I'm not sure where or how to generate this file so that it shows up in `${APPDIR}/usr/share/mime/`.

To summarize:
**Current State**: I see no way to integrate a custom mime-type into Linux, nor how to associate it with my extension defined in `fileAssociation`.

**Desired State**: Define both a file extension `ext` and a custom mimetype in `fileAssociation` field, and have both of those be integrated into Linux systems on installation, so that double-clicking a file with the mime type and extension get run by my application.

**Next Steps**: I'm willing to figure this out, but I'd like some advice on how to proceed. Here's my current thinking:
1. Add `mime-type` field to `fileAssociations` somehow
2. Add `--add-mime-type=application/x.mycustommime` from that field, if it exists
3. Create (or copy, if created by developer already), an XML to a location that will resolve to `${APPDIR}/usr/share/mime/` once packaged into an AppImage

Thoughts?

PR #2873, opened 2018-05-07
fix: add mime-type handling for linux AppImage

AdrianoFerrari:
Initial attempt to handle custom mime-types. Based on how copyIcons
works in AppImageTarget.ts. Related to issue #2872.

This is just a rough starting point, to get some concrete advice on how to proceed.

For instance, I'm not sure if we even need `xdg-mime install` command in `AppImageTarget.ts`, since `AppRun.sh` template contains the same command and looks for XML file in `/usr/share/mime`


develar:
Cool! Thanks. LGTM. I will test manually and merge this week.

Issue #2875, opened 2018-05-07
Cannot extract publisher name from code signing certificate, please file issue. (Windows)


iffy:
* **electron-builder Version**: 20.8.1

* **electron-updater Version**: 2.21.4

* **Target**: Windows

I'm building on a Windows 10 VM with a code signing certificate.  Here's what I've tried:

1. I first tried to build with just environment variables set:

```
set CSC_LINK=Y:\csc_link.p12
set /P CSC_KEY_PASSWORD=< Y:\csc_key_password.txt
```

and got this error (which is why I'm filing this issue):

```
Error: Cannot extract publisher name from code signing certificate, please file issue. As workaround, set win.publisherName: Error: Exit code: ENOENT. spawn openssl ENOENT
```

2. So next I tried with `win.publisherName` set:

Here's the relevant portion of `package.json`:

```
"build": {
    "win": {
      "publisherName": "MY COMPANY NAME"
    },
  },
```

This produced the following error:

```
Error: Exit code: 1. Command failed: Y:\cache\electron-cache\winCodeSign\winCodeSign-2.0.0\windows-10\ia32\signtool.exe sign /t http://timestamp.verisign.com/scripts/timstamp.dll /f Y:\csc_link.p12 /d Buckets /du https://www.budgetwithbuckets.com /p REDACTED (sha256 hash) C:\proj\app\dist\win-unpacked\Buckets.exe
SignTool Error: No certificates were found that met all the given criteria.

SignTool Error: No certificates were found that met all the given criteria.
```

iffy:
I ran the command electron-builder ran and got a password error:

<img width="800" alt="screen shot 2018-05-07 at 11 55 07 am" src="https://user-images.githubusercontent.com/161727/39716450-a44022fc-51ed-11e8-9292-9a8edfb05c42.png">



But then I ran the command with my actual password (rather than the sha256 hash) and it successfully signed:

```
y:\cache\electron-cache\winCodeSign\winCodeSign-2.0.0\windows-10\ia32\signtool.exe sign /t http://timestamp.verisign.com/scripts/timstamp.dll /f Y:\csc_link.p12 /d Buckets /du https://www.budgetwithbuckets.com /p MYREALPASSWORD /debug c:\proj\app\dist\win-unpacked\Buckets.exe
```

iffy:
Let me know what else you need to debug this.  I'm confused why running the same command electron-builder is running works for me but fails for electron-builder.

[...]

PR #2881, opened 2018-05-08
Always run code signing with /debug

iffy:
When there's a problem with codesigning, it would be nice if the /debug output were available.  As the output to this command doesn't seem to be displayed unless there's a problem, always running with /debug output seems reasonable.


iffy:
Sort of fixes #2875 (at least would have helped me know if the problem was with Electron or not).

Issue #1551, opened 2017-05-16
Switch /D is ignored


rozumny:
* **17.5.0**: 

* **WIN64**: 

Adding switch /D to set custom install directory when running generated installer from command line is ignored. Any idea how can this be done?

develar:
Please see https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/templates/nsis/multiUser.nsh `InstDir`.

develar:
Moved to backlog to keep issue list clear.

bennol:
@rozumny were you able to get `/D` working for you?

rozumny:
I had to change multiUser.nsh as following:

`!include FileFunc.nsh
!include UAC.nsh

!define FOLDERID_UserProgramFiles {5CD7AEE2-2219-4A67-B85D-6C9CE15660CB}
!define KF_FLAG_CREATE 0x00008000

!define INSTALL_REGISTRY_KEY "Software\${APP_GUID}"
!define UNINSTALL_REGISTRY_KEY "Software\Microsoft\Windows\CurrentVersion\Uninstall\${APP_GUID}"
!define UNINSTALL_DISPLAY_NAME "${PRODUCT_NAME} ${VERSION}"

# current Install Mode ("all" or "CurrentUser")
Var installMode

!ifndef INSTALL_MODE_PER_ALL_USERS
  !ifndef ONE_CLICK
    Var hasPerUserInstallation
    Var hasPerMachineInstallation
  !endif
  Var PerUserInstallationFolder

  !macro setInstallModePerUser
    StrCpy $installMode CurrentUser
    SetShellVarContext current

    # сhecks registry for previous installation path
    ReadRegStr $perUserInstallationFolder HKCU "${INSTALL_REGISTRY_KEY}" InstallLocation
    ${if} $perUserInstallationFolder != ""
      StrCpy $INSTDIR $perUserInstallationFolder
    ${else}
      StrCpy $0 "$LocalAppData\Programs"
      # Win7 has a per-user programfiles known folder and this can be a non-default location
      System::Call 'Shell32::SHGetKnownFolderPath(g "${FOLDERID_UserProgramFiles}",i ${KF_FLAG_CREATE},i0,*i.r2)i.r1'
      ${If} $1 == 0
        System::Call '*$2(&w${NSIS_MAX_STRLEN} .r1)'
        StrCpy $0 $1
        System::Call 'Ole32::CoTaskMemFree(ir2)'
[...]

PR #2885, opened 2018-05-09
fix: nsis installer /D installation path override

bennol:
The default /D parameter behavior (change installation path for silent install/software distribution) was not present in the NSIS installer. Default NSIS behavior translates /D path into an initial $INSTDIR value. 

The NSIS restriction that /D has to be the last parameter still exists, however due to use of StdUtils.GetParameter quotes are now supported.

Issue: #1551


develar:
Thanks!

Issue #2895, opened 2018-05-10
Application starts before NSIS installer is closed.


dscalzi:
* **Version**:  20.13.3

* **election-updater**: 2.21.10

* **Target**: nsis x64

After an auto-update, the application starts immediately, while the NSIS installer is still open. On the installer, you are given an option to open the application when finished. This option is enabled by default. Closing the installer with this option enabled will open another instance of the application.

Before the auto-update, I'm only calling
```javascript
autoUpdater.quitAndInstall()
```

I don't think this is expected behavior. I'm currently on Windows 10.

develar:
@dscalzi @computerquip-streamlabs Do you use assisted installer or one-click?

dscalzi:
Assisted installer

develar:
Duplicates #2619

Issue #2619, opened 2018-02-21
autoUpdater.quitAndInstall causes app to re-open twice with nsis


kspearrin:
* **Version**: electron-builder 20.0.7 and electron-updater 2.20.1

* **Target**:  build --win --x64 --ia32

We use `nsisWeb` with:

    "nsisWeb": {
      "oneClick": false,
      "perMachine": true,
      "allowToChangeInstallationDirectory": true,
      "artifactName": "${productName} Installer ${version}.${ext}",
      "uninstallDisplayName": "${productName}",
      "deleteAppDataOnUninstall": true
    },

We also handle updates manually with `autoUpdater` and prompt the user when an update is ready to install. If the user selects to "Restart Now", we call `autoUpdater.quitAndInstall();`. This causes the installer to run, and install the update.

After the update is installed the app automatically starts back up. This is the desired behavior.

The problem is that the "completed" installer window is still shown with the "Run" checkbox checked.

![image](https://user-images.githubusercontent.com/1190944/36564871-c7944246-17ec-11e8-9adc-bf8485bd6763.png)

If a user closes the installer window another instance of the app opens. We would like the ability to stop the installer window from staying open on the "completed" step whenever updates are ran from `autoUpdater.quitAndInstall();` since the app is going to open back up already anyways. I tried `autoUpdater.quitAndInstall(false, true);` as well, but it didn't seem to change anything.

akashnimare:
I was about to open an issue for this.

PR #2902, opened 2018-05-12
fix: Don't force run if not silent

computerquip-streamlabs:
Fixes https://github.com/electron-userland/electron-builder/issues/2895
Fixes https://github.com/electron-userland/electron-builder/issues/2619

I don't see a reason to force-run if the installer isn't set to silent mode. A quick test on Windows seems to work properly. I've not tested on other platforms.

### DIFF
--- a/packages/electron-updater/src/BaseUpdater.ts
+++ b/packages/electron-updater/src/BaseUpdater.ts
@@ -14,7 +14,7 @@ export abstract class BaseUpdater extends AppUpdater {
 
   async quitAndInstall(isSilent: boolean = false, isForceRunAfter: boolean = false): Promise<void> {
     this._logger.info(`Install on explicit quitAndInstall`)
-    const isInstalled = await this.install(isSilent, isSilent ? isForceRunAfter : true)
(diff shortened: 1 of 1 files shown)


dscalzi:
Hoping this can be merged, simple fix to annoying issue. Thanks

@develar

develar:
Reason — please answer to https://github.com/electron-userland/electron-builder/issues/2895#issuecomment-389114607 It seems you use assisted installer and indeed, this flag should be applicable only for one-click installer.

dscalzi:
Perhaps the behavior is supposed to force the application so start if isSilent is true? If I understand correctly a silent installer will not be visible so the application must be started when done.

Currently, the application is being force started **every time** when the application is not updated silently. The installer gives the option to start the installer, so this cannot be intended behavior.

Is this correct?

develar:
Correct fix is different. Thanks a lot for spotting this serious bug for assisted installers.

SourceBoy:
I'm using electron-updater 4.0.6, and I'm still seeing `this.install(isSilent, isSilent ? isForceRunAfter : true)`
Screenshot_2019-03-09_00-20-57

Issue #2859, opened 2018-05-02
Electron Updater occasionally uninstalls the app.


jonzlin95:
* **Version**: 20.11.0

* **Electron Updater Version** *: 2.21.8

* **Target**: NSIS / NSIS-Web Windows

Right now there is a case where the packageFile in electron-updater will be non-null, but doesn't exist on the filesystem. This is causing a bunch of uninstalls on update for some of our users.

I've proposed a very simple fix (to just re-download the file by not including --package-file arg) in the NSISUpdater. (https://github.com/electron-userland/electron-builder/pull/2858)

This uses the fs-extra package, but can be changed to whatever you want to use for filesystem ops.

**Further Investigation**

Upon further investigation, I think I've found the cause of the bug. The issue is that `downloadedUpdateHelper.validateDownloadedPath` only checks for the existence of the "installer" file, but not the package file. It just assumes the package file exists. However if the package file doesn't exist, it will never be caught and cause an issue where the `--package-file` passed to NSIS doesn't exist.

A similar fix can be placed in DownloadedUpdateHelper to catch this on the cache checking side.

PR #2929, opened 2018-05-21
windows nsis: prevent a missing package.7z from nuking the application.

hinaria:
when running setup with `--package-file`, electron-builder's nsis setup may occasionally delete all application files if that package can't be read.

related (but separate) is a bug in `electron-updater` that sometimes deletes the `package.7z`, thereby surfacing this bug. but we shouldn't nuke an app just because `package.7z` is missing in an update - we should fail with a message saying so.

## summary

let's say we have an application called "starcall".

during installation, re-installation, or an upgrade, setup first uninstalls any previous versions of starcall, if any exist.

only once that is done does setup try to access and extract `package.7z` - the package specified in `--package-file`. `electron-update` specifies this, for example.

however, if that package can't be read (for example, because it doesn't exist), the extraction obviously fails, but the installer continues as if everything succeeded. that's a problem because in the previous step we just nuked the entire application off people's computers , so now we've got a "completed" installation with shortcuts that point to an empty application directory.

## the fix

this pull request resolves by far the most common scenario that i've seen in production where people report electron-builder uninstalling their application. (#2859 and #2858 describe the same problem).

after this pull request, nsis will check that `--package-file` exists, and if it doesn't, falls back to installing starcall from the internet. the package url that it uses is version-specific, so `install-v0.0.1.exe` will download the `v0.0.1` package.

the guiding principle behind this change is that we want to keep the application in a working state - when upgrading, if the `--package-file` doesn't exist, our failure case shouldn't be that we nuke the entire application off people's computers (except with dead shortcuts). i'll go into why `package.7z` might not exist later, but tl;dr there is a bug in `electron-updater` that may cause it to occasionally delete the package if it thinks the update installer failed to launch.

## related

there are a few other related bugs that surfaced this bug. i plan to submit a pr to those in the near future, but this one resolves (or works around) the most common issue i've seen in production.

### DIFF
--- a/packages/electron-builder-lib/templates/nsis/include/installer.nsh
+++ b/packages/electron-builder-lib/templates/nsis/include/installer.nsh
@@ -73,6 +73,8 @@
   !else
     !ifdef APP_PACKAGE_URL
       Var /GLOBAL packageFile
+      Var /GLOBAL packageFileExplicitlySpecified
+
       ${StdUtils.GetParameter} $packageFile "package-file" ""
       ${if} $packageFile == ""
         !ifdef APP_64_NAME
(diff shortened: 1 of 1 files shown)


alexlee85:
@hinaria it works!!! Tank you so much.

develar:
@hinaria Thanks a lot! You are hero :)

PR #2952, opened 2018-05-25
fix: closes #2892

### LINKED ISSUE #2892: Build stucks building native addon
* **Version**: 20.13.3

* **Target**: Windows x64 (zip)

Similar to #2882
Build stucks when rebuilding native addon.
Build works well with `$env:DEBUG="electron-builder"`
Looks like there are a lot of warning output from msbuild and they are stuck

* log
```
> git clone https://github.com/akimasa/electron-quick-start
> cd .\electron-quick-start\
> npm i
> npm run build.win32

> electron-quick-start@1.0.0 build.win32 C:\Users\Akimasa\Documents\work\tmp\electron-quick-start
> electron-builder --windows=zip --x64 ./ sample

  • electron-builder version=20.13.3
  • writing effective config file=dist\electron-builder-effective-config.yaml
  • rebuilding native production dependencies platform=win32 arch=x64
```
stucks here.

akimasa:


develar:
It is already ignored.


```
if (options.stdio == null) {
    const isDebugEnabled = debug.enabled
    // do not ignore stdout/stderr if not debug, because in this case we will read into buffer and print on error
    options.stdio = [extraOptions != null && extraOptions.isPipeInput ? "pipe" : "ignore", isDebugEnabled ? "inherit" : "pipe", isDebugEnabled ? "inherit" : "pipe"]
  }
```

Or do you observe another behaviour?

akimasa:
https://github.com/electron-userland/electron-builder/issues/2892#issue-321909751
Could you try building https://github.com/akimasa/electron-quick-start on Windows?
I don't know why but native addon build stucks without this patch.

akimasa:
```
options.stdio = [extraOptions != null && extraOptions.isPipeInput ? "pipe" : "ignore", isDebugEnabled ? "inherit" : "pipe", isDebugEnabled ? "inherit" : "pipe"]
```
When I run the build, option.stdio will be `[ 'ignore', 'pipe', 'pipe' ]`.
That means ignore input, create pipes for stdout and stderr.
However, there are nothing reading output from pipes.
This will cause build process stuck.
We need to read pipe or ignore output or maybe another solution.

akimasa:
I confirmed fix in the latest version.
Thank you!

Issue #2951, opened 2018-05-25
[NSIS] runAfterFinish not working when `oneClick: true`


wojtkowiak:
* **Version**: 
`20.14.7`
* **Target**: 
NSIS

For quite some time I had:
```
oneClick: true,
runAfterFinish: false
```
as I want to build a simple installer but I need the app to not start at the end.
I see that from `20.13.5` the `runAfterFinish(RUN_AFTER_FINISH)` is not taken into account: https://github.com/electron-userland/electron-builder/commit/edb82afd328e51fe736da92cd389b5061bdf6318

I did check the corresponding tasks for this PR but I do not get the picture what exactly was fixed there and whether removing `runAfterFinish` was intended or not.

Anyway let me know if we can restore it, or how can I achieve the previous effect 🙂

develar:
It is bug, fixed, will be released soon.

wojtkowiak:
Thanks a lot!

wojtkowiak:
Sorry to bothering you once again but I have two problems after updating to 10.5.2.
When building with my `meteor-desktop` now the installation is stuck at `packaging` - it wait there indefinitely (waited 30 min) - it seems to be a issue with the new `copyNodeModules` - I will check it out more in detail in a moment.
I did remove that commit and tested the fix but the installer now does not close itself, I guess the

```
!ifdef ONE_CLICK
  !ifdef RUN_AFTER_FINISH
    ${ifNot} ${Silent}
    ${orIf} ${isForceRun}
      !insertmacro doStartApp
    ${endIf}

    !insertmacro quitSuccess
  !endif
!else
```

should be this way:

```
!ifdef ONE_CLICK
  !ifdef RUN_AFTER_FINISH
    ${ifNot} ${Silent}
    ${orIf} ${isForceRun}
      !insertmacro doStartApp
    ${endIf}
  !endif
  !insertmacro quitSuccess
!else
```

wojtkowiak:
About the first issue, it gets stuck here:
`await executeAppBuilder(["node-dep-tree", "--dir", platformPackager.info.appDir])` 

Probably this should not be invoked at all as I do not have `node_modules` at the point.
[...]

PR #2975, opened 2018-06-04
fix: Do not run node modules file matcher when node_modules are handl…

wojtkowiak:
…ed externally

Resolves https://github.com/electron-userland/electron-builder/issues/2951#issuecomment-392489184

### DIFF
--- a/packages/electron-builder-lib/src/packager.ts
+++ b/packages/electron-builder-lib/src/packager.ts
@@ -57,6 +57,12 @@ export class Packager {
     return this._metadata!!
   }
 
+  private _nodeModulesHandledExternally: boolean = false
+
+  get areNodeModulesHandledExternally(): boolean {
+    return this._nodeModulesHandledExternally
+  }
+
   private _isPrepackedAppAsar: boolean = false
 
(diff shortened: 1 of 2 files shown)


fubhy:
Checking if there is a node_modules directory or not within the app directory is not good enough. In case of more complex setups e.g. with yarn workspaces, the node modules directory might be entirely hoisted to the workspace root, resulting in no node_modules within the app directory.

wojtkowiak:
Ok, I get it, good point. I do not use `yarn workspaces` so I was not aware of this possibility. I rewrote it to use the same information I used before and that is resolving `beforeBuild` hook to `false`.
I am already using that to prevent `electron-builder` from installing dependencies because they are handled externally so I already have this information passed to electron-builder so it probably better to just reuse that.

wojtkowiak:
@devlar sorry for pinging you, but I need to get it working as this is a blocker for next `meteor-desktop` as the integration is not working at all without this.

develar:
Sorry for delay, will be addressed tomorrow morning CET.

wojtkowiak:
Thanks a lot 👍

PR #2984, opened 2018-06-05
fix(archive): compute safe artifact name using standard util

nikku:
This ensures the used defined artifact name (`-c.artifactName`) is taken
in consideration when publishing files built with the artifact target.

Without this commit files would always be uploaded as
`${name}-${version}-${arch}.${ext}` for Linux targets with no way to
override it, i.e. to add a `-linux` suffix.


develar:
Ouch, so embarrassing :), indeed.

Issue #1838, opened 2017-07-18
Mac Installer do not create .app folder in /Applications when launching .pkg


Mic75:
* **Version**: 19.16.0

* **Target**: pkg (macos)

Let's say the application name is myApp and the build config in the package.json is the following:
```javascript
"build": {
    "appId": "com.myorg.myapp",
    "nsis": {
      "include": "nsis/refreshDesktop.nsh",
      "artifactName": "MyAppSetup.${ext}"
    },
    "publish": {
      "provider": "generic",
      "url": "${env.MYAPP_UPDATE_URL}"
    },
    "mac": {
      "target": [
        "pkg",
        "zip"
      ],
      "artifactName": "MyAppSetup.${ext}"
    }
  }
```

### Reproduction
Ok, so repro steps are:
1. build myApp
2) launch the .pkg
3) remove myApp.app from /Applications
5) relaunch pkg

### Expected

myApp installed into /Applications

### Observed 

myApp not installed into /Applications

### Diagnostic 

Having something already called myApp.app somewhere in the disc like the one generated by electron-builder in the mac folder prevent the myAppl.app folder to be re-created into the /Applications folder.

This is related to this [stackoverflow](https://stackoverflow.com/a/19537586)
One of the comment suggest to have the BundleIsRelocatable set to false in the plist file.
Could this be added as a new option for the .pkg target ?

etherElric:
It seems like the generated pkg detects that it should do an upgrade instead of a fresh install. But even though I did two subsequent installs with a different name and build number, the second time, nothing changed.

**Repro:** 
1. build the app with the following [relevant] configs in package.json:
{
  "name": "Test1",
  "version": "0.3.0-dev",
  "build": {
    "appId": "com.test.test",
    "buildVersion": "1000",
    "mac": {
      "target": [
        "pkg",
        "zip"
      ],
      "artifactName": "TestSetup.${ext}"
    }
  },
}
2. Launch the pkg
3. Remove the folder myproject/dist/mac
4. Build again with the following [relevant] config in package.json:
{
  "name": "Test2",
  "version": "0.3.0-dev",
  "build": {
    "appId": "com.test.test",
    "buildVersion": "1002",
    "mac": {
      "target": [
        "pkg",
        "zip"
      ],
      "artifactName": "TestSetup.${ext}"
    }
  },
}
5. Launch the pkg

**Observed:** 
The app installed in /Applications is still called Test1.

**Expected:**
[...]

Issue #1711, opened 2017-06-20
Cannot downgrade installed pkg


p120ph37:
* **Version**: 19.5.1
* **Target**: pkg

`BundleIsVersionChecked` seems to default to `true`.  It would be nice to be able to override this default to allow for downgradable pkg files.

See: https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/pkgbuild.1.html

p120ph37:
After doing some more research and testing, it seems that the `BundleIsVersionChecked` option that is mentioned in the `pkgbuild` manpage is only applicable when using `pkgbuild` in its stand-alone `--root` form, and not applicable for the `--component` form as used in `electron-builder`.

The effect of this option is to populate or clear the contents of the `<bundle-version>` tag in the `PackageInfo` which is contained in the inner `pkg`.  As far as I can tell, there doesn't seem to be a command-line option to `pkgbuild --component` which directly controls this tag.

If I manually use `pkgutil --expand`, edit the inner `PackageInfo` to remove the contents of the `<bundle-version>` tag, and then use `pkgutil --flatten` to repackage, the resulting installer is capable of downgrading an installation.

One really confusing "feature" is that the installer does not give any indication when an attempted downgrade is prevented - it still displays the success screen, and the only way to tell that the downgrade was not actually applied is to examine the installed application itself.

develar:
Not clear right now how to fix it.

p120ph37:
Yeah, I'm not sure how to fix it properly either.  It's kind of an unusual use-case.  I think Apple expects people who might need this feature to use the `pkgbuild --root` form which accepts the `BundleIsVersionChecked` option in the plist file.

I currently have a fixup script that expands the pkg, empties the `<bundle-version>` tag in the inner `PackageInfo` and re-flattens.  If the `pkg` target becomes complex enough someday that it needs to do an expand/flatten for other reasons, this would be a good feature to include at that point, but I can't see any way to include it cleanly in the current build process.

And maybe Apple will add another command-line option to `pkgbuild --component` someday to control this?  (heh, I doubt it).

develar:
Still not moved to backlog — it not windows, so, probably I will return to this task when other urgent issues will be fixed.

develar:
Moved to backlog to keep issue list clear.

p120ph37:
It looks to me like this has been fixed by now.  The `isVersionChecked` option appears to be working correctly.

PR #2991, opened 2018-06-07
Add feature to override the Mac PKG component plist

pronebird:
This PR adds ability to override the component plist provided to `pkgbuild`, discussed in #1838 

The example YML configuration:

```yml
pkg:
  isRelocatable: true
  isVersionChecked: true
  hasStrictIdentifier: true
  overwriteAction: "upgrade"
```

The new options that I've added correspond to the options in component plist file saved and consumed by `pkgbuild`, this is the `man` page excerpt:

> COMPONENT PROPERTY LIST
>      When you package a destination root, you can use a component property list to specify how bundles in
>      that root should be handled by the OS X Installer. This property list should be an array of dictionar-
>      ies, where each dictionary specifies a single bundle. The dictionary keys are as follows:
> 
>      Key                            Description
>      RootRelativeBundlePath         Path to bundle relative to the destination root (string)
>      BundleIsRelocatable            Install bundle over previous version if moved by user? (bool)
>      BundleIsVersionChecked         Don't install bundle if newer version on disk? (bool)
>      BundleHasStrictIdentifier      Require identical bundle identifiers at install path? (bool)
>      BundleOverwriteAction          How to treat existing on-disk version of bundle (string)
>      BundlePreInstallScriptPath     Relative path to bundle-specific preinstall script
>      BundlePostInstallScriptPath    Relative path to bundle-specific postinstall script
>      ChildBundles                   Bundles under this bundle (array of dictionaries)

I ran `yarn run docs` but It seems that docs haven't been updated in a while so I'll leave it up to somebody else to send a PR with the generated docs. I don't want to commit that many changes, some of which didn't look right :)


pronebird:
The relevant test passed on Travis, however the unrelated job went into infinite execution.

develar:
Thanks a lot, sorry for delay.

pronebird:
@develar that's alright! Could you please make a release or pre-release? I really need this feature.

p120ph37:
This addressed #1711

Issue #3003, opened 2018-06-11
Downloading builds(updates) more than once even if downloaded already


developeryashraj:
* **Version**: 20.15.1

* **Target**: Windows

I am facing issues with electron-updater. Currently electron updater download new version of builds every time when it checks for update and even that new version has been downloaded already. So considering application size is 50 MB then everytime when it checks for update, it uses 50MB of client's data and server bandwidth.

I know that github issues are not for questions so i have already asked this on stackoverflow as well and the detail of this is available here :- https://stackoverflow.com/questions/50744254/electron-auto-updater-downloading-multiple-builds

My other problem is similar to #3000 . So i haven't specify details here but you can get it on stack.

My package json is 
```json
{
  "name": "electorn_application",
  "version": "10.1.1",
  "private": true,
  "main": "main.js",
  "author": "Yashraj",
  "homepage": "http://xyz.com",
  "description": "A Good Software",
  "dependencies": {
    "body-parser": "~1.18.3",
    "dateformat": "^3.0.3",
    "debug": "~3.1.0",
    "electron-is-dev": "^0.3.0",
    "electron-updater": "^2.21.10",
    "express": "~4.16.3",
    "jade": "^1.11.0",
    "jquery": "^3.3.1",
    "lodash": "^4.17.10",
    "mosca": "^2.8.1",
    "mqtt": "^2.18.0",
    "ncp": "^2.0.0",
    "request": "^2.87.0",
    "serve-favicon": "~2.5.0",
    "socket.io": "^2.1.1",
    "sqlite3": "^4.0.0",
    "sqlite3-transactions": "0.0.5",
    "xtend": "^4.0.1"
  },
  "build": {
    "publish": [
      {
        "provider": "generic",
        "url": "http://xyz.com"
      }
    ],
    "win": {
      "icon": "./icons/icon.ico"
    }
  },
  "scripts": {
    "start": "electron .",
    "test": "echo \"Error: no test specified\" && exit 1",
    "build": "electron-builder",
    "build:linux": "electron-builder --linux",
    "postinstall": "electron-builder install-app-deps"
  },
  "devDependencies": {
    "electron": "^2.0.2",
    "electron-builder": "^20.15.1"
  }
}
```

developeryashraj:
[...]

PR #3007, opened 2018-06-12
Auto update issue #3003

developeryashraj:
Improved comparison with only file info


develar:
You are right — no need to compare URL since `versionInfo` is already compared.

PR #3015, opened 2018-06-16
Support differentialPackage for dmg builds

c960657:
Building the block map takes a while. If you don't need it, there is no reason to spend time building it.

This PR adds support for `differentialPackage` for DMG builds.


develar:
Log is async. I doubt that block map generation takes a lot of time. Why do you think so?

c960657:
You are right :-)

Generating the block map takes less than a second. I was confused that it took a long time between the “building block map” line and the “Done in 123s”. It turns out that it is the zip generation that takes a long time.

So instead I have created a PR #3032.

PR #3032, opened 2018-06-24
perf(zip): Use default compression level

c960657:
Building the `zip` target takes a long time, because the file is compressed with level 9, `-mx=9`.

I suggest `-mx=9` is only used when `compression` is set to `maximum`. For `normal` compression, we should use the default level, 5.

In my application, adding `-mx=9` increases the run time by a factor 6, but the size of the compressed file is only reduced with 1.2%. This seems like a bad trade-off.

### DIFF
--- a/packages/electron-builder-lib/src/targets/archive.ts
+++ b/packages/electron-builder-lib/src/targets/archive.ts
@@ -80,20 +80,16 @@ export function compute7zCompressArgs(format: string, options: ArchiveOptions = 
   let storeOnly = options.compression === "store"
   const args = debug7zArgs("a")
 
-  let isLevelSet = false
   if (process.env.ELECTRON_BUILDER_COMPRESSION_LEVEL != null) {
     storeOnly = false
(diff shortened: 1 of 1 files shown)


develar:
Size of your app (unpacked and packed)? Do you have media assets (video, images)?

c960657:
```
Scanning the drive:
120 folders, 230 files, 146231734 bytes (140 MiB)
```

Compressed size is 50 MB. Few media assets. One native depencency, [jabra](https://www.npmjs.com/package/jabra).

develar:
> 120 folders, 230 files, 146231734 bytes (140 MiB)

It means that your case is not representative, because it is standard size of Electron app (minimal). Need to test compression in case of large Electron app.

But... we use 7z and 7z is very, very slow compared to native zip without any really visible difference.

> This seems like a bad trade-off.

👍

develar:
For 6 difference is 512kb in 7 seconds (42 sec win). For 7 difference is 164KB in 19 seconds (30 sec win).

Since for most developers better will be better "wait but get the best compression", I changed default level for **zip only** and to 7, not to 6. 

Thanks for spotting this.

PR #3057, opened 2018-06-29
Code sign .dlls on Windows

pronebird:
Currently windows packager only signs `.exe` files. This PR enables code signing for `.dll`s too.


pronebird:
Seems like the test on appveyor failed because the timestamp server is not reachable:

```
The specified timestamp server either could not be reached or
599      returned an invalid response.
600
```

develar:
Thanks. And now I see that current implementation doesn't conform our quality standards. I will reimplement to avoid current ugly `readdir` call (it is expensive).

jonzlin95:
This update causes some (unintended?) consequences where if you have an already signed .dll file packaged, it will cause the build to fail. Suggestion is probably to add a way to exclude signing certain files or opt into .dll signing.

PR #3065, opened 2018-07-01
Remove unnecessary volume binding

MacroChip:
PWD is already bound to /project which includes node_modules.


develar:
It must be separated named volume (not bind mount) because node_modules on your machine maybe not suitable for docker OS (linux) due to native dependencies.

Issue #3091, opened 2018-07-09
Blockmap Comparision Fails for AWS S3


developeryashraj:
* **Version**: 20.19.2

* **Target**: Windows

I found that when we use S3 as a provider then it fails to compare and check blockmap files [at this place](https://github.com/electron-userland/electron-builder/blob/v20.19.2/packages/electron-updater/src/NsisUpdater.ts#L152). This is the log of `fileInfo`, `newBlockMapUrl` and `oldBlockMapUrl` Object.

**Fileinfo :-**
```
{
  url: URL{
    href: 'https://AWS_S3_HOST//bucket-yashraj/electron%20Setup%2011.0.3.exe',
    origin: 'https://AWS_S3_HOST',
    protocol: 'https:',
    username: '',
    password: '',
    host: 'AWS_S3_HOST',
    hostname: 'AWS_S3_HOST',
    port: '',
    pathname: '//bucket-yashraj/electron%20Setup%2011.0.3.exe',
    search: '',
    searchParams: URLSearchParams{
      
    },
    hash: ''
  },
  info: {
    url: 'electron Setup 11.0.3.exe',
    sha512: 'k3seJ1BoSsaxAxODNJqhf7xUh6vfQUflvU+qu0mXXiuBFb8gpvt0oDS2rJm6gP+4lEilw6Gt8bt7BbimXftVNQ==',
    size: 52021199
  }
}
```
**newBlockMapUrl :-**
```
{
  url: URL{
    href: 'https://bucket-yashraj/electron%20Setup%2011.0.3.exe',
    origin: 'https://bucket-yashraj',
    protocol: 'https:',
    username: '',
    password: '',
    host: 'bucket-yashraj',
    hostname: 'bucket-yashraj',
    port: '',
    pathname: '/electron%20Setup%2011.0.3.exe',
    search: '',
    searchParams: URLSearchParams{
      
    },
    hash: ''
  }
}
```
**oldBlockMapUrl :-**
```
{
  url: URL{
    href: 'https://bucket-yashraj/electron%20Setup%2011.0.3.exe',
    origin: 'https://bucket-yashraj',
    protocol: 'https:',
    username: '',
    password: '',
    host: 'bucket-yashraj',
    hostname: 'bucket-yashraj',
    port: '',
    pathname: '/electron%20Setup%2011.0.3.exe',
    search: '',
    searchParams: URLSearchParams{
      
    },
    hash: ''
  }
}
```
[...]

PR #3092, opened 2018-07-09
Solve issue with blockmap comparision for amazon S3

developeryashraj:
* Solve issue #3091


develar:
Cannot be merged as is because you have found a critical issue and all affected places must be revisited. Thanks a lot for spotting, I will check and add missed tests today evening CET.

Issue #2921, opened 2018-05-17
Web installer forces application to open before hits the last screen


mastergberry:
* **Version**: 19.53.6

* **Target**: nsis-web

I cannot find a mistake in our nsh file we use for our electron-builder project. When upgrading our application using nsis-web it seems to force open the application after the installation/patching is done before they can even get to the last screen to choose whether they want to run it or not.

Seems like a mistake with the --force-run stuff.

```
"nsis": {
      "oneClick": false,
      "perMachine": true,
      "allowElevation": true,
      "allowToChangeInstallationDirectory": true,
      "runAfterFinish": false,
      "unicode": true,
      "differentialPackage": true
    },
```

Relevant nsis information is above.

develar:
Should be fixed now.

PR #3093, opened 2018-07-09
fix(nsis): respect --force-run when RUN_AFTER_FINISH is false

wojtkowiak:
This is a follow up for:

https://github.com/electron-userland/electron-builder/issues/2951
https://github.com/electron-userland/electron-builder/pull/2974

```
oneClick: true,
runAfterFinish: false
```
My configuration is now working as expected but the auto update is not starting the app after the update.
Please take a look at it but it should now be ok for all cases.

### DIFF
--- a/packages/electron-builder-lib/templates/nsis/installSection.nsh
+++ b/packages/electron-builder-lib/templates/nsis/installSection.nsh
@@ -85,7 +85,10 @@ ${endIf}
 !ifdef ONE_CLICK
   !ifdef RUN_AFTER_FINISH
     ${ifNot} ${Silent}
-    ${orIf} ${isForceRun}
+      !insertmacro doStartApp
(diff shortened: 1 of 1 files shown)


develar:
Hmm... You want to build one-click installer that doesn't start app on finish, but electron-updater behaviour must be independent and allows you to start app if you pass corresponding flag....

Good catch. Makes sense. One correction — if `RUN_AFTER_FINISH` defined, we still need to respect `isForceRun` even if `silent`.

wojtkowiak:
Thanks!

> if RUN_AFTER_FINISH defined, we still need to respect isForceRun even if silent

Right, I've misinterpreted that conditional.

mastergberry:
Is this related to #2921 also?

develar:
Yes.

Issue #1488, opened 2017-04-21
Incorrect comparison of version numbers (electron-updater)


mikecao:
* **Version**: 
electron-builder 17.01
electron-updater 1.14.0

* **Target**: 
Windows x64

I pushed out a version with the number `1.0.0-beta.2`. When checking for updates with my local version `1.0.0-beta.3`, it says a an update is available even though `1.0.0-beta.3` should be newer than `1.0.0-beta.2`. If I change the version to simply `1.0.0`, then it correctly says there is no update available.

PR #3111, opened 2018-07-12
feat(updater): add vertical upgrading for channels

vincentbriglia:
Hi, 

in one of the projects I'm working on I have to adhere to gitflow for our git branching model, we have enhanced it with gitversion to automatically tag our releases with a semver that makes sense across branches. 

from the following example graphs, take note that I only build electron apps from develop branches (alpha channel), master branches (latest channel), hotfix and release branches (beta channel) as per 
http://gitversion.readthedocs.io/en/latest/git-branching-strategies/gitflow-examples/

the current alpha/beta/latest setup doesn't allow for switching between release channels in such a way that it always takes the latest from that channel.

the option `generateUpdatesFilesForAllChannels` set to `true` doesn't work for us for the following reason: I release `alpha` versions far more frequently than beta releases and in turn I release `beta` versions far more frequently than `latest` ones, setting `generateUpdatesFilesForAllChannels` to true would result in a tiny window where my `alpha` channel and `beta` channel have a non-alpha and a non-beta release assigned to it. 

I want to avoid this

I tried a combination of `allowPrerelease` and `allowDowngrade` and made sure to understand the codebase before making this change. 

I added a flag as a non-breaking change in the `AppUpdater` class: `alwaysDownloadLatestForChannel` that would allow downgrading or upgrading to a newer or older version on a particular channel. 

the use case is that currently I cannot switch between a `latest` 2.0.0 and a beta `2.0.0-beta.123` if I wanted to switch back to the `beta` channel and update to that version. I am aware that this sounds somewhat strange, but for QA this is a must have. In the current setup what has to be done is the following: 

I am on `latest` 2.0.0, I have an `alpha` 3.0.0-alpha.34 and a `beta` 2.0.0-beta.23. When `allowDowngrade` and `allowPrerelease` are set I can `upgrade` from 2.0.0 to 3.0.0-alpha.34 but I can not `downgrade` to 2.0.0-beta.23. Only after upgrading to 3.0.0-alpha.34 can i downgrade to 2.0.0-beta.23 and then back to 2.0.0. 

adding the `alwaysDownloadLatestForChannel` option just ensures that when I switch to a channel, I download the version stated in that channel, unless if it's the version I am currently on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ electron-builder-*.d.ts
 /scripts/renderer/out/
 
 __publish.sh
+
+.vscode
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -49,6 +49,13 @@ export abstract class AppUpdater extends EventEmitter {
   allowDowngrade: boolean = false
 
   /**
+   * Whether to always allow downgrading or upgrading for a channel (0.2.0 >
+   * 0.3.0-alpha.234 > 0.2.1-beta.22 > 0.2.0)
+   * @default false
+   */
+  alwaysDownloadLatestForChannel: boolean = false
+
+  /**
    * The current application version.
    */
   readonly currentVersion: string
@@ -317,7 +324,17 @@ export abstract class AppUpdater extends EventEmitter {
     }
 
     const isStagingMatch = await this.isStagingMatch(updateInfo)
-    if (!isStagingMatch || ((this.allowDowngrade && !hasPrereleaseComponents(latestVersion)) ? isVersionsEqual(latestVersion, this.currentVersion) : !isVersionGreaterThan(latestVersion, this.currentVersion))) {
+    const isSameVersion = isVersionsEqual(latestVersion, this.currentVersion)
+    const isLatestVersionNewer = isVersionGreaterThan(latestVersion, this.currentVersion)
+
+    const disallowDowngrade =
+      this.allowDowngrade && !hasPrereleaseComponents(latestVersion)
+        ? isSameVersion
+        : !isLatestVersionNewer
+
+    const forceUpdate = !isSameVersion && this.alwaysDownloadLatestForChannel
+
+    if (!isStagingMatch || (!forceUpdate && disallowDowngrade)) {
       this.updateAvailable = false
       this._logger.info(`Update for version ${this.currentVersion} is not available (latest version: ${updateInfo.version}, downgrade is ${this.allowDowngrade ? "allowed" : "disallowed"}).`)
       this.emit("update-not-available", updateInfo)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  // not all of us have IntelliJ IDEA
+  "editor.formatOnSave": false,
```
develar: Yes, but sorry, I doubt that this file will be merged :)

vincentbriglia: you ok if I add it to .gitignore ? I use vscode but don't want to mess up the formatting standards you have (semicolons, etc )

vincentbriglia: I removed the file from the pull request, but added a line in .gitignore

develar:
Do you mean that you sometimes downgrade version specified in a channel file? You publish beta channel with `2.0.0` but for some reasons decided to revert it to `2.0.0-beta.23`?

develar:
Boo... `hasPrereleaseComponents` prevents `allowDowngrade` to suit your needs. Interesting, why, we have this restriction :)

develar:
You don't use `generateUpdatesFilesForAllChannels` because "would result in a tiny window where my alpha channel and beta channel have a non-alpha and a non-beta release assigned to it"

And you don't like this "window" because in this case your QA team can get unwanted (unnecessary) update from `2.0.0-beta.123` to `2.0.0`. This update is unnecessary because you will release 2.0.1-beta.1` soon.

Ok, I will check why we have `hasPrereleaseComponents` and if will be a strong reasons to keep current behaviour, suggested option will be added (goal: reduce number of options :)).

vincentbriglia:
correct! thanks

develar:
This restriction is added by https://github.com/electron-userland/electron-builder/issues/1488

develar:
And now I see that that fix was strange — it is clearly stated that if you want to revert bad version, you **must** publish a new version — "you must increment the version number higher than your broken release"

develar:
> develop branch is always ahead (if master / latest is version 1.6.0, then develop is 1.7.0-alpha.numberofcommits) and beta releases are only used to harden a release before it goes out, so the latest beta at this point would be 1.6.0-beta.numberofcommits, if 1.7.0 is still in development and not branched off to get ready for release

In short: latest beta is always equals to published release. Published release is nothing more that "beta version is good enough, promote it as release (latest in terms of electron-builder channels)"

develar:
So, `allowDowngrade` behaviour will be changed — downgrade will be never allowed for the same channel (channel inferred from version pre-release component). So, `hasPrereleaseComponents` check will be removed and original issue will be solved.

> I am on latest 2.0.0, I have an alpha 3.0.0-alpha.34 and a beta 2.0.0-beta.23. 

As result, if you are on `2.0.0` and channel `latest`, but channel switched to `2.0.0-beta.23`, app version will be downgraded to `2.0.0-beta.23`.

@vincentbriglia Sorry, but for me still not really clear why you want to update to `2.0.0-beta.23` if current version `2.0.0` To be really sure that beta version works? To be really sure that you test beta version? But why and for what if 2.0.0 is already released? Or because you want to downgrade from your dev `3.0.0-alpha.34`?

Ok, ok, cleared up, I read again:

>  am on latest 2.0.0, I have an alpha 3.0.0-alpha.34 and a beta 2.0.0-beta.23. When allowDowngrade and allowPrerelease are set **I can upgrade from 2.0.0 to 3.0.0-alpha.34** *but* **I can not downgrade to 2.0.0-beta.23**. Only after upgrading to 3.0.0-alpha.34 can i downgrade to 2.0.0-beta.23 and then back to 2.0.0.

---

Ok, now all is clear :)

develar:
Although maybe `allowDowngrade` by default logically should be `true` and by our conditions we require and explicitly state that you **must** ""you must increment the version number higher than your broken release".", it cannot be `true` by default because behaviour can lead to unsafe result — old version can be not compatible with your current (db or app configuration). So, to the sake of safety, default value is a little bit "silly".

PS. Ouch, I see that you are from LEGO :) 50% of my room occupied by 2 LEGO trains, railways and railway infrastructure :)

develar:
electron-updater 3.0.0 is released as pre-release (tagged as `next` on npm). Please try and thanks a lot for PR.

vincentbriglia:
I will try it as soon as I have access to my laptop. Thanks for your support through the weekend. I hope you enjoy the trains, are they perhaps the new ones? That remote is great fun.

develar:
> are they perhaps the new ones

Yes, my daughter and me enjoys it. Hooray, new trains!

vincentbriglia:
works like a charm 👍 I will keep contributing as much as possible!

Issue #2559, opened 2018-02-07
Private github unable to update


DmitryAstafyev:
* **Version of builder**: 19.56.0

* **Version of updater**: 2.18.2

* **Target**: mac, win, linux

Hello, 

I have some troubles with updating. 

Here is my build configuration (as example for mac-target):

`appId           : 'com.xx.de',
        productName     : 'myapp',
        copyright       : 'Copyright © 2018 year xx',
        directories     :{
            output  : 'dist'
        },
        publish: [{
            provider        : "github",
            owner           : "owner",
            repo            : "myapp",
            private         : true,
            token           : GIT_TOKEN
        }],
        npmRebuild      : true`

And here, a way how I'm trying update it:

`        
        const electronUpdater = require("electron-updater");

        _autoUpdater = electronUpdater.autoUpdater;

        _autoUpdater.setFeedURL({
            provider        : "github",
            owner           : "owner",
            repo            : "myapp",
            token           : GIT_TOKEN,
            private         : true
        });
        _autoUpdater.checkForUpdates();`

In addition I've include token into env and headers:

> _autoUpdater.requestHeaders    = { "PRIVATE-TOKEN": GIT_TOKEN };
>         process.env.GH_TOKEN                = GIT_TOKEN;

Updates are detected, but I have next error:

> Error: Cannot download "https://api.github.com/repos/owner/MyApp/releases/assets/6124771": Error: Cannot download "https://api.github.com/repos/owner/MyApp/releases/assets/6124771", status 400: Bad Request
>     at /Applications/MyApp.app/Contents/Resources/app.asar/node_modules/electron-updater/src/MacUpdater.ts:74:22

Full logs:

> [02.06 23:27:15.596][Electron][[0;37minfo[0m]: Start update checking.
> [02.06 23:27:15.683][Electron][[0;37minfo[0m]: Checking for update
> [02.06 23:27:15.683][Electron][[0;37minfo[0m]: Checking for update...
> [02.06 23:27:17.287][Electron][[0;37minfo[0m]: Found version 0.3.12 (url: MyApp-0.3.12.dmg)
> [02.06 23:27:17.288][Electron][[0;37minfo[0m]: Update available.
> [02.06 23:27:17.290][Electron][[0;37minfo[0m]: Downloading update from MyApp-0.3.12.dmg
> [02.06 23:27:17.372][Electron][[0;37minfo[0m]: / requested
[...]

Issue #2540, opened 2018-01-31
MacOS Auto-Updater Stopped Working with Github releases (public) returns 400 


Nederby:
* electron-builder version: **19.55.2**

* electron-updater: **2.20.1**

*  Target: **Mac**

**The auto-updater works fine on Windows (nsis)** - but on mac I start getting `status 400: Bad Request` in the log file. When i push out a new version. I thought it was a temporary Github issue - but it has been broken for little month.   

My update script
``` javascript
import { autoUpdater } from "electron-updater"

autoUpdater.checkForUpdatesAndNotify() 

```
If I do a normal `GET` request to https://api.github.com/repos/dixahq/dixa-desktop-app-release/releases/assets/6032765 everything looks fine. 

```
[2018-01-31 09:34:20.668] [info] Checking for update...
[2018-01-31 09:34:21.394] [info] Found version 2.0.6 (url: dixa-2.0.6-mac.zip, dixa-2.0.6.dmg)
[2018-01-31 09:34:21.396] [info] Update available.
[2018-01-31 09:34:21.396] [info] 2.0.6
[2018-01-31 09:34:21.397] [info] Downloading update from dixa-2.0.6-mac.zip, dixa-2.0.6.dmg
[2018-01-31 09:34:21.409] [info] / requested
[2018-01-31 09:34:21.412] [info] /app.zip requested
[2018-01-31 09:34:21.853] [warn] Error: Update download failed. The server sent an invalid response. Try again later.

[2018-01-31 09:34:21.853] [error] Error: Error: Update download failed. The server sent an invalid response. Try again later.
[2018-01-31 09:34:21.853] [info] Error in auto-updater.
[2018-01-31 09:34:21.853] [info] Error: Update download failed. The server sent an invalid response. Try again later.

[2018-01-31 09:34:21.854] [error] Error: Error: Cannot download "https://api.github.com/repos/dixahq/dixa-desktop-app-release/releases/assets/6032765": Error: Cannot download "https://api.github.com/repos/dixahq/dixa-desktop-app-release/releases/assets/6032765", status 400: Bad Request
    at /Applications/Dixa.app/Contents/Resources/app.asar/node_modules/electron-updater/src/MacUpdater.ts:74:22
    at ClientRequest.downloadRequest.httpExecutor.doRequest.downloadResponse (/Applications/Dixa.app/Contents/Resources/app.asar/node_modules/electron-updater/src/MacUpdater.ts:101:11)
[...]

PR #3168, opened 2018-07-25
Potential redirect fix

adrianmateoaea24:
I was having a problem with downloading updates from Github and some other people as well (#2540, #2559, etc.).

I found out that the problem was that the `redirect` function was not firing when downloading the update, therefore not deleting the `Authentication` field from the header and sending an invalid key to AWS.

### DIFF
--- a/packages/electron-updater/src/MacUpdater.ts
+++ b/packages/electron-updater/src/MacUpdater.ts
@@ -94,7 +94,8 @@ export class MacUpdater extends AppUpdater {
   }
 
   private doProxyUpdateFile(nativeResponse: ServerResponse, url: string, headers: OutgoingHttpHeaders, sha512: string | null, cancellationToken: CancellationToken, errorHandler: (error: Error) => void) {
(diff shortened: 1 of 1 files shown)


develar:
Thanks a lot. Proper solution will be get rid of this code — download using common util method and then pass local file to squirrel.mac. I will finally do it tomorrow.

develar:
Please fix electron-updater 3.1.0

sostenesapollo12:
> I was having a problem with downloading updates from Github and some other people as well (#2540, #2559, etc.).
> 
> I found out that the problem was that the `redirect` function was not firing when downloading the update, therefore not deleting the `Authentication` field from the header and sending an invalid key to AWS.

How can I use it instead my electron-updater ?

sostenesapollo12:
> Thanks a lot. Proper solution will be get rid of this code — download using common util method and then pass local file to squirrel.mac. I will finally do it tomorrow.



> 3.1.1

When i executed this ...

 ` loaded configuration file=package.json ("build" field)
At least electron-updater 4.0.0 is recommended by current electron-builder version. Please set electron-updater version to "^4.0.0"`

sostenesapollo12:
Someone can help me ?

PR #3198, opened 2018-08-01
electron-updater

Zendo93:
If the user uses custom nsis script from the scratch and he want's to avoid differential update when he is putting "differentialPackage": false into package.json under 
"build" : {
           "nsis" : {
               "differentialPackage": false
             }
}


develar:
In your custom NSIS script please use 

```
CopyFiles /SILENT "$EXEPATH" "$APPDATA\YOUR_APP_NAME\__installer.exe"
```

where `YOUR_APP_NAME` is app name that used as app data file name.

Issue #3254, opened 2018-08-18
--mac regression in 20.28.0 adding root folder to zip file


lutzroeder:
* **Version**: 20.28.0
* **Target**:  macOS

`npx electron-builder --mac` started producing zip files which contain `mac/Foo.app` instead of `Foo.app` in the root folder. 

Looks like this this started happening after updating from `20.27.1` to `20.28.0` and breaks `brew cask`. See Homebrew/homebrew-cask#50985.

@develar, efb40dabc608f9787e6ef9420e0fe71cfa97acf7 seems to be the only change in that release -- can you have a look?

PR #3277, opened 2018-08-26
Fix --mac ZIP root folder regression (#3254)

lutzroeder:
efb40da regressed `--mac` builds by adding a root folder to ZIP archives.


develar:
Thanks a lot.

PR #3300, opened 2018-09-05
snap: do not set environment for classic

sergiusens:
When setting an environment for classically confined snaps,
process that are spawned from within this environment will
contaminate the loading of the spawned applications on the
system causing symbol mismatches, for this reason it is
disabled, making library discovery rely on each binaries
RPATH.

Signed-off-by: Sergio Schvezov <sergio.schvezov@canonical.com>

--
I am not really much of a javascript developer, running `yarn test` produces this error which seems unrelated:
      ⨯ part download request failed with status code 403

### DIFF
--- a/packages/app-builder-lib/src/targets/snap.ts
+++ b/packages/app-builder-lib/src/targets/snap.ts
@@ -15,7 +15,7 @@ const defaultStagePackages = ["libasound2", "libgconf2-4", "libnotify4", "libnsp
 const defaultPlugs = ["desktop", "desktop-legacy", "home", "x11", "unity7", "browser-support", "network", "gsettings", "pulseaudio", "opengl"]
 
 export default class SnapTarget extends Target {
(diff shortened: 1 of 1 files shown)


develar:
Test added, verified manuallly:

```
cp: '/home/develar/.config/user-dirs.locale' and '/home/develar/.config/user-dirs.locale' are the same file
Cannot load module /snap/cl-co-app/x1/usr/lib/x86_64-linux-gnu/gtk-3.0/3.0.0/immodules/im-ibus.so: libibus-1.0.so.5: cannot open shared object file: No such file or directory
/snap/cl-co-app/x1/usr/lib/x86_64-linux-gnu/gtk-3.0/3.0.0/immodules/im-ibus.so does not export GTK+ IM module API: libibus-1.0.so.5: cannot open shared object file: No such file or directory
```

on Xubuntu. But app works and started.

develar:
20.28.4 released, thanks!

PR #3371, opened 2018-10-08
Added electron 3(Node 10) support to nisis updater

### LINKED ISSUE #3367: EACCES when using quitAndInstall electron 3
* **Version**: 
  - Builder: 20.28.4
  - Updater: 3.1.2

* **Target**: 
Windows

```
C:\Users\tiguerin\AppData\Roaming\BatchExplorer\__update__\BatchExplorer Setup 0.18.3-stable.99.exe EACCES
        at Process.ChildProcess._handle.onexit (internal/child_process.js:229:19)
        at onErrorNT (internal/child_process.js:406:16)
```

I believe the issue is that the `spawn` command doesn't actually throw an error that is catchable here https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/NsisUpdater.ts#L115

After testing locally I think the `try catch` needs to be changed to `on("error"` might have been a change to node 10 vs 8

This mock code for example works fine when using `.on("error"`
```ts
const { spawn } = require("child_process");

const installerPath = "C:\\Users\\tiguerin\\AppData\\Roaming\\BatchExplorer\\__update__\\BatchExplorer Setup 0.18.3-stable.99.exe";
const args = [];
const spawnOptions = {
    detached: true,
    stdio: "ignore",
};

const elevatePath = "C:\\Program Files\\BatchExplorer\\resources\\elevate.exe";

spawn(installerPath, args, spawnOptions).on("error", (e) => {
    // yes, such errors dispatched not as error event
    // https://github.com/electron-userland/electron-builder/issues/1129
    if ((e).code === "UNKNOWN" || (e).code === "EACCES") { // Node 8 sends errors: https://nodejs.org/dist/latest-v8.x/docs/api/errors.html#errors_common_system_errors
        console.log("Access denied or UNKNOWN error code on spawn, will be executed again using elevate")

        try {
            spawn(elevatePath, [installerPath].concat(args), spawnOptions).unref()
        } catch (e) {
            console.error("Elevate failed too", e);
        }
    } else {
        console.error("Error here", e);
    }
})
```

timotheeguerin:
fix #3367 

Looks like that in node 10 (Electron 3 use node 10) `spawn` is not throwing errors when the process fail to execute. This means the updater doesn't catch the `EACCES` error emitted when installer need admin permissions.

The fix will try to catch error both from node 8 and node 10 by listening to the error event for node 10 and using try catch for node 8. This should keep support for electron 2 while enabling electron 3.

### DIFF
--- a/packages/electron-updater/src/AppImageUpdater.ts
+++ b/packages/electron-updater/src/AppImageUpdater.ts
@@ -74,7 +74,7 @@ export class AppImageUpdater extends BaseUpdater {
     })
   }
 
-  protected doInstall(installerPath: string, isSilent: boolean, isRunAfter: boolean): boolean {
+  protected async doInstall(installerPath: string, isSilent: boolean, isRunAfter: boolean): Promise<boolean> {
     const appImageFile = process.env.APPIMAGE!!
     if (appImageFile == null) {
       throw newError("APPIMAGE env is not defined", "ERR_UPDATER_OLD_FILE_NOT_FOUND")
--- a/packages/electron-updater/src/BaseUpdater.ts
+++ b/packages/electron-updater/src/BaseUpdater.ts
@@ -33,7 +33,7 @@ export abstract class BaseUpdater extends AppUpdater {
     })
   }
 
-  protected abstract doInstall(installerPath: string, isSilent: boolean, isRunAfter: boolean): boolean
+  protected abstract doInstall(installerPath: string, isSilent: boolean, isRunAfter: boolean): Promise<boolean>
 
   protected async install(isSilent: boolean, isRunAfter: boolean): Promise<boolean> {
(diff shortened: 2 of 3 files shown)


dscalzi:
@develar 🙏

AhadCove:
This is a crucial PR. All updates are failing atm with the new Electron

dscalzi:
Yea this is really a show-stopper bug. Entire module not working with electron 3.

develar:
@timotheeguerin Thanks a lot. Sorry for delay. 

electron-updater 3.1.3 released. It is `next` release, please try it.

timotheeguerin:
Looks to work great! Thanks.

AhadCove:
Whooo, thank you all

PR #3400, opened 2018-10-17
fix(nsis): create appdata directory before copying installer

zhaoterryy:
First update gives error due to __installer.exe missing. 

This is due to it failing to copy over on first install because of no appdata folder yet.


develar:
Thanks!

Issue #2955, opened 2018-05-26
Error: Application entry file "main.js" in the ".../Contents/Resources/app.asar" does not exist. Seems like a wrong configuration.


nkint:
* **Version**: 
    "electron": "^2.0.2",
    "electron-builder": "^20.15.1",
    "electron-webpack": "^2.1.2"

    node v10.0.0
    yarn 1.6.0
    mac sierra 10.13.4

* **Target**: 
mac

I'm following this gist:
https://gist.github.com/matthewjberger/6f42452cb1a2253667942d333ff53404

Everything is ok, development is ok, but on `yarn dist`
I get this error:
```
Error: Application entry file "main.js" in the ".../Contents/Resources/app.asar" does not exist. Seems like a wrong configuration.
```

I found somewhere a workaround as add `"extends": null` in `build` in package.json but I get a compiled file (.app) that open a white page.
If I start the electron app created that shows only a white page with the inspector opened I see `app.js not found error`

My package.json build right now is:
```
"build": {
      "appId": "com.electron.electron-with-create-react-app",
      "win": {
          "icon": "https://cdn2.iconfinder.com/data/icons/designer-skills/128/react-256.png"
      },
      "directories": {
          "buildResources": "public"
      }
  },
```

I've tried to copy my `src/electron-starter.js` into `main.js` (into main project directory) and in `src/main.js` and into `public/main.js` as a workaround but the error is still there.

Do you have any hint?
Do I have to post something more about my actual project?

alice-em:
I'm getting a similar issue also.

```
> wonder-reader@1.0.0-a electron-pack /home/alice-mod/Documents/git/wonder-reader
> electron-builder -c.extraMetadata.main = build / electron.js

  • electron-builder version=20.15.1
  • loaded configuration file=package.json ("build" field)
  • loaded parent configuration preset=react-cra
  • writing effective config file=dist/electron-builder-effective-config.yaml
  • rebuilding native production dependencies platform=linux arch=x64
  • packaging       platform=linux arch=x64 electron=2.0.2 appOutDir=dist/linux-unpacked
Error: Application entry file "=" in the "/home/alice-mod/Documents/git/wonder-reader/dist/linux-unpacked/resources/app.asar" does not exist. Seems like a wrong configuration.
    at error (/home/alice-mod/Documents/git/wonder-reader/node_modules/electron-builder-lib/src/asar/asarFileChecker.ts:7:12)
[...]

PR #3434, opened 2018-10-31
docs: add note for `files` field

g-plane:
I encountered a similar error reported by electron-builder: https://github.com/electron-userland/electron-builder/issues/2955 . And @pcicho gave a solution ([link](https://github.com/electron-userland/electron-builder/issues/2955#issuecomment-393524832)), so I think this should be noted at the quick start guide.


develar:
Thanks.

PR #3448, opened 2018-11-05
 Electron builder yaml is now discovered through JS file

VictorLeach96:
Depends on https://github.com/electron-userland/electron-webpack/pull/221

Electron webpack does not respect the dist directory specified in `directories.output` so this PR allows the config to be passed through correctly.

### DIFF
--- a/packages/app-builder-lib/src/util/config.ts
+++ b/packages/app-builder-lib/src/util/config.ts
@@ -53,7 +53,7 @@ export async function getConfig(projectDir: string, configPath: string | null, c
       config.extends = extendsSpec
     }
     else if (devDependencies != null && "electron-webpack" in devDependencies) {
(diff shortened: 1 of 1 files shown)


packages/app-builder-lib/src/util/config.ts
```diff
@@ -53,8 +53,8 @@ export async function getConfig(projectDir: string, configPath: string | null, c
       config.extends = extendsSpec
     }
     else if (devDependencies != null && "electron-webpack" in devDependencies) {
-      extendsSpec = "electron-webpack/electron-builder.yml"
-      config.extends = extendsSpec
+      const electronWebpack = require("electron-webpack")
+      deepAssign(config, await electronWebpack.electronBuilderConfig())
```
develar: Old versions must be supported — please check that new logic is used only if `electronWebpack.electronBuilderConfig` is not null. `!= null`

PR #3449, opened 2018-11-05
feat: add more pkg configuration options

tlrobrn:
Add the ability to set a custom background image, welcome message, and
conclusion message to PKG installers.


develar:
Thanks! We should provide ability to use convention over configuration (default location for these files), but I will do it.

PR #3457, opened 2018-11-07
feat(nsis): add possibility to force an install mode programmatically

julianhille:
New custom macro in the multiui installer, which
allows to set two variables.
* $forceMachineInstall
* $forceCurrentInstall

Both allow to programmatically set '/currentuser'
or '/allusers'

I was unable to find the files for these docs: https://www.electron.build/configuration/nsis#custom-nsis-script and so the documentations are currently missing for that new custom macro.


develar:
Thanks!

PR #3503, opened 2018-11-27
icons.md: Refer to buildResources

aknuds1:
Be explicit about the `buildResources` directory instead of referring to its default value ("build").

### DIFF
--- a/icons.md
+++ b/icons.md
@@ -8,20 +8,20 @@ Files
 * *Optional* `background.png` (macOS DMG background).
 * *Optional* `background@2x.png` (macOS DMG Retina background).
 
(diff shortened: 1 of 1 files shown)


develar:
Why it is better? Currently, user don't need to guess what's default build dir.

aknuds1:
@develar You've got a point about people not having to look up the reference of `buildResources`, but referring to the `buildResources` directory as simply `build` is misleading on two accounts:
1. The link points to the `buildResources` configuration property, but the link text is `build` which makes you think electron-builder always find icons in the 'build/' directory. At least I did, because I didn't know any better.
2. 'build/' is merely the default for `buildResources`, it can be something else, which it was in my case. This made me very confused as to why my icon wasn't picked up, until I dug for a while and found out it was actually controlled via `buildResources`, which was set to something else.

For clarity, I will update my patch to point out that the default `buildResources` directory is build/.

aknuds1:
@develar I made the change clarifying that the default value of `buildResources` is build/.

develar:
Danke!

Issue #2534, opened 2018-01-29
Linux auto-update ERR_CONTENT_LENGTH_MISMATCH


logidelic:
* **Version**: 19.55.1
* **electron-updater**: 2.18.2
* **Target**: Linux

See attached screenshot. When one of my computers attempts to auto-update my electron app, the following **popup error dialog** displays on that computer:

![img_0511](https://user-images.githubusercontent.com/18287271/35536133-f8be5320-0513-11e8-8ee7-c173b3c11696.jpeg)

Aside from the fact that I don't know why this error is occurring (the file on the server seems correct and it works from another computer), I am much more concerned with the fact that:

- It popped up an error dialog. Is there a way to avoid this behavior?
- It did not log a corresponding error to autoUpdater.logger

afshing:
Mac also

iffy:
Yes, the title should be changed to also include macOS.

EtienneLem:
> It popped up an error dialog. Is there a way to avoid this behavior?

I believe the only way is to use `uncaughtException`:
```js
// Don’t show error popup to users
process.on('uncaughtException', (error) => {
  console.error(error)
})
```

I think that’s because errors are re-thrown (see #2398)

PR #3531, opened 2018-12-09
make "checkForUpdatesAndNotify()" promise actually catchable

vladimiry:
It makes the promise returned by checkForUpdatesAndNotify() call **actually** catchable.

Without this change possible checkForUpdatesAndNotify() promise rejection becomes `unhandled rejection` and so only global `process.on('unhandledRejection', ...` handler is able to catch the error which is not an acceptable solution.

The simplest way to reproduce the unhandled rejection caused by checkForUpdatesAndNotify() call is to turn off the network and call the checkForUpdatesAndNotify() method. Event if you setup `.catch` handler for this call you will still get unhandled rejection with Error: net::ERR_INTERNET_DISCONNECTED error.

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -246,8 +246,7 @@ export abstract class AppUpdater extends EventEmitter {
       return Promise.resolve(null)
     }
 
-    const checkForUpdatesPromise = this.checkForUpdates()
-    checkForUpdatesPromise
+    return this.checkForUpdates()
       .then(it => {
         const downloadPromise = it.downloadPromise
         if (downloadPromise == null) {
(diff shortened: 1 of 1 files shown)


vladimiry:
@EtienneLem re your comment https://github.com/electron-userland/electron-builder/issues/2534#issuecomment-376716890. I confirm that `checkForUpdatesAndNotify()` call is currently uncatchable and the only way to catch it is using global `process.on('uncaughtException' ` handler. Hope this PR gets merged, so we could simply do `checkForUpdatesAndNotify().catch` thing.

vladimiry:
Hi @develar, I'd appreciate any reaction to this simple PR, so I know if there a chance it gets merged in a near future. Rejection also works.

develar:
Thanks! electron-updater 4.0.7 published as pre-release version (`next`). Please verify.

vladimiry:
I just have tested no internet connection case running electron-updater@4.0.7 on Windows and I confirm that promise returned by `checkForUpdatesAndNotify()` call is now catchable, thanks a lot.

siebertm:
I can also confirm the same on linux:

* channel URL returns a 404,
* `checkForUpdatesAndNotify()` is catchable and does not crash the whole process anymore

Thanks!

Issue #3535, opened 2018-12-10
Nsis Updater - Custom Headers Not Sent when Downloading Blockmap


neilrees:
* Electron-Updater: 4.0.4
* Target: Windows 10 x64
* Updater: Nsis

When setting custom headers using `autoUpdater.requestHeaders` the custom headers are not sent on the request to download the `blockmap` file.

https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/NsisUpdater.ts#L146 the `requestHeaders` parameter is not used for the blockmap download.

The `downloadUpdateOptions` object has a property named `requestHeader` whereas `httpExecutor` expects a property named `headers`.

In our scenario we are setting an `Authorization` header and are seeing it fall back to doing a full download every time as it is not able to download the blockmap from our update server without providing that header.

PR #3536, opened 2018-12-10
Pass requestHeaders to httpExecutor

neilrees:
Proposed fix for https://github.com/electron-userland/electron-builder/issues/3535


stuartcusackie:
Does this change take into account the S3 region?

https://github.com/electron-userland/electron-builder/issues/4030

PR #3605, opened 2019-01-14
feat: add new option allowDowngradeOnlyIfChannelDiffers

Cselt:
With this small change you will be able to enable downgrade on channels.

This would be very welcomed in our application because we need to provide different version of the application in different networks. So let's say a user connects to Wifi channel "App1" we need to use application version 1.2.0. But if he connects to Wifi channel "App2" we need to change (downgrade) to version 1.1.0. But we don't wan't to use semver channels.

If you want to enable downgrade on all channels all you need to do is:
```
autoUpdater.allowDowngradeOnlyIfChannelDiffers = false;
autoUpdater.allowDowngrade = true;
```

Thanks


Cselt:
Hi @develar,

any thoughts about this PR? I would really appreciate if you could merge it, or just share your thought with me.

Thanks!

Issue #2057, opened 2017-09-11
NSIS "deleteAppDataOnUninstall": true doesn't work [BUG]


S74nk0:
* **Version**: electron-builder 19.27.7

* **Target**: Windows x64

With the **deleteAppDataOnUninstall** set to **true** it has no effect when uninstalling the aplication. 
This [line](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/templates/nsis/uninstaller.nsh#L83) in the NSIS templates has no effect, but when changing `RMDir /r "$APPDATA\${APP_FILENAME}"` to `RMDir /r "$APPDATA\package_name"` it deletes the app-data successfully. This was also an issue with the older electron-builder versions (v19.16.3 was the last one I have tried). 

"productName" and "name" are set. This is the nsis snippet: 

```
"nsis": {
      "include": "build/nsisExtra.nsh",
      "allowElevation": false,
      "perMachine": false,
      "oneClick": false,
      "multiLanguageInstaller": false,
      "warningsAsErrors": true,
      "deleteAppDataOnUninstall": true
    },

# nsisExtra.nsh
 !macro customInstall
   ExecWait "$INSTDIR\vc_redist_2013_x64.exe /install /passive /norestart /silent"
   ExecWait "$INSTDIR\vc_redist_2015_x86.exe /install /passive /norestart /silent"
   ExecWait "$INSTDIR\vc_redist_2015_x64.exe /install /passive /norestart /silent"
 !macroend
```

Regards.

develar:
You use API app.getPath, right?

develar:
Yes, default value of deleteAppDataOnUninstall is false.

genbandbrent:
For others that have encountered this issue, hope this helps. I found that my issue was that my productName and build.productName values didn't match, and that's why my AppData\Roaming\<app> folder was getting left behind on uninstall. 

To test it, I built an app with package.json name: "name", productName: "productName", build.productName: "build productName". Both build.mac and build.win sections had artifactName: "${productName}.${ext}"

Results
In my output directory (dist/) I had:
dist/mac/build productName.app, 
dist/win-unpacked/build productName.exe
In dist i had build productName.dmg, build productName.exe, build productName.zip

Installed the app on Windows. 
Executables went into C:\Users\<user>\AppData\Programs\name\ (i.e. package.json name).
Executables were called build productName.exe, Uninstall build productName.exe, etc.
(i.e. matching build.productName).

[...]

PR #3649, opened 2019-01-29
fix(nsis): when app is uninstalling, uninstaller does not remove app cache folder

sahinyyurt:
when app is uninstalling, uninstaller does not remove app cache folder which created by electron


sahinyyurt:
Fixes #2057

develar:
Thanks!

le4onardo:
Might I ask why was this closed? 
At my company we have the same [#2057](https://github.com/electron-userland/electron-builder/issues/2057#issuecomment-332661214
) issue
Changing `appId` to match `productName` is not an option because our product is already in prod.
Maybe I am missing something? we are using electron-builder 23.4.0

PR #3748, opened 2019-03-13
fixed unhandled error

uihoh0:
fix: Add a catch to due with the Error thrown in the method<checkAndNotifyPromise> of AppUpdater when the network is down(offline)

We don't want our users disturbed by an  unhandled error which showing up with the system-dialog. <checkAndNotifyPromise> is just a daily checking job.It should always hide in the background.


uihoh0:
Seems like your bucket is out of use.
> Error:Cannot find channel "latest.yml" update info: HttpError: 404 Not Found
"method: undefined url: https://electron-builder-test.nyc3.digitaloceanspaces.com/light-updater-test/latest.yml?noCache=1d5r0l3ks\n\nPlease double check that your authentication token is correct. Due to security reasons actual status maybe not reported, but 404.\n"

uihoh0:
Found that's a fixed issue.Close.

Issue #3681, opened 2019-02-11
EPERM error on windows during update process


scottnonnenberg-signal:
* **Version**: 
  electron-builder v20.38.5
  electron-updater v4.0.6
  https://github.com/signalapp/Signal-Desktop/blob/development/package.json#L116

* **Target**: 
macOS

We've had quite a few of our users run into this, a new issue appearing with a recent upgrade of Electron Builder/Updater. Maybe the file isn't fully downloaded and closed before we try to do something with it?

It seems to happen before the prompt is shown asking the user to update - you can see our update code here:

- Checking for new version: https://github.com/signalapp/Signal-Desktop/blob/84bb9be6a9c8bcd520301cc0220f4b7109cb0923/app/auto_update.js#L21-L27
- Listening for 'update-downloaded' event: https://github.com/signalapp/Signal-Desktop/blob/84bb9be6a9c8bcd520301cc0220f4b7109cb0923/app/auto_update.js#L77-L79
- Kicking off the upgrade with the downloaded file: https://github.com/signalapp/Signal-Desktop/blob/84bb9be6a9c8bcd520301cc0220f4b7109cb0923/app/auto_update.js#L49-L58

The callstack looks like this: 

```
Unhandled Promise Rejection
Error: EPERM: operation not permitted, open 'C:\Users\XXX\AppData\Local\signal-desktop-beta-updater\pending\temp-signal-desktop-beta-win-1.21.0-beta.4.exe'
    at ElectronHttpExecutor.download (C:\Users\XXX\AppData\Local\Programs\signal-desktop-beta\resources\app.asar\node_modules\electron-updater\src\electronHttpExecutor.ts:23:73)
    at Object.<anonymous> (C:\Users\XXX\AppData\Local\Programs\signal-desktop-beta\resources\app.asar\node_modules\electron-updater\src\NsisUpdater.ts:35:35)
From previous event:
    at Object.task (C:\Users\XXX\AppData\Local\Programs\signal-desktop-beta\resources\app.asar\node_modules\electron-updater\src\NsisUpdater.ts:31:11)
    at C:\Users\XXX\AppData\Local\Programs\signal-desktop-beta\resources\app.asar\node_modules\electron-updater\src\AppUpdater.ts:610:25
From previous event:
    at NsisUpdater.executeDownload (C:\Users\XXX\AppData\Local\Programs\signal-desktop-beta\resources\app.asar\node_modules\electron-updater\src\AppUpdater.ts:545:68)
[...]

PR #3845, opened 2019-04-21
fix(builder-util-runtime): temp-update file is not closed correctly when update is cancelled 

StaZhu:
temp-update file is not closed correctly when update is cancelled or an error has hanpened which makes this temp-update cannot be removed during the update process and causes `EPERM: operation not permitted ` error
#3681

### DIFF
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -431,6 +431,7 @@ function configurePipes(options: DownloadCallOptions, response: IncomingMessage)
   let lastStream = response
   for (const stream of streams) {
     stream.on("error", (error: Error) => {
+      fileOut.close()
       if (!options.options.cancellationToken.cancelled) {
         options.callback(error)
       }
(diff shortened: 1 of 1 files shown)


packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -439,7 +440,7 @@ function configurePipes(options: DownloadCallOptions, response: IncomingMessage)
   }
 
   fileOut.on("finish", () => {
-    (fileOut.close as any)(options.callback)
```
develar: Why callback is not passed anymore?

jsheffers:
Still receiving this error.  Can anyone verify that this fixes the issue?

ShawnKuan:
Thanks @StaZhu for your changes! I've been facing this issue for 2 days, and finally use your change do solve the problem. very helpful! Even though the owner didn't merged your branch

mmaietta:
This was marked stale because this wasn't answered:
>Why callback is not passed anymore?
```
-    (fileOut.close as any)(options.callback)
+    fileOut.close()
```
Seems like that could break previous functionality.

Issue #3751, opened 2019-03-13
install-app-deps: Configuring yargs through package.json is deprecated


ivancuric:
* **Version**: 20.39.0

* **Target**: 4.0.8 on platform=win32 arch=x64

I have a few local native modules linked via `file:` in `package.json`.

```
> electron-builder install-app-deps

Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.
Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.
  • electron-builder version=20.39.0
  • loaded configuration file=package.json ("build" field)
  • rebuilding native production dependencies platform=win32 arch=x64
```

What are the yargs errors about?

nathanlesage:
I just discovered them myself. It appears that this pertains to the command line arguments. For instance, what I've been doing in my `package.json` for the last year was put these commands in the `scripts` section:

```json
...
"scripts": {
        "postinstall": "electron-builder install-app-deps",
        "start": "electron .",
        "build:quick": "electron-builder --dir",
        "release:this": "electron-builder",
        "release:mac": "electron-builder --mac",
        "release:win": "electron-builder --win",
        "release:linux": "electron-builder --linux"
    },
```

And `yargs` is basically a JavaScript package that parses command line arguments.

So I'm pretty sure these warnings mean: "Please don't use `--dir`, `--mac` etc. anymore, but use the Javascript API directly!"

I.e.: Run electron-builder from within a node.js-script.

ivancuric:
https://www.electron.build/api/electron-builder
I don't see any way to run it using the provided API.

nathanlesage:
@ivancuric It's just weirdly documented.

As it says in the API docs:

![image](https://user-images.githubusercontent.com/17251683/54532521-37ad6080-4988-11e9-9da1-476666da04bf.png)

Raw Options refer to the CLI options and there you'll find this:

![image](https://user-images.githubusercontent.com/17251683/54532571-514ea800-4988-11e9-9617-bde6b4024683.png)

So CLI equals API arguments, i.e. you'd have to pass the command line arguments into the `package.json`s `build`-property and hope you place them at the right position in the tree.

-----

Downside of this is obviously that you'd need several different scripts to pass data to the builder so that it does not always trigger a full chain-rebuild if you just want to drop a testing app into the release directory, so you'd have to do something like this to avoid this:

```javascript
// Some script file, i.e. build.js
const builder = require('electron-builder')

[...]

PR #3852, opened 2019-04-24
fix(electron-builder): Configuring yargs through package.json is deprecated

firede:
Issue: #3751

1. remove `yargs` field from `package.json`.
2. upgrade `@types/yargs` for better type definitions.
3. move `parserConfiguration` to the entry point for every "cli".


develar:
Thanks, it was tricky for me :)

PR #3858, opened 2019-04-25
feat(mac): Add option to enable hardened-runtime

loremattei:
This PR adds the option to enable the _hardened runtime_ option during code sign for Mac builds. 
The option is already supported by _electron-osx-sign_, so it's just a matter of passing the flag down. 

Hardened runtime is a requirement for app notarization.

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -1847,6 +1847,11 @@
                     "description": "Whether a dark mode is supported. If your app does have a dark mode, you can make your app follow the system-wide dark mode setting.",
                     "type": "boolean"
                 },
+                "hardenedRuntime": {
+                    "default": false,
+                    "description": "Whether your app has to be signed with hardened runtime.",
+                    "type": "boolean"
+                },
                 "detectUpdateChannel": {
                     "default": true,
                     "description": "Whether to infer update channel from application version pre-release components. e.g. if version `0.12.1-alpha.1`, channel will be set to `alpha`. Otherwise to `latest`.",
@@ -2330,6 +2335,11 @@
                 "darkModeSupport": {
                     "default": false,
                     "description": "Whether a dark mode is supported. If your app does have a dark mode, you can make your app follow the system-wide dark mode setting.",
+                    "type": "boolean"
+                },
+                "hardenedRuntime": {
+                    "default": false,
+                    "description": "Whether your app has to be signed with hardened runtime.",
                     "type": "boolean"
                 },
                 "detectUpdateChannel": {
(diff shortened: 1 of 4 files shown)


sindresorhus:
I would recommend defaulting it to `true` as most will want to enable it since notarization (which requires hardened runtime) is required for apps with a new Developer ID and will be required for everyone in a future macOS release.

https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution

loremattei:
> I would recommend defaulting it to `true` as most will want to enable it since notarization (which requires hardened runtime) is required for apps with a new Developer ID and will be required for everyone in a future macOS release.

Yep! It's true that it will be required for everyone, though I defaulted it to false because: 
- _codesign_ and _electron-osx-sign_ default it to `false`, so it may be a bit confusing to have a different default here.
- It seems that a lot of Electron apps break when hardened runtime is enabled if proper entitlements are not provided, so I think that setting the default to true would be disruptive for many people. 

Does it make sense?

thelifenadine:
Thanks @loremattei, I'll be using this as soon as it gets released!

NickTikhonov:
This is both excellent and exciting @loremattei

easyCZ:
This feature is a must. Without it, `electron-builder` likely loses users as it would no longer provide the ability to support MacOS apps, which is its tagline...

Please consider merging this feature soon to continue enabling awesome cross platform packaging tools.

lottec:
Thanks for this @loremattei 💯 What's holding up merging? Keen to start using this...

develar:
Thanks @loremattei, sorry for delay.

martani:
Setting this flag to true causes the app to crash on startup. It's mentioned above that this could be the case of the proper entitlements are not set. I'm not setting any, are there examples of what entitlements need to be set to use this flag?

Thanks.

loremattei:
Hi @martani! You can find some info about the Hardened Runtime Entitlements here: https://developer.apple.com/documentation/security/hardened_runtime_entitlements. 

I'd suggest to start with looking at `allow-jit`, `allow-unsigned-executable-memory` and `allow-dyld-environment-variables`.

martani:
Thanks @loremattei. Adding those entitlements did the trick when Hardened Runtime is enabled.

1009466147:
hi I see you said you added a option, is  electron-osx-sign?, Where should I use it.

loremattei:
Hey @1009466147! It's `hardenedRuntime`. It defaults to `false` and you can enable it if you need to sign your app with the Hardened Runtime, for example if you want to notarize the app.

PR #3882, opened 2019-05-09
menuCategory enhancement

spikefu:
Added enhancements so menuCategory does what the code comments suggest it is supposed to do:
* Enable sub-directory in program files based on menuCategory or Author value in package.json
* Added Directory attribute to RemoveFolder tag in MSITarget so it passes wix MSI validation.


develar:
Thanks!

Issue #3504, opened 2018-11-27
afterSign hook is triggered before code signing happens


davej:
* **Version**:  20.37.0

* **Target**: Mac

Config option looks like:
```js
afterSign() {
  console.log("after sign handler called");
}
```

Console output looks like this.

```
after sign handler called
  • signing         file=/{snipped}/dist/mac/TestApp.app identityName=Developer ID Application: Dave Jeffery (SN1PP3D) identityHash=446B886736EC179EF88B902E0515A166202A5FFF provisioningProfile=none
```

Kilian:
Still the case in 20.41.0

Kilian:
It's probably because the call here: https://github.com/electron-userland/electron-builder/blob/ceafebbfda6c68dfb1da85879fd635ad1aae7f25/packages/app-builder-lib/src/platformPackager.ts#L250 

links to a function that resolves immediately: https://github.com/electron-userland/electron-builder/blob/ceafebbfda6c68dfb1da85879fd635ad1aae7f25/packages/app-builder-lib/src/platformPackager.ts#L332-L334

whereas the signing function for mac seems to be called just 'sign', and not 'signApp':
https://github.com/electron-userland/electron-builder/blob/ceafebbfda6c68dfb1da85879fd635ad1aae7f25/packages/app-builder-lib/src/macPackager.ts#L121

What happens is that in the following code, it checks for non-MAS, or target of more than one. If you're building just a DMG, then MAS will be false but it will also be creating a ZIP, setting the target count to two.
https://github.com/electron-userland/electron-builder/blob/09726952278c9977f81a530a10e9f87dcbfcc756/packages/app-builder-lib/src/macPackager.ts#L86-L91

In this function, the entire doPack function is called first (that function includes the `afterSign` call and the `await this.signApp` that immediately resolves) and only after that function is resolved, calls the local `sign` function.

It seems to me the macPackager file needs to be updated: the sign function needs to be renamed to signApp so it's correctly called in the doPack function, and `.then(() => this.sign(appPath, null, null))` on line 89 should be removed.

Kilian:
This ticket can be closed, now that #3912 is merged 🎉

Issue #3828, opened 2019-04-15
Code signing rejected - unnotarized Developer ID


mattgogerly:
* **Version**:  20.39.0

* **Target**:  platform=darwin arch=x64 electron=4.1.4

I've generated a Developer ID Application certificate using Xcode and I can see it in login on Keychain. When I run electron_builder I get the following error:

> Error: Command failed: spctl --assess --type execute --verbose --ignore-cache --no-cache /Users/mattgogerly/Documents/GitHub/MailBean/frontend/electron_dist/mac/MailBean.app
> /Users/mattgogerly/Documents/GitHub/MailBean/frontend/electron_dist/mac/MailBean.app: rejected
> source=Unnotarized Developer ID

I can't seem to find anything about unnotarized IDs with electron-builder, hence this issue. Any ideas how to fix?

martani:
I have the same issue after updating to `OSX 10.14.5`. This error does not happen on previous versions, notably `10.14.4`. Any ideas about what's causing this and how to fix?

dorokei:
Unfortunately, I also updated to OSX 10.14.5 and have the same problem😭

martani:
You can use [electron-notarize](https://github.com/electron-userland/electron-notarize) but the `afterSign` hook is broken on Mac, so this is still an issue. See [#3908](https://github.com/electron-userland/electron-builder/issues/3908).

You can also try to notarize the app manually: https://stackoverflow.com/questions/53112078/how-to-upload-dmg-file-for-notarization-in-xcode

wildcodeman:
I also updated to OSX 10.14.5 and then...😂

Here is my solution:
----
Edit file node_modules/electron-osx-sign/sign.js, find function verifySignApplicationAsync, comment section below, and every thing goes right!

```
  // Additionally test Gatekeeper acceptance for darwin platform
  //if (opts.platform === 'darwin' && opts['gatekeeper-assess'] !== false) {
  //  promise = promise
  //    .then(function () {
  //      debuglog('Verifying Gatekeeper acceptance for darwin platform...')
  //      return execFileAsync('spctl', [
  //        '--assess',
  //        '--type', 'execute',
  //        '--verbose',
  //        '--ignore-cache',
  //        '--no-cache',
  //        opts.app
  //      ])
  //    })
  //}
```

That code looks just test the app files, so remove it should be ok, I don't know why...but that fixed my problem...

will-stone:
[...]

PR #3912, opened 2019-05-21
fix: implement signApp function for macPackager

Kilian:
As per #3504, afterSign was called before the sign function was called on mac. I tracked this down to a logic error where signApp wasn't implemented for macPackager and thus immediately resolved. Mac then called the sign function directly, leading to the incorrect behaviour.

I was able to successfully build with this, and have the afterSign function called at the right time, with the changes below. I'm however not familiar with typescript (or electron-builders code standards) so likely I'll need to make some changes. This can serve as a first setup.

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -1,7 +1,7 @@
 import BluebirdPromise from "bluebird-lst"
 import { deepAssign, Arch, AsyncTaskManager, exec, InvalidConfigurationError, log, use } from "builder-util"
 import { signAsync, SignOptions } from "electron-osx-sign"
-import { ensureDir } from "fs-extra-p"
+import { ensureDir, readdir } from "fs-extra-p"
 import { Lazy } from "lazy-val"
 import * as path from "path"
 import { copyFile, unlinkIfExists } from "builder-util/out/fs"
@@ -9,7 +9,7 @@ import { orIfFileNotExist } from "builder-util/out/promise"
 import { AppInfo } from "./appInfo"
 import { appleCertificatePrefixes, CertType, CodeSigningInfo, createKeychain, findIdentity, Identity, isSignAllowed, reportError } from "./codeSign/macCodeSign"
 import { DIR_TARGET, Platform, Target } from "./core"
-import { ElectronPlatformName } from "./index"
+import { AfterPackContext, ElectronPlatformName } from "./index"
 import { MacConfiguration, MasConfiguration } from "./options/macOptions"
 import { Packager } from "./packager"
 import { chooseNotNull, PlatformPackager } from "./platformPackager"
@@ -306,16 +306,13 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       Object.assign(appPlist, extendInfo)
     }
   }
-  
+
   protected async signApp(packContext: AfterPackContext, isAsar: boolean): Promise<any> {
     const appFileName = `${this.appInfo.productFilename}.app`
-    if (this.platformSpecificBuildOptions.signAndEditExecutable === false) {
-      return
-    }
 
     await BluebirdPromise.map(readdir(packContext.appOutDir), (file: string): any => {
       if (file === appFileName) {
-        return this.sign(path.join(packContext.appOutDir, file))
+        return this.sign(path.join(packContext.appOutDir, file), null, null)
       }
       return null
     })
@@ -327,7 +324,7 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     const outResourcesDir = path.join(packContext.appOutDir, "resources", "app.asar.unpacked")
     await BluebirdPromise.map(orIfFileNotExist(readdir(outResourcesDir), []), (file: string): any => {
       if (file.endsWith(".app")) {
-        return this.sign(path.join(outResourcesDir, file))
+        return this.sign(path.join(outResourcesDir, file), null, null)
       }
       else {
         return null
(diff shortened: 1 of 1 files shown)


martani:
I'm interested in testing this, how can I fork a local version of electron builder and reference in from my app?

It doesn't seem to work when I simply reference the package this way in package.json:
```
"devDependencies": {
    "electron-builder": "file:../electron-builder/packages/electron-builder"
}
```

delusive:
I tried this and I am still receiving "Unnotarized Developer ID" from this line https://github.com/electron-userland/electron-builder/pull/3912/files#diff-2f245a25ceda557ecf2530ec7e88c805R315, before it even runs the afterSign hook.  Maybe I am doing something wrong here.

delusive:
> I tried this and I am still receiving "Unnotarized Developer ID" from this line https://github.com/electron-userland/electron-builder/pull/3912/files#diff-2f245a25ceda557ecf2530ec7e88c805R315, before it even runs the afterSign hook. Maybe I am doing something wrong here.

I believe this might be due to the "hardenedRuntime": true functionality here: https://github.com/electron-userland/electron-builder/pull/3858

martani:
@delusive how were you able to test this? I can notarize my app manually but need to test doing it in the afterSign step.

delusive:
> @delusive how were you able to test this? I can notarize my app manually but need to test doing it in the afterSign step.

I cloned Kilian:patch-1, and then compiled the code, I then replaced packages/app-builder-lib/src/macPackager.ts in my local node_modules with the new one.

linuxjuggler:
> > I tried this and I am still receiving "Unnotarized Developer ID" from this line https://github.com/electron-userland/electron-builder/pull/3912/files#diff-2f245a25ceda557ecf2530ec7e88c805R315, before it even runs the afterSign hook. Maybe I am doing something wrong here.
> 
> I believe this might be due to the "hardenedRuntime": true functionality here: #3858

@delusive  You will need to use the pre-released version of electron-builder which is version 20.41.0 and add the `"hardenedRuntime": true,` option to your mac build section.
let me know if it will work with you, am stuck with the rejection all the time now.

Thanks

develar:
Explanation: `sign` is removed from the promise chain after `doPack`, but `signApp` is called as part of `doPack`.

develar:
Our code sign tests not executed locally due to security reasons, buy I set `CSC_KEY_PASSWORD` and verify it manually in addition to CI tests on Travis.

Thank you! I will prepare release today.

linuxjuggler:
Hey @Kilian may I ask you to show your `afterSign` script? for some reason now that am using the prerelease version of electron-builder with your fix, I can't any result from the script, always get :

```shell
/Users/zaher/Sites/electron/my app.app: rejected
source=Unnotarized Developer ID
```

delusive:
I built the newest version from master and installed from the local filesystem into my project.  I have added "afterSign": "./after_sign_hook.js" in the build section of package.json along with "hardenedRuntime": true under the mac build section of package.json.  My after_sign hook contains the following code:

```
const enotarize = require("electron-notarize");

exports.default = async function(context) {
    options = {
        appBundleId: "***",
        appPath: context.appOutDir,
        appleId: "********",
        appleIdPassword: "******"        
    };

    console.log("Notarizing application: " + options)

    await enotarize.notarize(options);

};
```

When I run electron-builder I get the Unnotarized Developer ID error.  The interesting thing is that I don't get my output message "Notarizing application: " in the output, meaning it hasn't tried to notorize the application before returning the error.

Any help/example would be greatly appreciated.

Thanks.

Kilian:
@delusive your app path needs the actual app in addition to the directory! so something like
```
AppPath: `${context.appOutDir}/myApp.app`
```

martani:
This is the `afterSign` hook I use for notarization. It works well with the `.app` output. (Side note: it also appears that notarizing the `dmg` also notarizes the `.app` inside.)

```ts
const fs = require('fs');
const path = require('path');
var electron_notarize = require('electron-notarize');

module.exports = async function (params) {
    // Only notarize the app on Mac OS only.
    if (process.platform !== 'darwin') {
        return;
    }
    console.log('afterSign hook triggered', params);

    // Same appId in electron-builder.
    let appId = '<Change this to your appId ad defined in your electron-builder config>'

    let appPath = path.join(params.appOutDir, `${params.packager.appInfo.productFilename}.app`);
    if (!fs.existsSync(appPath)) {
        throw new Error(`Cannot find application at: ${appPath}`);
    }

    console.log(`Notarizing ${appId} found at ${appPath}`);

    try {
        await electron_notarize.notarize({
            appBundleId: appId,
            appPath: appPath,
            appleId: process.env.appleId,
            appleIdPassword: process.env.appleIdPassword,
        });
    } catch (error) {
        console.error(error);
    }

    console.log(`Done notarizing ${appId}`);
};
```

Kilian:
@martani it does, but I don't think it staples the .app, which means it can't be verified offline. So notarizing the app and stapling it before packaging it into the dmg (and then notarizing that) is probably best.

martani:
Thanks @Kilian, that's what I ended up doing. This adds a lot of latency to the build process, notarizing the app takes ~10 minutes + an extra ~10 minutes for the dmg.

Unfortunately, that does not seem to fully solve the issue of users not able to open the dmg on macOS 10.14.5 as I reported here: https://github.com/electron-userland/electron-builder/issues/3828#issuecomment-495860811

DuBistKomisch:
I think this has broken our use case of passing electron-builder a pre-built but unsigned .app and turning it into a .pkg, signing both the .app and .pkg. It no longer signs the .app at all now. I guess this is a bit of a dumb use case anyway, I can just run `electron-osx-sign` and `electron-notarize` first instead, but just mentioning anyway.

damianobarbati:
@martani @delusive where did you find the `appleId` and `applePassword`?

Issue #3870, opened 2019-05-02
Notarize app for macOS


pashvin:
Is there any way to notarize app using electron builder?
Looks like Apple is going to force it soon. see link below
https://developer.apple.com/news/?id=04102019a

There is one npm module available but it will be nice this is supported by electron-builder.
https://github.com/electron-userland/electron-notarize

martani:
I have users on OSX 10.14.5 reporting that the electron app can't be opened anymore.

Running `spctl -a -t open --context context:primary-signature -v <pathToDMG>` returns the following error:
```
Downloads/<pathToDMG>.dmg: rejected
source=Unnotarized Developer ID
```

It would be great if we have some option to notarize the app with `electron-builder`.

Kilian:
ticket #3908 mentions running `electron-notarize` yourself in the `afterAllArtifactBuild` step, but probably due to timing, this interferes with the publish step so is less than ideal. Having this integrated would be super helpful.

Kilian:
Also, this is blocked by #3504, otherwise you could roll your own with `electron-notarize` in the `afterSign`(not in the `afterAllArtifactBuild` step).

lane-c-wagner:
Yeah this is a big issue :( all our Mac installs started breaking. Trying to figure out how we can do notarization using electron builder on a ci server (Travis) if anyone figures it out let me know!

Kilian:
I have a fix in #3912, once that lands you can call `electron-notarize` in the `afterSign` step and everthing works as expected. I'll write a blogpost about it once it's merged in/published :)

lane-c-wagner:
You are a hero sir.

lane-c-wagner:
> I have a fix in #3912, once that lands you can call `electron-notarize` in the `afterSign` step and everthing works as expected. I'll write a blogpost about it once it's merged in/published :)

You're a a hero, I'm excited.

Kilian:
There's a few more things that need to happen, it seems. The DMG is also signed by electron-builder. and anything that gets signed needs to be notarized or it won't be allowed to install. So we need one of two options:

* A setting to turn off signing the DMG. This is less than ideal, but workable. This creates an unsigned, un-notarized DMG that installs a signed and notarized app. You'll likely get the unknown developer message.
* An extra hook called after creating the dmg but before any of the `electron-updater` files (blockmap, latest-mac.yml) are created. The notarization process "staples" the dmg so the file changes, so it needs to run before any of the updater files are created.

[...]

PR #3941, opened 2019-06-05
Adding ‘sign’ dmg option to scheme.json to fix build errors

JoshRiser:
This solves a build issue with the 'sign' dmg configuration mentioned in #3870

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -804,7 +804,7 @@
                     "description": "The DMG windows position and size."
                 },
                 "sign": {
-                    "default": true,
+                    "default": false,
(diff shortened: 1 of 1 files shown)


packages/app-builder-lib/scheme.json
```diff
@@ -802,6 +802,11 @@
                 "window": {
                     "$ref": "#/definitions/DmgWindow",
                     "description": "The DMG windows position and size."
+                },
+                "sign": {
+                    "default": true,
```
rajivshah3: According to https://github.com/electron-userland/electron-builder/commit/ef0028d#diff-95f823016d05ece1f904f6c59a352ebaR190 the default is `false`. That file might need to be updated

Issue #3962, opened 2019-06-12
It's not possible to include $schema property into electron-builder.json file


Eshva:
* **verson**: 20.43.0

* **target**: any 

There is a schema for ```electron-builder.json``` file located in https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/scheme.json but it has a bug. You can't include this file as the value for ```$schema``` property of a JSON file because no additional files are allowed and in the schema no property ```$schema``` defined.
I've fixed this bug and make a pull request #3963 to add the fix to this repo.
As well I plan to publish this updated schema on http://schemastore.org/json/.

Eshva:
I've pushed a PR but it broke the ```doExtraResourcesTest``` test on AppVeyor. I believe I couldn't break that test.

PR #3963, opened 2019-06-12
fix: allow $schema property in electron-builder.json

Eshva:
It should be allowed to include $schema property in JSON file.

Issue: #3962


Eshva:
I just updated a JSON schema file with a new property.
I don't understand how my commit could break this test ```doExtraResourcesTest```. Seams like it's an infrastructure problem.
Please check it out and apply my PR.

PR #3966, opened 2019-06-13
Improve DMG icon coordinates documentation

ezzatron:
When configuring the coordinates of icons in a DMG, I had a hard time figuring out how the X and Y values actually related to the output I was seeing. I had to do quite a bit of searching to find [this documentation](https://metacpan.org/pod/distribution/Mac-Finder-DSStore/DSStoreFormat.pod) (emphasis mine):

> `'Iloc'`
>
> 16-byte blob, attached to files and directories. The file's icon location. Two 4-byte values **representing the horizontal and vertical positions of the icon's center (not top-left)**. (Then, 6 bytes 0xff and 2 bytes 0?) For the purposes of the center, the **icon only is taken into account, not any label**. The icon's size comes from the icvo blob.

In this PR, I've distilled the information and added it to the DMG configuration schema that drives the relevant documentation.

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -650,9 +650,11 @@
                     "type": "string"
                 },
                 "x": {
+                    "description": "The device-independent pixel offset from the left of the window to the **center** of the icon.",
                     "type": "number"
                 },
                 "y": {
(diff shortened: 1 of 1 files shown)


develar:
Thanks a lot.

Issue #3163, opened 2018-07-24
Electron Updater: Error checking for updates on Private GitHub repo


hicaro:
* **electron-builder**: 20.25.0

* **electron-updater**: 3.0.3

* **Target**: macOS (High Sierra)

I have encountered an error running electron-updater against GitHub enterprise:

```js
[11:02:13.525] [error] Error: TypeError: Cannot read property 'assets' of undefined
    at /Users/user/projects/becks/prodsuite/MyProject/out/mac/MyProject.app/Contents/Resources/app.asar/node_modules/electron-updater/src/providers/PrivateGitHubProvider.ts:31:31
```

This basically happened due to the following facts:

* All releases were published to GitHub using `alpha` in the version number, e.g. `v0.1.2-alpha`;
* Releases were not published as pre-releases (did not check the `This is a pre-release` checkbox);
* electron-update automatically sets `allowPrerelease` to true in case of containing an pre-release component such as `alpha` in the version name, according to the [documentation](https://www.electron.build/auto-update/#appupdater-eventemitter);

Whenever there is a check for updates, the function `getLatestVersion()` is called (in my case since I am using a private repo), which will call `getLatestVersionInfo`.
> line 26 of electron-builder/packages/electron-updater/src/providers/PrivateGitHubProvider.ts 

```js
const releaseInfo = await this.getLatestVersionInfo(cancellationToken)
const asset = releaseInfo.assets.find(it => it.name === channelFile) // error happens here
```

As it was detected the `alpha` in my version number, setting `allowPrerelease` to true, in `getLatestVersionInfo`, it fetched all releases not just the latest one. Then, within the function, upon receiving the http response, the following happened:

```js
if (allowPrerelease) {
    version = version.find((v: any) => v.prerelease)
}
```

As I did not have any release with `prerelease` set to true, `version` became `undefined`. This value was returned and assigned to `releaseInfo` in getLatestVersion.
When the code tried to extract assets from releaseInfo, it errored out.

The way I see the fix for this is to either throw an error if version becomes `undefined` within getLatestVersionInfo or make it check for undefined in `releaseInfo` in getLatestVersion.

I can make the change a submit a PR, if you guys prefer. I just wanted to check what would be the expected behavior for this.

Full functions for reference:
```js
async getLatestVersion(): Promise<PrivateGitHubUpdateInfo> {
    const cancellationToken = new CancellationToken()
    const channelFile = getChannelFilename(getDefaultChannelName())

[...]

PR #3999, opened 2019-06-27
fix(electron-updater): update from prerelease to prerelease crash

alexandrricov:
Updater is checking current app version, and if it contains alfa or beta etc., allowPrerelease became true.
If no one update contains prerelease github flag – app crashes.
If next release (sorted by semver) contains prerelease flag – then if will be installed, but I can't see how to exit from prerelease circle using current implementation.

closes: https://github.com/electron-userland/electron-builder/issues/3163


develar:
Indeed, thanks!

develar:
electron-updater 4.0.14 released.

PR #4002, opened 2019-06-29
corrected path for binary in package

Toothless204:
Debian packages should place binaries in /usr/bin rather than /usr/local/bin see [this](https://www.debian.org/doc/manuals/maint-guide/modify.en.html#destdir) for details


develar:
Ok, as we soon will release 21 (breaking release).

PR #4027, opened 2019-07-08
Add fileAssociation via mimeType for fpm target

patsimm:
I added support for fileAssociation in Linux via the fpm target. Therefore I followed the instructions from freedesktop.org: https://www.freedesktop.org/wiki/Specifications/AddingMIMETutor/

What is not covered by my PR is adding Icons to MIME types. But as stated in the current documentation I defaulted the icon to `x-office-document`

> - `icon` String - The path to icon (`.icns` for MacOS and `.ico` for Windows), relative to `build` (build resources directory). Defaults to `${firstExt}.icns`/`${firstExt}.ico` (if several extensions specified, first is used) or to application icon.
>
>   Not supported on Linux, file issue if need (default icon will be `x-office-document`).


patsimm:
Since we would really like to see this feature in our app, I would appreciate any information about what it is that is holding back from merging this PR.

develar:
Sorry for delay, I will take a look in 2 days.

patsimm:
@develar Do you see any other way that we can take to be able to generate .deb packages that add their fileAssociations?

develar:
@patsimm What do you mean? Your solution works for me and release will be in several minutes.

patsimm:
> @patsimm What do you mean? Your solution works for me and release will be in several minutes.

Sorry, I've only seen 'closed' and saw that it was actually added a little too late. Thanks for your work!

Issue #4068, opened 2019-07-18
NSIS Uninstaller registry entry format change 


abhishekmatcha:
* **Version**:  21.1.1

* **Target**:  Windows

In windows Registry entry for uninstaller is located at this path `Software\Microsoft\Windows\CurrentVersion\Uninstall\`  with just the guid  as `Example: \SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\ddd366r9-367b-5d51-b552-9754365`  but the convention to store the key with id in registry  is  `\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{ddd366r9-367b-5d51-b552-9754365}`. It was not a bigger issue and won't cause any impact but following may be better.
`Note:- Any suggestions from yous side is also appreciated`.

abhishekmatcha:
please review the PR: `https://github.com/electron-userland/electron-builder/pull/4069`

PR #4069, opened 2019-07-18
Fix-[#4068]:NSIS Uninstaller registry entry format change

abhishekmatcha:
1) In windows Registry entry for uninstaller is located at this path 
`Software\Microsoft\Windows\CurrentVersion\Uninstall\` with just the guid as Example: `\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\ddd366r9-367b-5d51-b552-9754365` but the convention to store the key with id in registry is `\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\\{ddd366r9-367b-5d51-b552-9754365}`.

### DIFF
--- a/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
+++ b/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
@@ -167,7 +167,7 @@ export class NsisTarget extends Target {
       APP_PACKAGE_NAME: appInfo.name
     }
     if (uninstallAppKey !== guid) {
-      defines.UNINSTALL_REGISTRY_KEY_2 = `Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\${guid}`
+      defines.UNINSTALL_REGISTRY_KEY_2 = `Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{${guid}}`
     }
 
     const commands: any = {
--- a/packages/app-builder-lib/templates/nsis/multiUser.nsh
+++ b/packages/app-builder-lib/templates/nsis/multiUser.nsh
@@ -6,7 +6,7 @@
 
 # allow user to define own custom
 !define /ifndef INSTALL_REGISTRY_KEY "Software\${APP_GUID}"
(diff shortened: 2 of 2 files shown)


mastergberry:
@abhishekmatcha this MR caused an issue it seems where there is a glitched entry now for the old GUID still in the Programs & Features section.

New format:

`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{1de14785-dd8c-5cd2-aae8-d4a376f81d78}`

Old format:

`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\1de14785-dd8c-5cd2-aae8-d4a376f81d78`

Unfortunately any users that upgraded from the old to new format have now 2 "installations" in their Programs & Features on Windows. They can't remove the second one though, because the first actually already did the uninstallation.

For now I am writing a one time script in nsis to clean up this old bugged entry for my application...but this should probably be fixed for everyone too.

pronebird:
Yeah what a bloody mess. Why would this be merged without a migration path?

andrienko:
This basically means that we can not update `electron-builder` beyond `21.1.2`, because it corrupts the nsis updates (because of the reg keys changed and it does not clean the installation folder prior to installation).

Also, we can't use the electron `v6`, because it requires at least `electron-builder` version 22, otherwise it crashes on MacOS.

Nice.

damianobarbati:
@andrienko I'm stuck as well: can't downgrade because of hardenedRuntime needed by Catalina: did you find solution?

menelike:
@andrienko 

We're in the same situation. https://github.com/electron-userland/electron-builder/pull/4443 addresses this, but afaik the only proper solution to this is to backport that change in (your own fork) `21.2.0`. Then compile, submodule & local npm install `app-builder-lib` from that. You could also publish an backports version. Whatever you/we currently choose, it's a mess...

laurent22:
Why was this pull request merged in such a careless way? No review, no question asked, just merge without thinking of the consequences. What a complete mess, and not a word from the maintainers on how to fix this.

PR #4111, opened 2019-07-31
fix electron mac : add needed helpers for electron 6.0.0

Evoks:
add needed helpers (Renderer, Plugin, GPU) to fix the crash of electron 6.0.0 app at start up


neilrees:
@Evoks The tests are failing on cicleci since this was merged: https://github.com/electron-userland/electron-builder/blob/master/test/src/mac/macArchiveTest.ts

matheo:
my App is crashing on start because the `Helper (GPU)` throws `Could not set sandbox profile data: Operation not permitted (1)`... this was supposed to detect it right? what can I do to force its support?

PR #4131, opened 2019-08-08
fix electron-forge-maker-appimage

Marcus10110:
AppImage does not seem to work from Electron Forge. issue here:
https://github.com/electron-userland/electron-forge/issues/26

The problem seems to be that the 'electron-forge-maker-appimage' package exports a function, when electron-forge expects a class that implements a specific interface. Details in the issue link above.

I'm not sure if this is the best way to solve this. In the electron-builder repository, all four of the 'electron-forge-maker-* targets are implemented in the same way.

The right solution might be to leave this as-is, and implement this class in the electron-forge repository in this directory:
https://github.com/electron-userland/electron-forge/tree/master/packages/maker

Another possible solution would be to keep this class out of both repositories, and maintain it in a separate repository entirely.

Feedback welcome.

### DIFF
--- a/packages/electron-forge-maker-appimage/.gitignore
+++ b/packages/electron-forge-maker-appimage/.gitignore
@@ -1,0 +1,1 @@
+/node_modules/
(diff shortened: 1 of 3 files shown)


Marcus10110:
Updated the PR include '@electron-forge/maker-base',, so the proper base class can be used.

Marcus10110:
Closing because this is probably not the best solution, it appears that makers in this repository that target electron-forge use `buildForge`, which is part of `app-builder-lib` and doesn't look like it works quite properly.

Issue #4151, opened 2019-08-12
Invalid Bundle Identifier Electron Helper (GPU, Plugin, and Renderer) Mac App Store


dimitarnestorov:
* **Version**: 21.2.0
* **Target**: macOS App Store
* **Electron Version**: 6.0.1

Couldn't find a duplicate of my issue which I feel is odd. Can anyone confirm  Electron 6+ on the App Store using electron-builder?

![image](https://user-images.githubusercontent.com/8790386/62866022-91883b80-bd18-11e9-84f8-2f6055e5a0f2.png)

Pretty sure the problem is the brackets, I believe the uppercase letters isn't.

gaodeng:
related https://github.com/electron-userland/electron-builder/issues/4104

gaodeng:
workaround - change node_modules/app-builder-lib/out/electron/electronMac.js line 178
to 
```
helper.CFBundleIdentifier = `${helperBundleIdentifier}.${postfix.replace(/[^a-z0-9]/gim,"")}`;

```

dangeredwolf:
I had the same issue, thankfully @gaodeng's solution seems to have worked.

PR #4154, opened 2019-08-13
fix Invalid Bundle Identifier Electron Helper (GPU, Plugin, and Rende…

gaodeng:
fix https://github.com/electron-userland/electron-builder/issues/4151


dimitarnestorov:
I can confirm that it works

alibosworth:
This seems really important.  Is there a reason it has been sitting this long?

Issue #3872, opened 2019-05-03
Packaging Electron v5 with sandbox (linux)


thomasnordquist:
With electron 5 the sandbox is enabled by default. The sandbox requires elevated rights to run.
This can be achieved if the kernel has `kernel.unprivileged_userns_clone` enabled or a SUID flag is set on `chrome-sandbox`.

If `unprivileged_userns_clone` is disabled (default for Arch) the SUID-flag is required.

To disable / enable said kernel feature:
```
sudo sysctl kernel.unprivileged_userns_clone=0
```

Snaps cannot use the SUID flag (automatic app review) and seem to prevent access to the kernel feature.

One can disable the sandbox by running electron with the `--no-sandbox` argument.

As a workaround, I currently use an after-pack hook to wrap the electron binary and append the `--no-sandbox` argument, which seemed to be to be the only way to add the sandbox argument. (`app.commandLine.appendSwitch` does not work since electron has already launched)
https://github.com/thomasnordquist/MQTT-Explorer/blob/master/scripts/afterPack.ts

Electron issue concerning this: https://github.com/electron/electron/issues/17972

burtonator:
Can't the old behavior be resolved by specifying ```sandbox: false``` in the browser window constructor?

Seems like it could be a viable workaround. I've disabled our snaps for now but would be nice to have a workaround.

burtonator:
sandbox: false doesn't seem to fix thigns. I still get:

"Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted
Trace/breakpoint trap (core dumped)
"

burtonator:
This might be possible if we pass --no-sandbox to the electron command line but I can't figure out how to set this in electron-builder.

thomasnordquist:
To my knowledge, as of now, one can not disable the sandbox/sandbox-helper with an electron builder configuration.

@burtonator see the initial comment for the workaround I chose to use for snaps and AppImage builds, (afterPack hook/electron wrapper script).
I believe there is still a lot to do for different parties (electron, snapcraft) until the sandbox feature will be available for snaps in snapcraft.

An electron-builder build-option to disable/configure the sandbox would be better, but until the situation is resolved I'll use the workaround.

burtonator:
I implemented a hack where I renamed the main script to myapp.bin and then added the --no-sandbox argument manually and it worked fine.  Not pretty but at least we have code pushed!

[...]

PR #4163, opened 2019-08-16
fix(deb): chmod SUID bit on chrome-sandbox for debs

jryans:
Add SUID bit to packaged `chrome-sandbox` so that it works out of the box on
distros like Debian.

Part of https://github.com/electron-userland/electron-builder/issues/3872


jryans:
The CI failures appear to be unrelated to this change. It seems like the past few commits were failing as well.

Issue #4213, opened 2019-09-04
updater: allowDowngrade doesn't allow downgrades on the same channel


neilrees:
* **electron-builder**: 20.44.4
* **electron-updater**: 4.1.2
* **Target**: Windows NSIS

Version downgrades on the same channel are currently blocked even when `allowDowngrade` is true.
`Update for version 1.0.3 is not available (latest version: 1.0.2, downgrade is allowed).`

allowDowngrade will only install the lower version number if the prerelease channel is different between the version number specified by the information downloaded from the .yml and the currently installed version.

Is there any reason for this rule?  Would you accept a PR that deleted this check? https://github.com/electron-userland/electron-builder/blob/ff242ab865f65987da4d95183a99c6fbaa7d3b78/packages/electron-updater/src/AppUpdater.ts#L327

PR #4218, opened 2019-09-06
When AllowDowngrade is true and latest version is older, take the update

neilrees:
https://github.com/electron-userland/electron-builder/issues/4213

Change to take an update to an older version when `AllowDowngrade` is true, and the latest version is less than the currently installed version.


pianomanjh:
Please merge this, we're suffering from the same issue

customcommander:
@neilrees Thanks for this. However the doc here https://www.electron.build/auto-update doesn't seem to reflect that change. I understand that this has lifted the "different channel" restriction. Is that right? Do you think you could submit a PR for this as well? :)

PR #4244, opened 2019-09-17
Bug fix for App sandbox not enabled (ITMS-90296)

vaibhavtel:
The issue in detail is explained here
https://github.com/electron/electron-osx-sign/issues/192


lutzroeder:
@vaibhavtel this change is crashing existing macOS apps (#4390). Why was this not validated?

#4414 to revert the change until a more complete fix becomes available.

Issue #1175, opened 2017-01-26
Feature request: Custom Request Headers for Electron Updater


kevinphelps:
I'd like to be able add custom request headers (specifically, an authorization header) to the HTTP requests sent by electron updater. Perhaps as in config parameter passed to `autoUpdater.checkForUpdates()`?

develar:
`requestHeaders` option should be added to [AppUpdater](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/AppUpdater.ts) base class.

kevinphelps:
Thanks for that info. I am happy to send a pull request.

PR #4250, opened 2019-09-19
Feature/export updaters for configurability

jeffreypriebe:
Expose Updaters downstream so configs (especially requestHeaders) can be passed in.

Custom request header supports was added in issue #1175 / commit dd1320d but those are _not_ exposed to callers.
This PR exposes those by:
1. Making the updaters available directly (0330632)
2. Setting the value `requestHeaders` in the `AppUpdater` constructor (bcae4e8)


jeffreypriebe:
Thanks @develar! Would love a new release once you have time.

Also note, want to push some docs in for this once I get past current release (of our downstream app).

PR #4284, opened 2019-10-01
docs(contrib): Add mention of missing mkdocs dep

jeffreypriebe:
Trying to build docs, missing a python dep (added to contrib readme) and additional info for getting the docs built.

Should we just move the mkdocs.yml from master branch to docs?


develar:
> Should we just move the mkdocs.yml from master branch to docs?

No, it should be in the root. Well, there is a way to move to docs, but for simplicity it is in the root.

Thanks for fix.

PR #4285, opened 2019-10-01
 docs(auto-update): Add docs for custom config

jeffreypriebe:
With example of request headers
Add example code

Docs related to #4250 

Once this gets merged, would be great to get a new release as [the current release](https://www.npmjs.com/package/electron-updater) doesn't yet have the exposed classes this refers to in order to add custom request headers.


develar:
Thanks for improving docs.

PR #4321, opened 2019-10-15
Remove redundant signing op for mas and mas-dev targets

oNaiPs:
I noticed that running the packager on a basic vanilla repo was yielding two signing operations for mas/mas-dev targets:
```
mkdir test && cd test
yarn init
yarn add --dev electron electron-builder
echo "console.log('hello world')" > index.js
yarn electron-builder -m mas
  • electron-builder  version=21.2.0 os=18.7.0
  • description is missed in the package.json  appPackageFile=/Users/onaips/test/package.json
  • author is missed in the package.json  appPackageFile=/Users/onaips/test/package.json
  • packaging       platform=mas arch=x64 electron=6.0.12 appOutDir=dist/mas
  • signing         file=dist/mas/test.app identityName=Developer ID Application: -redacted- identityHash=-redacted- provisioningProfile=none
  • signing         file=dist/mas/test.app identityName=3rd Party Mac Developer Application:-redacted- identityHash=-redacted- provisioningProfile=none
```

I think this is an artifact from [this](https://github.com/electron-userland/electron-builder/commit/f4398a7225c88880f9dee755b6c14699adf91c27) PR after this [other one](https://github.com/electron-userland/electron-builder/pull/3912) was implemented

This PR removes the unneeded intermediate package signing.

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -88,10 +88,14 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     }
   }
 
+  private hasMasTarget(targets: Array<Target>) {
+    return targets.length !== 0 && targets.some(it => it.name === "mas" || it.name === "mas-dev")
+  }
+
   async pack(outDir: string, arch: Arch, targets: Array<Target>, taskManager: AsyncTaskManager): Promise<any> {
     let nonMasPromise: Promise<any> | null = null
 
-    const hasMas = targets.length !== 0 && targets.some(it => it.name === "mas" || it.name === "mas-dev")
+    const hasMas = this.hasMasTarget(targets)
     const prepackaged = this.packagerOptions.prepackaged
 
     if (!hasMas || targets.length > 1) {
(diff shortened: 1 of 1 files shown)


oNaiPs:
It looks like these signing operation are indeed needed, since the `signApp` is called with any MAS options.
Perhaps way to go would be to instead skip the regular signing called on `doPack` from `platformPackager.ts`?

oNaiPs:
OK should be fixed now.

tflack:
> OK should be fixed now.

CI is failing because of an extra semi colon...

oNaiPs:
@tflack should be fixed now.

develar:
Thanks!

patrickmichalina:
FYI: This new code has broken "DMG" signing.

quanglam2807:
I agree, @patrickmichalina. This PR has regression. It disables signing altogether if `mas` is in `targets`. So if `targets` include non-mas like `dmg`, `zip` together with `mas`, non-mas builds won't be signed.  Would you mind providing a fix, @oNaiPs?

gaodeng:
please revert this commit https://github.com/electron-userland/electron-builder/commit/56922819f0be39fffb31bbb50437ad4fcdde7365  @develar

develar:
Reverted.

Issue #4234, opened 2019-09-12
Please adjust snap packaging template to plugs audio-playback


jdstrand:
* applies to all supported electron-builder versions 
* target: snap

The new audio-playback interface in snapd 2.41 will be the preferred interface for audio moving forward. On systems with a mediating pulseaudio, it provides playback access (with audio-record providing recording), otherwise full access. The interface allows for updating snapd to allow other sound servers, such as pipewire, without the need for snaps to change their plugs policy. While audio-playback is preferred, keep pulseaudio for now until everyone has updated to snapd 2.41.

I'm not an electron-developer, but am a snapd developer. In the interest of time I created a branch against master for this: https://github.com/jdstrand/electron-builder/tree/snap-plugs-audio-playback. I haven't run any tests or verified it beyond visual inspection.

Thanks!

References:
- https://github.com/snapcore/snapd/pull/5644
- https://forum.snapcraft.io/t/the-audio-playback-interface/13089
- https://forum.snapcraft.io/t/the-audio-record-interface/13090

jdstrand:
FYI, (cc @tom-james-watson), https://github.com/snapcore/snapd/pull/7631. This is milestoned for snapd 2.43.

jdstrand:
Yes this issue is still relevant. snapd 2.43 is now in beta. Once released, pulseaudio will not auto-connect. Electron not adding audio-playback will cause friction for your developers.

PR #4331, opened 2019-10-17
snap: plugs audio-playback by default

tom-james-watson:
@jdstrand has already implemented the changes needed to fix https://github.com/electron-userland/electron-builder/issues/4234. I came across this same deprecation and was about to make the changes myself, but I saw that the changes made here are already what's needed.

The new audio-playback interface in snapd 2.41 will be the preferred
interface for audio moving forward. On systems with a mediating
pulseaudio, it provides playback access (with audio-record providing
recording). The interface allows for updating snapd to allow other sound
servers, such as pipewire, without the need for snaps to change their
plugs policy. While audio-playback is preferred, keep pulseaudio for now
until everyone has updated to snapd 2.41.

References:
- https://github.com/snapcore/snapd/pull/5644
- https://forum.snapcraft.io/t/the-audio-playback-interface/13089
- https://forum.snapcraft.io/t/the-audio-record-interface/13090

Closes https://github.com/electron-userland/electron-builder/issues/4234


tom-james-watson:
Yes, this is still relevant.

felicien-brochu:
Pulseaudio interface has become deprecated:
https://forum.snapcraft.io/t/upcoming-pulseaudio-interface-deprecation/13418

Electron builder should add "audio-playback" interface as default in "plugs" in snap build configuration. For now we could leave pulseaudio in default plugs conf for backward compatibility purpose:

> If your snap currently plugs: [ pulseaudio ], you may want to opt into future-proofing and adjust it to use plugs: [ audio-playback, pulseaudio ] (adding audio-record as desired). Again, this is not required for existing snaps since they will be grandfathered into auto-connection.

Issue #4324, opened 2019-10-15
Pass -k keychain for getProvisioningProfileAsync. Bamboo/Travis builds do not have access to unlock this file


JohnTendik:
* **Version**: 21.1.1, 20.44.0 (tried both)

* **Target**: Mas, Mas-dev, dmg

**Sorry if this doesnt belong here** I created an issue on the[ electron-osx-sign ](https://github.com/electron/electron-osx-sign/issues/206)repo as well. Im happy to create a PR for the proposed changes. 

We're using electron-builder to build our application and are running into issues during signing. Specifically we have this error:
```
Unhandled rejection Error: Command failed: security cms -D -i ./embedded.dev.provisionprofile
security: cert import failed: Write permissions error.
security: problem decoding
```
I tracked down the command that is failing and its part of electron-osx-sign, specifically the getProvisioningProfileAsync function. When this function runs, it only finds system keychain file (locked). I tried unlocking the login.keychain but it is still not seen by this command.

Trying to automate build on Bamboo have been proving difficult due to permission issues. Whenever we sign the app security cms -d -i ... etc etc fails with a write permission issue. That is because the child node process does not have access to the same keychain file. Outputting security list-keychains shows only system keys.

I also tried passing in CSC_KEYCHAIN value but that also is not working, which makes sense, since in the getProvisioningProfileAsync function it isnt being surfaced.

Is it possible to add a -k flag to this
```
return execFileAsync('security', [
    'cms',
    '-D', // Decode a CMS message
    '-i', filePath // Use infile as source of data
    '-k', keychainfile // to specify which keychain to use when trying to decode provision
  ])
```

JohnTendik:
I've created a PR for the proposed changes. 🙏'ing that it goes through :D

JohnTendik:
Merged! Yay!! Thanks Develar!

PR #4332, opened 2019-10-17
added keychain flag to getProvisioningProfileAsync (#4324)

JohnTendik:
## What is this PR fixing?
Fixes (#4324) - When the command `security cms` runs, it will default to system keychains during a bamboo build even if a keychain is passed in using the CSC_KEYCHAIN environment variable.

## What has changed?
Inside of getProvisioningProfileAsync function (found within electron-osx-sign package) I added a keychain arg to surface the CSC_KEYCHAIN environment variable as a flag to the `security` command.

### How to test?
This is best reproduced on a bamboo build agent machine. If you do not have a bamboo server to test with, follow these instructions:
1) Toggle debug mode for electron osx sign `debug=electron-osx-sign` 
2) set `CSC_KEYCHAIN` environment variable
3) build and ensure that during the signing process that the correct keychain file is being used (instead of login or system keychain)

### Notes
1) This is my first contribution to this repo. If I've done anything wrong please let me know I will fix :p

2) Few tests failed for me on the mac (related to windows extra builds), Im assuming its safe to skip those failed tests since its for windows and im on mac?

### DIFF
--- a/packages/app-builder-lib/electron-osx-sign/util-provisioning-profiles.js
+++ b/packages/app-builder-lib/electron-osx-sign/util-provisioning-profiles.js
@@ -55,14 +55,21 @@ Object.defineProperty(ProvisioningProfile.prototype, 'type', {
  * Returns a promise resolving to a ProvisioningProfile instance based on file.
  * @function
  * @param {string} filePath - Path to provisioning profile.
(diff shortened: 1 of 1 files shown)


develar:
Thanks!

Issue #4305, opened 2019-10-09
Can not build the windows installer on macOS Catalina version 10.15


lingxf2002:
• electron-builder  version=21.2.0 os=19.0.0
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist/builder-effective-config.yaml
  • packaging       platform=win32 arch=x64 electron=6.0.12 appOutDir=dist/win-unpacked
  • signing         file=dist/win-unpacked/BUZZ.exe certificateFile=/Users/xling/buzzcodesigning/dialogic2018.pfx
  • building        target=nsis file=dist/buzz-installer-4.1.0.exe archs=x64 oneClick=true perMachine=false
  • signing         file=dist/win-unpacked/resources/elevate.exe certificateFile=/Users/xling/buzzcodesigning/dialogic2018.pfx
  ⨯ Exit code: ENOENT. spawn prlctl ENOENT  stackTrace=
                                                        Error: Exit code: ENOENT. spawn prlctl ENOENT
                                                            at /Users/xling/workspace/git_bit/main/buzz_desktop/electron_app/node_modules/builder-util/src/util.ts:125:16
                                                            at exithandler (child_process.js:295:5)
                                                            at ChildProcess.errorhandler (child_process.js:307:5)
                                                            at ChildProcess.emit (events.js:182:13)
                                                            at Process.ChildProcess._handle.onexit (internal/child_process.js:236:12)
                                                            at onErrorNT (internal/child_process.js:407:16)
                                                            at process._tickCallback (internal/process/next_tick.js:63:19)

pashvin:
I think to build Windows app you need Window machine and to build mac app you need Mac machine.

lingxf2002:
it works on older macOS

jsgv:
@pashvin I have been able to build Windows app on my macOS until the Catalina update.

ukmarn:
Same here.. Worked fine till Catalina update, now throwing same error as above.

jsgv:
Catalina removed support for 32 bit apps (you can still create 64 bit builds).
I have to do my builds on a Window VM because of this.

jsgv:
https://support.apple.com/en-us/HT208436

lingxf2002:
@jsgv Can not build with 64 bit either.
 electron-builder  version=21.2.0 os=19.0.0
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist/builder-effective-config.yaml
[...]

PR #4355, opened 2019-10-28
NSIS uninstaller reader (#4305)

lutzroeder:


develar:
You are hero. Does it mean that we can build NSIS uninstaller on any OS as before on macOS Catalina?

lutzroeder:
It's a pretty targeted workaround that will only work for simple NSIS files. One idea would be to run both code paths each time for a while and flag an throw an error if the resulting files aren't 100% equal. If users are fine switch over. Currently the code throws aggressively on anything unknown and falls back to the existing code paths and only tested it on Catalina.

PR #4356, opened 2019-10-28
fix(nsis): fix ignore arch in artifactName when build

codeudan:
arch in artifactName will be ignore because  packager.expandArtifactNamePattern(options, "exe", null, this.installerFilenamePattern). I just fixed it.


develar:
NSIS supports both archs in one installer. That's why arch is ignored and passed as `null`.

codeudan:
@develar But when I set "artifactName": "${name}-${version}-${arch}.${ext}" in config, the ${arch} will be ignore. installer file name include arch maybe better.

PR #4364, opened 2019-10-29
feat(app-builder-lib): support to set executableArgs in linux config

embbnux:
Would like to set executableArgs with executableName in builder config. So it will generate desktop.Exec with args.

For example:

```yml
linux:
  target:
    - deb
  category: Telephony
  executableName: my-app-name
  executableArgs:
    - --no-sandbox
```

And will generate desktop file:

```
[Desktop Entry]
Name=My Name
Exec="/opt/My Name/my-app-name" --no-sandbox %U
Terminal=false
```

Can be used to fix https://github.com/electron-userland/electron-builder/issues/3872 https://github.com/electron/electron/issues/17972

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -29,6 +29,20 @@
         "desktop": {
           "description": "The [Desktop file](https://developer.gnome.org/integration-guide/stable/desktop-files.html.en) entries (name to value)."
         },
+        "executableArgs": {
+          "anyOf": [
+            {
+              "items": {
+                "type": "string"
+              },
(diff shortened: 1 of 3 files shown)


catalin-luntraru:
> besides, where do i put it if i use the package.json instead of yml?

`"build": {
"linux": {
"executableArgs": [
        "--no-sandbox"
      ]
}
}`

embbnux:
> > besides, where do i put it if i use the package.json instead of yml?
> 
> `"build": { "linux": { "executableArgs": [ "--no-sandbox" ] } }`

Yes

PR #4369, opened 2019-10-30
drop code duplicates in downloadPlanBuilder.ts

artspb:


develar:
First implementation was quite complex, but then I realized that no need to complicate, because differential algo rocks and no need to know file format.

Issue #4380, opened 2019-11-01
[feat]: The pkg interface adds the must-close attribute


BlackHole1:
* **Version**: `22.1.0`

* **Target**:  `pkg`

currently, some of the interfaces are provided when packaging the pkg application, but not all.

In the Apple developer documentation, there is a [must-close](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW77) attribute. I think this property is very necessary to join in.

Apple's introduction to `must-close`:

> Identifies applications that must be closed before the package is installed.
> Applications are considered only if the package that contains this element has been selected for installation.

Because the current `electron-builder` does not provide this property, it will not exit some apps that must be quit when installing pkg. You must use `postinstall` or `preinstall` to force the process to be killed. This is a very bad user check.

BlackHole1:
I added the relevant PR: https://github.com/electron-userland/electron-builder/pull/4382

Tomyail:
I need this feature。 👏👏

PR #4382, opened 2019-11-01
feat(app-builder-lib): the pkg interface adds the must-close attribute (#4380)

BlackHole1:
Add mustClsoe key in pkg
Identifies applications that must be closed before the package is installed.
must-close: https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW77

issues: https://github.com/electron-userland/electron-builder/issues/4380

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -3831,6 +3831,20 @@
             "null",
             "string"
           ]
+        },
+        "mustClose": {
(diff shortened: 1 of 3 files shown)


BlackHole1:
This PR will eventually achieve the following effects:

image

Issue #3972, opened 2019-06-13
Indicator of portable app loading


campbellhennessy:
* **Version**: 20.43.0

N/A

* **Target**:  Windows portable

I have packaged a portable exe that is running off a USB. There are a lot of static media files in the project so my app.asar is several hundred MB in size. The application itself loads quickly in development (like 2-3s) but when opened from the exe it takes about a minute to load on some computers from the time the user clicks. I assume that this is the app.asar being copied to a temporary directory and decompressed.

My issue isn't with the delay, this is the price we pay for wanting it portable on a USB. The issue we are having is that there is no indication to the user that the application is still loading. The cursor spins for a few seconds but then goes back to a normal mouse and the only way you can tell the program is loading is that there is an entry in the task manger. This leads to some users clicking the executable many times before many windows start to open.

Is there a way to launch a small modal or other indication to the user that the application is still alive before the app.asar is transferred?

develar:
Please try to use `useZip` option for portable target.

```
"portable": {
  "useZip": true
}
```

campbellhennessy:
Sorry, I was a bit slow.  I'm not sure if I entered it correctly or not, but I did not find any improvement with the following:

 ```
"win": {
      "target": "portable",
      "icon": "build/icon.png"
    },
    "portable": {
      "useZip": true
    }
```

Did I put it in correctly?  Though wouldn't it still need to unzip when it copies the files to the temporary directory?  Again, I am okay if it takes a long time to load, it is just that I would like to make sure the user is aware that the application is loading and that they shouldn't keep clicking at the icon.

campbellhennessy:
I had some success in getting something like my desired outcome by modifying the portable.nsi file in the app-builder-lib package to set the installer to silent only after initialization.  This way it shows the "verifying installer..." message during the CRCCheck but then goes silent before the install.  I found the bulk of my startup delay is the CRCCheck and at least with this the user knows that the application is in the process of starting.  

The top of my portable.nsi file now looks like this:
`!include "common.nsh"
!include "extractAppPackage.nsh"

WindowIcon Off
AutoCloseWindow True
RequestExecutionLevel ${REQUEST_EXECUTION_LEVEL}

;SilentInstall silent

Function .onInit
  !insertmacro check64BitAndSetRegView
  SetSilent silent
FunctionEnd`

develar:
I think `CRCCheck off` is a good value for portable — doesn't make a lot of sense, since in any case it is a part of some package.

[...]

Issue #2548, opened 2018-02-05
[Feature Request] Splash Screen for Windows Portable exe


connollykeyjoint:
* **Version**: 19.45.4

* **Target**: Windows Portable

My understanding is that the portable executable extracts the application to a temp folder to run it.

The problem I have is that the users are stuck waiting for the extraction process before they see the application has launched. The process is rather fast on my flash PC with its SSD and ridiculous CPU clock speeds but when you start running on a PC that is still running a 3rd Gen i3 CPU and HDD that's equally as old, you find that people are impatient and often end up opening the executable 4-5 times before the initial one opens.

Therefore, my proposal is that there be an option to include a splash screen/image that shows when the portable application executable is opened to indicate that something is actually happening in the background.

develar:
Instead of splash screen, we _should_ use ZSTD instead of LZMA2 compression.

Samuel3:
+1

develar:
Upcoming electron-builder versions allows you to use ZIP compression, so, decompression will be very fast.

```
"portable": {
  "useZip": true
}
```

connollykeyjoint:
Are we not using ZSTD now?

What is the reasoning behind this?

ZIP would generate a much larger file than the current.

develar:
ZSTD: 42MB
ZIP: 54MB

Size difference: 12MB

Speed: even if ZSTD decompression speed ~ equals to ZIP decompression  (or even faster, see https://raw.githubusercontent.com/facebook/zstd/master/doc/images/Dspeed4.png), ZSTD implemented as tar + zstd. It means, that NSIS will write all 42MB  data to a temp file, and then this file will be extracted to real files. So, in case of ZSTD, it is additional time to copy 42MB (can be relatively slow on HDD).

So, I decided to simply use built-in NSIS ZIP compression instead. Also, it is portable — it means that user downloads it and then uses on some portable device (USB?).

Do you disagree?

develar:
BTW — 7z rocks not only because of lzma2, but just because 7z is excellent. XZ also uses LZMA2, but sucks in all aspects. For example, 7z automatically applies filters and supports compression chains. That's why 34 MB :)

connollykeyjoint:
Thanks for the explanation. Makes sense. Great Work! :1st_place_medal:

develar:
[...]

PR #4425, opened 2019-11-15
feat(portable): adding splash image to portable nsis package

catdad:
This PR is meant to address #2548 and #3972 by allowing a developer to add a splash image that is displayed while the portable Windows executable is extracting files. This lets the user know that the application is doing work and will soon open.

### DIFF
--- a/test/src/windows/portableTest.ts
+++ b/test/src/windows/portableTest.ts
@@ -1,6 +1,6 @@
 import { Platform, Arch } from "electron-builder"
 import * as path from "path"
-import { app, copyTestAsset } from "../helpers/packTester"
+import { app, copyTestAsset, getFixtureDir } from "../helpers/packTester"
 
 // build in parallel - https://github.com/electron-userland/electron-builder/issues/1340#issuecomment-286061789
 test.ifAll.ifNotCiMac("portable", app({
@@ -53,5 +53,17 @@ test.ifNotCiMac("portable - artifactName and request execution level", app({
 }, {
   projectDirCreated: projectDir => {
     return copyTestAsset("headerIcon.ico", path.join(projectDir, "build", "foo test space.ico"))
+  },
+}))
+
+test.ifNotCiMac("portable - splashImage", app({
+  targets: Platform.WINDOWS.createTarget(["portable"]),
+  config: {
+    publish: null,
+    portable: {
+      //tslint:disable-next-line:no-invalid-template-strings
+      artifactName: "${productName}Portable.${version}.${ext}",
+      splashImage: path.resolve(getFixtureDir(), "installerHeader.bmp"),
+    },
   },
 }))
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -3928,6 +3928,13 @@
             }
           ]
         },
+        "splashImage": {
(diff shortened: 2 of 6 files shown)


catdad:
I am opening this PR early as a draft in order to get the conversation started and get the benefit of CI. I am still working through it, but I welcome any feedback, help, or tips.

test/src/windows/portableTest.ts
```diff
@@ -54,4 +54,16 @@ test.ifNotCiMac("portable - artifactName and request execution level", app({
...
+}))
+
+test.ifNotCiMac("portable - splashImage", app({
+  targets: Platform.WINDOWS.createTarget(["portable"]),
+  config: {
+    publish: null,
+    portable: {
+      //tslint:disable-next-line:no-invalid-template-strings
+      artifactName: "${productName}Portable.${version}.${ext}",
+      splashImage: path.resolve(getFixtureDir(), "installerHeader.bmp"),
```
catdad: I am not a fan of this one. However, I tried tracing through the code quickly and did not find an appropriate location to translate relative input (`some-dir/my-image.bmp`) to the full path (`C:\\Users\\Me\\my-project\\some-dir\\my-image.bmp`). I am not sure if NSIS will use a relative path (most likely yes), but it didn't work when I just followed the examples of existing tests.

I'd appreciate any help here. Where would be the best place in the code to resolve the path for this image?

develar:
I applied your changes. But functionality doesn't work correctly — image disappears too quickly. Please address this issue.

catdad:
Can you give me more details? What project did you apply this to? I tested it on two projects, and the image always disappeared immediately before the app started. How can I repro the issue you saw?

Also, why is this closed? Is there really no chance to work on improvements for PRs submitted to this projects before they are rejected?

develar:
@catdad This PR accepted and merged into master — https://github.com/electron-userland/electron-builder/commit/91e6183bfd27191083872bcf07a0db51f9470e1e But I didn't mention it in 22.2.0 release notes. Please try 22.2.0 — does it work for you?

develar:
> What project did you apply this to?

Added by you test :)

montoyaaguirre:
Is this feature working? If so, can we update the documentation?

So far I can set the splashImage. I can see that the portable exe contains the bmp image (inspecting with 7z) but the image is never displayed. (I wouldn't miss it, the extraction takes about 20 seconds)

catdad:
@montoyaaguirre As far as I know, this is not actually working yet. I implemented the functionality in the NSIS script and validated that, and I had implemented part of passing a variable into the NSIS script. However, I never figured out how the internal module that runs the scripts gets options provided by the publicly-documented module. I had asked for help in a comment on the MR, but instead it was just merged in. I got distracted by other projects so I never looked back to implement the rest of the functionality. If you know more about how the options passing works, I'm happy to help implement the rest.

So to summarize, currently the NSIS scripts are working but there is no way for you to actually pass in an image.

BotellaA:
@catdad @develar Is there still someone trying to implement this feature? My app takes 30s to extract so a splash screen would be a great feature.

catdad:
@BotellaA I created #4990 with the remaining work in order to allow providing a relative path to a bitmap image in your config. Feel free to track that MR

PR #4441, opened 2019-11-21
Remove incorrect parsing of electron version

momakes:
Semver.coerce removes `-beta.0` part of a version, thus failing to build for those using beta or alpha versions of electron.


develar:
Hmmm... looks strange. Could you please add tests?

momakes:
@develar I can! Can you please guide me in which file I should add tests? A new test file only for `electronVersion` or add a case to an existing file?

develar:
See BuildTest.ts Like `"electron version from electron-prebuilt dependency"`

Issue #4126, opened 2019-08-07
NSIS uninstaller now has two entries in windows registry


shnassar:
* **Version**:  21.2.0
electron-updater: 4.1.2

* **Target**: nsis

With the changes in https://github.com/electron-userland/electron-builder/pull/4069 I'm now seeing the previous version's entry in the uninstall list after an update. Inspected my windows registry and am seeing two entries for the app - the new one with `{}` and the old one without.

jeremyspiegel:
Looks like a duplicate of https://github.com/electron-userland/electron-builder/issues/4092.

shnassar:
Closing in favor of https://github.com/electron-userland/electron-builder/issues/4092

Issue #4092, opened 2019-07-25
Upgrade Leaves Duplicate Uninstall Entry


mockorp:
* **Version**: 21.1.3

* **Target**: Windows

When upgrading from a version of an application built with Electron Builder prior to 21.1.3 to a version built with Electron Builder after 21.1.3, there will be a duplicate "orphaned" uninstall entry in the registry and shown in Apps & Features.

Steps to reproduce:
* Create an installer for an application with version 21.1.2 of Electron Builder
* Install the application
* Create an installer for the same application with version 21.1.3 of Electron Builder
* Install this version

Expected Result:
The application updates and leaves a single uninstall option

Actual Result:
The application updates but leaves an uninstall record for both versions.
There are two records in the registry under "Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Uninstall"

The 21.1.3 version introduced a change (https://github.com/electron-userland/electron-builder/pull/4069/commits/6341b235dbc9aeaf3036a827f4394f3fe8914ee9) which updates the uninstall entries in the Windows registry.  Prior to 21.1.3, these entries were simply "GUID".  In 21.1.3 and greater, these entries are now "{GUID}".  When upgrading from an install prior to 21.1.3, the old "GUID" entry is not removed, but a new "{GUID}" entry is added.  This leaves the user with two uninstall records for the application, one for the old version and one for the new version.  Uninstalling either of these versions will remove the application, but leave the other version.  Attempting to then uninstall the remaining version will result in an error since the application was already uninstalled.

AleksMeshkov:
#4126 is closed in favour of this ticket which is closed too. Did we lost the case?

jahraphael:
We're also experiencing this issue with our electron builder app.

AleksMeshkov:
> We're also experiencing this issue with our electron builder app.

Hi @jahraphael! In case it might help you. As a workaround I'm currently having different versions in Mac and Win package.json

`21.1.2` for Win project and `21.2.0` for OSX

tzarebczan:
Seeing the same on our project as well. Can this be cleaned up in a future installer?

gniezen:
I'm also hoping that this can somehow be cleaned up in a future installer.

hoefling:
We are also affected by the change. FWIW, you can implement a workaround via custom NSIS script include, e.g. detect an old installation and abort with a message:

```
[...]

Issue #4247, opened 2019-09-18
[nsis] use custom install/uninstall keys in windows registry


pronebird:
* **Version**: 21.2.0

* **Target**: Windows

Currently `electron-builder` only provides the customisation for windows registry installation and uninstall keys via `nsis.guid` field. if not provided, the GUID is produced automatically based on the `appId` field.

However I would like to use a product name instead of GUID. That wasn't a problem until very recently, I used to have `nsis.guid` assigned to my product name and that worked fine.

Since #4069 got merged, the uninstall key under `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Product Name` is now being wrapped in curly braces `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{Product Name}`.

So not only this is a breaking change, but it also prevents me from abusing the `nsis.guid` in such manner. Removing the `nsis.guid` leaves me with the GUID being used for the installation key, i.e `HKLM\SOFTWARE\0000-0000-0000-0000` instead of `HKLM\SOFTWARE\Product Name`. 

Based on [documentation](https://docs.microsoft.com/en-us/windows/win32/msi/uninstall-registry-key), that's fine for uninstall key, but I couldn't find the same mention for the `HKLM\SOFTWARE` so I assume in the sake of backward compatibility Microsoft Windows supports both GUIDs or Product names.

Based on a quick survey conducted by my colleagues on a very boring windows machine, we couldn't find a single software product using GUID instead of Product name under `HKLM\SOFTWARE`, besides apps built with `electron-builder`.

Additionally, there is a discrepancy in how GUID is being used since #4069, only uninstall key is wrapped in curly braces:

```
HKLM\SOFTWARE\0000-0000-0000-0000
HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{0000-0000-0000-0000}
```

We've also noticed that some Antivirus products get triggered when they stumble upon a GUID key created under `HKLM\SOFTWARE\{GUID}`.

Given all of the reasoning above, it would be amazing to be able to use the `productName` for the `HKLM\SOFTWARE` entry name. This is a breaking change so there must be a migration path.

Based on the source code, it already seems that there is some sort of sanitisation going on in `nsisTarget.ts`, so allowing something like `options.registryInstallerKey || options.guid || UUID.v5(...)` would do it.

[...]

Issue #4057, opened 2019-07-15
NSIS Installer stuck at "Installing, please wait..."


ticarexbbl:
* **electron-builder Version**: 21.0.15

* **Target**: win10 nsis

Package.json:
```
"build": {
    "asar": false,
    "win": {
      "target": "nsis"
    },
    "nsis": {
      "runAfterFinish": false,
      "artifactName": "${productName}-${version}-instalador.${ext}",
      "oneClick": true
    }
  }
```

installer.nsh:
```
!macro preInit
  SetRegView 64
  WriteRegExpandStr HKLM "${INSTALL_REGISTRY_KEY}" InstallLocation "C:\${COMPANY_NAME}\${PRODUCT_NAME}"
  WriteRegExpandStr HKCU "${INSTALL_REGISTRY_KEY}" InstallLocation "C:\${COMPANY_NAME}\${PRODUCT_NAME}"
  SetRegView 32
  WriteRegExpandStr HKLM "${INSTALL_REGISTRY_KEY}" InstallLocation "C:\${COMPANY_NAME}\${PRODUCT_NAME}"
  WriteRegExpandStr HKCU "${INSTALL_REGISTRY_KEY}" InstallLocation "C:\${COMPANY_NAME}\${PRODUCT_NAME}"
!macroend
```

Installer gets successfully built but when executed gets stuck at "Installing, please wait..." forever

curlymon:
Generally the only time I see this behavior is when an existing install has been deleted rather then explicitly uninstalled via Add Remove Programs or the uninstaller itself.

ticarexbbl:
Yes, the existing install was indeed deleted instead of being uninstalled properly, but the installer should be able to handle this fairly common situation and install it anyway or at least show some kind of error

kebugcheckex:
> Generally the only time I see this behavior is when an existing install has been deleted rather then explicitly uninstalled via Add Remove Programs or the uninstaller itself.

For whoever run into this situation, use CCleaner to manually remove the entry from Windows Program list. Then re-run the installer.

For reference, see [this page](https://www.howtogeek.com/314734/how-to-manually-remove-programs-from-the-windows-uninstall-program-list/)

freewind:
This is quite confusing for end users, is it able to force re-install new versions?

gomorizsolt:
We've also encountered with the exact same situation. After removing the app's folder manually, the installer kept hanging around nearly half of the process.

[...]

PR #4443, opened 2019-11-21
Revert "fix(nsis): NSIS Uninstaller registry entry format change (#4069)"

elvisvoer:
It looks like the change of the registry format has not been well thought through and it causes lots of issues (#4092 and #4126).

The change was made for pure pedantic reasons and didn't consider backwards compatibility with older installer. In our case, we are wrapping the NSIS package with an MSI wrapper that has a bit more functionality than the one provided by electron-builder which no longer works after this change.

My suggestion is to revert the commit all together and leave the registries as they were before.

This reverts commit 7518aee8e1abe0516071b350748e84dc47e35cf7.

### DIFF
--- a/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
+++ b/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
@@ -167,7 +167,7 @@ export class NsisTarget extends Target {
       APP_PACKAGE_NAME: appInfo.name
     }
     if (uninstallAppKey !== guid) {
-      defines.UNINSTALL_REGISTRY_KEY_2 = `Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{${guid}}`
+      defines.UNINSTALL_REGISTRY_KEY_2 = `Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\${guid}`
     }
 
     const commands: any = {
--- a/packages/app-builder-lib/templates/nsis/multiUser.nsh
+++ b/packages/app-builder-lib/templates/nsis/multiUser.nsh
@@ -6,7 +6,7 @@
 
 # allow user to define own custom
 !define /ifndef INSTALL_REGISTRY_KEY "Software\${APP_GUID}"
-!define /ifndef UNINSTALL_REGISTRY_KEY "Software\Microsoft\Windows\CurrentVersion\Uninstall\{${UNINSTALL_APP_KEY}}"
+!define /ifndef UNINSTALL_REGISTRY_KEY "Software\Microsoft\Windows\CurrentVersion\Uninstall\${UNINSTALL_APP_KEY}"
 
 # current Install Mode ("all" or "CurrentUser")
 Var installMode


stilettk:
There is also another issue regarding this commit:
https://github.com/electron-userland/electron-builder/issues/4057

damianobarbati:
@elvisvoer was this released? Or do you know when this is going to be released? 
I have 2k users (free application) with broken auto-update (in infinite loop!) 😱

elvisvoer:
@damianobarbati pre-release 22.2.0 has it

menelike:
Folks, IMHO this should end up in `21.2.x` as long as `22.x.x` is not production-ready. 

Second, it should explain how already affected and released applications should be cleaned up. Or even better https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/nsis/include/installer.nsh should handle that cleanup.

Don't get me wrong, this is an open-source project and I am very grateful for all the efforts which have been put into this project, but this change has made the life of many developers already miserable https://github.com/electron-userland/electron-builder/issues/4092 and just reverting this for *a soon to come release* is simply not enough. 

I would have contributed myself, but I just have no experience with NSIS scripts (and the risk of making the situation worse is just too high) 

Just my 2 ¢.

mastergberry:
So...what are people supposed to do who migrated to the newer version?

We need a path back from new -> old...

elvisvoer:
@mastergberry we do this in an init script that we include in our windows yaml file
```; the preInit hook is executed even when packing the app so we use customInit 
; to make sure we execute this code only when the installer is invoked
!macro customInit
DeleteRegKey HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\{ProductID}"
DeleteRegKey HKLM "Software\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\{ProductID}"
!macroend
```

mastergberry:
Ok thanks @elvisvoer I will attempt this the next time we upgrade...just petrified because I had to fix like 3 different "bugs" because of the original chnage with hacky nsis scripts...hope I can just nuke those 2 new keys and then watch everything work...

damianobarbati:
@elvisvoer are you using that NSIS include with the pre-release `22.2.0` or with the current release `21.2.0`?
Is it fixing the install/autoupdate problem for you?

elvisvoer:
@mastergberry yeah, it should work. We are already in alpha with our app and I got to verify the update from the faulty version to the pre-release one.

@damianobarbati we use the pre-release but we've never had problems with the update (or none that we know of) so maybe if you give me some more details I can help.

damianobarbati:
@elvisvoer your script seems to work! 
It prevents my installer from breaking when the GUID changed from `xxx-xxx-xxx` to `{xxx-xxx-xxx}`.
I'm currently testing upgrades from several versions of my previous app to the one running your `include.nsh`.

Is it safe to always keep that script in place? To clean previous uninstall entries from registry / control panel?

elvisvoer:
@damianobarbati I think it is. This is what we do at the moment. However, the scripts will pile up so a way to mitigate is to make a "required" release: a version that all old applications will see as last update and which knows where the new updates are, this requires a bit of infra changes but it is a way to clean up technical debt

jmeinke:
@elvisvoer Wouldn't it make sense to extend your script for the `HKCU` current user installation entries?
```
; related to:
; https://github.com/electron-userland/electron-builder/issues/4092
; https://github.com/electron-userland/electron-builder/pull/4443#issuecomment-562962960
!macro customInit
  DeleteRegKey HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\${UNINSTALL_APP_KEY}"
  DeleteRegKey HKCU "Software\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\${UNINSTALL_APP_KEY}"
  DeleteRegKey HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\{UNINSTALL_APP_KEY}"
  DeleteRegKey HKLM "Software\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\{UNINSTALL_APP_KEY}"
!macroend
```

elvisvoer:
@jmeinke we use the oneClick installer which installs for all users but I think your script cover more cases so it's better

faern:
This change broke apps again. As soon as the GUID changes, an upgrade won't find the old app and you end up with two apps in the Windows list of installed programs. We have previously had to work around this by custom code renaming registry entries ourselves in the installer, and now we needed one more of those hacks (https://github.com/mullvad/mullvadvpn-app/pull/2023).

This issue is a thread about the problem, and how the best would be to be able to specify the GUID as a custom string. Virtually every program uses the program name instead of a generated UUID anyway: #4247. Sadly a bot closed the issue, but it's still highly relevant. Please re-open if you have the access to do so.

PR #4457, opened 2019-11-23
Make electron-builder find the nightly electron builds

ShadixAced:
The nightly version of Electron is named "electron-nightly". Therefore, using the actual version, electron-builder cannot find it. 
This change is to make electron-builder compatible with the nightly builds.


develar:
Thanks, but PR contains a lot of unrelated changes, please create a new one.

ShadixAced:
@develar Noted, i'm gonna try to make both PR's different.

ShadixAced:
This PR should be OK for now. The personal one contains a lot of TS modifications (due to a massive update. I do not know if it affects a lot. To avoid any confusion, the second PR will be deleted.

PR #4473, opened 2019-11-27
Make electron-builder compatible with electron-nightly.

ShadixAced:
(as @develar asked, this is a clean PR)

The nightly version of Electron is named "electron-nightly". Therefore, using the actual version, electron-builder cannot find it.
This change is to make electron-builder compatible with the nightly builds.


sandeep1995:
@develar Can it be released as a pre-release? We would like to use the electron's nightly build.

PR #4496, opened 2019-12-02
fix(AppImage): Add default argument --no-sandbox

### LINKED ISSUE #4495: Packaging Electron v5+ with sandbox (AppImage)
* **Version**: 21.2.0

* **Target**: AppImage

See: https://github.com/electron-userland/electron-builder/issues/3872

This issue was closed, but the solution only fixes the problem for `.deb` builds. AppImages are still broken by default, and there's currently no solution (other than asking people to [manually patch their AppImage](https://github.com/ssbc/patchwork/issues/1217#issuecomment-559609983)). Is there any recommended course of action for broken AppImage releases?

Thanks!

christianbundy:
Resolves #4495 by changing default `Exec` to  `AppRun --no-sandbox` on Electron v5+.

Is there a better way to do this? The solution is inspired by the solution for `.snap` files.


christianbundy:
Or maybe it should be supporting `executableArgs` instead? https://github.com/electron-userland/electron-builder/pull/4364

crafter999:
This has nothing to do with AppImage,  some Linux distributions does not support unprivileged user namespaces enabled by default. Tweaking the kernel may fix the problem but there must be an option for applying certain arguments to the electron. For example I can only run my electron app using `--no-sandbox` which makes electron-builder useless because there's no option for that.

christianbundy:
It sounds like the best solution would be to make the AppImage target (and others) respect `executableArgs`, but getting someone to look at your PR is impossible.

crafter999:
Yeah because disabling sandbox by default is not the right solution for security reasons. There must be an option "no-sandbox" and users must get a warning message during the building.

christianbundy:
maintainers plz

tzarebczan:
Can we get this merged? It's affecting our lbry-desktop app users on debian also. See https://github.com/electron/electron/issues/17972

christianbundy:
cc: @develar

develar:
@christianbundy Could you please resolve merge conflicts?

develar:
Thanks! I will release in several days.

jsg2021:
Is there a way to turn this off? / opt-out?

devinbinnie:
Is there any reason this still needs to be here? I understand that some people need it but why is it default? Shouldn't the args be configurable? Is that something we can change and perhaps just have `--no-sandbox` as a default argument such that no one's current workflow is broken.

PR #4528, opened 2019-12-12
fix(nsis): fix cli config error when use Boolean value in -c.nsis config

codeudan:
Currently, when I config in cli like -c.nsis.createDesktopShortcut=true will not work, because it will parsed to String instead of Boolean. 
I just fixed by using coerceTypes.

### DIFF
--- a/packages/electron-builder/src/builder.ts
+++ b/packages/electron-builder/src/builder.ts
@@ -155,8 +155,12 @@ export function normalizeOptions(args: CliOptions): BuildOptions {
       coerceValue(config.mac, "identity")
     }
 
+    // fix Boolean type by coerceTypes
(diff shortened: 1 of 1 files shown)


kzimny:
The same for `nsisWeb` config?

codeudan:
@kzimny  
nsisWeb config is  like  -c.nsisWeb
So need  can add 
`
 if (config.nsisWeb != null) {
      coerceTypes(config.nsisWeb);
 }
`
to fix it.

kzimny:
Can you please fix it for nsisWeb in the same way?

codeudan:
> Can you please fix it for nsisWeb in the same way?
Done

kzimny:
Great! Many thanks!

PR #4565, opened 2019-12-30
Fix old blockmap URL for differential download

### LINKED ISSUE #4564: Wrong blockmap URL for NSIS differential download
* **Version**: electron-builder@21.2.0

* **Version**: electron-updater@4.2.0

* **Target**: NSIS

Differential update always fails because it always fetch the wrong URL for old blockmap. Specifically in this line:
https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/NsisUpdater.ts#L140

```
const oldBlockMapUrl = newUrlFromBase(`${fileInfo.url.pathname.replace(new RegExp(downloadUpdateOptions.updateInfoAndProvider.info.version, "g"), this.currentVersion.version)}.blockmap`, fileInfo.url)
```

it builds the blockmap URL with the version from `this.currentVersion.version` while `this.currentVersion` must be a string and therefore `this.currentVersion.version` is obviously `undefined`.

Qrysto:
Fixes #4564

### DIFF
--- a/packages/electron-updater/src/NsisUpdater.ts
+++ b/packages/electron-updater/src/NsisUpdater.ts
@@ -137,7 +137,7 @@ export class NsisUpdater extends BaseUpdater {
       }
 
       const newBlockMapUrl = newUrlFromBase(`${fileInfo.url.pathname}.blockmap`, fileInfo.url)
-      const oldBlockMapUrl = newUrlFromBase(`${fileInfo.url.pathname.replace(new RegExp(downloadUpdateOptions.updateInfoAndProvider.info.version, "g"), this.currentVersion.version)}.blockmap`, fileInfo.url)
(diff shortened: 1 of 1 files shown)


develar:
Why? currentVersion here it is object.

Qrysto:
@develar How come can it be an object? Here in the documentation https://www.electron.build/auto-update#api

> `currentVersion` SemVer - The current application version.

I always assign a string to `currentVersion` (for example `autoUpdater.currentVersion = "1.0.1"`) and it works for most features except this differential update. I also debugged the code and saw `this.currentVersion` was a string (what I assigned) not an object.

Qrysto:
If you say it should be an object, what schema should it has? and where does that object come from?

PR #4574, opened 2020-01-03
fix(typings): add yargs and fs-extra types as dependencies

### LINKED ISSUE #4556: [typescript] `fs-extra` an `yargs` missing types error
* **Version**:  22.2

* **Target**: 

Hey there, you use `fs-extra` and `yargs` to provide typings, but those dependencies aren't provided with the production build (aka they are listed under `devDependencies` instead of `dependencies`).

This causes a `Could not find a declaration file for module 'fs-extra'|'yargs'` error when compiling with TS and your types are referenced in some ways.

See [this issue](https://github.com/microsoft/types-publisher/issues/81) for the reasons behind putting `@types/*` dependencies into `dependencies`.

Sometimes this could cause errors in production builds and is not feasible, but in your case it seems to be harmless to add the two missing dependencies

IlCallo:
Adding them directly on  `packages/electron-builder/package.json` solves the issue, but I preferred to add them in the `package.json`s of the sub-packages which it uses.

I wasn't able to test in a local environment if the issue is solved properly with this second way, because of version compatibility problems.

I also excluded new dependencies into `checkDeps` helper as I saw it was done with other similar dependencies

Closes #4556


IlCallo:
Don't really know what failing tests are referring to, but I see all other PR's are failing on those, so I guess it isn't my fault

develar:
Problem — in this case these dependencies will be installed for production. I want to avoid that.

IlCallo:
There's no way to avoid that as far as I know, except telling the devs to add them manually (which is a terrible DX for all TS devs)

develar:
Thanks! Released as `next` version — `22.3.3`.

Issue #2451, opened 2018-01-08
Updater causes uncaught exception in the main process if download interrupted


JonathanLydall:
* **Version**: 
electron-builder: `19.52.1`
electron-updater: `2.18.2`

* **Target**: 
`platform=win32 arch=x64`

While the updater has a download in progress, have the download interrupted. I was testing locally using the `http-server` npm package, so I just terminated the server, but other interruptions like unplugging your network cable, disconnecting WiFi, putting your computer to sleep, should have the same effect.

The following dialog shows:
![image](https://user-images.githubusercontent.com/1290262/34661713-25863f10-f454-11e7-88da-5db228ec65e4.png)
```
[Window Title]
Error

[Main Instruction]
A JavaScript error occurred in the main process

[Content]
Uncaught Exception:
Error: net::ERR_CONNECTION_RESET

[OK]
```

I have tried to catch the error unsuccessfully through the following methods:
- `updater.downloadUpdate(...).then(...).catch(function(error) { ... }`
- `updater.signals.emitter.addListener("error", function(error) { ... }`

It can be caught by putting `process.on('uncaughtException', function (error) { ... }`, in NodeJS's entry `.js` file, but there is no more context to the error than what you see above, so there is no way to determine that the error came from the updater.
![image](https://user-images.githubusercontent.com/1290262/34662257-ef505e7c-f457-11e7-9982-c4f140a8ea61.png)

marceloromaomultilaser:
Same here!!!

OsamaAamer:
+1

btibarra:
Does someone figured it out how to solve it?

JeongJun-Lee:
+1

tkleinke:
Same here.

amits1995:
+1  
"electron-updater": "^2.21.10"  
Auto updater from Github  
Happened on 2 Mac's but couldn't reproduce using @JonathanLydall steps on my Ubuntu machine

jmvallejo:
This happens only on mac for me, I don't seem to be able to catch the error. I have tried using a try/catch block as well as listening to the error event... nothing really works. Anyone has had any luck?
![image](https://user-images.githubusercontent.com/14879937/51355891-81f88b00-1a86-11e9-9bc3-c413cc43e801.png)
This error is intentional, I'm using a bad URL, but I'd like to be able to catch it.

alanning:
[...]

PR #4586, opened 2020-01-09
fix:(electron-updater) not rejecting with error when response stream gets an error

kevinlinv:
# Intent

Fix a bug where network error thrown when invoking `updater.downloadUpdate()` is not catchable.

Related issue: https://github.com/electron-userland/electron-builder/issues/2451

#### Problem

Network error such as `net::ERR_NETWORK_CHANGED`, etc could happen when a `response` stream is in progress. We do not seem to handle any cases related to this, which resulted in the error being thrown globally and the original `updater.downloadUpdate()` not resolving.

#### How to reproduce

```ts
// assumes that we have a new version ready to be downloaded
// on a cold cache
try {
  await autoUpdater.downloadUpdate() // When the download is in progress, disconnect
                                     // from the internet, this will not reject
                                     // and never settles. Additionally, a `net::ERR_...` 
                                     // depending on the network error will be thrown
                                     // and not caught causing an `uncaughtException`.
} catch (e) {
  console.error(e)
}

```

#### Solution

This PR is attaching `response.on("error", handler)` where appropriate to make sure we reject with the error so the consumer can catch and react to the error without attaching a global `uncaughtException` handler manually.

### DIFF
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -155,6 +155,7 @@ Please double check that your authentication token is correct. Due to security r
     response.setEncoding("utf8")
 
     let data = ""
+    response.on("error", reject)
     response.on("data", (chunk: string) => data += chunk)
     response.on("end", () => {
       try {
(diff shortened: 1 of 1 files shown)


develar:
Thanks a lot.

kevinlinv:
No worries... Thanks for reviewing and also appriciate the work you've done maintaining this project. 👍

Issue #4392, opened 2019-11-04
building snap is causing unexpected cli output


yafp:
* **Version**: 22.1.0

* **Target**:  Linux / .snap

I am building for several linux variants (.deb, .rpm, .pacman, AppImage ...and snap.)
Lately the output while processing snap looks strange (i assume since the last update to 22.1.0)

```
 npm run build-linux

> ttth@1.8.0-alpha-1-20191104 build-linux /home/X/ttth
> electron-builder --linux

  • electron-builder  version=22.1.0 os=4.15.0-55-generic
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist/builder-effective-config.yaml
  • packaging       platform=linux arch=x64 electron=6.1.3 appOutDir=dist/linux-unpacked
  • building        target=zip arch=x64 file=dist/ttth-1.8.0-alpha-1-20191104-linux-x64.zip
  • building        target=AppImage arch=x64 file=dist/ttth-1.8.0-alpha-1-20191104-linux-x86_64.AppImage
  • building        target=snap arch=x64 file=dist/ttth-1.8.0-alpha-1-20191104-linux-amd64.snap
{
  "base": "core18",
  "grade": "stable",
  "confinement": "strict",
  "plugs": {
    "gnome-3-28-1804": {
      "interface": "content",
      "target": "$SNAP/gnome-platform",
      "default-provider": "gnome-3-28-1804"
    },
    "gtk-3-themes": {
      "interface": "content",
      "target": "$SNAP/data-dir/themes",
      "default-provider": "gtk-common-themes"
    },
    "icon-themes": {
      "interface": "content",
      "target": "$SNAP/data-dir/icons",
      "default-provider": "gtk-common-themes"
    },
    "sound-themes": {
      "interface": "content",
      "target": "$SNAP/data-dir/sounds",
      "default-provider": "gtk-common-themes"
    }
  },
  "name": "ttth",
  "version": "1.8.0-alpha-1-20191104",
  "summary": "ttth",
  "description": "ttth (talk to the hand) is an electron based desktop app for online services like WhatsApp, Threema, Telegram, Twitter, Google and several others.",
  "architectures": [
    "amd64"
  ],
  "apps": {
    "ttth": {
      "command": "command.sh",
      "plugs": [
        "desktop",
        "desktop-legacy",
        "home",
        "x11",
        "wayland",
        "unity7",
        "browser-support",
        "network",
        "gsettings",
        "pulseaudio",
[...]

PR #4589, opened 2020-01-13
remove snap debugging

tom-james-watson:
This debugging was accidentally introduced by me in https://github.com/electron-userland/electron-builder/pull/4237


tom-james-watson:
Fixes https://github.com/electron-userland/electron-builder/issues/4392

PR #4619, opened 2020-01-27
always include a secure timestamp with your code-signing signature

Igor-Techsee:
> Include a secure timestamp with your code-signing signature. (The Xcode distribution workflow includes a secure timestamp by default. For custom workflows, include the --timestamp option when running the codesign tool.)

https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution?language=objc

### DIFF
--- a/packages/app-builder-lib/electron-osx-sign/sign.js
+++ b/packages/app-builder-lib/electron-osx-sign/sign.js
@@ -141,6 +141,7 @@ function signApplicationAsync (opts) {
 
       const args = [
         '--sign', opts.identity.hash || opts.identity.name,
+        '--deep',
         '--force'
       ]
       if (opts.keychain) {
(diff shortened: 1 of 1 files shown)


Jferra:
Would also need that fix :)

packages/app-builder-lib/electron-osx-sign/sign.js
```diff
@@ -141,6 +141,7 @@ function signApplicationAsync (opts) {
 
       const args = [
         '--sign', opts.identity.hash || opts.identity.name,
+        '--deep',
```
rajivshah3, line 144: Is this necessary? The Apple [docs](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Procedures/Procedures.html#//apple_ref/doc/uid/TP40005929-CH4-SW10) seem to recommend against it:

> **Important**: While you use the `--deep` option for *verification* to mimic what Gatekeeper does, it is not recommended for *signing*. During signing, if you have nested code, and if you are signing manually, you sign nested code in stages (as Xcode does automatically), starting with the most deeply embedded components first. You then sign code at the next level of hierarchy, and so on. You work your way outward, finally signing the top level entity that contains all the others. Signing all the components in one shot with `--deep` is for emergency repairs and temporary adjustments only. Note that signing with the combination `--deep --force` will forcibly re-sign all code in a bundle.

develar:
`timestamp` was added, thanks

Issue #4618, opened 2020-01-27
Differential update request is not timed out if it hangs


clusterberries:
* **Version**: 21.1.5

* **electron-updater**: 4.1.2

* **Target**:  nsis

Hello!
I faced with issue described here https://github.com/electron-userland/electron-builder/issues/3973
I added some more logs to electron-updated in order to understand why it is happening. The initial case is hard to define, but I found out that timeout should be set when request is created:
```
differentialDownloader.httpExecutor.addErrorAndTimeoutHandlers(request, reject);
```
(multipleRangeDownloader file)
But the request is never timed out, it just hangs.
I can create a PR with fix, but I am not sure how it should be fixed. @develar need you advice. Thanks!

PR #4649, opened 2020-02-05
fix(builder-util-runtime): set timeout when request is closed

clusterberries:
"Socket" event is not fired if Electron "net" version of request is used. It lead to issue that request will never be timed out if something went wrong. I didn't find how to set timeout request in Electron net module. I'm open to proposals of more clear approach.
This PR fixes issue https://github.com/electron-userland/electron-builder/issues/4618

### DIFF
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -285,11 +285,16 @@ Please double check that your authentication token is correct. Due to security r
   }
 
   private addTimeOutHandler(request: any, callback: (error: Error) => void) {
+    const timeoutCb = () => {
+      request.abort()
+      callback(new Error("Request timed out"))
+    }
+
     request.on("socket", (socket: Socket) => {
-      socket.setTimeout(60 * 1000, () => {
-        request.abort()
-        callback(new Error("Request timed out"))
-      })
+      socket.setTimeout(60 * 1000, timeoutCb)
+    })
+    request.on("close", () => {
+      setTimeout(timeoutCb, 30 * 1000)
     })
   }
 


pverscha:
This PR causes some issues when downloading a file takes longer than the timeout that's currently set. It seems that the close-event is fired almost immediately, then the timeout is set and after 30 seconds the request is aborted.

When updating my application, electron-updater downloads a reasonably small file which takes longer than 30 seconds to download and causes the update to fail, because the download-request is aborted when it shouldn't.

pverscha:
@clusterberries @develar Okay, so I figured out what is going wrong. The `close`-event is being fired immediately (long before the whole file is done downloading), which causes the `setTimeout` in `request.on("close", ...)` to be started which kills the request after 30 seconds.

The electron [documentation](https://www.electronjs.org/docs/api/client-request#event-close) for the net-module states for the `close` event that it is ```Emitted as the last event in the HTTP request-response transaction. The close event indicates that no more events will be emitted on either the request or response objects.``` One would expect that this means that the `close` event is only fired after downloading the complete file is done, as you'd expect there to be HTTP-communication until the end.

This is, however, NOT the case. I've investigated the behaviour using WireShark when trying to download a large file in Electron over HTTP. Only one HTTP-request is made at the start of the communication, after which the file download itself proceeds over TCP. Thus indeed all HTTP-communication is over, and the `close`-event is fired. BUT there is still data being transfered over TCP, which is why it's not a good solution to just abort the request 30 seconds after receiving the `close` event.

KillingJacky:
Agree with @pverscha , the close timeout callback will kill the file downloading after 30 seconds. That means if the updater needs to download a larger file that takes more than 30 seconds, the auto update will never succeed.

image

KillingJacky:
The `close` event of the ClientRequest is fired immediately when the request has been sent to the server, it's not the one when we're at the end of response stream.

develar:
Guys, could you please send PR to fix it? Or should I just revert this change? cc @clusterberries

clusterberries:
Hello guys! 
In general I agree that this solution has bad side-affects, but unfortunately I didn't find a solution how to time out the request correctly, so it just gets stuck in my case without this code. 
It would be great if you can propose clearer solution, this fix is very important for my application.

pverscha:
@clusterberries @develar I honestly don't know how we can fix this when using the electron net-module. All events that somewhat seem to indicate the end of a data-transfer all fire near immediately. I've tried to see what axios does in this case. Axios returns a promise when you instruct it to fetch some data, and this promise only resolves when the transfer is indeed completely finished which is what we want.

We might need to switch to axios instead of electron net for this to work, unless someone with more knowledge about electron net knows how to fix this?

KillingJacky:
Sorry I'm lack of TypeScript skill, I can't start debugging immediately though I've a few ideas. I'm assuming that we can try listening to the `close` / `error` event of the `response` object. From the [testing code](https://github.com/electron/electron/blob/master/spec-main/api-net-spec.ts#L740) of the electron project, the `response` object does have some events that are not documented well.  The second idea is, we can pipe the response stream to another stream to monitor if the stream is cut off https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util-runtime/src/httpExecutor.ts#L418 , e.g. the readable event doesn't happen for a specific time period.

develar:
@pverscha Problem is that no one will be happy about Axios, we must use electron net module to use more robust and better proxy / certs.

Looks like I will simply revert this change.

develar:
Until proper solution is not found, change reverted and electron-updater 4.2.5 published.

Sergey-Churilin:
Hi @develar, during debugging I noticed that when data for differential updates received, for some reason nothing more happens - I mean no events about successful download are fired.  I prepared another pull request, could you please have look
https://github.com/electron-userland/electron-builder/pull/4817

PR #4657, opened 2020-02-07
fix(builder-util): Retry flaky builder operations

abettadapur:
There are various operations that we have found flaky and unreliable on Windows due to AntiVirus issues. 
For example, the RCEdit.exe command to edit the binary information often fails 

```
 ⨯ cannot execute  cause=exit status 1
                    errorOut=Fatal error: Unable to commit changes
```

But will succeed immediately after

This PR adds support to `executeAppBuilder` to retry operations that have failed

### DIFF
--- a/packages/app-builder-lib/src/winPackager.ts
+++ b/packages/app-builder-lib/src/winPackager.ts
@@ -317,7 +317,7 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
     const timer = time("wine&sign")
     // rcedit crashed of executed using wine, resourcehacker works
     if (process.platform === "win32" || this.info.framework.name === "electron") {
-      await executeAppBuilder(["rcedit", "--args", JSON.stringify(args)])
+      await executeAppBuilder(["rcedit", "--args", JSON.stringify(args)], undefined /* child-process */, {}, 5 /* retry five times */)
     }
 
     await this.sign(file)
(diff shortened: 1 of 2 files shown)


darkguy2008:
Oh God, and I thought I was the only one suffering from this!

I found an easy fix for RCEdit.exe, no need to retry, a simple sleep() works. Just replace this fragment, tested & working with the latest version as of date (so it should pass Travis I guess):

```
const sleep = (ms) => { return new Promise(resolve => setTimeout(resolve, ms)); }
function executeAppBuilder(args, childProcessConsumer, extraOptions = {}) {
  return new Promise((resolve, reject) => {
    const command = _appBuilderBin().appBuilderPath;

    const env = Object.assign(Object.assign({}, process.env), {
      SZA_PATH: _zipBin().path7za,
      FORCE_COLOR: _chalk().default.level === 0 ? "0" : "1"
    });
    const cacheEnv = process.env.ELECTRON_BUILDER_CACHE;

    if (cacheEnv != null && cacheEnv.length > 0) {
      env.ELECTRON_BUILDER_CACHE = path.resolve(cacheEnv);
    }

    if (extraOptions.env != null) {
      Object.assign(env, extraOptions.env);
    }

    sleep(1000).then(_ => {
      const childProcess = doSpawn(command, args, Object.assign({
        env,
        stdio: ["ignore", "pipe", process.stdout]
      }, extraOptions));

      if (childProcessConsumer != null) {
        childProcessConsumer(childProcess);
      }

      handleProcess("close", childProcess, command, resolve, error => {
        if (error instanceof ExecError && error.exitCode === 2) {
          error.alreadyLogged = true;
        }

        reject(error);
      });
    });
  });
}
```

Issue #3335, opened 2018-09-19
devDependencies


wells1013:
* **Version**: 20.28.4  

* **Target**:  Mac, Windows

```
const deps = ["electron", "electron-prebuilt", "electron-rebuild"]
  if (process.env.ALLOW_ELECTRON_BUILDER_AS_PRODUCTION_DEPENDENCY !== "true") {
    deps.push("electron-builder")
  }
  for (const name of deps) {
    if (name in dependencies) {
      errors.push(`Package "${name}" is only allowed in "devDependencies". `
        + `Please remove it from the "dependencies" section in your package.json.`)
    }
  }
```

check Dependencies at 'devDependencies', how about 'optionalDependencies'?

develar:
Not clear what do you mean. Should we check also optionalDependencies for these deps?

s524797336:
The optionalDependencies will overwrite dependencies, add it to optionalDependencies is equal as dependencies.
But i have an other suggest.
I have a multi platform project which runs on web and electron, i use docker to build web version, so i added it to optionalDependencies, in docker i use `npm install --no-optional` to avoid install electron stuff, but in electron version, i have to move optionalDependencies to devDependencies, can you gays only check if electron-builder only in devDependencies instead of check it in dependencies?
Because i need add electron-builder in devDependencies and optionalDependencies, then `npm install --no-optional` will not install electron-builder

Issue #2481, opened 2018-01-16
Include devDependencies


mk010101:
I need to include all devDependencies when packaging the app. However, I get this: 
**Error: Package "electron-builder" is only allowed in "devDependencies".**

The reason I need to include electron-builder is that the app **is** the builder (GUI of sorts). 

What would be the correct approach? Also, is it possible to avoid making users install node/npm and just include them in the App?

develar:
Set env `ALLOW_ELECTRON_BUILDER_AS_PRODUCTION_DEPENDENCY=true`

> is it possible to avoid making users install node/npm and just include them in the App?

You can use electron as node.

mk010101:
Thanks for the quick response.
The env worked fine, however I'm getting another issue: the rebuilding part is not rebuilding native dependencies (7zip-bin-win is one of them). Here's my gulp code:

`gulp.task("mkexe",  ()=> {

    let config = {
        win: ["nsis"],
        mac: ["dmg"],
        config: {
            asar: false,
            npmRebuild: true,
            appId: "m.k.builder",
            files: [
                "**/*",
            ],
            directories: {
                app: "./",
                output: "./../__RELEASE_test",
                buildResources: "build"
            },
        },
    };

    process.env.ALLOW_ELECTRON_BUILDER_AS_PRODUCTION_DEPENDENCY=true;

    builder.build(config)
        .then(() => {
           console.log("done.")
        })
        .catch((error) => {
           console.log(error);
        });


});`

I'm building on osx.

PR #4659, opened 2020-02-09
fix(app-builder-lib): remove lock to devDependencies

### LINKED ISSUE #4563: Package "electron-builder" is only allowed in "devDependencies".
Hi, I want to build an electron app for ARMv7 in production. The command:

    npm --arch=arm --target_arch=arm --production --no-registry install

gives me this error:

> Package "electron-builder" is only allowed in "devDependencies". Please remove it from the "dependencies" section in your package.json.

How to build the app with electron-builder in prod?

sapkra:
This code was blocking the installation in a CI pipeline without installing all devDependecies which is against the official documentation of npm.

https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

- **"dependencies"**: Packages required by your application in production.
- **"devDependencies"**: Packages that are only needed for local development and testing.

Electron, electron-builder etc. are all required to build the production version and not just for local development or testing. So they should be normal dependencies inside the package.json.

Related issues:
#2330
#2481
#3335
Fixes #4563

### DIFF
--- a/packages/app-builder-lib/src/util/packageMetadata.ts
+++ b/packages/app-builder-lib/src/util/packageMetadata.ts
@@ -109,15 +109,4 @@ function checkDependencies(dependencies: { [key: string]: string } | null | unde
   if (swVersion != null && !versionSatisfies(swVersion, ">=20.32.0")) {
     errors.push(`At least electron-builder-squirrel-windows 20.32.0 is required by current electron-builder version. Please set electron-builder-squirrel-windows to "^20.32.0"`)
   }
-
-  const deps = ["electron", "electron-prebuilt", "electron-rebuild"]
-  if (process.env.ALLOW_ELECTRON_BUILDER_AS_PRODUCTION_DEPENDENCY !== "true") {
-    deps.push("electron-builder")
-  }
-  for (const name of deps) {
-    if (name in dependencies) {
-      errors.push(`Package "${name}" is only allowed in "devDependencies". `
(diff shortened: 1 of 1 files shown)


develar:
In most cases including electron as dependency it is an error. Maybe electron-builder can somehow detect that it is safe (say, some field in package json, I don't know). If no such indicators, than I think better to introduce option to allow.

l1bbcsg:
Even if it is an error, it's really outside of electron-builder's realm of responsibility to correct application developer's mistakes in host project setup. The build should still work if app setup is sub-optimal, the current behaviour is just intrusive and opinionated.

I believe at most it could produce a warning, but definitely not impose such restrictions on users. I know what I am doing and I need both electron and electron-builder in `dependencies`, but for arbitrary reasons electron-builder just refuses to work.

If there are cases that lead to erroneous behaviour, those particular cases should fail and those errors should be reported, perhaps with a suggestion to change package.json structure.

develar:
If electron will be in your production deps, it will be packed into your app dist. And that's not expected by users. Any reason why you want to include electron into deps?

develar:
Closed due to inactivity. Feel free to reopen.

Kirpich634:
@develar Could you attach a link or tell how the electron decides which dependencies to include in the app dist?

develar:
@Kirpich634 Please see https://github.com/develar/app-builder/blob/master/pkg/node-modules/nodeModuleCollector.go#L56 `devDependencies` are not processed at all.

Issue #4691, opened 2020-02-21
Can I customize CFBundleDisplayName for the Renderer helper?


aguynamedben:
* **Version**: 22.3.2
* **Target**: macOS, Electron 7

Hi, first off, we love electron-builder and we will donate $100 after submitting this issue. Thank you for the time you spend working on this project, it makes our lives a lot easier.

My app is suffering an unfortunate real-world side effect rooted in the renderer process name changing `MyApp Helper` to `MyApp Helper (Renderer)` in Chromium/Electron 6. More on our specific problem [here](https://github.com/atom/node-keytar/issues/250), but the summary is:
* Then, from an Electron 5 renderer process, my app set macOS keychain values using keytar
* Now, from an Electron 7 renderer process, my app gets macOS keychain values using keytar, but macOS now prompts the user for a password for every keytar.get of a key.

I believe this is because the renderer name changed. Or possible the renderer bundle ID changing.

Can I specify `CFBundleDisplayName` (and similar values) for my renderer helper .plist file? I've had success defining a custom value for `CFBundleDisplayName` using `extendInfo`, but that value is for the main process. Is there something to akin to `extendInfo` to alter the name of the renderer helper?

In the code it looks like this is somewhere around [here](https://github.com/electron-userland/electron-builder/blob/ebbd9f796e2d8d5b0720b2b699ba24dc159ee692/packages/app-builder-lib/src/electron/electronMac.ts#L58), but I'm not sure the plist values for **helpers** can be extended via the electron-builder API.

What I want to do is customize `CFBundleDisplayName` back to `MyApp Helper` to see if macOS will hopefully not freak out and ask for passwords to access keychain values it created from Electron 5 renderers. It seems like macOS makes its decisions based on either `CFBundleDisplayName`, `CFBundleExecutable`, `CFBundleIdentifier`, or `CFBundleName`, but I'm not sure, so it'd be nice to play with all of those.

Thanks again for the excellent work you do on this project.

aguynamedben:
I'm not sure, but it looks like a similar issue may have happened before with the history of Electron, and the solution could have been to add the option `helperBundleId`. https://github.com/electron-userland/electron-builder/blob/ebbd9f796e2d8d5b0720b2b699ba24dc159ee692/packages/app-builder-lib/src/electron/electronMac.ts#L88

I wonder if the right way to expose this post Electron 6 is `helperRendererBundleId`. I'll play with adding that option to see if it solve my specific problem.

[...]

PR #4692, opened 2020-02-21
Allow customization of bundleIdentifier for the new Electron 6+ renderer processes

aguynamedben:
Addresses https://github.com/electron-userland/electron-builder/issues/4691

Allow customization of `bundleIdentifier` for the new Electron 6+ renderer processes. This change addresses backward-compatibility issues arising from the Electron 6 renderer process name change ([learn more](https://www.electronjs.org/blog/electron-6-0#electron-helper-rendererapp-electron-helper-gpuapp-and-electron-helper-pluginapp)). Without this, upon upgrading to Electron 6+ and building your app, macOS sees the renderer process as having a different `bundleIdentifier` than it saw before.

**Default `bundleIdentifier` values for Renderer process**
Electron 5 and before: `com.acme.CoolApp.helper`
Electron 6+: `com.acme.CoolApp.helper.Renderer`

macOS uses `bundleIdentifier` for things like keychain accesses, so apps that previously accessed things in macOS using the value `com.acme.CoolApp.helper` need to be able to keep that `bundleIdentifier` for their Electron renderer process.

Example usage in my app is:
```
"helperBundleId": "io.hypertools.Command-E.helperGeneric",
"helperRendererBundleId": "io.hypertools.Command-E.helper",
```

This maintains the `bundleIdentifier` value across Electron 5 and Electron 6+.

The real-world issue for app was this: In a new version of our app, our users saw a macOS permission dialog for every single `keytar.get()`. macOS was not trusting our renderer process because the `bundleIdentifier` had changed.


aguynamedben:
@develar I'm working on this more... how many versions back of Electron must the current electron-builder support? I'm looking at options like [this](https://github.com/electron-userland/electron-builder/pull/4692/files#diff-332d71e64017661fb6f3699af2d56323R89) that were deprecated in January 2018.

Also, would it be bad to drop support for Electron 4 support? The current version is Electron 8, and as a user of Electron from Electron 3 (?) I've always found upgrading Electron to be fairly straightforward.

develar:
Supporting v5+ only is ok. Just major version of electron builder will be
released.

aguynamedben:
@develar This is ready for review. I updated the PR description to make it more clear why I added this. It looks like there was probably some history with `helperBundleId` needing to be specified. This PR allows power users to customize the `bundleIdentifier` for Helpers as they need, but still provides good defaults for new users of electron-builder.

It should work across all version of Electron, and it's backward compatible (no breaking changes).

Let me know if there's anything else I can do to help get this merged!

aguynamedben:
I'm hoping this can be merged, is there any reason it's not likely to be merged? It's backwards-compatible, uncomplicated, and really just provides a config variable that was already available—setting the bundleId of the renderer—in a way that works with Electron 6+.

I understand you're extremely busy, I just want to understand if this is likely to be merged. We're in no rush, we have a branch working on a local machine, but given the compiled nature of electron-builder it'd be nice to have these config values in a compiled, packaged version out there the rest of my team can trust.

I appreciate your work. Oh yeah, we just donated!!! (Command E) 🙏

develar:
Before merge I have to release a new version to mark it a production ready (https://github.com/electron-userland/electron-builder/releases/tag/v22.3.6). And then release another version with this change as pre-release.

develar:
Thanks a lot of for this improvement, it moves electron-builder forward.

develar:
22.4.0 was released.

aguynamedben:
Thanks @develar, you rock 💪

PR #4707, opened 2020-02-25
feat(appx): add support for custom extensions in appmanifest.xml

brownbl1:
Add option for specifying a path relative to the app root that gives the user the ability to specify custom xml that should be merged into the <Extensions/> node in the appmanifest.xml. This gives the user the ability to add custom extensions that don't make sense as individual options, without breaking existing support for the options that fall under the <Extensions/> node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "yargs": "^15.1.0"
   },
   "devDependencies": {
+    "typescript-json-schema": "^0.42.0",
     "@babel/core": "^7.8.4",
     "@babel/plugin-proposal-class-properties": "^7.8.3",
     "@babel/plugin-proposal-decorators": "^7.8.3",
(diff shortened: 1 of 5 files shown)


brownbl1:
I regenerated the scheme.json using `yarn schema` so that is why some of the nodes in that document are re-arranged a bit. Otherwise, there are a few small whitespace changes that are unrelated but that seem to be in the right direction, so I left them in. :)

brownbl1:
Also, when running the `yarn schema` command, it failed at first, until I added the following to package.json:
image

Figured that should be a separate PR, which I'm happy to do if you think it's worth it.

brownbl1:
I am not sure why the tests are sometimes failing.. looks like they pass in 3 out of the 4 runs.

brownbl1:
Thanks for taking this into master.. looking forward to the next release so we can start using it!

develar:
@brownbl1 Thanks for feature, pre-release version is available for use — https://github.com/electron-userland/electron-builder/releases/tag/v22.4.0

PR #4731, opened 2020-03-04
emit appimage-filename-updated event

gaodeng:
The appimage-filename-updated event is emitted here, so that if the appimage filename is updated, the app will have the opportunity to update the path in the .desktop file

### DIFF
--- a/packages/app-builder-lib/src/electron/electronMac.ts
+++ b/packages/app-builder-lib/src/electron/electronMac.ts
@@ -121,7 +121,7 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
     helper.CFBundleDisplayName = `${appInfo.productName} Helper ${postfix}`
     helper.CFBundleIdentifier = userProvidedBundleIdentifier
       ? filterCFBundleIdentifier(userProvidedBundleIdentifier)
-      : `${helperBundleIdentifier}.${postfix.replace(/[^a-z0-9]/gim, "")}`
+      : filterCFBundleIdentifier(`${helperBundleIdentifier}.${postfix}`)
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -121,7 +121,7 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
     helper.CFBundleDisplayName = `${appInfo.productName} Helper ${postfix}`
     helper.CFBundleIdentifier = userProvidedBundleIdentifier
       ? filterCFBundleIdentifier(userProvidedBundleIdentifier)
-      : `${helperBundleIdentifier}.${postfix.replace(/[^a-z0-9]/gim, "")}`
+      : filterCFBundleIdentifier(`${helperBundleIdentifier}.${postfix}`)
```
develar, line 124: What's the reason for this change?

gaodeng: https://github.com/electron-userland/electron-builder/pull/4154
In this pr, I didn't realize that this filterCFBundleIdentifier function existed at that time, now I cleaned it up and replaced it with calling this filterCFBundleIdentifier function to avoid duplicate code

gaodeng: @develar  any news?

gaodeng:
Yes it's still relevant.

gaodeng:
any news?

gaodeng:
Yes it's still relevant.

gaodeng:
Yes it's still relevant.

PR #4743, opened 2020-03-06
change Russian locale from "game" to "application"

BusinessDuck:
Why "game" is default word for russian? I think "Приложение" (the Application) is much better and more adaptive for most case of usages electron-builder


develar:
Не проверил внимательно. Спасибо.

PR #4798, opened 2020-03-18
Fixed splash screen in windows portable

montoyaaguirre:
The splash screen is now working I've tested it with the "electron-webpack-quick-start"
I added a bmp image in a new folder "assets" which is located at the project root.
This is an example on how to configure the build in package.json

```
  "build": {
    "appId": "app.philips-dicom-viewer",
    "portable": {
      "splashImage": "assets\\splash.bmp"
    },
    "win": {
      "target": "portable"
    }
  }
```

### DIFF
--- a/packages/app-builder-lib/templates/nsis/portable.nsi
+++ b/packages/app-builder-lib/templates/nsis/portable.nsi
@@ -1,15 +1,12 @@
 !include "common.nsh"
 !include "extractAppPackage.nsh"
 
-# https://github.com/electron-userland/electron-builder/issues/3972#issuecomment-505171582
 CRCCheck off
 WindowIcon Off
 AutoCloseWindow True
 RequestExecutionLevel ${REQUEST_EXECUTION_LEVEL}
 
 Function .onInit
-  SetSilent silent
-
   !insertmacro check64BitAndSetRegView
 FunctionEnd
 
--- a/packages/app-builder-lib/templates/nsis/portable.nsi
+++ b/packages/app-builder-lib/templates/nsis/portable.nsi
@@ -7,6 +7,9 @@ AutoCloseWindow True
 RequestExecutionLevel ${REQUEST_EXECUTION_LEVEL}
 
 Function .onInit
+  !ifndef SPLASH_IMAGE
+    SetSilent silent
+  !endif
   !insertmacro check64BitAndSetRegView
 FunctionEnd
 


packages/app-builder-lib/templates/nsis/portable.nsi
```diff
@@ -1,15 +1,12 @@
 !include "common.nsh"
 !include "extractAppPackage.nsh"
 
-# https://github.com/electron-userland/electron-builder/issues/3972#issuecomment-505171582
```
montoyaaguirre: This does not seem relevant

packages/app-builder-lib/templates/nsis/portable.nsi
```diff
@@ -1,15 +1,12 @@
...
 CRCCheck off
 WindowIcon Off
 AutoCloseWindow True
 RequestExecutionLevel ${REQUEST_EXECUTION_LEVEL}
 
 Function .onInit
-  SetSilent silent
```
montoyaaguirre: In silent mode, the splash image will NOT be shown

develar: Without specified splash should be silent, so, here we need some condition.

montoyaaguirre: oh, I missed this

montoyaaguirre: Ohh, that's correct. I'm thinking something like this:
``` 
!ifndef SPLASH_IMAGE
  SetSilent silent
!endif
```

montoyaaguirre:
@develar  I've fixed the issue and tested again the two use cases for a portable windows app (with splash-screen and without it).

RoderickQiu:
Sounds great.

catdad:
@montoyaaguirre I am the one who implemented the original splash page work in #4425, where it was left incomplete (the path resolution issue specifically that you wrote about in your blog). I recently decided to finish that work and created #4990 for that. I did see that even in the latest `electron-builder`, I still had to resolve the path relative to the project to get it working. I don't see that in your PR though. Was it at any point allowing images to be specified relative to the project directory?

ietabhi:
I am eagerly waiting for this feature to be available for electron. Please merge so that we can use same.

ShenKelly:
why i use it in cordova-electron，the splashscreen  just in a flash

ietabhi:
How can i use this directly in my local? I want to generate same as it's taking long to merge to develop. 
Can you please share all steps, or publish your branch to npm with some different pacakge name for sometime

debjava:
I want to use it, when can we expect this splash image integration in electron-builder? It is really a nice feature to incorporate.

catdad:
I tested splash images in the 22.9.1 pre-release version and they work correctly there now, since #4990 was merged. Seems this PR is not necessary anymore as it is a duplicate

montoyaaguirre:
Closing this PR as #4990 contains this functionality and was Merged.

PR #4817, opened 2020-03-26
fix(multipleRangeDownloader): when response ends nothing more happens

Sergey-Churilin:
Nothing more happen when response ends. It leads to issue that no events are pushed outside: e.g. download is successful. In successful cases - I'm not an expert in streams - but as I understand somewhere resolve is called. I didn't find more proper solution, so just wait for 10 seconds and call reject. If resolve was already called previously in success case, this will not lead to any issues.
This PR fixes issue #[4618](https://github.com/electron-userland/electron-builder/issues/4618)
@develar could you please have a look

### DIFF
--- a/packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts
+++ b/packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts
@@ -93,6 +93,13 @@ function doExecuteTasks(differentialDownloader: DifferentialDownloader, options:
     const dicer = new DataSplitter(out, options, partIndexToTaskIndex, m[1] || m[2], partIndexToLength, resolve)
     dicer.on("error", reject)
     response.pipe(dicer)
+	
+	response.on('end', () => {
+	  setTimeout(() => {
+	    request && request.abort()
+	    reject(new Error("Response ends without calling any handlers"))
+	  }, 10000)
+	})
   })
   differentialDownloader.httpExecutor.addErrorAndTimeoutHandlers(request, reject)
   request.end()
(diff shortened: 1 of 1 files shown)


develar:
I don't like this timeout :( Not clear for me why it is required and why we cannot fix the issue without it.

Sergey-Churilin:
> I don't like this timeout :( Not clear for me why it is required and why we cannot fix the issue without it.

Sergey-Churilin:
Hi @develar , it is not the best option, I agree. Let me add more details according to my investigations (I added logs almost everywhere during testing) and maybe you can propose better solution.

When response.on('end') event is called in multipleRangeDownloader inside doExecuteTasks function, there are two scenarios:

**The first is a bad case:** logs in these event were the last point in download chain. 
If I call resolve in ten seconds, the code logic continues in DifferentialDownloader file.
Inside **async doDownloadFile**(tasks, fdList) function  - fileOut.on("finish") event is called.
Then **fileOut.close** function is called. Inside its callback  in try section **digestTransform.validate()** function leads to catch and reject is called.
Error that is passed to reject is the following: "Cannot download differentially, fallback to full download: Error: sha512 checksum mismatch, expected..."
I thought that maybe some errors were in writable stream, so in tests I also added subscription to **fileOut.on("error")** event.  It was never called.

In **success** case **digestTransform.validate()** goes smoothly and all father logic works as expected.

The main question is: why **fileOut.on("finish")** event was not called in bad case at all. 
Due to that issue the only one solution that I found was to add timeout() and call reject()

Sergey-Churilin:
Hi @develar can you propose some solutions, this fix it rather important for my application

Sergey-Churilin:
Hi @develar , could you please have a look on the fix details above

Sergey-Churilin:
Hi @develar, could you please assist on that pr?

Issue #3637, opened 2019-01-25
Electron builder resigning other binaries included in extrafiles


msdos621:
* **Version**:  20.28.4
Today in our application we bundle a partner's exe as part of extraFiles.    
```json
"build": {
    "mac": {
      "category": "public.app-category.utilities",
      "extraFiles": [
        "bin/mac/**/*"
      ]
    },
    "win": {
      "publisherName": "Redacted Technologies, Inc",
      "certificateSubjectName": "Redacted Technologies, Inc",
      "extraFiles": [
        "bin/win/**/*"
      ]
    },
```

The problem we have is that the code signing process seems to blow away the signature from our partners .exe file and replaces it with our code certificate.  Here is a screenshot of the signature before we build and a picture of the signature of the version that ends up installed
<img width="818" alt="beforeafter" src="https://user-images.githubusercontent.com/882850/51757611-d3f66d80-2091-11e9-87fc-8ed63561604d.png">

Is there anyway to exclude files form being signed or perhaps a way to ignore extraFiles completely during signing?

hrishabhkumar:
Is there any update on how to skip the signing of already signed binaries?

PR #4833, opened 2020-04-01
feat: win custom sign should have same signature of doSign

link89:
- The win custom sign script should be able to use `WinPackage` as `doSign` does.
- Export `doSign` method to allow users to wrap it in their script instead of rewrite whole code signing logic. For example, to just skip some files to sign.


link89:
@develar Can you help to review and approve this? 
We found that after upgrade from v20 to v22,  all binary files under `app.asar.unpacked` are signed, which would break some third party SDK, like this one https://github.com/electron-userland/electron-builder/issues/3637.  So we hope that `electron-builder` can have some easy way to skip some files to sign.

PR #4873, opened 2020-04-12
Change archive names

BusinessDuck:
Web installer, should be configured by options "archiveName". That be able to use only one installer for all versions of app. For example if i want to remove version from archive name that prevent web installer from downloading "holded" versions, and each one installation would be latest every time.


BusinessDuck:
build stalled, rerun please @

BusinessDuck:
Yes is still relevant

PR #4915, opened 2020-04-30
chore: enable NPM_NO_BIN_LINKS for yarn 1.x.x

Strangerxxx:
Enables NPM_NO_BIN_LINKS by pushing `--no-bin-links` parameter to yarn exec args
Since yarn 2 was released `--no-bin-links` parameter was deprecated


linegel:
Would be really nice to have this one, we use upgrading currently [wojtkowiak/meteor-desktop](https://github.com/wojtkowiak/meteor-desktop) and looks like without these change there is no way for us to use yarn and otherwise we will be forced to use npm

PR #4929, opened 2020-05-05
fix: cached update validation failing on undefined filename

### LINKED ISSUE #4928: No update-info.json file in update cache folder 😶
* **electron-builder**: "22.5.1"
* **Electron Version**: 8.2.2
* **Electron Type**: current
* **electron-updater**: "4.2.5"
* **Target**: windows

I'm seeing this error after uploading a new version to an S3 bucket.

The upload went fine and all of the artifacts are there as far as I can tell.

Here is the log of the local application (output to AppData log.log from `electron-log` logger)

```log
[2020-05-05 22:41:04.824] [info] Checking for update
[2020-05-05 22:41:05.017] [info] Generated new staging user ID: 586a9abf-80b7-5cb4-a8db-9af47cc1d5a7
[2020-05-05 22:41:06.100] [info] Found version 0.0.2-beta (url: MyApp Setup 0.0.2-beta.exe)
[2020-05-05 22:41:06.252] [info] Downloading update from MyApp Setup 0.0.2-beta.exe
[2020-05-05 22:41:06.273] [debug] updater cache dir: C:\Users\zurg\AppData\Local\app.com.org-updater
[2020-05-05 22:41:07.510] [error] Error: TypeError: Cannot read property 'fileName' of undefined
    at t.DownloadedUpdateHelper.cachedUpdateFile [as getValidCachedUpdateFile] (C:\Users\zurg\AppData\Local\Programs\app.com.org\resources\app.asar\.webpack\main\webpack:\node_modules\electron-updater\out\DownloadedUpdateHelper.js:149:20)
    at t.DownloadedUpdateHelper.cachedUpdateFile [as validateDownloadedPath] (C:\Users\zurg\AppData\Local\Programs\app.com.org\resources\app.asar\.webpack\main\webpack:\node_modules\electron-updater\out\DownloadedUpdateHelper.js:88:11)
    at b.executeDownload (C:\Users\zurg\AppData\Local\Programs\app.com.org\resources\app.asar\.webpack\main\webpack:\node_modules\electron-updater\out\AppUpdater.js:726:11)
[2020-05-05 22:41:07.522] [info] Updating **with** synchronization
[2020-05-05 22:41:07.832] [error] TYPE should be object
```

After looking at the source code the only place the `fileName` property is accessed is here:
```ts
    if (cachedInfo.fileName == null) {
      logger.warn(`Cached update info is corrupted: no fileName, directory for cached update will be cleaned`)
      await this.cleanCacheDirForPendingUpdate()
      return null
    }
```

I think this actually means that for some reason the `update-info.json` file that the cachedInfo object is supposed to draw from does not exist.  Indeed I checked my directory and there is nothing.

## Are there more logging steps that can be done?

b-zurg:
Closes #4928

In my setup I was encountering a situation where there was never an `update-info.json` file created and the `cachedInfo` object in `DownloadedUpdateHelper.getValidCachedUpdateFile()` was `undefined`.

This effectively meant that an error was thrown and the update never downloaded.

This makes the code a little bit more resilient by using optional chaining and nullish coalescing from typescript to ensure that the cachedInfo object is not undefined and then that the `fileName` is not null.   If it is undefined it will fall back to the `cleanCacheDirForPendingUpdate` function.

I also changed another instance of the `==` operator to `===` as that is generally preferred.

### Update

The root cause of this issue is that the try/catch that is supposed to check if the file exists does not work.  **It does not enter the catch block**.  I discovered this by a large set of console logs and multiple packages and tests.  

What is preferable is an early existence check and early return if the cachedInfo file does not exist.

### DIFF
--- a/packages/electron-updater/src/DownloadedUpdateHelper.ts
+++ b/packages/electron-updater/src/DownloadedUpdateHelper.ts
@@ -106,7 +106,8 @@ export class DownloadedUpdateHelper {
       return null
     }
 
-    if (cachedInfo.fileName == null) {
+    const isCachedInfoFileNameValid = cachedInfo?.fileName !== null ?? false
+    if (isCachedInfoFileNameValid) {
       logger.warn(`Cached update info is corrupted: no fileName, directory for cached update will be cleaned`)
       await this.cleanCacheDirForPendingUpdate()
       return null
--- a/packages/electron-updater/src/DownloadedUpdateHelper.ts
+++ b/packages/electron-updater/src/DownloadedUpdateHelper.ts
@@ -3,7 +3,7 @@ import { createHash } from "crypto"
 import { createReadStream } from "fs"
 import isEqual from "lodash.isequal"
 import { Logger, ResolvedUpdateFileInfo } from "./main"
-import { pathExists, readJson, emptyDir, outputJson, unlink } from "fs-extra"
+import { pathExists, readJson, emptyDir, outputJson, unlink, pathExistsSync } from "fs-extra"
 import * as path from "path"
 
 /** @private **/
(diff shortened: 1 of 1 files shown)


b-zurg:
I don't think this is a bad change but it not address the root cause.  Which was that the `cachedInfo` object itself was undefined.

PR #4945, opened 2020-05-08
Squirrel.Windows documentation

koush:
The app must be able to handle Squirrel.Windows startup events that occur during install and uninstall. Fixes missing shortcut icons, double startup, etc.


koush:
I"m not sure why a pull on docs would fail.

b-zurg:
lol. I don't imagine why there would need to be a job on docs PR either 😂

PR #4990, opened 2020-05-19
feature(portable): allowing splash image to be defined relative to the project directory

catdad:
This finishes the work started in #4425, for the Windows portable splash image feature requests in  #2548 and #3972. With these changes, the splash image no longer needs to be an absolute path. Now, paths can be defined relative to the project. For example, I have this in my package.json:

```json
{
  "name": "my-app",
  "build": {
    "win": {
      "target": [
        "portable"
      ],
      "icon": "assets/icon.ico"
    },
    "portable": {
      "artifactName": "${name}-v${version}-portable.${ext}",
      "splashImage": "assets/splash.bmp"
    }
  }
}
```

When provided, the splash image will show while the portable package is extracting, and disappear right before the Electron window is launched. When not provided, the extraction process will be silent. Note that only bitmap images are supported.


Timespace7:
I manually modify NsisTarget.js and portable.nsi according to your commits in my project. It did not work. Just something flash and disappear within one second after running the app.

catdad:
Does the app you are testing on take a long time to unzip. I had a comment in #4425 that explains this already. The splash page is not for suitable for every app. It the app is small (like a sample) it will unzip so fast that the user will immediately see a response to running the portable executable. However, large apps, like the one in my case or the ones reported by users in the mentioned issues above, it can take 30 seconds or longer for all the files in the portable executable to unzip. In those cases, the user doesn't see a response to running the executable for a while, and I've found that users will think they didn't run it correctly and double-click the executable several times while waiting impatiently.

This implementation of a splash page is meant for those developers that have found out that their application is very large and takes a long time to unzip. In those cases, it's not a "quick flash" but rather an image that shows up immediately and shows for the entire time it takes to unzip the application.

Timespace7:
It usually takes about 10s to see the application window after I open it. I saw something like a window but not the splash screen flashes just after opening the app. Actually, I found [this blog](http://dev.to/montoyaaguirre/adding-a-splash-screen-to-portable-electron-builder-apps-34d3) showing something very similar to your commits. It demonstrates the effect of the splash screen. However, it just doesn't work in my project.

catdad:
@Timespace7 do you mind sharing your application? Is it something on GitHub that we can see? I'd like to help with your concern, but saying "it just doesn't work" doesn't provide any information I can act on.

Timespace7:
During making a sample project, I finally found the reason why the splash screen doesn't work. It is because my splash.bmp has problem but it can be viewed as usual. Finally, I just save it again using Paint and the splash can be shown now.

DanielMcAssey:
Any luck getting this merged? Got an app that takes 20-30 seconds to unpack.

ietabhi:
It's frustrating when click on portable app and same not opening. user click multiple times as he don't have any clue. 
I am eagerly waiting for this feature to be available for electron. Please merge so that we can use same.

catdad:
This is still relevant and necessary, and the community has shown interest in this functionality. I have not heard from members of the project, so I am not sure what I can do to move it forward. Perhaps there is feedback that I missed?

develar:
Thanks for PR, sorry for stale bot :)

DanielMcAssey:
When will a new version that includes this be released?

PR #5012, opened 2020-05-27
Fix issue where we are unable to manipulate build contents once created

varnikaj:
Right now, after the build is created we cant manipulate the contents of archive created which we need for our project


develar:
Not clear for me why temp file is required. Could you please elaborate?

varnikaj:
Basically after the zip file is created using electron builder, we need to rename the app file created based on some parameters which are decided after the build is created.

Screenshot 2020-06-22 at 3 39 49 PM

lukeapage:
@varnikaj 
I think I can fill in the gaps here.

This PR is what our patch does to make electron builder usable as part of our process, but it cannot be merged as-is.

The patch is a hack to stop a blackcap being created on the zip - the blockmap stops us from updating the zip after electron builder, which we do during our own customisation process.

To make something electron-builder could merge, we probably need an option to stop creating a blockmap in order that we can configure electron builder to skip it for us.

See also this issue: https://github.com/develar/app-builder/issues/24
which *I think* might be the same issue.

I don't know why a blockmap is appended to the zip (what's it used for?) I also don't know what the updateInfo is that is returned by the command - so I'm not saying I have all the answers but hopefully this fills you in a bit more.

PR #5032, opened 2020-06-04
Docs: Fix typo in "Multi Platform Build"

adam-lynch:


janbrasna:
@develar This typo is still present in current published docs, is there a workflow different than the CI pipeline so that it'll get fixed for the site sometime later? Or the distribution failed due to the CircleCI failing?

PR #5071, opened 2020-06-15
fix(windows): fix encoding issues while verifying signatures

orzFly:
Our code signing certificate contains Chinese company name in the CN field. Encoding issues may be complex on Windows so I just encode the powershell output as base64-ed UTF-8 bytes.


hezhuojie:
Holyshit. 
It convert CN **from gbk encoded to utf-8** in checkForUpdate. 
But in publish stage, `windowsCodeSign` automatically extracts publisher name of  **gbk encoded** from code signing certificate in. The gbk encoded publish name is wrote to `app-update.yml` and is used for checkForUpdate.
So `AppUpdater` must fail to verify signature due to compare utf-8 encoded publish name and gbk encoded name.

develar:
@hezhuojie Thanks for finding, PR welcome :)

PR #5073, opened 2020-06-15
Snap: Add appPartStage and layout options

adam-lynch:
And here's the PR for the docs: #5074

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -4289,6 +4289,12 @@
           ],
           "description": "Specifies any [parts](https://snapcraft.io/docs/reference/parts) that should be built before this part.\nDefaults to `[\"desktop-gtk2\"\"]`.\n\nIf list contains `default`, it will be replaced to default list, so, `[\"default\", \"foo\"]` can be used to add custom parts `foo` in addition to defaults."
         },
+        "appPartStage": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
         "artifactName": {
           "description": "The [artifact file name template](/configuration/configuration#artifact-file-name-template).",
           "type": [
@@ -4417,6 +4423,9 @@
             "null",
             "string"
           ]
+        },
+        "layout": {
+          "type": "object"
         },
         "mimeTypes": {
           "anyOf": [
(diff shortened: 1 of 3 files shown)


rwwagner90:
We need this functionality as well. `layout` specifically fixes some binaries we ship.

rwwagner90:
@adam-lynch is there a way to try this PR out in our app?

adam-lynch:
@rwwagner90 `npm i -D @adam-lynch/app-builder-lib@^22.8.0-3`

rwwagner90:
@adam-lynch use that instead of having electron-builder as a dependency?

rwwagner90:
@adam-lynch also, is there an example of how to pass layout? We need to add this:

```yml
layout:
  /usr/lib/x86_64-linux-gnu/imlib2:
    bind: $SNAP/usr/lib/x86_64-linux-gnu/imlib2
```

adam-lynch:
The documentation is updated in https://github.com/electron-userland/electron-builder/pull/5074. It would be:

```
"layout": {
  "/usr/lib/x86_64-linux-gnu/imlib2": {
    "bind": "$SNAP/usr/lib/x86_64-linux-gnu/imlib2"
  }
}
```

rwwagner90:
@adam-lynch thanks so much! I tried installing your package, but I am not sure how to run it. I tried `electron-forge`, `app-builder-lib` and various other things. Once I install the package, how do I run it?

adam-lynch:
Ah sorry, I'm so used to how I use it. I'm not sure you can test my PR 🤔. app-builder-lib is an internal package that can be used programmatically but it's not documented and it works differently / accepts different config

rwwagner90:
@adam-lynch is there a way we can build the entire `electron-builder` package including your changes and publish it somewhere?

rwwagner90:
@adam-lynch looks like we need to update https://github.com/electron-userland/electron-builder/blob/7b335dbd7e5ca754598ca93318be1404fc518485/test/snapshots/BuildTest.js.snap#L1184 to be `5320` according to the Circle failure. Guessing the size is a bit different since we added some things.

@develar is anything else needed before this can be merged and published? I am very unclear on how to run this without a full publish, since the repo has a lot of packages, I cannot use a yarn resolution or install a fork, it needs a built release.

rwwagner90:
@adam-lynch @develar any updates here? I desperately need the ability to pass `layout` and due to this being a monorepo, I don't think there is any way I can build this myself and use it.

rwwagner90:
Glad to see this merged! Is it released yet?

Issue #4701, opened 2020-02-25
[electron-updater] Update is installed even though signature verification fails


rajivshah3:
* **Version**: 
`electron-builder`: 22.3.5
`electron-updater`: 4.2.4

* **Target**: Windows (NSIS)

dead150 is only a partial fix for the [signature verification bypass issue](https://blog.doyensec.com/2020/02/24/electron-updater-update-signature-bypass.html) recently disclosed by Doyensec. While it is no longer possible to trigger the parse errors with single or double quotes as of dead150, there are other ways to cause them. 

From the report:
> we believe that other attack payloads for the same vulnerable code path still [exist] in Electron-Builder.

In my opinion, the root cause of the vulnerability lies in the fact that even though signature verification is failing, the update is still installed:
https://github.com/electron-userland/electron-builder/blob/caebf37282a17786cff7c948eeec8fbb21b37a68/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts#L37-L38

https://github.com/electron-userland/electron-builder/blob/caebf37282a17786cff7c948eeec8fbb21b37a68/packages/electron-updater/src/NsisUpdater.ts#L42

So, even though an error is encountered, `null` is resolved and the update is installed anyway. I opened this issue because I was hoping to start a discussion on the following:

- Is there a reason why we would want the update to be installed despite the fact that signature verification failed (i.e. is there a valid use case for the current behavior)?
- We would prefer that any errors in signature verification cause the update process to abort. Would a PR for the ability to opt-in to stricter signature verification (e.g. `strictSignatureVerification: true` in the `electron-builder` config) be accepted?

rajivshah3:
This is still relevant

huntr-helper:
‎‍🛠️ A fix has been provided for this issue. Please reference: https://github.com/418sec/electron-builder/pull/1

🔥 This fix has been provided through the https://huntr.dev/ bug bounty platform.

huntr-helper:
### Bug Bounty

We have opened up a bounty for this issue on our bug bounty platform. Want to solve this vulnerability? Go to https://huntr.dev/

We will submit a pull request directly to your repository with the fix as soon as possible. Want to learn more? Go to https://github.com/418sec/huntr 📚

_Automatically generated by @huntr-helper..._

huntr-helper:
### Bug Bounty

We have opened up a bounty for this issue on our bug bounty platform. Want to solve this vulnerability? Go to https://huntr.dev/

[...]

PR #5082, opened 2020-06-18
Security Fix for Signature Validation Bypass - huntr.dev

huntr-helper:
https://huntr.dev/app/users/Mik317 has fixed the Signature Validation Bypass vulnerability 🔨. Mik317 has been awarded $25 for fixing the vulnerability through the huntr bug bounty program 💵. Think you could fix a vulnerability like this?

Get involved at https://huntr.dev/

Q | A
Version Affected | ALL
Bug Fix | YES
Original Pull Request | https://github.com/418sec/electron-builder/pull/1
GitHub Issue URL | https://github.com/electron-userland/electron-builder/issues/4701
Vulnerability README | https://github.com/418sec/huntr/blob/master/bounties/npm/electron-builder/1/README.md

### 📊 Metadata *

_Please enter the direct URL for this bounty on huntr.dev. This is compulsory and will help us process your bounty submission quicker._

### ⚙️ Description *

The issue has been fixed through 2 different ways:
1. Restricting the characters that could be inside the `filename` and `package` * (not only quotes like in the bypassed fix which is mentioned inside the https://github.com/electron-userland/electron-builder/issues/4701 issue)
2. Deleting all the `resolve(null)` lines initialized inside every `unsuccessful "if statement"`, which is suggested by the report it-self inside this comment:

> Screenshot from 2020-06-16 17-59-29

### 💻 Technical Description *

Removing the `resolve(null)` code when the `if` statements fail, makes us able not to `update` the packages when an error occurs (removing the root problem, so the fact the packages are still updated even if the `signature` isn't valid at all). At the same time, the increased checks done inside the characters that could figure inside the `filename/packagename` make us able to avoid malicious crafted `filenames` which could lead to secondary behaviors not correctly handled by the software. 

### 🐛 Proof of Concept (PoC) *

I've not been able to fully make a working POC, and that's why I based my fix mostly on comments I found on the issues created inside the original repo and code review. That said, the fixes introduced are really few lines of code, so this won't certainly break any functionality of the application and at the same time it __should__ fix correctly the issue. In negative cases, I suggest to publish/attach to every bounty a PoC which could be tested by anyone (in certain bounties is still present, but not in all of them), in order to be sure the fix has been patching the highlighted issue as well.

### 🔥 Proof of Fix (PoF) *

Read above


packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
```diff
@@ -16,8 +16,8 @@ export function verifySignature(publisherNames: Array<string>, tempUpdateFile: s
       try {
         if (error != null || stderr) {
           handleError(logger, error, stderr)
-          resolve(null)
-          return
+          logger.warn(`Can't verify the signature: ${error}`)
+          throw error
```
develar, line 20: Please add option to preserve `warn` as before`.

toufik-airane: I'm sorry, I don't understand your comment.
Can you illustrate that with an example or details?
Thank you.

JamieSlome:
@develar - could you explain your requested changes in more detail?

packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
develar: Should be a way to ignore error. I am afraid that in case of internal powershell error we cannot block auto-update.

toufik-airane: @Mik317 can you take care to replace `throw error` with `return`?

packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
```diff
@@ -35,8 +35,7 @@ export function verifySignature(publisherNames: Array<string>, tempUpdateFile: s
       }
       catch (e) {
         logger.warn(`Cannot execute Get-AuthenticodeSignature: ${error}. Ignoring signature validation due to unknown error.`)
-        resolve(null)
-        return
+        throw error
```
toufik-airane, line 38: @Mik317 can you take care to replace `throw error` with `return`?

JamieSlome:
@develar - I have made the requested changes - let me know if you are happy! 🍰 

Thanks @toufik-airane for pointing this out in more detail! 🪁

packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
JamieSlome: @develar - see comment in below thread.

rajivshah3:
If there is no error thrown, won't the auto-update silently fail? If the app does not have logging enabled with something like `electron-log`, I don't really know how people will be able to tell what's preventing the app from updating.

packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
Mik317: Hi @toufik-airane :)
I've updated on my `master` branch as you suggested previously :).
I'm not sure if it gets reflected directly here or I need to do something else, please let me know in case :)

Regards,
Mik

JamieSlome: @Mik317 - I have already added these changes in your commit in this pull request for you - nothing for you to do! 🍰

packages/electron-updater/src/NsisUpdater.ts
```diff
@@ -235,6 +235,7 @@ async function _spawn(exe: string, args: Array<string>): Promise<any> {
...
 }
 
-function hasQuotes(name: string): boolean {
-  return name.includes("'") || name.includes('"') || name.includes('`')
-}
\ No newline at end of file
+function isInvalid(name: string): boolean {
+  return name.includes("'") || name.includes('"') || name.includes('`') || name.includes('#') || name.includes('|') || name.includes('/') || name.includes('\') || name.includes(':') || name.includes('*') || name.includes('<') ||
+  name.includes('>') || name.includes('?')
+}
```
kevinlinv, line 241: Would checking for `name.includes(':')` not be problematic here?  I was under the impression that `destinationFile` and `packageFile` are an absolute path that includes the harddrive id, e.g.
```
C:\\Users\\User\\AppData\\Local\\myApp-updater
```
Matching `:` will cause the above path to be considered as invalid while it really should not be?

---
Also the same goes with `name.includes('\\')` where the path on windows will always contains `\\` 🤔 
```
C:\\Users\\User\\AppData\\Local\\myApp-updater
```

JamieSlome: @kevinlinv - I have reflected the requested changes!

Cheers! 🍰

JamieSlome:
Any updates on this?

packages/electron-updater/src/NsisUpdater.ts
b-zurg: A comment on code quality, having an array of characters and using an array function would be much more readable and maintainable. 

e.g. 
```ts
...
const disallowedChars = [':','//', ...];
if(disallowedChars.some(char => name.includes(char)) return false;
return true;
...
```

JamieSlome: @b-zurg - I have made the requested changes! 🍰

JamieSlome:
@kevinlinv @develar @b-zurg - all the requested changes have been made on code quality and array filtering.

Are we able to get this PR merged?

kevinlinv:
There seems to be a conflict in the file, this PR probably needs to be rebased. As for

> Are we able to get this PR merged?

I believe @develar would be the go-to person who can merge any PR in this project.

JamieSlome:
@kevinlinv - resolved the merge conflicts!

We can wait to hear from @develar! 🍰

packages/electron-updater/src/NsisUpdater.ts
```diff
@@ -27,6 +27,11 @@ export class NsisUpdater extends BaseUpdater {
       downloadUpdateOptions,
       fileInfo,
       task: async (destinationFile, downloadOptions, packageFile, removeTempDirIfAny) => {
+        if (isInvalid(destinationFile) || (packageFile != null && isInvalid(packageFile))) {
+          throw newError(`destinationFile or packageFile contains illegal chars`, "ERR_UPDATER_ILLEGAL_FILE_NAME")
+        }
```
kevinlinv, line 32: I believe we no longer to check for this invalid path, I've since opened a PR to escape these characters. See: https://github.com/electron-userland/electron-builder/pull/5116

Basically, throwing here means that we prevent user with username `.../User's/...` (_which is pretty common naming convention_) from updating and escaping would make more sense.

kevinlinv (commented):
Sorry for the back and forth comments, I forgot to mention this before!

JamieSlome:
@develar @stefanjudis - any updates on the status of this?

Cheers! 🍰

stefanjudis:
I wasn't involved with this project for years. :) Good luck!

Issue #4889, opened 2020-04-20
Quotes in updater path


Laruxo:
* **Version**: since `electron-updater` 4.2.0

* **Target**: Windows

Since [this commit](https://github.com/electron-userland/electron-builder/commit/dead150769463adbff38b19e9099df6547db0e24) quotes in destinationFile or packageFile throw an error during update. In windows it is possible to have account named `Jesse's` or `Joy's`, which results in download path being `C:\Users\Jesse's\AppData\Local\myApp-updater`.

I am not sure why this restriction was introduced, but it is obvious it should not crash on update for some users.

Laruxo:
Related to https://github.com/electron-userland/electron-builder/issues/4701

rajivshah3:
This was an incomplete fix for a vulnerability. Your best option is to use `patch-package` to revert this change if your users are running into issues. To patch the vulnerability, I think [this](https://github.com/iotaledger/trinity-wallet/blob/cf6e1099de10452a43cf59b144ff759788e77c72/src/desktop/patches/electron-updater%2B4.0.14.patch) is more appropriate.

PR #5116, opened 2020-07-03
fix(electron-updater): escape quotes instead of throwing an error

kevinlinv:
## Descriptions

This https://github.com/electron-userland/electron-builder/commit/dead150769463adbff38b19e9099df6547db0e24 commit was added as a security measure to avoid command injection when we use downloaded file path that contains quotes when running powershell command.

e.g: `Get-AuthenticodeSignature 'C:\\path\\my-bad-';calc;'filename.exe'` will start a calculator app.

Banning the quotes does stop the command injection, however it also introduce unwanted behavior when performing an update, an example, when the user's path contains a quote, update will never be installed. See: https://github.com/electron-userland/electron-builder/issues/4889

### This PR

Fixes https://github.com/electron-userland/electron-builder/issues/4889 by escaping the quotation as per windows powershell documentation.

e.g. 
```ts
// Unescaped
`Get-AuthenticodeSignature 'C:\\Users\\MyName's\\AppData\\Local\\MyApp.exe'`

// Escaped
`Get-AuthenticodeSignature 'C:\\Users\\MyName''s\\AppData\\Local\\MyApp.exe'`
```

Also when there is a command injection
```ts
// Unescaped
`Get-AuthenticodeSignature 'C:\\Users\\MyName';calc;'s\\AppData\\Local\\MyApp.exe'`

// Escaped
`Get-AuthenticodeSignature 'C:\\Users\\MyName'';calc;''s\\AppData\\Local\\MyApp.exe'`
```

The calculator app will not be opened.

### DIFF
--- a/packages/electron-updater/src/NsisUpdater.ts
+++ b/packages/electron-updater/src/NsisUpdater.ts
@@ -27,10 +27,6 @@ export class NsisUpdater extends BaseUpdater {
       downloadUpdateOptions,
       fileInfo,
       task: async (destinationFile, downloadOptions, packageFile, removeTempDirIfAny) => {
-        if (hasQuotes(destinationFile) || (packageFile != null && hasQuotes(packageFile))) {
-          throw newError(`destinationFile or packageFile contains illegal chars`, "ERR_UPDATER_ILLEGAL_FILE_NAME")
-        }
-
         const packageInfo = fileInfo.packageInfo
         const isWebInstaller = packageInfo != null && packageFile != null
         if (isWebInstaller || await this.differentialDownloadInstaller(fileInfo, downloadUpdateOptions, destinationFile, provider)) {
@@ -233,8 +229,4 @@ async function _spawn(exe: string, args: Array<string>): Promise<any> {
       reject(error)
     }
   })
-}
-
-function hasQuotes(name: string): boolean {
-  return name.includes("'") || name.includes('"') || name.includes('`')
 }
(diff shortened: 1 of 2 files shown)


packages/electron-updater/src/NsisUpdater.ts
```diff
@@ -27,10 +27,6 @@ export class NsisUpdater extends BaseUpdater {
       downloadUpdateOptions,
       fileInfo,
       task: async (destinationFile, downloadOptions, packageFile, removeTempDirIfAny) => {
-        if (hasQuotes(destinationFile) || (packageFile != null && hasQuotes(packageFile))) {
-          throw newError(`destinationFile or packageFile contains illegal chars`, "ERR_UPDATER_ILLEGAL_FILE_NAME")
-        }
```
kevinlinv, line 32: I am moving the check to the verifier, since I think that is the more appropriate to do this!

kevinlinv:
@develar could you help take a look at this PR please 🙏  

We have a subset of user that can no longer automatically update their app because their username contains a `'` character.

kevinlinv:
Friendly 🏓 @develar

kevinlinv:
A 🏓 @develar

kevinlinv:
Ping @develar

kevinlinv:
Thanks for merging 👍

sandeep1995:
Thank you very much.

Issue #5024, opened 2020-06-01
Protocol doesn't work on Linux


mahnunchik:
* **Version**: 22.7.0

* **Electron Version**: 8.3.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: linux - snap, AppImage

Deep linking doesn't work on linux.

https://www.electronjs.org/docs/api/app#apprequestsingleinstancelock

Possible fixes:

* https://github.com/electron-userland/electron-builder/pull/5001
* https://github.com/electron-userland/electron-builder/pull/4909

mahnunchik:
Related: https://github.com/electron-userland/electron-builder/issues/4035

mahnunchik:
If `exec` is null then `%U` is added: https://github.com/electron-userland/electron-builder/blob/7b335dbd7e5ca754598ca93318be1404fc518485/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts#L95-L105

mahnunchik:
AAAAAAAAAAAAAAAAAA!

gcascio:
I agree that this is still relevant. There are currently three PRs related to this problem, if there is anything I can do to get them merged I would be happy to help. In the meantime I applied a workaround, if anybody is interested [here](https://github.com/gcascio/electron-custom-url-workaround) is a minimal example.

anteeek:
Yeah, same here. Hovewer for some reason snap works for me, while AppImage doesn't.

lishid:
@mahnunchik Might want to close this issue as #5001 and #4909 have been merged.

mahnunchik:
https://github.com/electron-userland/electron-builder/issues/5024#issuecomment-739013018

julien-boucly-needone:
+1

mmaietta:
Just to confirm, is this still occurring on latest prerelease version v22.10.5?
I see PRs were merged in 22.9.1
https://github.com/electron-userland/electron-builder/releases/tag/v22.9.1

lishid:
Should be safe to close since the two fixes have been merged. (see https://github.com/electron-userland/electron-builder/issues/5024#issuecomment-703473136)

mahnunchik:
https://github.com/electron-userland/electron-builder/issues/5024#issuecomment-855159867

mahnunchik:
https://github.com/electron-userland/electron-builder/issues/5024#issuecomment-907656735

mmaietta:
[...]

PR #5001, opened 2020-05-22
add missing %U in desktop file of snap builds

timothylui:
Without this, using protocols in snaps don't work properly as the app is opened by the url isn't passed to the app.

PR #4909, opened 2020-04-28
Add missing %U in desktop file of appimages

Simon-Laux:
this should fix #4035

Add missing %U in .desktop file of appimages, which is missing because the `Exec` var is set, so the code to add it to the other desktop files is not active: 
https://github.com/electron-userland/electron-builder/blob/f938de5aae09cb832a6f7f742e64a0d6c025fb9f/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts#L95-L105

Issue #4035, opened 2019-07-09
app.setAsDefaultProtocolClient  doesn’t work on Linux .AppImage


brunolemos:
* **electon-builder**: 20.44.4 and tried 21.0.13

* **electon-updater**: 4.0.14 and tried 4.1.2

* **Target**: `linux` (`.AppImage`, `Ubuntu`)

`app.setAsDefaultProtocolClient` works for macOS and Windows, but not for Linux, at least when using `.AppImage`, which is the only one I use. 

The custom protocol / schema is never registered and `app.isDefaultProtocolClient` will always return false.

brunolemos:
Based on what I read I believe it should add the custom schema to `/usr/share/applications/defaults.list`, but it’s not doing that.

roaringUnicorn:
I ran into the same problem. I had to workaround using :
childProcess.exec(xdg-mime default myApp.desktop x-scheme-handler/myapp)

brunolemos:
@roaringUnicorn how to run this command from electron 5? I don’t see it in the docs

brunolemos:
@roaringUnicorn do I need to create a .desktop file or electron-builder handle that? I haven’t created one.

roaringUnicorn:
You do need to add "x-scheme-handler/MyApp;" as a known mimetype in your .desktop file. 

You don't need to specify the full path of your .desktop file, as `~/.local/share/applications/` is one of the default paths for them.

Desktop file are generated by electron-builder, and managed automaticaly by the OS.
You can customize it through your package.json:
```
"linux": {
  "desktop": {
    "MimeType": "x-scheme-handler/MyApp;" 
  }
}
```
The electron-builder doc might not be extremely explicit: https://www.electron.build/configuration/appimage


BTW, don't put a space in your appimage name ! There is an escaping probleme somewhere...
see: https://github.com/electron-userland/electron-builder/issues/3547

roaringUnicorn:
The bug is still present in electron 6.0.9

brunolemos:
@roaringUnicorn unfortunately that also didn't work for me.

brunolemos:
Still a problem.

brunolemos:
🙄

brunolemos:
.

jgredecka:
[...]

PR #5175, opened 2020-07-26
Pass URL argument %U in desktop file of snap and appimage builds

gcascio:
Currently the URL argument `%U` is not passed to the executable in the desktop files for snaps and appimages.
As a result the app does not receive the URL when custom protocol handlers are used. This PR passes the ad `%U` to the executable of both appimages and snaps.

Related PRs:
- #5001 (fix for snap)
- #4909 (fix for appimage)

Realted issues:
- #4035 
- #5024

### DIFF
--- a/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
+++ b/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
@@ -101,8 +101,9 @@ export class LinuxTargetHelper {
         exec += " "
         exec += executableArgs.join(" ")
       }
-      exec += " %U"
     }
+
+    exec += " %U"
 
     const desktopMeta: any = {
       Name: appInfo.productName,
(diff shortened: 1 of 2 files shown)


gcascio:
Regarding the failing CI tests, do I just have to adjust the sizes in the snapshot? I get that some file sizes might change but why would the size of `package.json` change with the changes I introduced?

```
"package.json": Object {
-               "size": 549,
+               "size": 817,
},
```

gcascio:
For the meantime I added a patch to my build process. I created a minimal [example](https://github.com/gcascio/electron-custom-url-workaround) for anyone who struggles with this problem.

sapkra:
@develar Would it be possible to review and merge it in the near future? There are several users which have this problem and it is a quite easy fix.

danhowitt:
I too would like to see this merged.

develar:
I merged PRs for snap and appimage. This PR is a little bit unsafe, as it is too generic place of code.

Issue #2074, opened 2017-09-13
Publishing with gitHubPublisher to an existing release always results in, the publisher thinking the release is more than 2 hours old


githubbob42:
* **Version**:: v19.28.4: 
* **Node**: v6.11.0
* **NPM**: 5.3.0

* **Target**:  windows 
(but I don't think it matters which target)
Not sure if it matters, but I'm running/building on macOS.

Publishing to an existing release always results in `Release with tag vX.X.X published at Tue Sep 12 2017 13:45:30 GMT-0500 (CDT), more than 2 hours ago`

Note that this was encountered after fixing Issue #2072.

Looking at the code:
```js
const publishedAt = release.published_at == null ? null : new Date(release.published_at)
if (publishedAt != null && (Date.now() - publishedAt.getMilliseconds()) > (2 * 3600 * 1000)) {
  // https://github.com/electron-userland/electron-builder/issues/1183#issuecomment-275867187
  warn(`Release with tag ${this.tag} published at ${publishedAt.toString()}, more than 2 hours ago`)
  return null
}
```
it appears that problem is caused because we are trying to get the milliseconds from the last published date/time (`publishedAt.getMilliseconds()`) but this results in the milliseconds for the given date/time (a 3 digit number), whereas `Date.now()` results in a the number of milliseconds since January 1, 1970, 00:00:00 (e.g. `1505315340692`).  Obviously, `1505315340692 - 253 > (2 * 3600 * 1000)` will always be true.

I think the correct logic should be to convert the `published_at` date to a number from the get go (`+new Date(release.published_at)`) and not use the `getMilliseconds` method at all (I think the person that added that thought is was going to give the number of milliseconds since January 1, 1970, 00:00:00 and not just the milliseconds of the given time).
```js
const publishedAt = release.published_at == null ? null : +new Date(release.published_at);
if (publishedAt != null && (Date.now() - publishedAt) > (2 * 3600 * 1000)) {
    // https://github.com/electron-userland/electron-builder/issues/1183#issuecomment-275867187
    (0, (_builderUtil || _load_builderUtil()).warn)(`Release with tag ${_this.tag} published at ${new Date(publishedAt).toString()}, more than 2 hours ago`);
    return null;
}
```

imolorhe:
[...]

PR #5179, opened 2020-07-28
make it possible to ignore release time when publish to github

wsmlby:
https://github.com/electron-userland/electron-builder/issues/2074#issuecomment-631292925

set EP_GH_IGNORE_TIME to true will allow us to ignore the release time on github to publish to latest release


stoefln:
this is not documented anywhere

Issue #5211, opened 2020-08-08
When packaging for Linux Snap - .desktop file for snap doesn't pass command line arguments. 


samnickolay:
Using versions:
"electron": "8.2.5",
"electron-builder": "^22.8.0",

In the generated .desktop file for a snap app, the Exec line doesn't have the %U that passes along arguments to the app.
For example, an app called Foo has an exec line like this:  `Exec=Foo`, when it should be like this `Exec=Foo %U`.

I believe the issue lies in:
app-builder-lib -> src ->targets ->LinuxTargetHelper.ts and snap.ts

In snap.ts, it generates the .desktop file by calling:
```
    await this.helper.writeDesktopEntry(this.options, packager.executableName, desktopFile, {
      // tslint:disable:no-invalid-template-strings
      Icon: "${SNAP}/meta/gui/icon.png"
    })
```
since it passes packager.executableName to this.helper.writeDesktopEntry (which is in LinuxTargetHelper.ts ),
then in computeDesktopEntry in LinuxTargetHelper.ts since exec is not null, then it doesn't add the executableArgs and " %U" to the exec string. 

To solve this, just add this code to snap.ts before calling writeDesktopEntry.

```
let exec = packager.executableName;
const executableArgs = this.options.executableArgs
if (executableArgs) {
        exec += " "
        exec += executableArgs.join(" ")
}
exec += " %U"    
await this.helper.writeDesktopEntry(this.options, exec, desktopFile, {
      // tslint:disable:no-invalid-template-strings
      Icon: "${SNAP}/meta/gui/icon.png"
})

```

samnickolay:
https://github.com/electron-userland/electron-builder/pull/5212

PR #5212, opened 2020-08-08
Add executableArgs and " %U" to executableName before calling writeDesktopEntry

samnickolay:
Add executableArgs and " %U" to packager.executableName before passing it writeDesktopEntry, since writeDesktopEntry() only adds executableArgs and " %U" if no exec string is passed. This fix works with the existing functions in LinuxTargetHelper.ts so it won't affect any other Linux targets.


samnickolay:
https://github.com/electron-userland/electron-builder/issues/5211

samnickolay:
Same issue as:
https://github.com/electron-userland/electron-builder/pull/5001/

https://github.com/electron-userland/electron-builder/pull/5175

PR #5235, opened 2020-08-19
Feat: Add support to pass along a value for the ‘strict verify’ flag …

### LINKED ISSUE #1480: Codesign: No such file or directory
* **Version**: 11.7.0

* **Target**: OSX

(This is really a comment so that others can find the solution if they are looking. Please close.) 

I was getting an error during the build process
```
codesign --verify --deep--verbose=2  --strict  /path/to/app/Test.app
...
/path/to/app/Test.app: No such file or directory
```
After a lot of debugging, it turns out this is a really terrible and cryptic codesign error message. 
If I change it to `--strict=symlinks`
I get better info: `/path/to/app/Test.app: invalid destination for symbolic link in bundle`
Sure enough, there was a symlink in my bundle pointing outside the bundle.

bedney:
…to the electron-osx-sign library

This change allows the author to provide a value for ‘strictVerify’ in the ’mac’ configuration. This value is then passed along to the electron-osx-sign library.

This is not a breaking change and it closes #1480

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -2284,6 +2284,11 @@
             "string"
           ]
         },
+        "strictVerify": {
+          "default": true,
+          "description": "Whether to require electron-osx-sign to strictly verify the package to be signed.",
+          "type": "boolean"
+        },
         "target": {
           "anyOf": [
             {
@@ -2830,6 +2835,11 @@
             "null",
             "string"
           ]
+        },
+        "strictVerify": {
+          "default": true,
+          "description": "Whether to require electron-osx-sign to strictly verify the package to be signed.",
+          "type": "boolean"
         },
         "target": {
           "anyOf": [
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -192,6 +192,8 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       // https://github.com/electron-userland/electron-osx-sign/issues/196
       // will fail on 10.14.5+ because a signed but unnotarized app is also rejected.
       "gatekeeper-assess": options.gatekeeperAssess === true,
+      // https://github.com/electron-userland/electron-builder/issues/1480
+      "strict-verify": options.strictVerify !== false,
       hardenedRuntime: isMas ? masOptions && masOptions.hardenedRuntime === true : options.hardenedRuntime !== false,
     }
 
--- a/packages/app-builder-lib/src/options/macOptions.ts
+++ b/packages/app-builder-lib/src/options/macOptions.ts
@@ -154,6 +154,12 @@ export interface MacConfiguration extends PlatformSpecificBuildOptions {
    * @default false
    */
   readonly gatekeeperAssess?: boolean
+
+  /**
+   * Whether to let electron-osx-sign verify the contents or not.
+   * @default true
+   */
+  readonly strictVerify?: boolean
 }
 
 export interface DmgOptions extends TargetSpecificOptions {


bedney:
@develar - I created this PR as best I could given that I've had a devil of a time actually getting my fork of `electron-builder` to install into a project to test it. Given that `electron-builder` is a monorepo, there are hoops to jump through to get a git fork of the repository to install and my knowledge of `yarn` just isn't that good - sorry. I hope this is a good starting point.

As an aside, upgrading the CONTRIBUTING document with specific instructions on how to install `electron-builder` from a git fork would go a long way to helping us help you develop this product.

Thank you for your hard work on this tool!

bedney:
So the check above is failing because I need to update the `macArchiveTest.js.snap` file that contains a snapshot of the options as they used to be.

I'm not familiar with `.snap` files or how to update them.

@aguynamedben, I see that you've done this before (updated the Mac options). Any idea here? I'm thinking it's just a command that I need to execute at the terminal to update the 'snap' files, but I'm a bit lost here. Any help you can give would be super appreciated - thank you!

aguynamedben:
@bedney not sure in this case, but we have this in our package.json's "builder" section, so maybe this helps you figure out how to tell macOS things from electron-builder world?

```json
    "mac": {
      "hardenedRuntime": true,
      "gatekeeperAssess": false,
      "entitlements": "configs/entitlements.mac.plist",
      "entitlementsInherit": "configs/entitlements.mac.plist",
      "helperBundleId": "io.hypertools.Command-E.helperGeneric",
      "helperRendererBundleId": "io.hypertools.Command-E.helper",
      "extendInfo": {
        "NSUserNotificationAlertStyle": "alert"
      }
    },
```

config/entitlements.mac.plist
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
    <true/>
    <key>com.apple.security.cs.disable-library-validation</key>
    <true/>
  </dict>
</plist>
```

bedney:
@aguynamedben -

Thanks so much for your reply! I do appreciate it.

I've figured out a few more things. As it turns out the `.snap` files thing was a canard. So I'm down to figuring out the primary problem here:

How to package my own fork of `electron-builder` to make it available for my projects to install from **git**.

Given that `electron-builder` is a constructed as a monorepo, and I've not dealt with them before, it is not clear to me how to do this. I've spent quite a few hours searching around to see if anyone has answers on how to do this. If you have any insights on that or articles you've read that show how that can be done, I would really appreciate it.

Thank you again for your reply.

aguynamedben:
@bedney I'm not sure on that, but I've wondered this before. I think what maintainers do is checkout code from the branch, run build commands (i.e. convert TypeScript to JavaScript, etc.) the upload THAT to npm. You can see some of that going on in [package.json](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/package.json) for the electron-builder package within the monorepo.

The git repo itself doesn't have the build/compiled version of the code (i.e. what you locally see when you run `ls node_modules/electron-builder/out/index.js`).

I'm not sure what the best practice is. I suppose you can modify electron-builder, then build it, zip it into a .tar.gz and host it somewhere you know about, and probably find a way to refer to it from your package.json... but that's all kind of a pain.

bedney:
@aguynamedben - Thank you for the further information. I figured out a hack to allow me to limp along for now.

As it turns out, in examining the `strict-verify` flag for the `electron-osx-sign` package, it isn't necessarily a Boolean, so I'm going to close this PR, update my patch and reissue it.

Thanks again for your help.

Issue #5163, opened 2020-07-22
How to set "strict-verify = false" via configuration


TimPietrusky:
* **Version**:  22.7.0
* **Electron Version**: 7.1.3
* **Electron Type (current, beta, nightly)**: current

* **Electron updater**: 4.3.1

* **Target**: Mac

I want to deactivate the `--verify` parameter when signing code for MacOS as this is causing problems for me. There is a parameter for `electron-osx-sign` called `strict-verify`, but I can't figure out how to pass this when using electron-builder.

2xAA:
This was also asked for here back in 2018.
https://github.com/electron-userland/electron-builder/issues/1480#issuecomment-423188723

DominikLevitsky:
It was asked, but never resolved. We need an option to set this parameter. Currently OSX builds are crashing because of this. @develar Any possible solutions?

TimPietrusky:
This ticket makes it impossible to use https://github.com/jensstigaard/grandiose with electron-builder, as some files in grandiose are sym-links.

TimPietrusky:
We will make sure to get this tested!

TimPietrusky:
This was tested and it is working. Thank you very much.

zjy369:
> This was tested and it is working since electron-builder 22.9.1:
> 
> ```js
> builderOptions: {
>   mac: {
>     strictVerify: false
>   }
> }
> ```
> 
> Thank you very much.

It's a fake

PR #5261, opened 2020-08-26
Added support for overriding ‘strict-verify’ electron-osx-sign property.

bedney:
This provides support for overriding the electron-osx-sign `strict-verify` property and supersedes a prior PR that only supported boolean values here.

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -2284,6 +2284,23 @@
             "string"
           ]
         },
+        "strictVerify": {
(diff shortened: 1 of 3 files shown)


bedney:
Note that this would close out https://github.com/electron-userland/electron-builder/issues/1480

TimPietrusky:
This would also close #5163

2xAA:
@develar Thanks for merging. Do you have a timeline for a new release?

develar:
I will release 22.9.1 today.

Issue #3790, opened 2019-03-29
Option to ignore files for signing


torjusb:
* **Version**:  20.39.0

* **Target**: OS X

We have a .app bundled in the app. I'm using the `asarUnpack` option for this. But I'm running into problems with the app signing, as it seems to be trying to sign files/folders which are not signable.

This is the error I'm getting:

```
Error: Command failed: codesign --sign *** --force /Users/***/Contents/Resources/app.asar.unpacked/resources/osx/***.app/Contents/Frameworks/Breakpad.framework/Breakpad: bundle format is ambiguous (could be app or framework)
```

Is there any way to ignore files which should not be signed? Or an option to specify exactly which files _should_ be signed? I've tried the `binaries` option, but it still tries to sign the other files.

xiaoping6688:
![image](https://user-images.githubusercontent.com/16394201/55271521-e3d83b00-52e8-11e9-9384-c2febac88475.png)

torjusb:
It's unclear to me how changing those settings helps. Even with `forceCodeSigning: false` it still exists when the `codesigns` fails with `bundle format is ambiguous`. I also still want to package my app files in an .asar file, but leave .app we package in the `app.asar.unpacked` folder.

hrishabhkumar:
There should be an option to specify list of files/folder or module which is not required to sign.

hrishabhkumar:
Is there any update on this please?

benoitlahoz:
Any update on this?

PR #5262, opened 2020-08-27
Added support for supplying a value for the ‘ignore’ electron-osx-sign property.

bedney:
This PR allows users of `electron-builder` to supply additional files to the `electron-osx-sign` utility to ignore during the signing process.

NOTE: This PR follows the `electron-osx-sign` utility's signature for this property. That is, it is either a single RegExp as a JSON String or an Array of JSON Strings of RegExps. It is **NOT** a 'glob' pattern, like many of the other config properties in `electron-builder`. That was more sophistication than I needed.

In any case, you use this patch in your config like this:

```
{
    "appId": "...",
    "mac": {
        ...
        "ignore": ".*\\.app$",
        ...
    }
}
```

**Note the escaping of the backslash there, as you might expect**.

In this case, we are filtering for files/directories ending with `.app` (I had one of those and the `electron-osx-sign` wants to think that they're embedded Mac OS X applications, but this wasn't).

Here's the Array version:

```
{
    "appId": "...",
    "mac": {
        ...
        "ignore": [".*\\.app$", ..., ...]
        ...
    }
}
```

Note that this closes https://github.com/electron-userland/electron-builder/issues/3790.

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -2183,6 +2183,21 @@
             "string"
           ]
         },
+        "ignore": {
+          "anyOf": [
+            {
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
+            },
+            {
+              "type": "string"
+            }
+          ],
+          "default": null,
+          "description": "Filter to tell electron-osx-sign to filter out certain files."
+        },
         "minimumSystemVersion": {
           "description": "The minimum version of macOS required for the app to run. Corresponds to `LSMinimumSystemVersion`.",
           "type": [
@@ -2729,6 +2744,21 @@
             "null",
             "string"
           ]
+        },
+        "ignore": {
+          "anyOf": [
+            {
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
+            },
+            {
+              "type": "string"
+            }
+          ],
+          "default": null,
+          "description": "Filter to tell electron-osx-sign to filter out certain files."
         },
         "minimumSystemVersion": {
           "description": "The minimum version of macOS required for the app to run. Corresponds to `LSMinimumSystemVersion`.",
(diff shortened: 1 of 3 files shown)


bedney:
@develar -

Can you merge this PR? If you feel like more work needs to be done on it, I can help. Otherwise, this is blocking us - we need to be able to exclude files/directories from the signing process.

Thanks!

develar:
Thanks for PR, I renamed option from `ignore` to `signIgnore`. Release will be today.

ym13349140:
> This PR allows users of `electron-builder` to supply additional files to the `electron-osx-sign` utility to ignore during the signing process.
> 
> NOTE: This PR follows the `electron-osx-sign` utility's signature for this property. That is, it is either a single RegExp as a JSON String or an Array of JSON Strings of RegExps. It is **NOT** a 'glob' pattern, like many of the other config properties in `electron-builder`. That was more sophistication than I needed.
> 
> In any case, you use this patch in your config like this:
> 
> ```
> {
>     "appId": "...",
>     "mac": {
>         ...
>         "ignore": ".*\\.app$",
>         ...
>     }
> }
> ```
> 
> **Note the escaping of the backslash there, as you might expect**.
> 
> In this case, we are filtering for files/directories ending with `.app` (I had one of those and the `electron-osx-sign` wants to think that they're embedded Mac OS X applications, but this wasn't).
> 
> Here's the Array version:
> 
> ```
> {
>     "appId": "...",
>     "mac": {
>         ...
>         "ignore": [".*\\.app$", ..., ...]
>         ...
>     }
> }
> ```
> 
> Note that this closes #3790.

@bedney Hello, I recently use signIgnore to ignore some .dll files in my projects, but it dosen't work.
these dll files are in ./dist_electron/mac/application/Contents directory, I tried many times of different value , including ./dist_electron/mac/application/Contents/xxx.dll , .dll$ and array value '[.dll$]', but it never make sense, is there any worng?
My config options and my dist_electron directory are as follows:
```
mac: {
    icon: './dist_electron/icons/icon.icns',
    signIgnore: '.*\\.dll$'
}
```
dist_electron


 I'd be really appreciateful if you can reply.

ferenc-a:
Apparently `signIgnore` does not apply to the `.app/Contents/` directory: https://github.com/electron-userland/electron-builder/issues/6085#issuecomment-886309893, as a limitation of `electron-osx-sign`.
I'm still looking for a solution to this, but haven't seen an actual one yet. @ym13349140 what did you end up doing?

Issue #4953, opened 2020-05-12
[auto-updater]: autoUpdater.download() downloads _and_ installs update


UberMouse:
22.6.0

* **Electron Version**: 8.2.0
* **Electron Type (current, beta, nightly)**: current

4.3.1

* macOS 

I've integrated with the auto updater events in electron-builder to build my own auto-update UI. I have separate steps for downloading and installing the update. Which I am performing by calling `autoUpdater.downloadUpdate()` and `autoUpdate.quitAndInstall()` respectively.

However, it appears that as soon as the update is done downloading after calling `downloadUpdate` the update is installed. Is it possible to change this behaviour on macOS?

I see that there is the `autoUpdater.autoInstallOnAppQuit` setting to turn this off, but it doesn't support macOS.

Are there any alternatives here? How come I can't disable this functionality on macOS?

UberMouse:
Related to #4308

gre:
sadly a bot automatically closed the PR #4308 , we kept maintaining this PR to be mergeable for months and it was not merged yet and not even commented by the owners. I hope we have the opportunity to look at it and get it merged because it seems pretty relevant and important regarding implementing a [**secure**](https://blog.doyensec.com/2020/02/24/electron-updater-update-signature-bypass.html) update.

cc @MortalKastor @develar @atali

UberMouse:
Seconded. I manually applied your PR with patch-package so at least one other company is using it! I can confirm it's been working without any problems for us.

UberMouse:
This is still relevant. It's blocked on someone merging the PR in #4308

cc @develar

gre:
Is there any way we can help on this topic? It is an important security-related topic, i'm not sure why is it ignored but we're here to help.

gre:
@MortalKastor can you recreate the PR? I hope it's going to be reviewed this time..

MortalKastor:
@gre OK

_Third time's a charm_, as the saying goes 🤞

UberMouse:
This is still relevant. It's blocked on someone merging the PR in #5271

gre:
YES

bjbk:
Any resolution to this issue?  There may be reasons for macOS users to delay update installs.  What is the issue with SquirrelMac that prevents `autoInstallOnAppQuit = false` from properly being implemented?

UberMouse:
[...]

PR #4308, opened 2019-10-10
feat(electron-updater): follow `autoInstallOnAppQuit = false` on macOS

MortalKastor:
Make `electron-updater` on macOS behave like other OSes regarding the `autoInstallOnAppQuit` flag by cheating a bit around Squirrel.Mac limitation of not being able to *not* apply an update after it has downloaded it.

I tried to make my changes as minimal as possible, and took advantage of the fact that `electron-updater` already downloads the update by itself first and only expose said update to Squirrel in a second time. The trick here is to only trigger Squirrel when we want to apply an update rather than after a successful download.

This PR succeed and replace the previous #3745 which was targeting an out of date version of `electron-builder`. The logic of both is more or less the same, but the implementation differs because of the change of `quitAndInstall()` method from `async` to sync.

### Context
This feature is required for our app ([Ledger Live](https://github.com/LedgerHQ/ledger-live-desktop)) to be able to do an additional cryptographic check on downloaded files and conditionally abort the update on a failed verification.

PR #5271, opened 2020-08-31
feat(electron-updater): follow `autoInstallOnAppQuit = false` on macOS

MortalKastor:
Make `electron-updater` on macOS behave like other OSes regarding the `autoInstallOnAppQuit` flag by cheating a bit around Squirrel.Mac limitation of not being able to *not* apply an update after it has downloaded it.

I tried to make my changes as minimal as possible, and took advantage of the fact that `electron-updater` already downloads the update by itself first and only expose said update to Squirrel in a second time. The trick here is to only trigger Squirrel when we want to apply an update rather than after a successful download.

This PR is basically a rebase of #4308 which never got reviewed and had been automatically closed by the stale bot.

### Context
This feature is required for our app ([Ledger Live](https://github.com/LedgerHQ/ledger-live-desktop)) to be able to do an additional cryptographic check on downloaded files and conditionally abort the update on a failed verification.

At least one other company is relying on this patch, cf #4953

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -26,8 +26,6 @@ export abstract class AppUpdater extends EventEmitter {
 
   /**
    * Whether to automatically install a downloaded update on app quit (if `quitAndInstall` was not called before).
-   *
-   * Applicable only on Windows and Linux.
    */
   autoInstallOnAppQuit: boolean = true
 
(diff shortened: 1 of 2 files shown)


gre:
YES YES YES

StaZhu:
this fix will be very helpful，please merge it

mmaietta (approved):
Nice work 👍

MortalKastor:
> Please update the PR to fix the merge conflict

Thanks for the review 🙏
I just fixed the conflict, rebased and updated the doc to reflect the change in behavior.

mmaietta (approved):

bjbk:
I really hope this gets pulled soon. 😄

castroCrea:
Hello, Thank for the PR @MortalKastor 
But now my auto update doesn't work anymore, it download it but doesn't install and doesn't restart. 
There is also no doc to be back on old behavior?
How should we do be back on the old behavior ?
Thanks

Issue #2227, opened 2017-10-25
Is there an option to force the installer to be executed only as administrator?


fttx:
* **Version**: 19.37.1

* **Target**: Windows

Hi, 

i'm trying to chain an installation of an msi package, in my app installer with this script:

```
!macro customInstall
${If} ${RunningX64}
  ExecWait '"msiexec" /a "${BUILD_RESOURCES_DIR}\Bonjour64.msi" /passive'   
${Else}
  ExecWait '"msiexec" /a "${BUILD_RESOURCES_DIR}\Bonjour.msi" /passive'   
${EndIf}  
!macroend
```
which requires administrative privileges, i guess that msiexec inherits the privileges from the main installer, so how do i force it to be executed as administrator?

develar:
Set `perMachine` to `true` (see https://www.electron.build/configuration/nsis).

fttx:
I've tried it, but even though it prompts the admin login popup, the msi package still fails to install.
Any idea why?

develar:
`BUILD_RESOURCES_DIR` macro expanded to path to build resources, but file is not included into installer. 

You need to include your file using `File /oname=$PLUGINSDIR\Bonjour64.msi "${BUILD_RESOURCES_DIR}\Bonjour64.msi"`

and then use `ExecWait '"msiexec" /i "$PLUGINSDIR\Bonjour64.msi" /passive' `

See usages of `PLUGINSDIR` in our code.

fttx:
It worked flawlessly, thank you so much!

PR #5308, opened 2020-09-24
docs(nsis): Add example for including installer resources

gpoole:
Add specific documentation about the the steps described in #2227 ([here](https://github.com/electron-userland/electron-builder/issues/2227#issuecomment-339662811)) for including nested installers and other install-time resources.


develar:
Sorry for delay, thanks a lot.

PR #5322, opened 2020-10-01
Codesign all binary-like files

adam-lynch:
This is an exact copy of https://github.com/electron/electron-osx-sign/pull/169


lutzroeder:
@adam-lynch, could this change have broken macOS `.dmg` builds? After updating to 22.10.3 and signing,  the `.dmg` target will fail running `hdiutil create` with an `Operation not permitted` error. The `.app` binary now includes more files with xattr attachments (the `.icns` and `app.asar` in `Contents/Resources` for example) and that seems to be causing `hdiutil create` to fail. Running the same build without this change succeeds.

See also #5465 and #5431 which might be related.

PR #5363, opened 2020-10-23
docs: clarify use of productName for the install directory

s4m0r4m4:
Add further documentation clarifying how/when the productName property is used for the install directory.

Unsure if this is only for NSIS or for all configurations.


s4m0r4m4:
Do I need to do something with CircleCI for this? It's such a minor change (to documentation, nonetheless) that I figured that would not be necessary.

Issue #5014, opened 2020-05-28
Generate Linux icons from png provided


mahnunchik:
* **Version**: 22.7.0

* **Electron Version**: any
* **Electron Type (current, beta, nightly)**: any

* **Target**: linux (any)

* Linux target allow us to generate Linux icons from `.icns` image container (source: https://github.com/electron-userland/electron-builder/blob/7b335dbd7e5ca754598ca93318be1404fc518485/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts#L61)
* macOS target allow us to generate `.icns` from single `.png` provided.

It would be really awesome to generate all necessary Linux icon sizes from single `.png` source image.

mahnunchik:
YEP!! RELEVENT!

mahnunchik:
![p70psfoihek51](https://user-images.githubusercontent.com/780935/94330919-ef352100-ffc8-11ea-8bb0-84579086826c.jpg)

mahnunchik:
qwerty

Issue #4617, opened 2020-01-27
AppImage icon not working


Coding-Kiwi:
* **Version**: 
21.2.0

* **Target**: 
linux, Appimage

As stated in the [docs](https://www.electron.build/configuration/appimage):

> Since electron-builder 21 desktop integration is not a part of produced AppImage file. AppImageLauncher is the recommended way to integrate AppImages.

I'm not a big fan of this decision but it for sure has reasons, idk.
I tried out linux mojave (arch linux) and installed the appimagelauncher to integrate the AppImage, Sadly there is no icon in the start menu.

 The AppImage [docs ](https://docs.appimage.org/reference/appdir.html) say the icon file has to be in `<root>/usr/share/icons/<theme>/<resolution>/apps/myapp.<ext>`, I extracted the AppImage generated by electron-builder and the icon file in there symlinks to `usr/share/icons/hicolor/scalable/myapp.svg`, as you see `/apps` is missing, this might be the cause?

Coding-Kiwi:
Yes it is still relevant.
Whats blocking it? well nobody seems to care.

digimezzo:
@Coding-Kiwi I have no idea how to resolve this on AppImage level, but I just applied a workaround to my project, which I found by looking at other open source projects.

1. Add a "static" folder to your Electron project

![image](https://user-images.githubusercontent.com/11945339/80907558-877bf200-8d18-11ea-85b4-70e5ab2d855e.png)

2. Add some icons to them in png format. I found that a 64x64 png icon looked best in the docks in Linux. Larger sizes are jagged.

3. In you main.ts, add this:

```
import * as os from 'os';

const globalAny: any = global; // If you use TypeScript, Global does not allow setting custom properties.

// Static folder is not detected correctly in production
if (process.env.NODE_ENV !== 'development') {
  globalAny.__static = require('path').join(__dirname, '/static').replace(/\\/g, '\\\\');
}

Then, when creating the BrowserWindow, you can reference the icons:

mainWindow = new BrowserWindow({
      'x': 50,
      'y': 50,
      'width': 500,
      'height': 500,
      backgroundColor: '#fff',
      icon: path.join(globalAny.__static, os.platform() === 'win32' ? 'icons/icon.ico' : 'icons/64x64.png'),
      show: false
    });
```

puresick:
I am having the same issue with AppImages not showing any Icon, except with the workaround described in https://github.com/electron-userland/electron-builder/issues/2269

[...]

Issue #4186, opened 2019-08-27
Linux Ubuntu AppImage target no icon


osenvosem:
* **Version**: 21.1.5

* **Target**: Linux AppImage

Icons are located in the default `build` directory. This is it's structure:

```
build
├── icon.icns
├── icon.ico
├── icon.png
└── icons
    ├── 128x128.png
    ├── 16x16.png
    ├── 256x256.png
    ├── 32x32.png
    ├── 512x512.png
    └── 64x64.png
```

I tried to specify path to icon in the configuration. Result is the same:

Ubuntu 19.04

![Screenshot from 2019-08-27 17-38-51](https://user-images.githubusercontent.com/6762673/63781386-5da13e80-c8f2-11e9-8c64-685ab5930016.png)

Any workaround?

BloodShura:
Same problem here.

BloodShura:
Yes, it **IS** relevant.

BloodShura:
Once again: it **IS** relevant.

BloodShura:
YES YES YES YES, ONCE AGAIN, YES

mahnunchik:
+1

ttomaszewski:
+1

mahnunchik:
![20200808_191450](https://user-images.githubusercontent.com/780935/91752238-56c4a000-ebc6-11ea-9a91-1e293d03352e.jpg)

BloodShura:
Hello darkness, my old friend. We've been abandoned once again.

Issue #2577, opened 2018-02-13
Linux icons issue


romangetman:
* **Version**: 20.0.4
* **Target**: deb, AppImage

No matter what I try I cannot set an icon on Linux. 
Building from a macOS Sierra.
I have dir's customized:
```
"directories": {
      "buildResources": "./build_res",
      "output": "./build",
      "app": "./dist"
    }
```
I've tried setting
```
"linux": {
      "icon": "..."
}
```
to all imaginable values, copying all icon folders to different locations in the root and build_res and I still get a question mark icon in Ubuntu 16.04.

I have an .icns file in `./build_res/icon.icns`. 

Is there a workaround? Or a fix?

develar:
Do you run app using file (AppImage) or using menu? Set env `DEBUG=electron-builder` and you will see stage dir for AppImage (`_appImage`), where some icon will be in the root — is it your icon?

romangetman:
Running the AppImage by clicking on it directly.
Yes there's a symlink that is linked to `__appImage-ia32/usr/share/icons/hicolor/512x512/apps/myicon.png`.

What about the `deb`?

develar:
> Running the AppImage by clicking on it directly.

Please see https://github.com/electron-userland/electron-builder/issues/748#issuecomment-342062462

> What about the deb?

Need to check.

gschier:
I've got the same issue. Defaults to Electron icon for both AppImage and deb builds. I also have custom dirs specified.

develar:
@gschier I am sure that you issue is different.  Please open a new issue and please set env `DEBUG=electron-builder` and attach log of the terminal output.

@gschier Or, if `__appImage` dir contains correct icon, your issue is the same :)

gschier:
Ya, looks like there is no icon at all in the root. Just an `app` and `usr` dir. Going to revert to a previous working version for now, but will create an issue next time I try and upgrade if I still have problems. 

Thanks! This project is the best 😄

develar:
@gschier I will try to find time to checkout your project and investigate what's wrong.

develar:
@gschier Are you sure that you use 20.0.4?

gschier:
From the testing I just did, the icon stopped working after upgrading from `19.47.1` to `19.53.0`. I also quickly tried 20.0.4 but ran into some different errors (vague message about JSON schema).

```
Error: no schema with key or ref "http://json-schema.org/draft-07/schema#"
```

[...]

PR #5385, opened 2020-11-01
fix(AppImage): Linux icon is not set if path is not explicitly defined in config

quanglam2807:
If icon path is not explicitly defined in config, `sources` would be an empty array. This leads `resolveIcon` not to send `--path` argument  to `app-builder-bin`. Then `app-builder-bin` fails to create the icon set properly.

I suppose we can set a default `--path` value in `resolveIcon` or `app-builder-bin`, but for safety, I chose to only modify the `LinuxTargetHelper`.

Fix https://github.com/electron-userland/electron-builder/issues/4617
Fix https://github.com/electron-userland/electron-builder/issues/4186
Fix https://github.com/electron-userland/electron-builder/issues/2577
Fix https://github.com/electron-userland/electron-builder/issues/5014

### DIFF
--- a/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
+++ b/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
@@ -61,6 +61,11 @@ export class LinuxTargetHelper {
       sources.push(icnsPath)
     }
 
+    // if no explicit sources are defined, default to buildResources directory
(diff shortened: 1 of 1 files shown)


quanglam2807:
Temporary workaround: https://github.com/electron-userland/electron-builder/issues/2577#issuecomment-659879082

Issue #4962, opened 2020-05-13
Add possibility for custom productFilename


raimonds-vilums:
* **Target**: all

As described here #2813, it would be nice to have other filename for executable rather than the product name via defining productFilename in configuration, e.g,

product name: My Super App
product filename: app.exe, not "My Super App.exe"

Issue #2813, opened 2018-04-19
Windows: Create exe with one name & use another for the add/remove programs list


warrenbuckley:
* **Version**: 
`"electron-builder": "^20.9.0"`

* **Target**: 
Windows: NSI

Hello,
I am after some help in creating my Electron application for windows using NSI which creates the installer for me.

I wish to have Electron to produce the exe as one name and to have a different name used for the add/remove programs uninstall list.

Here is a copy of my build section from my package.json
```
"build": {
    "appId": "lucene.index.explorer",
    "productName": "Lucene Index Explorer",
    "extraResources": {
      "from": "../LIE.Server/bin/dist/",
      "to": "LIE.Server/bin/dist/",
      "filter": [
        "**/*"
      ]
    },
    "win": {
      "target": [
        "nsis"
      ],
      "icon": "build/icon.png"
    },
    "nsis": {
      "deleteAppDataOnUninstall": true,
      "createDesktopShortcut": true,
      "createStartMenuShortcut": true,
      "shortcutName": "Lucene Index Explorer",
      "artifactName": "LIE.Setup.${version}.${ext}"
    }
  }
```

So if I set `productName` to `LIE` which will generate a `LIE.exe` and the Desktop and Start Menu shortcut shows up as `Lucene Index Explorer` by using `shortcutName` however in the Add or Remove Programs list it displays as LIE as opposed to Lucene Index Explorer.

Is there any config setting I have missed or can use, so I can achieve this please?

This will just make life easier for people viewing the running processes with Task Manager on Windows to see the running LIE.Server process and the main LIE application along side each in the list of processes IMO.

Look forward to hearing back from you.

Thanks,
Warren :)

develar:
I cannot reproduce. And `shortcutName` not required since it is equal to product name.

Please remove old app, restart windows and install again. Still no product name? Attach your exe or send me by email to test and attach here `dist/electron-builder-effective-config.yaml`.

warrenbuckley:
**dist/electron-builder-effective-config.yaml**
```
directories:
  output: dist
  buildResources: build
appId: lucene.index.explorer
productName: Lucene Index Explorer
extraResources:
  from: ../LIE.Server/bin/dist/
  to: LIE.Server/bin/dist/
  filter:
    - '**/*'
win:
  target:
    - nsis
  icon: build/icon.png
nsis:
  deleteAppDataOnUninstall: true
  createDesktopShortcut: true
  createStartMenuShortcut: true
[...]

PR #5409, opened 2020-11-10
feat: add support for custom executable file name

kagawarei:
Add support for custom executable file name in configuration.

FYI: #4962 #2813

In our project, the product name is in Japanese, but we want the executable file name to be in English.


packages/app-builder-lib/src/configuration.ts
```diff
@@ -32,6 +32,11 @@ export interface Configuration extends PlatformSpecificBuildOptions {
    */
   readonly productName?: string | null
 
+  /**
+   * The executable file name. Defaults to `productName`.
+   */
+  readonly productFilename?: string | null
```
mmaietta, line 38: This seems to be duplicating what already exists for "Any Linux Target" -> `executableName`
https://www.electron.build/configuration/linux

https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/options/linuxOptions.ts#L24-L28
I think the proper solution would be to move `executableName` out of `LinuxConfiguration` and into `PlatformSpecificBuildOptions`. That would keep it backward compatible for Linux options, while exposing it to each platform to uniquely define.

mmaietta (changes requested):

packages/app-builder-lib/src/configuration.ts
kagawarei: I'm sorry, I don't understand why I'm using executableName on Linux.
(Is this not good with AppInfo.productFilename?)

> I think the proper solution would be to move executableName out of LinuxConfiguration and into PlatformSpecificBuildOptions.

LinuxPackager.executableName will have the same value as AppInfo.productFilename.

Are you ok with this?

mmaietta: Your `productFilename` is a nomenclature misnomer as your documentation for it states it is the "executable file name", which is the exact same as `executableName` in `LinuxConfiguration`.

`LinuxConfiguration` inherits from `PlatformSpecificBuildOptions` which is where you're adding this new property to. Please do not do this, it duplicates existing functionality and is confusing as to what value to choose for Linux targets (which takes precedence for instance for Linux builds?)

You should move `executableName` to `PlatformSpecificBuildOptions` (plus update the doc comments on it) and it'll serve your purposes just fine. 
Another way to state this is: 
- Rename this variable to `executableName`
- Remove `executableName` from https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/options/linuxOptions.ts#L24-L28

kagawarei: Thanks for the detailed explanation. I'll give it a try.

https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts#L100

This will affect the value of the exe when it is called FpmTarget.build -> LinuxTargetHelper.writeDesktopEntry -> computeDesktopEntry.

kagawarei: > The executable name. Defaults to `productName`.

https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/linuxPackager.ts#L21

I'll going to leave it as it is, but I feel like the default for `LinuxPackager.executableName` is not `productName`. (I think it's only referring to metadata.name)

kagawarei: https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/util/macroExpander.ts#L20

productFilename affects not only the executable name, but also the `File Macro`.

The File Macro will keep its original value.

mmaietta:
Very nice work! Just need the mac/win tests since this is new functionality and I think we're set? 🙂

kagawarei:
Thanks for the feedback.

Sorry for the delay in adding the test code.

I'm not familiar with the code for this project and it is taking a little time. (In addition, Mac do not have a local test environment)

packages/app-builder-lib/src/configuration.ts
kagawarei: > `shortcutName` String - The name that will be used for all shortcuts. Defaults to the application name.

https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/options/CommonWindowsInstallerConfiguration.ts#L76

shortcutName should also use productName. (not filename)

The rest of the code looks fine.

kagawarei:
Added Win test code. Is this good?

I'm trying to add it to dmgTest, but I'm having trouble checking the snapshot.
I think you should check the env. What should I do?

mmaietta:
Here's the patch to apply for fixing the mac test 👍 
```
diff --git a/test/snapshots/mac/macArchiveTest.js.snap b/test/snapshots/mac/macArchiveTest.js.snap
index 48d9c95c..2244a81c 100644
--- a/test/snapshots/mac/macArchiveTest.js.snap
+++ b/test/snapshots/mac/macArchiveTest.js.snap
@@ -106,7 +106,7 @@ Object {
 exports[`invalid target 1`] = `
 "Invalid configuration object. electron-builder <appVersion> has been initialized using a configuration object that does not match the API schema.
  - configuration.mac should be one of these:
-   object { appId?, artifactName?, asar?, asarUnpack?, binaries?, bundleShortVersion?, bundleVersion?, category?, compression?, cscInstallerKeyPassword?, cscInstallerLink?, cscKeyPassword?, cscLink?, darkModeSupport?, detectUpdateChannel?, electronLanguages?, electronUpdaterCompatibility?, entitlements?, entitlementsInherit?, entitlementsLoginHelper?, extendInfo?, extraDistFiles?, extraFiles?, extraResources?, fileAssociations?, files?, forceCodeSigning?, gatekeeperAssess?, generateUpdatesFilesForAllChannels?, hardenedRuntime?, helperBundleId?, helperEHBundleId?, helperGPUBundleId?, helperNPBundleId?, helperPluginBundleId?, helperRendererBundleId?, icon?, identity?, minimumSystemVersion?, protocols?, provisioningProfile?, publish?, releaseInfo?, requirements?, signIgnore?, strictVerify?, target?, type? } | null
+   object { appId?, artifactName?, executableName?, asar?, asarUnpack?, binaries?, bundleShortVersion?, bundleVersion?, category?, compression?, cscInstallerKeyPassword?, cscInstallerLink?, cscKeyPassword?, cscLink?, darkModeSupport?, detectUpdateChannel?, electronLanguages?, electronUpdaterCompatibility?, entitlements?, entitlementsInherit?, entitlementsLoginHelper?, extendInfo?, extraDistFiles?, extraFiles?, extraResources?, fileAssociations?, files?, forceCodeSigning?, gatekeeperAssess?, generateUpdatesFilesForAllChannels?, hardenedRuntime?, helperBundleId?, helperEHBundleId?, helperGPUBundleId?, helperNPBundleId?, helperPluginBundleId?, helperRendererBundleId?, icon?, identity?, minimumSystemVersion?, protocols?, provisioningProfile?, publish?, releaseInfo?, requirements?, signIgnore?, strictVerify?, target?, type? } | null
    -> Options related to how build macOS targets.
    Details:
     * configuration.mac.target[0] should be one of these:

```

kagawarei:
Thanks! I fixed it.

Sorry, I didn't check the CI results properly. :bow:

mmaietta (approved):

mmaietta:
@develar would you mind taking a look at this when you have a chance? Ready for final review 🙂

Issue #5402, opened 2020-11-06
Apple Silicon arm64 support


gaodeng:
* **22.9.1**: 

* **Eelectron 11**:
* **Electron Type (current, beta, nightly)**:
beta

* **macOS**: 

With the release of Electron 11.0.0-beta.1, the Electron team is now shipping builds of Electron that run on the new Apple Silicon hardware 
more info about Apple Silicon Support  https://www.electronjs.org/blog/apple-silicon

ahmadwaliesipick:
> * **22.9.1**:
> 
> * **Eelectron 11**:
> * **Electron Type (current, beta, nightly)**:
>   beta
> 
> * **macOS**:
> 
> With the release of Electron 11.0.0-beta.1, the Electron team is now shipping builds of Electron that run on the new Apple Silicon hardware
> more info about Apple Silicon Support https://www.electronjs.org/blog/apple-silicon

@gaodeng  but electron-builder is not supporting ARM64.

ttomaszewski:
Please support ARM64!!!! Don't want to have to move to another package!

ggreco:
Apple will release to the user 3 models of ARM64 macs in 6 days, the support for this is planned/in progress?

alectrocute:
Currently stressing/investigating into this. I love electron-builder and would prefer not to switch.

gaodeng:
I've built a pkg file with arm64 binary and x64 binary, uploaded it to apple for review, and unfortunately the latest version of electron beta contains some private api calls that cause my app to be rejected. here's the details https://github.com/electron/electron/issues/26480

Anyway, here's how I built it.
1 Build the x64 version of the app normally first.
2 Using electron-rebuild -f --arch=arm64 to build native modules for arm64 architecture
3 Configure electron-builder's electronDownload.arch value to "arm64"
4 Run electron-builder to build the arm64 version of the app.
5 Merge x64 and arm64 versions using [@electron/universal ](https://github.com/electron/universal)
6 Manually signing the merged univeral app

warpdesign:
> I've built a pkg file with arm64 binary and x64 binary, uploaded it to apple for review, and unfortunately the latest version of electron beta contains some private api calls that cause my app to be rejected. here's the details [electron/electron#26480](https://github.com/electron/electron/issues/26480)
> 
> Anyway, here's how I built it.
> 3 Configure electron-builder's electronDownload.arch value to "arm64"
Could you elaborate?

[...]

Issue #5397, opened 2020-11-05
You cannot open the application because this application is not supported by this Mac, Big Sur Silicon mac 


ahmadwaliesipick:
I am using current latest Electron Beta build to create dmg at Big Sur Silicon mac. Build has successfully created and installed. Once i try to open app i see below error.
https://share.getcloudapp.com/Z4uyExDY

Once I run build through electron command then app is working fine but after dmg create app is not working.
Anyone has experience this issue?

Anil8753:
Is there any workaround available?

ahmadwaliesipick:
No yet @Anil8753

ahmadwaliesipick:
@Anil8753  see here they have put this ticket in backlog and marked closed even still electron-builder is not supporting ARM64 build. #5095

Anil8753:
@ahmadwaliesipick did you find any other alternative. Yesterday Apple released Silicon hardware and user can not launch Electron builder apps, such a painful situation.

ishowshao:
waiting...

ahmadwaliesipick:
> @ahmadwaliesipick did you find any other alternative. Yesterday Apple released Silicon hardware and user can not launch Electron builder apps, such a painful situation.

I was thinking to use electron-packager but it is lot of work.

Anil8753:
https://github.com/electron/electron/issues/26530#issuecomment-729163078
@MarshallOfSound  says, Confirmed that this will work just fine on the M1 machines. If you want to work around this on your DTK ensure you launch the app from the terminal directly. E.g. ./MyApp.app/Contents/MacOS/MyApp

ishowshao:
`electronDownload.arch = arm64` is ok

aeschylus:
https://github.com/beakerbrowser/beaker/issues/1861

Issue #5392, opened 2020-11-03
Plans to Support Apple Silicon builds?


eengoron:
Are there current plans to support Apple Silicon builds with `electron-builder`? We're currently in the process of making sure our electron app is ready for ARM support, so I wanted to make sure that we'd be able to build to those targets using `electron-builder`.

If so, what does the release timeline look like for that? 

Since https://github.com/electron-userland/electron-builder/issues/5095 was closed with no resolution, I figured it best to open a new issue.

ahmadwaliesipick:
any update on this?

ahmadwaliesipick:
> Are there current plans to support Apple Silicon builds with `electron-builder`? We're currently in the process of making sure our electron app is ready for ARM support, so I wanted to make sure that we'd be able to build to those targets using `electron-builder`.
> 
> If so, what does the release timeline look like for that?
> 
> Since #5095 was closed with no resolution, I figured it best to open a new issue.

@eengoron  see here they have put this ticket in backlog and marked closed even still electron-builder is not supporting ARM64 build. #5095

eengoron:
@ahmadwaliesipick -- I mentioned that ticket in my original message. That ticket was closed without resolution, so I opened a new one.

cc @develar

ahmadwaliesipick:
> @ahmadwaliesipick -- I mentioned that ticket in my original message. That ticket was closed without resolution, so I opened a new one.
> 
> cc @develar

Yes @eengoron . Do you have any information are they working on ARM64 build?

aabuhijleh:
Unfortunately, it looks like it's time to consider migrating to electron-packager for this issue :(

develar:
Anyone is willing to open a pull request? Fix should be easy, as far I see.

deathlyrage:
Can we get this issued pinned, saves all the duplicate issues being made and easier to focus for people helping out with pull requests and debugging.

mmaietta:
FWIW, I was able to build an `arm64` electron example app with the help of `patch-package` on two files.  Perhaps this will help/unblock people investigating further 🙂 

I was able to package via both `yarn electron-builder` to use package.json and cli `yarn electron-builder --dir --arm64`

Unfortunately, I'm actually unable to test the app though as I don't have a compatible MacOS version installed right now. Just to be verbose, here's the error alert
> You can’t use this version of the application “MyApp.app” with this version of macOS.
You have macOS 10.15.7. The application requires macOS 11.0 or later.

[...]

Issue #5095, opened 2020-06-24
Unable to build for ARM64


fabiospampinato:
* **Version**: 22.7.0

* **Electron Version**: 9.0.4
* **Electron Type (current, beta, nightly)**: Current

* **Target**: All targets (ARM64)

I'm trying to produce arm64 artifacts for my app, for all platforms, from macOS 10.14, but the process gets stuck and never finishes, no error is outputted. I can only see the `multipass` process spinning forever from the task amanger.

rankjie:
I believe the building process has hardcoded to be only building Arch.x64 for macOS right now.
https://github.com/electron-userland/electron-builder/blob/7b335dbd7e5ca754598ca93318be1404fc518485/packages/app-builder-lib/src/macPackager.ts#L103

Since Apple is about to release ARM-based Mac this year, we need to support building for ARM64.

fabiospampinato:
Still relevant.

deathlyrage:
Still relevant.

fabiospampinato:
I like how the stupid bot just closes issues like that, shouldn't there be an issue tracking compatibility with Apple Silicon? @develar

deathlyrage:
Apple silicon launching at the end of the year, google resolved ARM64 issues, Upstream is going to have ARM64 builds for apple silicon soon.  We should be tracking this one. Lots of electron builder apps have a lot of upgrade work to do to even open.

mmaietta:
Just for posterity's sake and future readers, we probably should consolidate Apple Silicon/arm64 discussions in this ticket: https://github.com/electron-userland/electron-builder/issues/5392#issuecomment-725441867
An investigation is ongoing there.

fabiospampinato:
Closing as this repo isn't issue-friendly, I don't want to receive notifications from the idiot bot anymore.

PR #5426, opened 2020-11-18
feat:(mac): Apple Silicon builds (arm64) 

mmaietta:
Apple Silicon support: Initial dev work for achieving arm64 builds for Mac. This PR enables packaging both arm64 and x64 on mac. Supports building native app dependencies for both archs as well. dmg assets will follow current electron-builder logic/nomenclature, with only x64 builds not having an arch suffix.

Opening PR for tracking issues:
#5095
#5392 
#5397
#5402 

Quick Notes:
Unrelated, PR also fixes build error with unused `semver` import

Copying here from #5392 for posterity's sake. I'm unable test on MacOS Big Sur as I'm blocked from installing it on my company laptop. I'd appreciate it if others could try it locally

For local development, I used [yalc](https://github.com/whitecolor/yalc) so I could test changes with the electron-quick-start app
Env setup from scratch (line by line for an easy copy-paste)
```
yarn global add yalc

git clone -b Feature/AppleSiliconARM64 https://github.com/mmaietta/electron-builder.git

( cd electron-builder && yarn && yarn compile && \
(cd ./packages/app-builder-lib && yalc push) && \
(cd ./packages/builder-util && yalc push) && \
(cd ./packages/builder-util-runtime && yalc push) && \
(cd ./packages/dmg-builder && yalc push) && \
(cd ./packages/electron-builder && yalc push) && \
(cd ./packages/electron-publish && yalc push) && \
(cd ./packages/electron-builder-squirrel-windows && yalc push) && \
(cd ./packages/electron-forge-maker-appimage && yalc push) && \
(cd ./packages/electron-forge-maker-nsis && yalc push) && \
(cd ./packages/electron-forge-maker-nsis-web && yalc push) && \
(cd ./packages/electron-forge-maker-snap && yalc push) && \
(cd ./packages/electron-updater && yalc push) )

git clone https://github.com/electron/electron-quick-start.git

cd electron-quick-start

npm install && npm install -D electron-builder

yalc link app-builder-lib builder-util builder-util-runtime dmg-builder electron-builder electron-publish electron-builder-squirrel-windows electron-forge-maker-appimage electron-forge-maker-nsis electron-forge-maker-nsis-web electron-forge-maker-snap electron-updater

yarn electron-builder --arm64 --x64
```
And now the magical brute-force one-liner for whenever you make changes! Rebuilds electron-builder, patches the npm module in electron-quick-start, and begins packaging.
```
cd electron-builder

yarn compile && \
(cd ./packages/app-builder-lib && yalc push) && \
(cd ./packages/builder-util && yalc push) && \
(cd ./packages/builder-util-runtime && yalc push) && \
(cd ./packages/dmg-builder && yalc push) && \
(cd ./packages/electron-builder && yalc push) && \
(cd ./packages/electron-publish && yalc push) && \
(cd ./packages/electron-builder-squirrel-windows && yalc push) && \
(cd ./packages/electron-forge-maker-appimage && yalc push) && \
(cd ./packages/electron-forge-maker-nsis && yalc push) && \
(cd ./packages/electron-forge-maker-nsis-web && yalc push) && \
(cd ./packages/electron-forge-maker-snap && yalc push) && \
(cd ./packages/electron-updater && yalc push) \
&& ( cd ../electron-quick-start && rm -rf dist && yarn electron-builder --arm64 --x64 )
```

If wanting to use a package.json, this is the config:
```
"devDependencies": {
    "electron": "^11.0.0",
    "electron-builder": "^22.9.1"
  },
  "build": {
    "mac": {
      "target": {
        "target": "dmg",
        "arch": [
          "arm64",
          "x64"
        ]
      }
    }
  }
```
If using native addons, it's likely they haven't supplied compiled assets yet for darwin-arm64, so you'll need to rebuild from source.  That also works with this PR, just add `"buildDependenciesFromSource": true` and presto, arm64 native addon.

Example output (with spacing for better legibility) if using sqlite3 and electron-quick-start app with the env described above:
```
$ electron-builder  • electron-builder  version=22.8.1 os=19.6.0  • loaded configuration  file=package.json ("build" field)

  • rebuilding native dependencies  dependencies=sqlite3@5.0.0 platform=darwin arch=arm64
  • rebuilding native dependency  name=sqlite3 version=5.0.0
  • packaging       platform=darwin arch=arm64 electron=11.0.0 appOutDir=dist/mac-arm64
  • skipped macOS code signing  reason=identity explicitly is set to null
  • building        target=DMG arch=arm64 file=dist/electron-quick-start-1.0.0-arm64.dmg

  • rebuilding native dependencies  dependencies=sqlite3@5.0.0 platform=darwin arch=x64
  • rebuilding native dependency  name=sqlite3 version=5.0.0

  • building block map  blockMapFile=dist/electron-quick-start-1.0.0-arm64.dmg.blockmap

  • packaging       platform=darwin arch=x64 electron=11.0.0 appOutDir=dist/mac
  • skipped macOS code signing  reason=identity explicitly is set to null
  • building        target=DMG arch=x64 file=dist/electron-quick-start-1.0.0.dmg

  • building block map  blockMapFile=dist/electron-quick-start-1.0.0.dmg.blockmap
✨  Done in 157.86s.
```


lutzroeder:
@mmaietta Do you have any thoughts on next steps to support building a single universal binary?

quanglam2807:
> @mmaietta Do you have any thoughts on next steps to support building a single universal binary?

That actually depends on `electron`, not just `electron-builder`

Source: https://www.electronjs.org/blog/apple-silicon
> In the future, we will release a package that allows you to "merge" your arm64 and x64 apps into a single universal binary, but it's worth noting that this binary would be huge and probably isn't ideal for shipping to users.

lutzroeder:
>> @mmaietta Do you have any thoughts on next steps to support building a single universal binary?
> That actually depends on electron, not just electron-builder

@quanglam2807 see electron/electron#26710. The package already exists.

elliottkember:
I'm trying to get this working, followed all the steps but seems to be hanging on this one:

```
$ DEBUG=electron-builder npx electron-builder install-app-deps --arch=arm64
[...snip]
  • rebuilding native dependencies  dependencies=@serialport/bindings@9.0.2, keytar@7.2.0 platform=darwin arch=arm64
  • execute command  command=/Users/elliott/.nvm/versions/node/v15.3.0/bin/node /Users/elliott/.nvm/versions/node/v15.3.0/lib/node_modules/npm/bin/npx-cli.js rebuild --verbose @serialport/bindings@9.0.2 keytar@7.2.0
                     workingDirectory=
```

It hangs on that command (I'm on an M1 MBP) - just runs forever. Doesn't matter if I specify `--arch`. Am I maybe running a broken version of node-gyp?

packages/dmg-builder/src/dmg.ts
```diff
@@ -115,15 +115,17 @@ export class DmgTarget extends Target {
...
 
-  computeVolumeName(custom?: string | null): string {
+  computeVolumeName(arch: Arch, custom?: string | null): string {
     const appInfo = this.packager.appInfo
     const shortVersion = this.packager.platformSpecificBuildOptions.bundleShortVersion || appInfo.version
+    const archString = getArchSuffix(arch)
 
     if (custom == null) {
-      return `${appInfo.productFilename} ${shortVersion}`
+      return `${appInfo.productFilename} ${shortVersion}${archString}`
```
quanglam2807, line 124: For backward compatibility, I think it's better if arch suffix is only added for `arm64` builds. If we add suffix to `x64` builds, would that break `electron-updater`?

mmaietta: Great callout, `getArchSuffix` actually already handles that logic:
```
export function getArchSuffix(arch: Arch): string {
  return arch === Arch.x64 ? "" : `-${Arch[arch]}`
}
```
https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util/src/arch.ts#L27-L29

test/snapshots/mac/macPackagerTest.js.snap
```diff
@@ -152,23 +168,43 @@ Object {
...
           },
           Object {
             "sha512": "@sha512",
             "size": "@size",
-            "url": "TestApp-1.1.0-mac.dmg",
+            "url": "TestApp-1.1.0-mac-arm64.dmg",
           },
         ],
-        "path": "TestApp-1.1.0-mac.zip",
+        "path": "TestApp-1.1.0-mac-x64.zip",
```
quanglam2807, line 190: So suffix for `x64` is not added, as defined in https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util/src/arch.ts#L27-L29, then I guess this test is not valid? Or am I missing something?

mmaietta: The other logic is for `dmg.ts` only. I'll double-check regarding this zip logic

mmaietta: Okay, the .zip logic by default includes an arch name
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/electron/ElectronFramework.ts#L151

The reason this test is updated is due to the custom artifact name template that had to be updated in the test.
https://github.com/electron-userland/electron-builder/blob/1acc32e6c3eb566c3c4e59f9098db05cead4656e/test/src/mac/macPackagerTest.ts#L19

The reason for the test update is that a custom artifact string requires an `${arch}` value otherwise it attempts to produce 2 dmg artifacts of the same name. I recall this behavior also exists for Windows builds if one wants to bundle 32bit and 64bit separately, so I assumed this was the equivalent approach here as well.
Example code from my build scripts:
```
const buildSuffix = `-${process.platform === 'win32' ? arch : "${arch}"}${isDebug ? "-Debug" : ""}`;
const artifactTemplate = "${productName}-${buildVersion}" + buildSuffix + ".${ext}",
```

Can you please advise on what you'd expect the behavior here to be?

quanglam2807: Thank you for taking your time, @mmaietta. Currently,  because the old `artifactName` doesn't include `arch`, the `zip` package for macOS doesn't have arch suffix. I doesn't bother me too much if the zip package file name changes. Also, I think currently, `electron-updater` only uses DMG instead of ZIP for updating (https://github.com/electron-userland/electron-builder/issues/2199).

I just worry we might end up breaking something in the process with this change. If not, this is no big deal.

@devlar Would changing `artifactName` breaks `electron-updater`?

mmaietta: @quanglam2807 I found an easy change to retain the previous naming style and only append the arch for mac arm64 builds.
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/platformPackager.ts#L475
Simply change that line to:
```
else if (this.platform !== Platform.MAC) {
```
It feels messy, and it's basically equivalent to the MAC platform<->arch hardcoding this PR is fixing, but I also don't want to break anything we couldn't foresee. The code comments there already exemplify previous backward-compatibility woes in issue https://github.com/electron-userland/electron-builder/issues/3510. An additional win, this approach should also be forward compatible with universal binaries, which by default should not have an arch suffix.

AFAICT, this approach also mirrors the Windows Packager logic with its universal installers too.

quanglam2807: Awesome! Thank you so much, @mmaietta.

quanglam2807 (approved):

quanglam2807:
@mmaietta @develar Is it possible to get this PR merged and cut a pre-release so we can all test it ?

mmaietta:
> @mmaietta @develar Is it possible to get this PR merged and cut a pre-release so we can all test it ?

Yes, the build and tests were passing previously, but master branch is currently broken. I'm hesitant to merge into something broken (currently working on this PR https://github.com/electron-userland/electron-builder/pull/5449 to fix master, almost done). If you'd like to merge now, then we can do so.

elliottkember:
One thing I've found is that I get an error if I run `electron-builder` natively:

      ⨯ Exit code: 1. Command failed: /usr/bin/python /Users/elliott/appname/.yalc/dmg-builder/vendor/dmgbuild/core.py
    Traceback (most recent call last):
      File "/Users/elliott/appname/.yalc/dmg-builder/vendor/dmgbuild/core.py", line 25, in <module>
        from mac_alias import *
      File "/Users/elliott/appname/.yalc/dmg-builder/vendor/mac_alias/__init__.py", line 1, in <module>
        from .alias import *
      File "/Users/elliott/appname/.yalc/dmg-builder/vendor/mac_alias/alias.py", line 16, in <module>
        from . import osx
      File "/Users/elliott/appname/.yalc/dmg-builder/vendor/mac_alias/osx.py", line 424, in <module>
        _statfs = libc['statfs$INODE64']
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ctypes/__init__.py", line 384, in __getitem__
        func = self._FuncPtr((name_or_ordinal, self))
    AttributeError: dlsym(0x100ed1690, statfs$INODE64): symbol not found

I think the reason for this is that the `python` running these python commands is arm64 and doesn't like `statfs$INODE64` - I'll add an issue to the `mac_alias` repo.

From reading [this link](https://assert.cc/posts/darwin_use_64_bit_inode_vs_ctypes/) - I think this has something to do with the `$INODE64` suffix. Perhaps it isn't needed on ARM python.

Anyway, running the command with `arch -x86_64` worked and built for both x86 and arm64.

elliottkember:
I think the MacUpdater will also need a new `findFile` function to determine which architecture to download.

The [current findFile](https://github.com/electron-userland/electron-builder/blob/f938de5aae09cb832a6f7f742e64a0d6c025fb9f/packages/electron-updater/src/MacUpdater.ts#L35) seems to use the first matching-suffix file it finds. So both arm64 and x86 will download whichever is listed first in the `files` array.

Maybe even

    const result = files.find(it => it.url.pathname.toLowerCase().endsWith(`${getArchSuffix(process.arch)}.${extension}`))

test/src/mac/macPackagerTest.ts
```diff
@@ -16,7 +16,7 @@ test.ifMac.ifAll("two-package", () => assertPack("test-app", {
       electronLanguages: ["bn", "en"]
     },
     //tslint:disable-next-line:no-invalid-template-strings
-    artifactName: "${name}-${version}-${os}.${ext}",
+    artifactName: "${name}-${version}-${os}-${arch}.${ext}",
```
elliottkember, line 19: From my testing I think this might actually be:

`artifactName: "${name}-${version}-${arch}-${os}.${ext}",`

and I think it's generated in https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/targets/ArchiveTarget.ts#L28

mmaietta: Hmm, that is true if an `artifactName` template is not provided in the build config. I assumed this test was also validating `artifactName` as opposed to using the `defaultPattern` you linked in ArchiveTarget.ts.
I updated this test to have the `${arch}` template explicitly specified as otherwise, the dmg/zip will overwrite itself when building for multiple targets.

mmaietta:
> I think there's a bug with the `findFile` function, which will need to determine which architecture to download.
> 
> The [current findFile](https://github.com/electron-userland/electron-builder/blob/f938de5aae09cb832a6f7f742e64a0d6c025fb9f/packages/electron-updater/src/MacUpdater.ts#L35) seems to use the first matching-suffix file it finds. So both arm64 and x86 will download whichever is listed first in the `files` array.
> 
> Because `x86` isn't included in the `.zip` filename, we have:
> 
> * MyApp-2.0.89-arm64-mac.zip
> * MyApp-2.0.89-mac.zip
> 
> They're both `.zip` and whichever one is listed first will be downloaded. We can't rely on `getArchSuffix` because x86 is "" and they'll both match an empty string.

Nice callout. I see `findFiles` is provided `Array<ResolvedUpdateFileInfo>`, which contains `UpdateInfo::url string` and `PackageFileInfo::path` and I belieeve should already have identified the correct file assets, so we may need to process the arch further upstream as you hinted at. I think we could rename `WindowsUpdateInfo` to `MultiArchUpdateInfo` and reuse a `packages[process.arch] || packages[process.platform == 'darwin' ? 'x64' : 'ia32'` approach. It already rewrites `result[0]` which is what `findFiles` will utilize.
https://github.com/electron-userland/electron-builder/blob/f938de5aae09cb832a6f7f742e64a0d6c025fb9f/packages/electron-updater/src/providers/Provider.ts#L149-L152
My worry is if/how we need to migrate between electron-updater schema versions? What are your thoughts?
I'm really not familiar with how electron-updater works as my team doesn't use it.

elliottkember:
That sounds good. Might also be nice to refactor the overwriting results[0] thing to just return a single file.

What about if we patched MacUpdater with the `.filter` here, then do the `MultiArchUpdateInfo` refactor in a separate PR? With the electron-updater fix, which I'm currently using from my own fork, it all seems to be working nicely on my end. Proposing this because it'd be great to get it released so people can test without jumping through yalc hoops!

> My worry is if/how we need to migrate between electron-updater schema versions? What are your thoughts?

Sorry, I don't really know what you mean by this. I don't use it very much either (I just use and enjoy electron-builder) but I'm sure we can deal with that if/when it comes up. What's your worry here?

mmaietta:
> What about if we patched MacUpdater with the `.filter` here, then do the `MultiArchUpdateInfo` refactor in a separate PR? With the electron-updater fix, which I'm currently using from my own fork, it all seems to be working nicely on my end. Proposing this because it'd be great to get it released so people can test without jumping through yalc hoops!

Completely agreed. Updated the PR accordingly.

> Sorry, I don't really know what you mean by this. I don't use it very much either (I just use and enjoy electron-builder) but I'm sure we could deal with that if/when it comes up. What's the concern?

I have no idea how to handle universal binaries as the build process is outside of electron-builder, so I have no idea how it'll integrate with electron-updater. It should be outside the scope of this PR though since that would require an entirely new npm module integrated 
https://github.com/electron/universal

Hopefully we can get a pre-release cut first :)

quanglam2807:
From my perspective: universal binary simply combines x64 and arm64 .app files into a single one. So technically, when a user runs the app, they still runs the same individual x64 or arm64 binary. Because of this, it wouldn’t break electron-updater as it still relies on arch-specific builds to update.

test/src/mac/macPackagerTest.ts
elliottkember: I think I've been seeing something like this. When I'm exporting and uploading DMGs, the -arm64 DMGs don't get published - it's weird. It doesn't seem to happen every time, but I can't figure out what's causing it to happen.

elliottkember:
I'm having trouble with DMG filenames. Sometimes they come out the same, without the ${arch}, sometimes they don't. It feels like a weird race condition, or maybe I've forgotten a command somewhere.

I'm going to try using artifactName to differentiate them.

```
version: 2.0.111
files:
  - url: Soulmate-2.0.111-mac.zip
    sha512: N9DU9tPg9YfB4OIdHIvTgN7e1C/KRCU1B71mDkTHx027AYMgHC2/c0KVi2u067QZYb4v7Us7lgl6/HT9HOgk8A==
    size: 149032867
    blockMapSize: 157632
  - url: Soulmate-2.0.111-mac.zip
    sha512: XlhTgqqi/NR6s1Rla4ipKij8gZR6NT6vpBArcm7/E0vSuRhL5AJgnUqEHECkLqRerZTu9Pt8HUzknxlkGjeSxw==
    size: 149625156
    blockMapSize: 156752
  - url: Soulmate-2.0.111.dmg
    sha512: 596iDNFGKqCN/EsBz9SwEA0Wi74P0+T5QQxMdVr4x7Ooyc15R8EbyeHdg++KEI9U4Qyn1neGpveUfoj4xiTUAQ==
    size: 155574955
  - url: Soulmate-2.0.111.dmg
    sha512: u6cQxajQS0b286xIf+dko0mw2oNKxsTxo/30YYZaveVSp0oUxvZkdzcK4ExnYJ2xNKpPSdYnNsQT3UIDIRG/Ow==
    size: 156393297
path: Soulmate-2.0.111-mac.zip
sha512: N9DU9tPg9YfB4OIdHIvTgN7e1C/KRCU1B71mDkTHx027AYMgHC2/c0KVi2u067QZYb4v7Us7lgl6/HT9HOgk8A==
releaseDate: '2020-12-09T01:46:27.980Z'
```

elliottkember:
My 2c on universal electron binaries is that they'll be HUGE. It might just be easier to have two download buttons, at least for now?

lutzroeder:
@elliottkember all 3 options should be supported: arm64, x86_64 and universal. The bits might be larger but another consideration is the complexity of choosing the right architecture. For some apps users shouldn't be exposed to this even if it means larger binaries. As discussed already auto-update is another aspect. Not so much now, but in the future when Intel is phased out apps might have to keep the `-arm64` suffix to keep update going. Universal makes this more seamless and there should be a choice for app authors to go either way. Obviously large 500 MB apps should think carefully before moving to a 1 GB download, but for smaller apps 50 MB vs. 100 MB might be bearable for a while.

elliottkember:
@lutzroeder That's a good point. I'm working with a 500MB app and it's easy to forget that there are many smaller apps around. Universal is definitely preferable.

quanglam2807:
> @lutzroeder That's a good point. I'm working with a 500MB app and it's easy to forget that there are many smaller apps around now. Universal is definitely preferable.

I agree. Also, as I mentioned in my previous comment, if `electron-updater` still downloads arch-specific builds for updating, then only the initial download is heavy. 

Anyway, I agree with @mmaietta that we should have this PR merged and tested it out first before moving to adding universal binary support.

mmaietta:
Tests in this PR will continue to be broken until https://github.com/electron-userland/electron-builder/pull/5449 is merged and I can pull it into my branch 😕
Trying to keep changes in this PR explicitly for M1 arm64

Who can we ping to move these forward?

elliottkember:
@mmaietta https://github.com/electron-userland/electron-builder/pull/5449 is now merged!

elliottkember:
I'm getting all kinds of issues with Github Actions today. Sometimes it packages arm64 with Intel files, other times it doesn't upload all the built assets.

```
  • building block map  blockMapFile=dist/Soulmate-2.0.131-mac.dmg.blockmap
  • building embedded block map  file=dist/Soulmate-2.0.131-mac-x64.zip
  • publishing      publisher=S3 (bucket: soulmate-ide-updates)
  • uploading       file=Soulmate-2.0.131-mac.dmg.blockmap provider=S3
  • uploading       file=Soulmate-2.0.131-mac.dmg provider=S3
  • uploading       file=Soulmate-2.0.131-mac-x64.zip provider=S3
Done in 451.29s.
```

It feels a bit like there's a race condition happening here - it doesn't get around to uploading the final arm64 DMG, and it also doesn't upload the `latest-mac.yml` file. I'm wondering whether there's some logic that is tracking all the steps to be completed, and isn't counting both steps.

I'd love to debug this with someone who knows more about electron-builder!

mmaietta:
@elliottkember what I've noticed is that node-gyp doesn't clean its build/Release/*.node directory between builds, so you're likely packaging whichever node-gyp arch was executed first.
I can reproduce this with both electron-builder, electron-rebuild, and node-gyp directly.

The only solution I can think of in order for electron-builder to be able to build multiple arch's at the same time is to have a list of targets each separately, and also clean the node addon build directory between node-gyp executions.  This may reduce how much of the compilation is in parallel, I do not know for certain.

For the interim, I suggest an approach like this:
```
  const build = async (arch) => {
    const targets = builder.createTargets([builder.Platform.MAC], 'dir', arch)
    const config: builder.Configuration = {
      ...baseConfig,
      afterSign: async (context: builder.AfterPackContext) => {
        if (context.electronPlatformName === "darwin") {
          await notarize(...)
        }
      },
    }
    return builder.build({ targets, config })
  }

  await build('x64')
  execSync('yarn node-gyp clean');
  await build('arm64')
```

elliottkember:
> This may reduce how much of the compilation is in parallel, I do not know for certain.

Right - I wouldn't mind the whole thing being synchronous as it happens in CI and would only add a few minutes to my build. I think it's an improvement if this would guarantee the correct architecture. Perhaps it could be a flag passed to `electron-builder`, something like `-clean-node-gyp-between-builds`.

Your code example seems to be using the programmatic API, which I haven't used yet - I'm just using `electron-builder -m` to publish.

elliottkember:
@mmaietta FYI, `node-gyp clean` deletes the `build` directory, which is the default `buildResources` directory for `electron-builder` 😅

mmaietta:
> Would an afterPack function work, do you think?

For universal packages, I think we could use `afterAllArtifactBuild` for stitching them together since we'd need both arm64 and x64 packages successfully built already. Might require another signing step for the new universal app though 🤷‍♂️ And somehow integrate another packaging step.
An `afterAllPack` could be handy for this use case. If an `afterAllPack` hook were to return an array of app filepaths, which would then each be signed+packaged into their own corresponding dmg/zip/etc. It'd be easily backward-compatible.

> FYI, node-gyp clean deletes the entire build directory, which is the default buildResources directory for electron-builder

Well that's just unfortunate, how did we end up using the default output directory of node-gyp `./build/Release/` as the default directory for `buildResources`? 😅

I'm not sure on the best route then seeing as `node-gyp clean` seems more future-proof and config-agnostic of an approach compared to `rm -rf build/Resources`.
On a related note, I still haven't been able to figure out how to clear previously built native modules during `install-app-deps` when `buildDependenciesFromSource: true`. 
Got any ideas by chance? Any vanilla `node-gyp` or `npm/yarn install` cmds we could leverage?

lutzroeder:
@mmaietta is there a way to run the lipo step before sign/notarize or skipping the signing during the initial build? Since that tends to be the weak link doing it once would be preferable and Apple should know how to handle just the universal binary.

elliottkember:
> Well that's just unfortunate, how did we end up using the default output directory of node-gyp ./build/Release/ as the default directory for buildResources? 😅

Great minds thinking alike!

> On a related note, I still haven't been able to figure out how to clear previously built native modules during install-app-deps when buildDependenciesFromSource: true.

I had to abandon that. It just hung on Github Actions and never finished. Which can be expensive when you're running macos workers. I know that there's `yarn run electron-builder --config.buildDependenciesFromSource=true` which I have as a separate yarn script.

All I can think of doing is `rm -rf node_modules && yarn`. I don't mind nuclear options like this, especially in CI.

> For universal packages, I think we could use afterAllArtifactBuild for stitching them together since we'd need both arm64 and x64 packages successfully built already. Might require another signing step for the new universal app though 🤷‍♂️ And somehow integrate another packaging step.

I was actually talking about cleaning out node-gyp between builds. So `afterPack` is called after the first build is signed and before the second one is started. This has worked great for me as far as I can tell.

mmaietta:
> All I can think of doing is rm -rf node_modules && yarn. I don't mind nuclear options like this, especially in CI.
>
> I was actually talking about cleaning out node-gyp between builds. So afterPack is called after the first build is signed and before the second one is started. This has worked great for me as far as I can tell.

A shotgun non-nuclear approach might be `rm -rf **/*.node`

Turns out there's another hook, `beforeBuild`, which cancels out the default behavior of electron-builder's install-production-dependencies step.  That might be optimal for running `electron-rebuild` to avoid the nuking, haven't tested my scripted pipeline yet
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/packager.ts#L503

I was debugging further into `buildDependenciesFromSource` and it seems that it's unable to bypass other configurations like electron-rebuild does.  leveldown and node-mac-permissions, for instance, always build x86_64 with node-gyp-build, whereas electron-rebuild properly builds arm64. Easy lipo search: `find . -name '*.node' | xargs -I {} lipo -info {} | grep -e Non-fat`
I think the rebuild breaks here:
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/util/yarn.ts#L156

mmaietta:
@develar @quanglam2807 just curious, is there a specific reason we use app-builder-bin for recompiling node modules instead of using electron's maintained electron-rebuild package? I see that electron-forge has it as a dependency.

Curious if we could allow electron-rebuild to be used if it is present in the project's package.json and/or allow it to be enabled via  anelectron-builder.js config var.

Eugeny:
@mmaietta electron-rebuild is terrible on Windows, and still overrides gyp config in a way that breaks compatibility with recent Visual Studio, even though gyp by itself supports it.

mmaietta:
@Eugeny quite unfortunate 😅

AFAICT, the issue right now is that app-builder-bin doesn't **consistently** respect the build config being passed to it. So we end up with the default arch always being built for some modules. See output below:
```
➜  builder-project git:(Maintenance) ✗ rm -rf node_modules/**/*.node; yarn node-gyp clean;
yarn electron-builder install-app-deps --arch=x64 && \
(find . -name '*.node' | xargs -I {} lipo -info {} | grep -e Non-fat)

yarn run v1.22.10
$ /builder-project/node_modules/.bin/electron-builder install-app-deps --arch=x64
  • electron-builder  version=22.8.1
  • /builder-project/electron-builder.js
  • rebuilding native dependencies  dependencies=leveldown@5.6.0, node-mac-permissions@2.2.0, sqlite3@5.0.0 platform=darwin arch=x64
  • rebuilding native dependency  name=node-mac-permissions version=2.2.0
  • rebuilding native dependency  name=leveldown version=5.6.0
  • rebuilding native dependency  name=sqlite3 version=5.0.0
✨  Done in 56.99s.

Non-fat file: ./node_modules/leveldown/build/Release/leveldown.node is architecture: x86_64
Non-fat file: ./node_modules/node-mac-permissions/build/Release/permissions.node is architecture: x86_64
Non-fat file: ./node_modules/sqlite3/build-tmp-napi-v3/Release/node_sqlite3.node is architecture: x86_64
Non-fat file: ./node_modules/sqlite3/lib/binding/napi-v3-darwin-x64/node_sqlite3.node is architecture: x86_64



➜  builder-project git:(Maintenance) ✗ rm -rf node_modules/**/*.node; yarn node-gyp clean;
yarn electron-builder install-app-deps --arch=arm64 && \
(find . -name '*.node' | xargs -I {} lipo -info {} | grep -e Non-fat)

yarn run v1.22.10
$ /builder-project/node_modules/.bin/electron-builder install-app-deps --arch=arm64
  • electron-builder  version=22.8.1
  • /builder-project/electron-builder.js
  • rebuilding native dependencies  dependencies=leveldown@5.6.0, node-mac-permissions@2.2.0, sqlite3@5.0.0 platform=darwin arch=arm64
  • rebuilding native dependency  name=node-mac-permissions version=2.2.0
  • rebuilding native dependency  name=leveldown version=5.6.0
  • rebuilding native dependency  name=sqlite3 version=5.0.0
✨  Done in 58.02s.

Non-fat file: ./node_modules/leveldown/build/Release/leveldown.node is architecture: x86_64
Non-fat file: ./node_modules/node-mac-permissions/build/Release/permissions.node is architecture: x86_64
Non-fat file: ./node_modules/sqlite3/build-tmp-napi-v3/Release/node_sqlite3.node is architecture: arm64
Non-fat file: ./node_modules/sqlite3/lib/binding/napi-v3-darwin-arm64/node_sqlite3.node is architecture: arm64
```

`node-gyp-rebuild` uses `node-gyp rebuild` directly:
https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/src/cli/cli.ts#L88

`install-app-deps` does something funky that I can't trace to any source code (it'd help a lot if I could! Have any pointers by chance?)
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/util/yarn.ts#L155-L156

I have some old code laying around that can run `node-gyp rebuild` in every working directory of a project's native modules that we could try out, but it'd basically be the same as https://github.com/electron/electron-rebuild/blob/master/src/module-rebuilder.ts#L78-L107

mmaietta:
Are there any remaining todo items or can we merge for a pre-release cut? It'd be great to get some stress-testing by the community.
All tests are passing. 🙂 
CC @develar

gnattu:
#5460 should be merged as `libc['statfs$INODE64']` is deprecated and actually no longer available on arm64, this will make the building fail.

#5465 should be addressed as HFS+ disk image may fail to create during copy of the .app file on Big Sur which all arm64 Macs are on, but using default setting(APFS), the dmg builds successfully.

mmaietta:
Good pointers.
#5460 needs the CI retriggered as master branch is now compile-able.
#5465 seems to just need a one-liner PR change, but I don't think letting it use the default setting will be backward compatible as APFS is only available High Sierra 10.13+. It would also require verifying that the default value is consistent across all recent OS versions, so an explicit `-fs` arg might be necessary.

Anyways, issues specifically associated with bundling on Big Sur and/or M1 will be progressively discovered (particularly more once this is merged I'd imagine), so I'd like to not stall this PR. As a baseline, we need arm64 compilation on Intel-based macs to start.

I'd advocate merging this PR as I think we could pre-release without those.

kaatt:
@mmaietta can we use your feature branch and test this without waiting for merge?

`yarn add https://github.com/mmaietta/electron-builder.git#Feature/AppleSiliconARM64` should work for installing

quanglam2807:
@develar Is it possible to take risk, just merge this PR and cut a pre-release? We'd love to test this out. If there're bugs, we can always fix them later.

kaatt:
Can someone fork and publish the packages temporarily?

I forked and merged here: https://github.com/kaatt/electron-builder but the monorepo makes this complex. I'm trying out https://github.com/whitecolor/yalc now

develar:
https://github.com/electron-userland/electron-builder/releases/tag/v22.10.0

quanglam2807:
We missed something. Builds for `mas x64` and `mas arm64`  (Mac App Store) are both created in the same output dir (`mas` or `mas-dev`) with the same file names (no arch prefix) and overwrite each other. I tried to look through the code but couldn't find where it's defined. Can you someone help with this?

mmaietta:
@quanglam2807 can you test it with this change?
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/macPackager.ts#L119
Change to:
```
const targetOutDir = path.join(outDir, `${targetName}${getArchSuffix(arch)}`)
```

lutzroeder:
@quanglam2807 @mmaietta have you successfully built a `.dmg` using a signed `.app` with 22.10.3?

quanglam2807:
> @quanglam2807 can you test it with this change?
> https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/macPackager.ts#L119
> Change to:
> 
> ```
> const targetOutDir = path.join(outDir, `${targetName}${getArchSuffix(arch)}`)
> ```

Thank you, @mmaietta. It works. What about `artificateName` for `mas` builds (pkg)? The filenames don't have `arch`.
`/dist/mas-arm64/Translatium 16.0.0.pkg`
`/dist/mas/Translatium 16.0.0.pkg`

Probably, `artifactName` for `mas` is hard-coded somewhere.

quanglam2807:
> @quanglam2807 @mmaietta have you successfully built a `.dmg` using a signed `.app` with 22.10.3?
Everything is working correctly for `dmg` :)

lutzroeder:
>>@quanglam2807 @mmaietta have you successfully built a .dmg using a signed .app with 22.10.3?
>Everything is working correctly for dmg :)

Are you running on Big Sur or Catalina, does Terminal have full disk access? See #5322.

quanglam2807:
> > > @quanglam2807 @mmaietta have you successfully built a .dmg using a signed .app with 22.10.3?
> > > Everything is working correctly for dmg :)
> 
> Are you running on Big Sur or Catalina, does Terminal have full disk access? See #5322.

I build using GitHub Actions, it's probably Catalina.

quanglam2807:
@mmaietta Found it: https://github.com/electron-userland/electron-builder/blob/2b116d138e9effa3314fe6ff4474b639f7fddf47/packages/app-builder-lib/src/macPackager.ts#L241

castroCrea:
I renew my comment here #5484
> Did you check that it will work with electron-updater ?

larrywalters:
Has anyone confirmed that the `${arch}` variable works when building for both `x64` and `arm64`? In the config file, I use something like: `"artifactName": "App-${version}-BUILD-${os}-${arch}.${ext}"`, but nothing gets substituted for the variable.

lutzroeder:
@mmaietta is there a fix for the Travis CI failure in `macPackagerTest:two-package`?

mmaietta:
@lutzroeder unfortunately, I'm not too sure. Oddly, all 3 of my PRs had all tests passing prior to merging.
Running tests locally pass as well for me. AFAICT, it's simply the ordering of assets being built, which seems to be async enough that they're not matching depending on the node building it.

lutzroeder:
@mmaietta Makes sense. Previously this was `x64` only. Maybe the ordering issue was added by adding `arm64` support? Is it possible to add sorting to the test code before matching?

mmaietta:
Yeah, that's most likely. Oddly, it seems that it's already trying to:
https://github.com/electron-userland/electron-builder/blob/master/test/src/helpers/packTester.ts#L194

lutzroeder:
> Yeah, that's most likely. Oddly, it seems that it's already trying to:

The sorting happens at the `ArtifactCreated` level, the mismatch is under `fileContent.files`?

davej:
> Has anyone confirmed that the ${arch} variable works when building for both x64 and arm64? In the config file, I use something like: "artifactName": "App-${version}-BUILD-${os}-${arch}.${ext}", but nothing gets substituted for the variable. All the other variables work though. This is for building a pkg on Mac.

@larrywalters I'm seeing the same issue, my DMG artifact name is: `"${productName} ${version}-${arch}.${ext}"` but the output filename is "ProductName 1.0.0.dmg". This seems to be a long-standing issue and pre-dates this PR though.

Would be interesting to know if others are experiencing similar and if this is a known issue.

Not difficult to rename the file after Electron Builder has done its magic but I don't want to do anything if this will be "fixed" later.

davej:
@develar @mmaietta @lutzroeder @quanglam2807 I'm happy to try and PR for the issue above (broken artifact names for DMG) but before I jump in, could you verify that:
1. It is an issue that you can reproduce.
2. This is not "by design" and is actually a bug that you want fixed.

davej:
@develar @mmaietta @lutzroeder @quanglam2807 I'm happy to try and PR for the issue above (broken artifact names for DMG) but before I jump in, could you verify that:
1. It is an issue that you can reproduce.
2. This is not "by design" and is actually a bug that you want fixed.

larrywalters:
@davej Thanks for sharing, that's interesting because when I tried it with DMG it worked fine. In fact, it also seemed to work fine for all the zipped formats I tried, but not for PKG. Therefore, I concluded it was an issue specifically with PKG. In looking at the code, I see that the call to `expandArtifactNamePattern` when using PKG does not pass the architecture. I think that should be added to fix this. The architecture is passed in most other places where this function is called.
https://github.com/mmaietta/electron-builder/blob/07ff09503457fe10c84da8989e8d5da94aade803/packages/app-builder-lib/src/targets/pkg.ts#L37
So I don't know why it wouldn't be working for DMG.

davej:
@larrywalters Interesting. The zip outputs work fine for me too and I haven't tried pkg.
What does your config look like? Here's the relevant part of my config:
```js
{
  "dmg": {
    "artifactName": "${productName} ${version}-${arch}.${ext}"
  },
  "mac": {
    "artifactName": "${productName} ${version}-${arch}-${os}-update.${ext}",
    // ...
    "target": [
      {
        "target": "zip",
        "arch": [
          "x64",
          "arm64"
        ]
      },
      {
        "target": "dmg",
        "arch": [
          "x64",
          "arm64"
        ]
      }
    ],
  }
}
```

And my terminal output looks like this:
```
  • building        target=macOS zip arch=arm64 file=/var/folders/6y/gy9gggt14379c_k39vwb50lc0000gn/T/2001112vht2b9h7/dist/App 1.5.0-arm64-mac-update.zip
  • building        target=DMG arch=arm64 file=/var/folders/6y/gy9gggt14379c_k39vwb50lc0000gn/T/2001112vht2b9h7/dist/App 1.5.0.dmg
  • building block map  blockMapFile=/var/folders/6y/gy9gggt14379c_k39vwb50lc0000gn/T/2001112vht2b9h7/dist/App 1.5.0.dmg.blockmap
  • building embedded block map  file=/var/folders/6y/gy9gggt14379c_k39vwb50lc0000gn/T/2001112vht2b9h7/dist/App 1.5.0-arm64-mac-update.zip
```

lutzroeder:
@davej could this be related to [skipArchIfX64](https://github.com/electron-userland/electron-builder/blob/eec01f7ba9ce5b17dfcf84ecb237dd4872d6c880/packages/app-builder-lib/src/platformPackager.ts#L502)?

davej:
>@davej could this be related to skipArchIfX64?

@lutzroeder Maybe, but I would have thought would only affect the x64 output so I would have expected the following filenames:
```
x64 = "App 1.5.0.dmg"
arm64 = "App 1.5.0-arm64.dmg"
```

What do you think?

lutzroeder:
> Thank you, @mmaietta. It works. What about artificateName for mas builds (pkg)? The filenames don't have arch.

@mmaietta @quanglam2807 looks like [pkg.ts](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/targets/pkg.ts) changes are not included?

elliottkember:
I'm seeing something weird. My builds in Github Actions aren't adding the `arch` to the filename, and my users are getting the wrong builds. 

While I'm debugging that, I'm also finding that on the command-line, `electron-builder -m` doesn't actually produce the targets I want.

```
  • rebuilding native dependencies  dependencies=@serialport/bindings@9.0.3, keytar@5.6.0 platform=darwin arch=x64
  • install prebuilt binary  name=keytar version=5.6.0 platform=darwin arch=x64
  • install prebuilt binary  name=@serialport/bindings version=9.0.3 platform=darwin arch=x64
  • build native dependency from sources  name=@serialport/bindings
                                          version=9.0.3
                                          platform=darwin
                                          arch=x64
                                          reason=prebuild-install failed with error (run with env DEBUG=electron-builder to get more information)
                                          error=prebuild-install info begin Prebuild-install version 5.3.3
    prebuild-install info looking for cached prebuild @ /Users/elliott/.npm/_prebuilds/091534-bindings-v9.0.3-electron-v85-darwin-x64.tar.gz
    prebuild-install http request GET https://github.com/serialport/node-serialport/releases/download/@serialport/bindings@9.0.3/bindings-v9.0.3-electron-v85-darwin-x64.tar.gz
    prebuild-install http 404 https://github.com/serialport/node-serialport/releases/download/@serialport/bindings@9.0.3/bindings-v9.0.3-electron-v85-darwin-x64.tar.gz
    prebuild-install WARN install No prebuilt binaries found (target=11.0.3 runtime=electron arch=x64 libc= platform=darwin)

  • build native dependency from sources  name=keytar
                                          version=5.6.0
                                          platform=darwin
                                          arch=x64
                                          reason=prebuild-install failed with error (run with env DEBUG=electron-builder to get more information)
                                          error=prebuild-install info begin Prebuild-install version 5.3.3
    prebuild-install info looking for cached prebuild @ /Users/elliott/.npm/_prebuilds/1078bd-keytar-v5.6.0-electron-v85-darwin-x64.tar.gz
    prebuild-install http request GET https://github.com/atom/node-keytar/releases/download/v5.6.0/keytar-v5.6.0-electron-v85-darwin-x64.tar.gz
    prebuild-install http 404 https://github.com/atom/node-keytar/releases/download/v5.6.0/keytar-v5.6.0-electron-v85-darwin-x64.tar.gz
    prebuild-install WARN install No prebuilt binaries found (target=11.0.3 runtime=electron arch=x64 libc= platform=darwin)

  • rebuilding native dependency  name=@serialport/bindings version=9.0.3
  • rebuilding native dependency  name=keytar version=5.6.0
  • packaging       platform=darwin arch=x64 electron=11.0.3 appOutDir=dist/mac
✨  Done in 6.69s.
yarn run electron-builder -m  10.11s user 2.31s system 182% cpu 6.794 total
```

It gets to the end of `packaging` and decides it's done. `dist/mac` has an `Electron.app` in it that doesn't include my application code or run. Not quite sure what happened - I was running the earlier patched electron-builder code and replaced it with the latest from master branch.

quanglam2807:
> > Thank you, @mmaietta. It works. What about artificateName for mas builds (pkg)? The filenames don't have arch.
> 
> @mmaietta @quanglam2807 looks like [pkg.ts](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/targets/pkg.ts) changes are not included?

mas-related changes are here: https://github.com/electron-userland/electron-builder/pull/5484. It works fine for me (using it in production).

larrywalters:
@davej Please pardon the delay. I tried it for example with this configuration. It works for all the formats there except pkg. I have tried other zip formats in the past successfully too.
```
"artifactName": "Name-${version}-BUILD-${os}-${arch}.${ext}"
...
"mac": {
    "target": [
      {
        "target": "pkg",
        "arch": [
          "x64",
          "arm64"
        ]
      },
      {
        "target": "dmg",
        "arch": [
          "x64",
          "arm64"
        ]
      },
      {
        "target": "tar.gz",
        "arch": [
          "x64",
          "arm64"
        ]
      },
      {
        "target": "7z",
        "arch": [
          "x64",
          "arm64"
        ]
      }
    ]
}
```
The output was like below; only one pkg file appeared and it wrapped the arm64 version, I think suspect means one architecture overwrote the other.
```
Name-1.2.3-alpha.0-BUILD-mac.pkg
Name-1.2.3-alpha.0-BUILD-mac.tar.gz
Name-1.2.3-alpha.0-BUILD-mac-arm64.tar.gz
Name-1.2.3-alpha.0-BUILD-mac.dmg
Name-1.2.3-alpha.0-BUILD-mac-arm64.dmg
Name-1.2.3-alpha.0-BUILD-mac.7z
Name-1.2.3-alpha.0-BUILD-mac-arm64.7z
```
So I think we need that `pkg.ts` change I mentioned.

Issue #2894, opened 2018-05-10
Improve Windows installer task kill process (related to child processes)


tzarebczan:
* **Version**: 20.13.3

electron-updater: 2.21.4

* **Target**: Windows

The upgrade process was changed (https://github.com/electron-userland/electron-builder/issues/2516) to use taskkill which works great for us but we are having trouble when we run our installer while the app is running. We get the prompt to close LBRY, which occurs, but it leaves one of our grandchild  processes running (which is not good because we are trying to update that running file). 

Can the installer process be updated to use the same mechanism for killing processes as the upgrade process?  Haven't dug into the code to see what the differences are, but I'm assuming it's not the same because of the different behavior. 

Donating...

develar:
Both uses the same logic to kill. Let's investigate #2896 and then maybe this issue will gone.

tzarebczan:
I'll give the newer updater a shot and report back, but I don't expect this behavior to change (it's always worked like this).

tzarebczan:
@develar this still exists even though https://github.com/electron-userland/electron-builder/issues/2896 was fixed by upgrading the updater version. This is fairly easy to reproduce in our environment. When I run the installer while the app is running, I'm left with a single lbrynet-daemon.exe process (it starts with 2, a shell processes which launches the main one) that is not killed off. The in-app upgrade process kills both. 

Any thoughts? Let me know if I need to give you more info/example.

tzarebczan:
@develar any updates?

tzarebczan:
@develar can you take a look at this one again? Were there any changes recently which may have fixed it? Should be easy to reproduce.

tzarebczan:
@develar any updates?

tzarebczan:
@develar this is still happening on 20.22.0. This is how the app normally starts up:
![install1](https://user-images.githubusercontent.com/8120721/43423699-1541340e-941b-11e8-8494-b0bfc8cd1409.jpg)

When clicking OK, LBRY.exe related processes shut down, but a single lbrynet-daemon.exe stays open. The result is:
![install2](https://user-images.githubusercontent.com/8120721/43423728-2928d620-941b-11e8-8c88-01f0f83adda3.jpg)

This causes a few issues:
1) The lbrynet-daemon.exe files does not get updated because it's in use (we frequently update it). 
2) If a user closes the LBRY app now, lbrynet-daemon.exe will continue running
[...]

PR #5445, opened 2020-11-29
fix(nsis): don't kill ourselves when upgrading

mastergberry:
Windows `taskkill.exe` will kill the actual installer when using the `/t` option it appears after some vigorous testing.

```
C:\Windows\system32>taskkill /t /im "Badlion Client.exe" /fi "PID ne 62548"
ERROR: The process with PID 65300 (child process of PID 57508) could not be terminated.
Reason: This process can only be terminated forcefully (with /F option).
ERROR: The process with PID 43552 (child process of PID 57508) could not be terminated.
Reason: This process can only be terminated forcefully (with /F option).
ERROR: The process with PID 59664 (child process of PID 57508) could not be terminated.
Reason: This process can only be terminated forcefully (with /F option).
SUCCESS: Sent termination signal to process with PID 62548, child of PID 57508.
ERROR: The process with PID 57508 (child process of PID 7964) could not be terminated.
Reason: One or more child processes of this process were still running.

C:\Windows\system32>taskkill /t /f /im "Badlion Client.exe" /fi "PID ne 62548"
SUCCESS: The process with PID 65300 (child process of PID 57508) has been terminated.
SUCCESS: The process with PID 43552 (child process of PID 57508) has been terminated.
SUCCESS: The process with PID 59664 (child process of PID 57508) has been terminated.
SUCCESS: The process with PID 62548 (child process of PID 57508) has been terminated.
SUCCESS: The process with PID 57508 (child process of PID 7964) has been terminated.

C:\Windows\system32>taskkill /f /im "Badlion Client.exe" /fi "PID ne 39848"
SUCCESS: The process "Badlion Client.exe" with PID 65496 has been terminated.
SUCCESS: The process "Badlion Client.exe" with PID 65756 has been terminated.
SUCCESS: The process "Badlion Client.exe" with PID 65472 has been terminated.
SUCCESS: The process "Badlion Client.exe" with PID 63528 has been terminated.
```

The first attempt the nsis installer's PID was 62548, and the second was 39848. Notice that with the `/t` flag it tries to kill itself, and does not listen to the `/fi` filter option which should ignore itself. The `/t` flag seems to take precedence here. 

I know that this was originally done due to this issue: https://github.com/electron-userland/electron-builder/issues/2894 but the reality here is that the user application should be responsible to clean up any spawned children before running the updating software.

I actually currently have a bug myself that is preventing proper shut down which I will fix, but this second layer bug of the installer killing itself is also a problem that should be resolved.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
@@ -58,7 +58,7 @@
       DetailPrint `Closing running "${PRODUCT_NAME}"...`
 
       # https://github.com/electron-userland/electron-builder/issues/2516#issuecomment-372009092
-      nsExec::Exec `taskkill /t /im "${APP_EXECUTABLE_FILENAME}" /fi "PID ne $pid"` $R0
+      nsExec::Exec `taskkill /im "${APP_EXECUTABLE_FILENAME}" /fi "PID ne $pid"` $R0
       # to ensure that files are not "in-use"
       Sleep 300
 
@@ -66,7 +66,7 @@
       ${if} $R0 == 0
         # wait to give a chance to exit gracefully
         Sleep 1000
-        nsExec::Exec `taskkill /f /t /im "${APP_EXECUTABLE_FILENAME}" /fi "PID ne $pid"` $R0
+        nsExec::Exec `taskkill /f /im "${APP_EXECUTABLE_FILENAME}" /fi "PID ne $pid"` $R0
         ${If} $R0 != 0
           DetailPrint `Waiting for "${PRODUCT_NAME}" to close (taskkill exit code $R0).`
           Sleep 2000


develar:
@tzarebczan What do you think about removing `t` flag?

mastergberry:
Just to clarify here further @develar I spent a lot of time testing, and as shown above the `/fi` is just being ignored and therefore with the `/t` flag it just kills the actual installer itself (since it is a child).

The `/t` flag should not have been introduced if was known already, but like I mentioned proper cleanup of any child processes the application makes should be done on the developer's side before kicking off an update.

mastergberry:
Is it possible to get this into the next pre-release @develar? I wanted to test it out to verify it's all working good still :) 

Had some other fixes I wanted to grab also.

develar:
I released a new pre-rerelease version for you:

electron-builder: `22.10.0`
electron-updater: `4.3.6`

develar:
electron-updater: 4.3.7

mastergberry:
Just to follow up here, I spent an hour now testing and trying to break this. It's working 100% as intended now @develar 

The installer does not die anymore when force killing the original process. Thanks :)

tzarebczan:
We'll check if this breaks our upgrade process. If so, we may need to call a shutdown manually on our sub processes.

PR #5424, opened 2020-11-17
fix(dmg): build under Big Sur is not possible if license file is used

k2s:
Fixes #5319

In Big Sur the long time deprecated `Rez` and `hdiutil unflatten` tools were removed.
I am  injecting DMG license with [dmg-license](https://github.com/argv-minus-one/dmg-license) package by keeping compatibility with TXT/RTF license and buttons files.

In this first version I keep all the old code.

PR #5449, opened 2020-12-02
fix: compiler error and test updates

mmaietta:
Fixing compiler errors listed below and updated yarn.lock to appropriately track `master` branch's package.json. Tests needed to be updated since dmg-license does not have a default 'ru' label. (Origin https://github.com/electron-userland/electron-builder/pull/5424)
```
/Users/travis/build/electron-userland/electron-builder/packages/dmg-builder/src/dmgLicense.ts
  12:98  error  Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead  @typescript-eslint/ban-types
  37:9   error  'label' is never reassigned. Use 'const' instead     prefer-const
```

Fixing buildResources directory "default" for Linux icons to be a fully defined path and applied as a fallback resource (Origin https://github.com/electron-userland/electron-builder/pull/5385)

Updating snapTest snapshots to include new title entry (Origin https://github.com/electron-userland/electron-builder/pull/5350)


packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
```diff
@@ -61,13 +61,11 @@ export class LinuxTargetHelper {
       sources.push(icnsPath)
     }
 
-    // if no explicit sources are defined, default to buildResources directory
-    if (sources.length < 1) {
-      sources.push('./')
```
quanglam2807, line 66: Actually, I was the one who made this change. The reason why I didn't put `packager.buildResourcesDir` in `fallbackSources` is because in this case, it's not actually fallback (we just assume the dev wants to use the globally defined `buildResourcesDir`). I think fallback means `electron-builder` find no defined icons and opt to use Electron icon instead (Ref: https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/platformPackager.ts#L590).

quanglam2807 (changes requested):

packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
mmaietta: Hmmm, that log warning seems to be assuming that only one fallback source is provided.

It's possible to provide multiple fallback sources, would that be sequentially read? To provide the functionality: 1st. icns -> 2nd. build resources -> 3rd. framework icon

With the code as-is, the `snapTests` need fixing. :/

quanglam2807: I just checked more carefully, it seems like as long as `fallbackSources` is used, `isFallback = true`. Ref: https://github.com/develar/app-builder/blob/1cceb8fd6fe68955cecb96fb48259087980cd826/pkg/icons/icon-converter.go#L64

quanglam2807: > With the code as-is, the `snapTests` need fixing. :/

You mean https://github.com/electron-userland/electron-builder/pull/5385 breaks `snapTests`?

quanglam2807: In this case, I would propose to keep `buildResourcesDir` in `sources` and make a small change:
`sources.push('./')` -> `sources.push(packager.buildResourcesDir)`

mmaietta: Yes #5385 broke the test suite. Seems `linuxPackagerTest` and/or `snapTest` looking at my terminal history.

I'm confused why we can't use fallback sources? 
The `isFallback` (from what you've linked) is simply used in the code for logging purposes, after which the value is discarded.  Functionally, nothing behaves differently. We can update the logged message and/or make the boolean flag more explicit as to what source was fallen back to

I just tried `sources.push(packager.buildResourcesDir)` and it doesn't work due to other build errors as everything expects a icon `sources` path to always exist, whereas fallback sources seem to behave as optional.
```
 icon directory /private/var/folders/xv/37hyp31j7tv32txvjnnly0p1qccy3t/T/et-a34345c0e457c6fd5301936e13688492/t-O76XSJ/test-project-4/build doesn't contain icons
```

quanglam2807: +1 If so, then feel free to keep your changes. I still feel it's not what `fallbackSources` is intended to be. But since it works perfectly and you have worked so hard on all those changes, it's unnecessary to make things harder.

packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
```diff
@@ -61,11 +61,13 @@ export class LinuxTargetHelper {
       sources.push(icnsPath)
     }
 
-    // If no explicit sources are defined, fallback to buildResources directory, then default framework icon
-    const fallbackSources = [packager.buildResourcesDir, ...asArray(packager.getDefaultFrameworkIcon())]
+    // if no explicit sources are defined, default to buildResources directory
+    if (sources.length < 1) {
+      sources.push(packager.buildResourcesDir)
```
quanglam2807, line 66: Please revert this commit. I'm sorry for making bad suggestion.

quanglam2807 (approved):

mmaietta:
@quanglam2807 thanks for calling out the `buildResourcesDir`. 🙂  It actually was still a bug and I've updated the PR accordingly (needed to use `config.directories?.buildResources`) 

I was able to confirm that `fallbackSources` must be used, `sources` requires icon files to exist at the entry.  `fallbackSources` just requires the dir to exist, but no icons needed. Since `buildResources` may not be specified by the user (or simply not contain the icon files), we'll need it to be a fallback.

I also consolidated some logic with shorthand filtering null/undefined values and validating paths.

Could you please re-review?

Note: Apologies for any commit notifications, some of the tests only seem to run on the CI, it made it a bit difficult to debug initially.

mmaietta:
@develar all tests now pass. Ready for review 🙂

quanglam2807 (approved):

elliottkember:
@develar Bumping this one to fix tests on master, which is blocking us testing other PRs <3

develar:
You are awesome, guys, thanks for help.

ahmadwaliesipick:
@mmaietta  @develar  when you are going to release new version?

Slidemagic:
The license gets added correctly in pre-release V22.10.3 now (Big Sur). The builder is very picky now when it comes to UTF-8 encoding. It took me some time to remove the left-over messy characters from a lawyer's Microsoft Word file. Maybe there is a tool that can be built in to do the cleaning automatically. Thx for the update,.

Issue #5383, opened 2020-10-30
Code can't be signed when playwright is added as a dependency


SentoxAIO:
* **Version**: 22.9.1
* **Electron Version**: 10.1.5
* **Target**:  macOS

I basically have the same error as this [issue](https://github.com/electron-userland/electron-builder/issues/2010)

It's just a different module doing the same so ignoring **playwright-firefox/.local-browsers/** should do the job.

For publishing the update I use the GitHub Action of [electron-builder](https://github.com/samuelmeuli/action-electron-builder), not sure if something has to be changed there too.

isaacaddis:
Same issue here! Have you figured this out?

FoonkG:
Same issue here!

hyped0001:
Same issue

SentoxAIO:
Getting fixed by https://github.com/electron-userland/electron-builder/pull/5451

PR #5451, opened 2020-12-02
fix: signing of playwright/playwright-firefox

SentoxAIO:
This PR is based on https://github.com/electron-userland/electron-builder/issues/5383

Now “/node_modules/playwright-firefox/.local-browsers” and “/node_modules/playwright/.local-browsers” also getting ignored like “/node_modules/puppeteer/.local-chromium”.


nwittwer:
@SentoxAIO Could you share your `builder.config.js`? Does this work out of the box now or are there some special settings like `asarUnpack` and `signIgnore` that are required in order to fix the codesign issues around Playwright's `.local-browsers`? I'm still hitting issues when trying to notarize a Mac app (#5500).

Issue #5431, opened 2020-11-19
hdiutil exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE


EdmarPereira:
I'm not able to generate builds for MacOs, I've made several attempts to change versions and the error persists:

`  ⨯ hdiutil exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE  stackTrace=
                                                                    Error: hdiutil exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
                                                                        at ChildProcess.<anonymous> (/.../node_modules/builder-util/src/util.ts:243:14)
                                                                        at Object.onceWrapper (events.js:417:26)
                                                                        at ChildProcess.emit (events.js:310:20)
                                                                        at maybeClose (internal/child_process.js:1021:16)
                                                                        at Process.ChildProcess._handle.onexit (internal/child_process.js:286:5)`

I did the tests with the following combinations
electron-builder: 20.36.2
electron: 3.0.9

electron-builder: 21.1.6
electron: 7.15.0

electron-builder: 21.1.5
electron: 5.0.8

and also with the latest version for both
electron-builder: 22.9.1
electron: 11.0.2

All tests were done using a Macbook Air Catalina version 10.15.7 and Node version 12.16.3. Sometimes I have some success using debug mode with `DEBUG=electron-builder node_modules/.bin/electron-builder build -m`
Additional information: I do not use any kind of antivirus.

Please help me, I create a big project but in every version released I have problems building versions for MacOs.

ahmetuysal:
I have the same issue.

btzsoft:
having the same issue, also it says:

```
Error output:
- Searching dependency tree
rm: ./Release/.deps/Release/obj.target/validation/src/validation.o.d.raw: No such file or directory
make: *** [Release/obj.target/validation/src/validation.o] Error 1
✖ Rebuild Failed
An unhandled error occurred inside electron-rebuild
node-gyp failed t
```

EdmarPereira:
I ask for help because in the company where we work we have created a very large project using electron, if we cannot solve this problem we will have to look for another option. I don't know what else to do, the only problem we found is to generate builds for MacOs, we bought a Macbook just for that, we don't have anything else installed to avoid conflicts, please if anyone has a solution for this I would really appreciate it.
Thank you.

mmaietta:
What version of xcode are you guys using? What version of python?

[...]

PR #5464, opened 2020-12-10
fix(mac): Workaround for hdiutil randomly failing (#5431)

chokladmal:
For some but not all Mac users building dmg files often fails.

After some investigations the cause seems to be that `hdiutil create` (when being run to create a temporary dmg file in createStageDmg) fail when trying to unmount the created volume due to "resource busy". As this only affects some Macs it seems to be caused by something on those machines. Suggested is antivirus software but that's not confirmed.

It turns out that after retrying the `hdiutil create` command a few times it usually succeeds and this workaround is based on that idea - to simply retry the command up to five times before throwing and failing the build. This does not fix the real issue!

Tracking issue: https://github.com/electron-userland/electron-builder/issues/5431


mmaietta (changes requested):
I'm not sure on what could be the actual solution here, as the random failures seem to be more OS-related from what I've searched and my personal experience.

As an improvement here, I'd recommend using `builder-util` `retry<T>` function. 
https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util/src/util.ts#L384
Just need to export it and it simplifies all your usage to 2 lines. 

Related notes: This may be misusing the logger's debug flag, but you could also conditionally add `-debug` since the retry function doesn't have any explicit logging info. I also increased duration to 1s as sometimes I had to wait that long before a retry when I had my own hdiutil bash scripts

```
if (log.isDebugEnabled) {
  imageArgs.push("-debug")
}
await retry(spawn("hdiutil", imageArgs), 5, 1000)
return tempDmg
```

BelinChung:
It seems to be the problem caused by this line
```
// imageArgs.push("-fs", "HFS+", "-fsargs", "-c c=64,a=16,e=16")
```
When I commented it out, the dmg package was successfully built on Big Sur.

mmaietta:
Out of curiosity, can you try just omitting the `-fsargs < >`? I'm wondering if all we need is
```
imageArgs.push("-fs", "HFS+")
```

From the man:
```
-fsargs newfs_args
                          additional arguments to pass to whatever newfs pro-
                          gram is implied by -fs.  newfs_hfs(8) has a number
                          of options that can reduce the amount of space
                          needed by the filesystem's data structures.  Sup-
                          pressing the journal with -fs HFS+ and passing argu-
                          ments such as -c c=64,a=16,e=16 to -fsargs will min-
                          imize gaps at the front of the filesystem, allowing
                          resize to squeeze more space from the filesystem.
                          For truly optimal filesystems, use makehybrid.
```

chokladmal:
Thank you for your comments! I really appreciate that you're helping out with this :)
As I mentioned in the issue thread I've tried running hdiutils with -attach with some success. It doesn't fail but it will have to be unmounted at a later stage instead. Right now I have somewhat limited time but I will keep trying, especially the suggested changes to the args.

lutzroeder:
@chokladmal did you track down the actual error by running with `export DEBUG_DMG=true` and repro by running this command in isolation. Which version of electron-builder and macOS are you running and does the Terminal have full disk access? This seems to be a bit too much "no idea why this is happening but a for loop might hide it"? Is there some effort to diagnose the underlying issue and drive a fix?

chokladmal:
@lutzroeder I saw that you had already commented in the issue but for the record it's this one https://github.com/electron-userland/electron-builder/issues/5431

I'm running electron-builder 22.9.1 and macOS Catalina 10.15.17

In https://github.com/electron-userland/electron-builder/issues/5431 we've narrowed it down to hdiutil failing due to "resource busy" when hdiutil create tries to unmount the volume.

chokladmal:
> Out of curiosity, can you try just omitting the `-fsargs < >`? I'm wondering if all we need is
> 
> ```
> imageArgs.push("-fs", "HFS+")
> ```
> 
> From the man:
> 
> ```
> -fsargs newfs_args
>                           additional arguments to pass to whatever newfs pro-
>                           gram is implied by -fs.  newfs_hfs(8) has a number
>                           of options that can reduce the amount of space
>                           needed by the filesystem's data structures.  Sup-
>                           pressing the journal with -fs HFS+ and passing argu-
>                           ments such as -c c=64,a=16,e=16 to -fsargs will min-
>                           imize gaps at the front of the filesystem, allowing
>                           resize to squeeze more space from the filesystem.
>                           For truly optimal filesystems, use makehybrid.
> ```

I've tried it. At first it seemed to be working but it finally failed with the same 
```

DIHLDiskImageCreate() returned 49168
hdiutil: create failed - Resource busy
```
after ten or so tries.

mmaietta:
This should be fixed (from a PR revert) in the latest prerelease cut `22.10.4`

lutzroeder:
@mmaietta this might be a different issue? The reverted PR was a recent change and repros reliably. This issue sounds like it succeeds on retry, shows a different error and also happened on previous versions? Next steps are probably either finding a reliable repro and tracking down why this is happening.

chokladmal:
If it is antivirus or something else maybe having something imitating that
behaviour? A script opening files on the volume as soon as it's mounted and
files are added could perhaps trigger the problem. I can try it out once
I'm back at my company laptop after the holidays.

lutzroeder:
@chokladmal, yes, the goal is to root cause the issue, less how we can recreate a behavior that would require adding a retry approach to the build pipeline. Masking fragile behavior that isn't well understood tends to lead more fragile and unpredictable behavior. For example, if this is caused by a virus scanner, the right fix is probably to suspend the scanner during build, add an exclusion or work with the vendor. If the issue is a bug in `hdiutil` the right fix would be fixing `hdiutil`.

shrpne:
@lutzroeder can it be released as a workaround to fix current issue and then be replaced with real fix, when it ready?

Because for now there is no workaround. I even can't install chokladmal's fork because it is monorepo. And since it is typescript it requires a lot of tooling to apply his patch to my own project.

I see no problems with releasing it. It should not cause performance issues or bugs to other users. The only problem with it, it is not beautiful, but it works.

mmaietta:
@shrpne here's a script that @lutzroeder put together for compiling any branch and patching the results. Run from your current app directory (with electron-builder one dir up. You can adjust paths as needed)
```
#!/bin/bash

pushd ../electron-builder
yarn compile
popd

rsync --recursive ../electron-builder/packages/app-builder-lib/out/ ./node_modules/app-builder-lib/out/
rsync --recursive ../electron-builder/packages/app-builder-lib/electron-osx-sign/ ./node_modules/app-builder-lib/electron-osx-sign/
rsync --recursive ../electron-builder/packages/builder-util/out/ ./node_modules/builder-util/out/
rsync --recursive ../electron-builder/packages/dmg-builder/out/ ./node_modules/dmg-builder/out/

cp ../electron-builder/packages/app-builder-lib/scheme.json ./node_modules/app-builder-lib/scheme.json

yarn patch-package app-builder-lib
yarn patch-package builder-util
yarn patch-package dmg-builder
```

lutzroeder:
@shrpne not my call. This is an open source project so the challenge isn't shipping workarounds. The question is more who would own getting to the bottom of this and drive the right fix.

lutzroeder:
@mmaietta the script looks like it got merged from different sources. Why is `patch-package` is needed?

mmaietta:
@lutzroeder `patch-package` allows changes to be persisted in the project repo as a `.patch` file to save your local changes.

shrpne:
@mmaietta thank you for sharing this script, but it still looks pretty complicated, as I understand it is can't be used as a drop-in solution.

@lutzroeder honestly I don't understand, why opensource project can't accept temporary workaround, while waiting for real fix, which then will replace this workaround. 
From my point of view, working opensource project with temporary workaround is better than temporary not working opensource project without workarounds.

lutzroeder:
@shrpne sure, not arguing that and it’s not my call to make for this repo. Why not come up with the real fix now? Are you signing up to get to the bottom of this later? What is going to change to make this easier when folks are hitting the issue less often or in more random circumstances due to the workaround masking the real issue. I’m more asking why the problem isn’t better understood and what can be done about that. You are asking for the cheapest way to make it not your problem.

jzybert:
Which version of electron will this be included in?

mmaietta:
Please update to latest electron-builder@22.10.5

PR #5473, opened 2020-12-14
fix(dmg-builder): fix dmg-license is incompatible with linux and win32

quanglam2807:
`dmg-license` `package.json` sets that the module is only compatible with mac. This breaks `yarn install` on `linux` and `win32`. This PR moves `dmg-license` to `optionalDependencies` so `yarn` can ignore the error.

Screen Shot 2020-12-14 at 7 51 38 PM


VitoBeijing:
How do I package dmg on Linux？

Issue #5475, opened 2020-12-14
Apple Silicon Universal Support


lutzroeder:
Tracking Universal support from electron-userland/electron-builder#5426.

See also electron/electron#26710 and [electron/universal](https://github.com/electron/universal).

@mmaietta @quanglam2807 @elliottkember @devlar

mmaietta:
Getting the universal package seems to work quite simply. 
```
const universal = require('@electron/universal');
const path = require('path');

universal.makeUniversalApp({
  x64AppPath: path.resolve('./dist/artifacts/mac/Electron.app'),
  arm64AppPath: path.resolve('./dist/artifacts/mac-arm64/Electron.app'),
  outAppPath: path.resolve('./dist/artifacts/mac-universal/Electron.app'),
});
```
Native addons are spliced together into a "fat" file.

Two things I don't know yet:
- Do we need to resign the universal app?
- Where do we insert this in EB? `macPackager.ts`? Everything I've seen so far is heavily tied to an `Arch` mapped to Target(s). This location seems ideal but we need all the output directories provided, and we'd want to abstract the functionality back to macPackager as well I think.
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/packager.ts#L425-L441
Any thoughts/input on where to start with incorporating `makeUniversalApp` from `@electron/universal`?

lutzroeder:
@mmaietta @develar How would this be initiated at command line? `npx electron-builder --x86 --arm64` builds separate x64 and arm64 targets. Would it make sense add another `Arch` called `--universal` or `--x64_arm64`? The goal would be to avoid building multiple `.dmg` or `.zip` files or signing before the merged binary is created. Does building the executables exist as separate step?

mmaietta:
There's actually already an `all` available via the programmatic API, so I'd imagine we could reuse that here.
https://github.com/electron-userland/electron-builder/blob/master/test/src/mac/macPackagerTest.ts#L9

Universal apps still require an x64 and arm64 build. I think electron-updater may require those separate builds as well, as opposed to the larger initial universal binary, I honestly don't know because I don't use it. It would make sense though from a user-perspective as they should only need to be downloading an arch-specific package. 

My plan is to just check if both mac x64 and arm64 builds are created, if so, also create a universal app.  From my testing, the universal app "stitching" together took only a few seconds to complete, so the impact should be minimal. We just need to figure out the resign and an additional dmg/zip packing.

[...]

PR #5481, opened 2020-12-16
feat: Apple Silicon Universal Support

lutzroeder:
#5475

@mmaietta

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -18,6 +18,7 @@ import { PkgTarget, prepareProductBuildArgs } from "./targets/pkg"
 import { createCommonTarget, NoOpTarget } from "./targets/targetFactory"
 import { isMacOsHighSierra } from "./util/macosVersion"
 import { getTemplatePath } from "./util/pathManager"
+import { promisify } from "util"
 
 export default class MacPackager extends PlatformPackager<MacConfiguration> {
   readonly codeSigningInfo = new Lazy<CodeSigningInfo>(() => {
@@ -87,6 +88,42 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
         default:
           mapper(name, outDir => name === "mas" || name === "mas-dev" ? new NoOpTarget(name) : createCommonTarget(name, outDir, this))
           break
+      }
+    }
+  }
+
+  protected async doPack(outDir: string, appOutDir: string, platformName: ElectronPlatformName, arch: Arch, platformSpecificBuildOptions: MacConfiguration, targets: Array<Target>): Promise<any> {
+    switch (arch) {
+      default: {
+        return super.doPack(outDir, appOutDir, this.platform.nodeName as ElectronPlatformName, arch, this.platformSpecificBuildOptions, targets);
+      }
+      case Arch.universal: {
+        const x64Arch = Arch.x64;
+        const x64AppOutDir = appOutDir + '-' + Arch[x64Arch];
+        await super.doPack(outDir, x64AppOutDir, platformName, x64Arch, platformSpecificBuildOptions, targets, false);
+        const arm64Arch = Arch.arm64;
+        const arm64AppOutPath = appOutDir + '-' + Arch[arm64Arch];
+        await super.doPack(outDir, arm64AppOutPath, platformName, arm64Arch, platformSpecificBuildOptions, targets, false);
+        const framework = this.info.framework
+        log.info({
+          platform: platformName,
+          arch: Arch[arch],
+          [`${framework.name}`]: framework.version,
+          appOutDir: log.filePath(appOutDir),
+        }, `packaging`)
+        const appFile = `${this.appInfo.productFilename}.app`;
+        const { makeUniversalApp } = require('@electron/universal');
+        await makeUniversalApp({
+          x64AppPath: path.join(x64AppOutDir, appFile),
+          arm64AppPath: path.join(arm64AppOutPath, appFile),
+          outAppPath: path.join(appOutDir, appFile),
+          force: true
+        });
+        const rmdir = promisify(require('fs').rmdir);
+        await rmdir(x64AppOutDir, { recursive: true });
+        await rmdir(arm64AppOutPath, { recursive: true });
+        await this.doSignAfterPack(outDir, appOutDir, platformName, arch, platformSpecificBuildOptions, targets);
+        break;
       }
     }
   }
--- a/packages/app-builder-lib/src/packager.ts
+++ b/packages/app-builder-lib/src/packager.ts
@@ -427,12 +427,6 @@ export class Packager {
           break
         }
 
-        await this.installAppDependencies(platform, arch)
-
-        if (this.cancellationToken.cancelled) {
-          break
-        }
-
         // support os and arch macro in output value
         const outDir = path.resolve(this.projectDir, packager.expandMacro(this._configuration!!.directories!!.output!!, Arch[arch]))
         const targetList = createTargets(nameToTarget, targetNames.length === 0 ? packager.defaultTarget : targetNames, outDir, packager)
@@ -477,7 +471,7 @@ export class Packager {
     }
   }
 
-  private async installAppDependencies(platform: Platform, arch: Arch): Promise<any> {
+  public async installAppDependencies(platform: Platform, arch: Arch): Promise<any> {
     if (this.options.prepackaged != null || this.framework.isNpmRebuildRequired !== true) {
       return
     }
--- a/packages/app-builder-lib/package.json
+++ b/packages/app-builder-lib/package.json
@@ -44,6 +44,7 @@
   "dependencies": {
     "7zip-bin": "~5.0.3",
     "@develar/schema-utils": "~2.6.5",
+    "@electron/universal": "1.0.4",
     "async-exit-hook": "^2.0.1",
     "bluebird-lst": "^1.0.9",
     "builder-util": "workspace:*",
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -4960,6 +4960,7 @@
                   "arm64",
                   "armv7l",
                   "ia32",
+                  "universal",
                   "x64"
                 ],
                 "type": "string"
@@ -4971,6 +4972,7 @@
                 "arm64",
                 "armv7l",
                 "ia32",
+                "universal",
                 "x64"
               ],
               "type": "string"
--- a/packages/app-builder-lib/src/platformPackager.ts
+++ b/packages/app-builder-lib/src/platformPackager.ts
@@ -158,8 +158,18 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions> 
     }
   }
 
-  protected async doPack(outDir: string, appOutDir: string, platformName: ElectronPlatformName, arch: Arch, platformSpecificBuildOptions: DC, targets: Array<Target>) {
+  protected async doPack(outDir: string, appOutDir: string, platformName: ElectronPlatformName, arch: Arch, platformSpecificBuildOptions: DC, targets: Array<Target>, sign: boolean = true) {
     if (this.packagerOptions.prepackaged != null) {
+      return
+    }
+
+    if (this.info.cancellationToken.cancelled) {
+      return
+    }
+
+    await this.info.installAppDependencies(this.platform, arch);
+
+    if (this.info.cancellationToken.cancelled) {
       return
     }
 
@@ -241,11 +251,26 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions> 
 
     const isAsar = asarOptions != null
     await this.sanityCheckPackage(appOutDir, isAsar, framework)
-    await this.signApp(packContext, isAsar)
+    if (sign) {
+      await this.doSignAfterPack(outDir, appOutDir, platformName, arch, platformSpecificBuildOptions, targets);
+    }
+  }
 
-    const afterSign = resolveFunction(this.config.afterSign, "afterSign")
+  protected async doSignAfterPack(outDir: string, appOutDir: string, platformName: ElectronPlatformName, arch: Arch, platformSpecificBuildOptions: DC, targets: Array<Target>) {
+    const asarOptions = await this.computeAsarOptions(platformSpecificBuildOptions);
+    const isAsar = asarOptions != null;
+    const packContext = {
+      appOutDir,
+      outDir,
+      arch,
+      targets,
+      packager: this,
+      electronPlatformName: platformName
+    };
+    await this.signApp(packContext, isAsar);
+    const afterSign = resolveFunction(this.config.afterSign, "afterSign");
     if (afterSign != null) {
-      await Promise.resolve(afterSign(packContext))
+      await Promise.resolve(afterSign(packContext));
     }
   }
 
--- a/packages/builder-util/src/arch.ts
+++ b/packages/builder-util/src/arch.ts
@@ -1,8 +1,8 @@
 export enum Arch {
-  ia32, x64, armv7l, arm64
+  ia32, x64, armv7l, arm64, universal
 }
 
-export type ArchType = "x64" | "ia32" | "armv7l" | "arm64"
+export type ArchType = "x64" | "ia32" | "armv7l" | "arm64" | "universal"
 
 export function toLinuxArchString(arch: Arch, targetName: string): string {
   switch (arch) {
@@ -38,7 +38,8 @@ export function archFromString(name: string): Arch {
       return Arch.arm64
     case "armv7l":
       return Arch.armv7l
-
+    case "universal":
+      return Arch.universal
     default:
       throw new Error(`Unsupported arch ${name}`)
   }
--- a/packages/electron-builder/src/builder.ts
+++ b/packages/electron-builder/src/builder.ts
@@ -19,6 +19,7 @@ export interface CliOptions extends PackagerOptions, PublishOptions {
   ia32?: boolean
   armv7l?: boolean
   arm64?: boolean
+  universal?: boolean
 
   dir?: boolean
 }
@@ -45,6 +46,9 @@ export function normalizeOptions(args: CliOptions): BuildOptions {
       }
       if (args.ia32) {
         result.push(Arch.ia32)
+      }
+      if (args.universal) {
+        result.push(Arch.universal)
       }
 
       return result.length === 0 && currentIfNotSpecified ? [archFromString(process.arch)] : result
@@ -120,6 +124,7 @@ export function normalizeOptions(args: CliOptions): BuildOptions {
   delete result.x64
   delete result.armv7l
   delete result.arm64
+  delete result.universal
 
   let config = result.config
 
@@ -198,7 +203,7 @@ export function coerceTypes(host: any): any {
 export function createTargets(platforms: Array<Platform>, type?: string | null, arch?: string | null): Map<Platform, Map<Arch, Array<string>>> {
   const targets = new Map<Platform, Map<Arch, Array<string>>>()
   for (const platform of platforms) {
-    const archs = (arch === "all" ? (platform === Platform.MAC ? [Arch.x64, Arch.arm64] : [Arch.x64, Arch.ia32]) : [archFromString(arch == null ? process.arch : arch)])
+    const archs = (arch === "all" ? (platform === Platform.MAC ? [Arch.x64, Arch.arm64, Arch.universal] : [Arch.x64, Arch.ia32]) : [archFromString(arch == null ? process.arch : arch)])
     const archToType = new Map<Arch, Array<string>>()
     targets.set(platform, archToType)
 
@@ -257,6 +262,11 @@ export function configureBuildCommand(yargs: yargs.Argv): yargs.Argv {
     .option("arm64", {
       group: buildGroup,
       description: "Build for arm64",
+      type: "boolean",
+    })
+    .option("universal", {
+      group: buildGroup,
+      description: "Build for universal",
       type: "boolean",
     })
     .option("dir", {
--- a/test/snapshots/mac/macPackagerTest.js.snap
+++ b/test/snapshots/mac/macPackagerTest.js.snap
@@ -146,6 +146,21 @@ Object {
       },
     },
     Object {
+      "arch": "universal",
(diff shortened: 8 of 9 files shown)


mmaietta (commented):
With the introduction of `--universal` as an arch, I think this will break as node-gyp won't understand `universal`.
```
await this.installAppDependencies(platform, arch)
```
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/packager.ts#L430

It can process `--arch=all`, however, that doesn't work for me on Mac and it just produces the system's `process.arch`.
```
$ electron-builder node-gyp-rebuild --arch universal

Rebuild own native code

Options:
  --version   Show version number                                      [boolean]
  --help      Show help                                                [boolean]
  --platform  The target platform
                       [choices: "linux", "darwin", "win32"] [default: "darwin"]
  --arch      The target arch
             [choices: "ia32", "x64", "armv7l", "arm64", "all"] [default: "x64"]

Invalid values:
  Argument: arch, Given: "universal", Choices: "ia32", "x64", "armv7l", "arm64", "all"
error Command failed with exit code 1.
```

[UPDATE]
I'm starting to think we should just add a property specific to the builder config, like below. Partly because the packager logic is too tied to a specific architecture, and "universal" is technically not an architecture, but rather just combining/stitching assets into a fat binary.
```
    mac: {
        identity: <non-null>,
        universal: true,
        target: {
            target: 'dir',
            arch: [
                'arm64',
                'x64', 
            ]
        },
```
In taking this approach, the `target` field corresponds directly to an arch AND signs it. `universal` just validates that both arm64 and x64 are built, but doesn't force each individually to get signed, only the final fat app.

packages/app-builder-lib/src/macPackager.ts
```diff
@@ -91,6 +91,58 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
...
+                };
+                return Promise.all([
+                  rmdir(x64AppOutDir, { recursive: true }),
+                  rmdir(arm64AppOutPath, { recursive: true }),
+                ]).then(() => {
+                  return this.doSignAfterPack(outDir, appOutDir, platformName, arch, platformSpecificBuildOptions, targets);
+                });
+              });
+            });
+        break;
```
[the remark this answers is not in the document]
lutzroeder: That's not how this is being used in practice. Universal exists to avoid the existence of platform specific builds, not to bring them back in a hidden way or special case updaters. If an app builds universal, it would not build or ship x86 or arm64 targets.

Using a flag, where in the pipeline would the universal code run? It would ultimately have to do all the same things a regular MacPackager does and hide the existence of the x86 and arm64 build results. 

Is there a sample project or repro for the node-gyp-rebuild issue?

I can abandon this PR if you want to start a new one if there are clear answers but sounds like it isn't all that simple and more trading one set of problems for another?

quanglam2807:
I have a better idea for this. Why don’t we introduce an option in build config. Something like `combine: true`. And if that is set, we will combine the arch-specific files together with lipo. 

I also agree with @mmaietta that universal binaries are only useful for initial downloads. For updates, `electron-updater` should just download arch-specific files.

If you go with this design, it will be much simpler as only need to add a hook to run after arch-specific builds are ready.

mmaietta:
> If you go with this design, it will be much simpler as only need to add a hook to run after arch-specific builds are ready.

Perhaps a hook over here? https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/packager.ts#L442
The only difficulty I'm running into is how to `map` `packager.pack(...) => Promise<string>` so that we return a target-arch-outputpath map/object of assets to lipo after (it doesn't return anything right now). Then it'd be easy to just pass that into the hook for `postProcessPackages(...)`

lutzroeder:
@quanglam2807 @mmaietta to get on the same page on this, there aren't apps out there shipping a universal app and then upgrading to a platform specific version. Download and update any universal app, run an update and the result will be universal. With the "flag" approach the platform specific targets would need to be signed, notarized and published (on GitHub Releases) which would make CI pipelines more fragile and add to complexity (now there are 3 targets to choose from which we tried to avoid, users at different stages might rund a different version of the app). I'm not trying to argue which sets of pros and cons are preferable, just this is how it seems to be done by pretty much all app vendors so the goal is to optimize for this model.

lutzroeder:
> Perhaps a hook over here? https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/packager.ts#L442

What would this hook do? Create another MacPackager that doesn't build but can take inputs and `universal` architecture to create targets, then special case MacPackager so it doesn't sign or produce targets for `x64` and `arm64` if `universal` is set? Might be easier to have another PR to have a look?

mmaietta:
Thanks so much for tackling this @lutzroeder! The work is fantastic and covers almost all the use cases. The main issue lies with native modules as `installAppDependencies` must be called prior to any packing and per-architecture. `universal` unfortunately isn't accepted by node-gyp itself (`all` is though). I think we shouldn't move `installAppDependencies` into `doPack` as it convolutes the build stage's description of just "packing" (I suppose a simple rename could change that explicit description though, perhaps `buildPackage`?)
I think moving `installAppDependencies` here might solve the issue such that `doPack` always installs/rebuilds the correct arch's native addons: https://github.com/electron-userland/electron-builder/blob/70708a470ba148787c110be37c482e12c52e9351/packages/app-builder-lib/src/platformPackager.ts#L161-L162
Haven't tested that locally yet, will need to stash my current investigations.

packages/app-builder-lib/src/macPackager.ts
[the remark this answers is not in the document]
mmaietta: Sorry for the delay on this one.
You're completely correct that the approachs will be trading one set of problems for another. 😅 

Quick note: `electron-updater` currently expects to find an arch-specific build. 
https://github.com/electron-userland/electron-builder/blob/2b116d138e9effa3314fe6ff4474b639f7fddf47/packages/electron-updater/src/MacUpdater.ts#L34-L35
After much time for users to upgrade to M1, the universal binaries will be dropped anyhow, and builds will just need to include arm64, which electron-updater will already be ready to handle.

lutzroeder:
@mmaietta updated to call `installAppDependencies` from `doPack`:

```
~/electron-quick-start: npx electron-builder --mac --universal
  • electron-builder  version=22.10.0 os=20.2.0
  • rebuilding native dependencies  dependencies=leveldown@5.6.0 platform=darwin arch=x64
  • packaging       platform=darwin arch=x64 electron=11.1.0 appOutDir=dist/mac-universal-x64
  • rebuilding native dependencies  dependencies=leveldown@5.6.0 platform=darwin arch=arm64
  • packaging       platform=darwin arch=arm64 electron=11.1.0 appOutDir=dist/mac-universal-arm64
  • packaging       platform=darwin arch=universal electron=11.1.0 appOutDir=dist/mac-universal
```

mmaietta:
Nice! Those logs look perfect.

For the interim, I think this is a good way to approach it without very large changes for a single hook. A larger refactor without an explicit need for a post-all-packed hook is just asking for bugs to be created. I'm still wary of the "universal" as its own arch as anyone building all 3 archs (x64, arm64, universal) with native addons rebuilding may take a very long time. Oh well 😅 

I gave it a whirl locally and it works perfect on the first run. lipo results:
```
$ find dist/artifacts build -name '*.node' | xargs -I {} lipo -info "{}"
Architectures in the fat file: dist/artifacts/local/mac-universal/Electron.app/Contents/MacOS/app.node are: x86_64 arm64 
```
It does throw an error on subsequent builds that is an easy fix:
```
  • packaging       platform=darwin arch=arm64 electron=11.0.3 appOutDir=dist/artifacts/local/mac-universal-arm64
  • packaging       platform=darwin arch=universal electron=11.0.3 appOutDir=dist/artifacts/local/mac-universal
  ⨯ The out path "/shell-builder/dist/artifacts/local/mac-universal/Electron.app" already exists and force is not set to true 
```
Just need to add `force: true` to the `makeUniversalApp` config to enable overwriting

The only remaining step AFAICT is to get this working with the electron-builder config so that this works:
```
    mac: {
        target: {
            target: 'dir',
            arch: 'universal'
        },
```

lutzroeder:
> The only remaining step AFAICT is to get this working with the electron-builder config so that this works

The config gives `Invalid configuration object. electron-builder 22.10.0 has been initialized using a configuration object that does not match the API schema.`?

mmaietta:
Yep! You'll need to update the scheme.json. Just gotta add `"universal",` here:
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/scheme.json#L4955-L4980

Now, how can we fix that test?

lutzroeder:
> Now, how can we fix that test?

That should probably be a separate PR as it happens for blank PRs like #5488 as well. We should fix the test first and merge vs. including it in this PR which has higher risk to be reverted. Can you have a look?

mmaietta:
Your tests pass here!
But locally they fail for me, I get:
```
Snapshot name: `two-package 1`

    - Snapshot  - 3
    + Received  + 3

    @@ -68,21 +68,21 @@
                  "url": "TestApp-1.1.0-mac-universal.zip",
                },
                Object {
                  "sha512": "@sha512",
                  "size": "@size",
    -             "url": "TestApp-1.1.0-mac.dmg",
    +             "url": "TestApp-1.1.0-mac-universal.dmg",
                },
                Object {
                  "sha512": "@sha512",
                  "size": "@size",
    -             "url": "TestApp-1.1.0-mac-arm64.dmg",
    +             "url": "TestApp-1.1.0-mac.dmg",
                },
                Object {
                  "sha512": "@sha512",
                  "size": "@size",
    -             "url": "TestApp-1.1.0-mac-universal.dmg",
    +             "url": "TestApp-1.1.0-mac-arm64.dmg",
                },
              ],
              "path": "TestApp-1.1.0-mac.zip",
              "releaseDate": "@releaseDate",
              "sha512": "@sha512",

      235 |   }
      236 | 
    > 237 |   expect(objectToCompare).toMatchSnapshot()
```

I feel like something is wrong with the build node as these results are way too inconsistent.
Yours fllows this order from packages/electron-builder/src/builder.ts `[Arch.x64, Arch.arm64, Arch.universal]`, whereas mine is in some weird format.

lutzroeder:
@mmaietta not sure how this explains why #5488 is failing? This PR doesn't reorder `x64` and `arm64` and other PRs are failing as well.

quanglam2807:
Thank you so much, @lutzroeder. I'm so excited for this. I just realize Mac App Store only accepts a single `pkg` file, so universal binary is a must. Hope this PR can be merged soon so we can then test in production.

packages/app-builder-lib/src/packager.ts
```diff
@@ -477,7 +471,7 @@ export class Packager {
     }
   }
 
-  private async installAppDependencies(platform: Platform, arch: Arch): Promise<any> {
+  public async installAppDependencies(platform: Platform, arch: Arch): Promise<any> {
```
mmaietta, line 474: Minor: Since `installAppDependencies` is isolated to `doPack` now, can we move this to `platformPackager` and use `protected` instead of `public`?

lutzroeder:
@mmaietta @develar can we merge this PR and #5488 and for another round of prerelease testing?

packages/app-builder-lib/src/macPackager.ts
```diff
@@ -91,6 +92,42 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
...
+        await makeUniversalApp({
+          x64AppPath: path.join(x64AppOutDir, appFile),
+          arm64AppPath: path.join(arm64AppOutPath, appFile),
+          outAppPath: path.join(appOutDir, appFile),
+          force: true
+        });
+        const rmdir = promisify(require('fs').rmdir);
+        await rmdir(x64AppOutDir, { recursive: true });
+        await rmdir(arm64AppOutPath, { recursive: true });
+        await this.doSignAfterPack(outDir, appOutDir, platformName, arch, platformSpecificBuildOptions, targets);
```
mmaietta, line 125: Minor and probably pedantic, but if we were to consolidate this logic into a `map` it might be less lengthy of a `doPack` override.
```
const appFile = `${this.appInfo.productFilename}.app`;
const [x64AppPath, arm64AppPath] = [Arch.x64, Arch.arm64].map(async arch => {
  const destDir = appOutDir + '-' + Arch[arch];
  await super.doPack(outDir, destDir, platformName, arch, platformSpecificBuildOptions, targets, false);
  return path.join(destDir, appFile)
})
const { version, name } = this.info.framework
log.info({
  platform: platformName,
  arch: Arch[arch],
  [name]: version,
  appOutDir: log.filePath(appOutDir),
}, `packaging`)
const { makeUniversalApp } = require('@electron/universal');
await makeUniversalApp({ x64AppPath, arm64AppPath, outAppPath: path.join(appOutDir, appFile), force: true });

const rmdir = promisify(require('fs').rmdir);
archOutput.forEach(async dir => { await rmdir(dir, { recursive: true }) });
await this.doSignAfterPack(outDir, appOutDir, platformName, arch, platformSpecificBuildOptions, targets);
```

Regardless, this is fantastic work! Everything checks out locally for me as well.

mmaietta (approved):
Fantastic work @lutzroeder! 🎉

packages/app-builder-lib/src/packager.ts
lutzroeder: `installAppDependencies` has a bunch of dependencies on `Packager` and not sure it should be moved.  If at all, this probably should be a separate refactoring.

mmaietta:
I think we should also merge these for the prerelease cut:
https://github.com/electron-userland/electron-builder/pull/5488 https://github.com/electron-userland/electron-builder/pull/5489 https://github.com/electron-userland/electron-builder/pull/5484
CC @quanglam2807 

@lutzroeder I, unfortunately, don't have merge permissions for this project. It'd be awesome if I did though!

mmaietta:
Quick follow up, can you please update this PR's `yarn.lock` 🙂 

Also, @lutzroeder I've pushed a "PreRelease" branch to my fork containing multiple fixes from open PRs we've worked on (dmg, universal, signing, unit tests). Feel free to generate patch files from it to use for the interim.
https://github.com/mmaietta/electron-builder/tree/PreRelease

Just compile, link, and use rsync to migrate all the files so you can run `patch-package` on your project.
Handy cmd: `rsync -avh --exclude=package.json --exclude=yalc.sig .yalc/ node_modules/`
You'll need to manually `yarn add -D istextorbinary @electron/universal` in your project for now because `yarn install` can't utilize a patched package.json to update the yarn.lock.

develar:
Thanks for your work, you are awesome :)

I cannot merge https://github.com/electron-userland/electron-builder/pull/5488 since not clear for me, is it safe just revert or it is possible just add required patterns to auto sign.

lutzroeder:
@develar note this PR is orthogonal to #5488. #5488 tracks an existing regression, this is a new feature.

lutzroeder:
> Quick follow up, can you please update this PR's yarn.lock

@mmaietta ran `yarn install` and added the updated `yarn.lock`. Is that all that is needed for this PR?

mmaietta (approved):
LGTM

lutzroeder:
@develar @mmaietta what does it take to push a new prerelease with the recent changes?

Is there a doc page how to build `electron-builder` locally and fully patch the `node_modules` of a project?

mmaietta:
I have no idea in terms of a release 😅 

In terms of patch package for tracking latest master, here's how I've done it:
Step 1
```
yarn compile && \
(cd ./packages/app-builder-lib && yalc push) && \
(cd ./packages/builder-util && yalc push) && \
(cd ./packages/builder-util-runtime && yalc push) && \
(cd ./packages/dmg-builder && yalc push) && \
(cd ./packages/electron-builder && yalc push) && \
(cd ./packages/electron-publish && yalc push) && \
(cd ./packages/electron-builder-squirrel-windows && yalc push) && \
(cd ./packages/electron-forge-maker-appimage && yalc push) && \
(cd ./packages/electron-forge-maker-nsis && yalc push) && \
(cd ./packages/electron-forge-maker-nsis-web && yalc push) && \
(cd ./packages/electron-forge-maker-snap && yalc push) && \
(cd ./packages/electron-updater && yalc push)
```
Step 2 in your project:
```
yalc remove --all
yalc retreat --all
yarn install --check-files --force --ignore-scripts
yalc link app-builder-lib builder-util builder-util-runtime dmg-builder electron-builder electron-publish electron-builder-squirrel-windows electron-forge-maker-appimage electron-forge-maker-nsis electron-forge-maker-nsis-web electron-forge-maker-snap electron-updater
rsync -avh --exclude=package.json --exclude=yalc.sig .yalc/ node_modules/
yarn patch-package app-builder-lib --exclude
yarn patch-package builder-util --exclude
yarn patch-package electron-builder --exclude
yarn patch-package electron-updater --exclude
```
If you hit a permission issue during `yarn install`, then just rerun the `retreat` and `remove` commands until it properly resets the permissions in node_modules. Note: the `--ignore-scripts` is necessary to avoid applying your current patch files as it'll break the subsequent attempts to create new patches.

develar:
`22.10.4` is released as pre-release.

mmaietta:
@lutzroeder EB provides an `electronDist` param for custom electron dists/forks, which is what my team uses and something I think we may have overlooked. Not a bug as part of this PR as this issue already exists for compiling 2+ archs on a single build run, as there will always be only one path to an arch-specific electron dist.

For building universal, we need to provide 2 different paths, one for each arch. However, `electronDist` only accepts one path to a corresponding `.app`.
What are your thoughts on us enabling `electronDist` to accept a dictionary of `${platform}-${arch} : <path>` key-value pairs? It'd retain backward compatibility while enabling custom forks for universal builds.

CC @quanglam2807

thaddeusm:
Has anyone else tried publishing these files to Github along with other targets?

I am able to build locally (22.10.4) but publishing does not trigger an upload of Mac builds, though Windows assets are being uploaded as usual for my configuration.

quanglam2807:
@mmaietta Instead of using dictionary, wouldn't it better to allow specifying `electronDist` as a function? I think that'd be more flexible.

quanglam2807:
@lutzroeder I just gave this a test in production. 

`@electron/universal` validates SHA of all non-binary files and `app.asar` to make sure that they are consistent between arm64/x64 builds:
https://github.com/electron/universal/blob/a05a5e6db87e88ffa60038de50a5ec930b8f64be/src/index.ts#L108

`Contents/Info.plist` SHAs are also inconsistent because `electron-builder`.
- adds `AsarIntegrity` field  (SHA of `app.asar`) to `Contents/Info.plist`.
- SHA of `Contents/Info.plist` is inconsistent between builds (we generate them separately for each arch-specific build).
This

quanglam2807:
I just checked again, it's ok for SHAs of `app.asar` to be different. So I think we need to strip `AsarIntegrity` field from `Info.plist`.

quanglam2807:
We can add a conditional check here: https://github.com/electron-userland/electron-builder/blob/ca20151c3416324d2413f451dea0c9e3853bab79/packages/app-builder-lib/src/platformPackager.ts#L229 but I'm not sure if removing `AsarIntegrity` would do any harms.

lutzroeder:
@quanglam2807 what is causing them to be different?

quanglam2807:
> @quanglam2807 what is causing them to be different?

We generate app.asar for each arch-specific build separately so they're different files with different SHAs.

lutzroeder:
@quanglam2807 do the `app.asar` files contain arch-specific binaries or do they contain something else that would make them different? Is the difference is arch-specific binaries not sure how that would work as `lipo` is expected to merge such binaries.

lutzroeder:
Follow-up #5495

quanglam2807:
> @quanglam2807 do the `app.asar` files contain arch-specific binaries or do they contain something else that would make them different? If the difference is arch-specific binaries not sure how that would work as `lipo` is expected to merge such binaries.

`app.asar` normally doesn't contain binaries. If it does, `electron-builder` will automatically unpack them from the asar file.

We generate x64 & arm64 builds separately before merging them together. So for each arch, we generate different app.asar file. These files are identical in contents, but could have different SHA because they differ in metadata (eg date created).

So in this case, we can opt to remove AsarIntegrity from Info.plist. Or we can unify the app.asar process for universal build.

lutzroeder:
@quanglam2807 do you have a repro or is there some extra logging that needs to be enabled? I've been running with 20.10.4, pushed a release and haven't seen this error.

quanglam2807:
@lutzroeder You can clone this repo: https://github.com/webcatalog/translatium

Then remove the patch: `patches/app-builder-lib+22.10.4.patch`. Run `yarn` and then `yarn dist-dev`.

packages/app-builder-lib/src/macPackager.ts
```diff
@@ -91,6 +92,42 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     }
   }
 
+  protected async doPack(outDir: string, appOutDir: string, platformName: ElectronPlatformName, arch: Arch, platformSpecificBuildOptions: MacConfiguration, targets: Array<Target>): Promise<any> {
+    switch (arch) {
+      default: {
+        return super.doPack(outDir, appOutDir, this.platform.nodeName as ElectronPlatformName, arch, this.platformSpecificBuildOptions, targets);
```
vadim-termius, line 98: Hi @lutzroeder, 

Thank you for the universal support.

Why do you overwrite `platformName` and `platformSpecificBuildOptions` instead of passing arguments directly? It breaks mas builds.

lutzroeder: Good catch, these should pass through. Can you create an issue to track and/or feel free to send a PR.

fabiospampinato:
Does this PR take into account `electron-updater`? 

Because when I switch from the `x64` to the `universal` architecture I'm not getting the build artifacts that I would need, just a single file at `mac-universal/MyApp.app` and this error in the terminal:

```
Error: spawn zip ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:264:19)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
```

I'm under Mojave.

lutzroeder:
@fabiospampinato see #5504 which is not included in a preview build yet and #5495.

fabiospampinato:
@lutzroeder Would that probably address my issue? 🤔 I can get a universal build of the app, but I can't get the dmg and accessory files. Seeing as an error, possibly related to packing the app, got thrown at build time I don't think that would address my issue.

Issue #5465, opened 2020-12-10
Operation not permitted during dmg building on macOS 11 Big Sur


gnattu:
* **Version**: electron-builder@22.9.1

* **Electron Version**:electron@11.0.4
* **Electron Type (current, beta, nightly)**:current

* **Electron updater:**  electron-updater@4.3.1

* **Target**: `dmg` for macOS

After setting `DEBUG=electron-builder DEBUG_DMG=true`, the console logs error during copy:

```
2020-12-10 11:45:43.585 copy-helper[54295:350710] About to copy "/path/app.app".
2020-12-10 11:45:44.255 copy-helper[54295:350710] copy error (canceling): /Volumes/img/app.app/Contents/Resources/app.asar.unpacked/node_modules/npm/node_modules/libnpmorg/.BC.T_K9LSTf: Operation not permitted
2020-12-10 11:45:44.255 copy-helper[54295:350710] Copy finished with error 1 (Operation not permitted).
................................Error 1 (Operation not permitted).
could not access /Volumes/img/app.app/Contents/Resources/app.asar.unpacked/node_modules/npm/node_modules/libnpmorg/.BC.T_K9LSTf - Operation not permitted
```

Then building ultimately failed with:

```
2020-12-10 11:45:45.386 diskimages-helper[54280:350655] *useEffectiveIDs**** euid/egid changed to 501,20 (uid/gid is 501,20)
DIHLDiskImageCreate() returned 1
hdiutil: create: returning 1
hdiutil: create failed - Operation not permitted

...

 ⨯ hdiutil exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE  stackTrace=
                                                                    Error: hdiutil exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
                                                                        at ChildProcess.<anonymous> (/Users/gnattu/src/poi/node_modules/builder-util/src/util.ts:243:14)
                                                                        at Object.onceWrapper (node:events:483:26)
                                                                        at ChildProcess.emit (node:events:376:20)
                                                                        at maybeClose (node:internal/child_process:1063:16)
                                                                        at Process.ChildProcess._handle.onexit (node:internal/child_process:295:5)
```

I can create an image with [appdmg](https://github.com/LinusU/node-appdmg) for the same app.

gnattu:
[...]

PR #5488, opened 2020-12-17
Revert "Codesign all binary-like files"

### LINKED ISSUE #5487: electron-builder 20.10.3 fails with Operation not permitted
* **Version**: 22.10.3
* **Electron Version**: 11.1.0
* **Target**: DMG
* **OS**: macOS Big Sur 11.1 (20C69)

**Repro steps**
1. `git clone https://github.com/electron/electron-quick-start.git`
2. `cd electron-quick-start`
3. `npm install --save-dev electron-builder@22.10.3`
4. `npm install`
5. `export DEBUG_DMG=true`
6. `npx electron-builder --mac`

```
~/electron-quick-start: npx electron-builder --mac
  • electron-builder  version=22.10.3 os=20.2.0
  • packaging       platform=darwin arch=x64 electron=11.1.0 appOutDir=dist/mac
  • signing         file=dist/mac/electron-quick-start.app identityName=*** identityHash=*** provisioningProfile=none
  • building        target=macOS zip arch=x64 file=dist/electron-quick-start-1.0.0-mac.zip
  • building        target=DMG arch=x64 file=dist/electron-quick-start-1.0.0.dmg
  • building embedded block map  file=dist/electron-quick-start-1.0.0-mac.zip
  ⨯ hdiutil exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
Output:

Output:
Initializing…
Creating…
Copying…
could not access /Volumes/electron-quick-start 1.0.0/electron-quick-start.app/Contents/Resources/.BC.T_OYJ2t1 - Operation not permitted
Finishing…
```

This regressed starting with #5322.

Related #5465 #5431

@adam-lynch @develar

lutzroeder:
Revert electron-userland/electron-builder#5322

Fix #5487

@adam-lynch @develar

### DIFF
--- a/packages/app-builder-lib/electron-osx-sign/util.js
+++ b/packages/app-builder-lib/electron-osx-sign/util.js
@@ -202,12 +202,23 @@ module.exports.walkAsync = async function (dirPath) {
       const stat = await fs.lstat(filePath)
       if (stat.isFile()) {
         switch (path.extname(filePath)) {
+          case '': // Binary
+            if (path.basename(filePath)[0] !== '.') {
+              return getFilePathIfBinaryAsync(filePath)
+            } // Else reject hidden file
+            break
+          case '.dylib': // Dynamic library
+          case '.node': // Native node addon
+            return filePath
           case '.cstemp': // Temporary file generated from past codesign
             debuglog('Removing... ' + filePath)
             await fs.unlink(filePath)
             return
           default:
-            return getFilePathIfBinaryAsync(filePath)
+            if (path.extname(filePath).indexOf(' ') >= 0) {
+              // Still consider the file as binary if extension seems invalid
+              return getFilePathIfBinaryAsync(filePath)
+            }
         }
       } else if (stat.isDirectory() && !stat.isSymbolicLink()) {
         const result = await _walkAsync(filePath)


mmaietta (approved):

lutzroeder:
Test failure is unrelated to this change. See #5426.

develar:
@lutzroeder If we revert, what's alternative? Maybe you can add more patterns to codesign to solve initial issue? I mean if original fix was intended to fix code signing of some files, let's just add such patterns. Because I cannot simply revert.

lutzroeder:
@develar Hello World regressed in 22.10.3 due to this change so we should start by reverting the change. I'm not an expert what the issues are with special cases involving signing native modules nor did the original pull request include an exact repro. The goal for this pull request is to undo the regression from 22.9.1 to 22.10.3.

#5465 is tracking one investigation for native modules. Maybe we should use that ticket to drive a better fix?

@mmaietta @gnattu @adam-lynch

mmaietta:
@develar thanks for merging.

https://github.com/electron-userland/electron-builder/pull/5493
I gave another swing at this for detecting binary files, however, I don't have a repo or exact description from the original #5322 PR on what else to test specifically for. It works locally in my projects but that's still limited

PR #5493, opened 2020-12-21
fix: binary detection signing

mmaietta:
Provides an alternative method of detecting binaries since `isBinaryFile` improperly detects binaries for a path
Switching from `isBinary` to `istextorbinary`. Exposing `getFilePathIfBinarySync` to app-builder-lib to consolidate binary detection logic
Check if each path exists for mac options `binaries`, otherwise resolve binary relative path from the artifact's app Contents path

This is a followup from #5488 that #5322

Relates to #5465 & potentially #5431


packages/app-builder-lib/electron-osx-sign/util.js
```diff
@@ -215,10 +212,7 @@ module.exports.walkAsync = async function (dirPath) {
             await fs.unlink(filePath)
             return
           default:
-            if (path.extname(filePath).indexOf(' ') >= 0) {
-              // Still consider the file as binary if extension seems invalid
-              return getFilePathIfBinaryAsync(filePath)
-            }
+            return getFilePathIfBinarySync(filePath)
```
gnattu, line 215: Dropping the extension test will code sign all binary-like files including pictures and icons eventually, we should still include `if (path.extname(filePath).indexOf(' ') >= 0) {` check.

gnattu (changes requested):

packages/app-builder-lib/electron-osx-sign/util.js
mmaietta: Nice catch! I've updated and also consolidated some of the logic into `getFilePathIfBinarySync` for `unpackDetector` to also leverage

lutzroeder:
@mmaietta @develar Travis is failing with `The command "node .yarn/releases/yarn-2.4.0.cjs pretest" exited with 1.` for all pull requests starting with this one?

mmaietta:
It's failing with `checkDeps` starting from the commit prior to this merge: https://github.com/electron-userland/electron-builder/commit/eec01f7ba9ce5b17dfcf84ecb237dd4872d6c880

mmaietta:
@lutzroeder opened https://github.com/electron-userland/electron-builder/pull/5523 to fix master branch

PR #5511, opened 2020-12-30
fix(mac): Pass platformName and options in doPack 

lutzroeder:
Follow-up for #5481 fixing mas support.

@vadim-termius @mmaietta @quanglam2807


quanglam2807:
Thank you, @lutzroeder. This might be what causes https://github.com/electron-userland/electron-builder/pull/5481#issuecomment-749898441.

quanglam2807 (approved):

vadim-termius (approved):

mmaietta (approved):

lutzroeder:
@devlar looks like this one is ready to go. We might want to take #5523 or a similar change first to unblock tests.

Issue #5341, opened 2020-10-12
electron-updater: when targetting nsis-web, the "download-progress" event returns inaccurate results


wartab:
* **Version**: 22.8.1

* **Electron Version**: 8.2.5
* **Electron Type (current, beta, nightly)**: current
* **electron-updater version **: 4.3.1

* **Target**: nsis-web

This has been reported before, but closed quickly after no reaction was given, and the follow-up description was ignored.
See here: https://github.com/electron-userland/electron-builder/issues/2535

It seems like when targetting nsis-web, the `download-progress` event only gives information about the download of the relatively small installer executable. It does not take into account the more substantial download of the 7z file containing the actual application. 
This results in the download progress hanging at 100% for a long time when the actual application is being downloaded. Our users often assume the download has failed and just proceed to kill the application.

I believe it would be good if the progress did reflect a meaningful indication of progress.

In the meantime, if anyone knows a hack to circumvent this problem, I am willing to hear you out.

utajum:
https://imjselectron.github.io/electron-differential-updater/#/?id=quick-start

utajum:
Still relevant

ubjw:
relevant

Issue #4919, opened 2020-05-02
Updater: download-progress event not firing


Christilut:
* **Version**: 
    "vue-cli-plugin-electron-builder": "1.4.4",
    "electron-builder": "^21.2.0",

* **Electron Version**: 
    "electron": "^6.1.7",

* **Electron Type (current, beta, nightly)**: current

* **Target**: Windows

From the autoUpdater logs:
```
[2020-05-02 10:59:11.283] [info] Checking for update
[2020-05-02T08:59:11.290Z] [verbose]: updater: checking-for-update
[2020-05-02T08:59:11.291Z] [verbose]: status message to renderer, win (true): {"loading":true,"message":"Checking for update..."}
[2020-05-02T08:59:11.291Z] [debug]: application_update event: {"loading":true,"message":"Checking for update..."}
[2020-05-02 10:59:14.436] [info] Found version 1.2.37 (url: Rekordcloud-Companion-App-Setup.exe)
[2020-05-02T08:59:14.441Z] [verbose]: updater: update-available
[2020-05-02 10:59:14.441] [info] Downloading update from Rekordcloud-Companion-App-Setup.exe
[2020-05-02T08:59:14.442Z] [verbose]: status message to renderer, win (true): {"loading":true,"message":"Update found, downloading..."}
[2020-05-02T08:59:14.442Z] [debug]: application_update event: {"loading":true,"message":"Update found, downloading..."}
[2020-05-02 10:59:14.448] [debug] updater cache dir: C:\Users\Alpha\AppData\Local\rekordcloud-companion-app-updater
[2020-05-02 10:59:14.456] [info] No cached update info available
[2020-05-02 10:59:14.462] [info] Download block maps (old: "https://github.com/rekordcloud/companion-app-releases/releases/download/v1.2.36/Rekordcloud-Companion-App-Setup.exe.blockmap", new: https://github.com/rekordcloud/companion-app-releases/releases/download/v1.2.37/Rekordcloud-Companion-App-Setup.exe.blockmap)
[2020-05-02 10:59:15.443] [debug] g37lYn5+Uv1hYjkjPz8yNqpB duplicated in blockmap (same size), it doesn't lead to broken differential downloader, just corresponding block will be skipped)
[2020-05-02 10:59:15.449] [debug] 9sp5PQVmpRErRDAGvBxKRxuc duplicated in blockmap (same size), it doesn't lead to broken differential downloader, just corresponding block will be skipped)
[...]

Issue #2521, opened 2018-01-24
download-progress does not emit proper values (nsis-web, differentialPackage)


kukiel:
I have problem with update events when using nsis-web. The thing is that **download-progress** event is only taking into consideration the **.exe** file which is 500KB and when I want to display some progressbar it is downloading 500KB and showing instantly 100%. After that it is downloading 7z file in the background and I cannot see progress of it - this is bad especially as I plan to have large files included in app.

package.json
```
  "scripts": {
    "publish": "build --x64 --ia32 --windows nsis-web -p always"
  },
  "devDependencies": {
    ...
    "electron": "1.7.10",
    "electron-builder": "19.54.0",
    "electron-publisher-s3": "^19.55.2"
  },
  "dependencies": {
    ...
    "electron-log": "2.2.14",
    "electron-updater": "2.20.0"
  }
```

electron-build.json

```
{
  "appId": "com.kkl.test",
  "publish": [
    {
      "provider": "s3",
      "bucket": "somebucket",
      "region": "eu-central-1"
    }
  ],
  "productName": "test",
  "directories": {
    "output": "app-builds"
  },
  "win": {
    "target": [
      "nsis-web"
    ]
  },
  "nsisWeb": {
    "differentialPackage": true,
    "oneClick": false,
    "perMachine": true,
    "allowElevation": true,
    "allowToChangeInstallationDirectory": true,
    "unicode": true,
    "runAfterFinish": true,
    "deleteAppDataOnUninstall": true
  }
}
```

log.log file:
```
[2018-01-24 21:29:57.751] [info] Found version 0.0.4-test (url: test Web Setup 0.0.4-test.exe)
[2018-01-24 21:29:57.751] [info] { info: 
   { version: '0.0.4-test',
     files: [ [Object] ],
     path: 'test Web Setup 0.0.4-test.exe',
     sha512: 'jcijdawbJlHYIm6vf43RQeMTm1QLMhjZ1J3rviGW6XRwvd/J7Q3RC6I4zihJjh97CjUe+KJ7X1H/5WVLICLlKw==',
     packages: { ia32: [Object], x64: [Object] },
     sha2: 'c9e920597c31b826e332614ac824a69dc30bd0b01586fce737fe4fe09eca4153',
     releaseDate: '2018-01-24T15:47:49.189Z' },
  type: 'update-available' }
[2018-01-24 21:29:57.753] [info] Downloading update from test Web Setup 0.0.4-test.exe
[...]

PR #5515, opened 2021-01-03
feat: send download-progress information when downloading diff packages

mastergberry:
Fixes a bunch of issues such as #2521 #4919 #5341...and so many other duplicate tickets

I have tested this with:

- nsis
- nsis-web
- AppImage

I have not gotten to test the AppImage system, but I will test that in the next day or two. Probably works fine though since it is just using the same logic as nsis-web

Some further notes:

1. I have gone ahead and tested this with and without a `onProgress` being supplied, it works for both
2. With these logic changes it will go from 0 -> 100% on the `download-progress` event two times on `nsis-web` (and other similar packaging systems) because it will first download the blockmap, and second download the actual changes. I figured it was better to leave like this, vs change it. This makes it more clear to the user that something is happening (instead of nothing happening which is the current problem/situation).
3. I do not have a way to test `useMultipleRangeRequest`, so I have not implemented this. I left a TODO for someone if they want to make a separate PR for it...doesn't seem too hard, but I did not want to touch the `DataSplitter` stuff, as I did not understand it all.

Overall this is a huge improvement for user experience, but not perfect still. Would love to have someone pick up and finish this and implement the missing functionality into the `executeTasksUsingMultipleRangeRequests` function (and sub-functions).

If you stumble across this PR, and are wondering why you still are not having progress shown, then try using the option `useMultipleRangeRequest` and setting it to `false` in your `package.json` under your `publish` configuration. This option will disable doing all of the downloads in a single HTTP request, and instead will use one HTTP request per download range. This does not work on AWS S3 anyways (which is where my stuff is currently hosted).

### DIFF
--- a/packages/electron-updater/src/AppImageUpdater.ts
+++ b/packages/electron-updater/src/AppImageUpdater.ts
@@ -5,7 +5,9 @@ import { unlinkSync } from "fs"
 import * as path from "path"
 import { DownloadUpdateOptions } from "./AppUpdater"
 import { BaseUpdater, InstallOptions } from "./BaseUpdater"
+import { DifferentialDownloaderOptions } from "./differentialDownloader/DifferentialDownloader"
 import { FileWithEmbeddedBlockMapDifferentialDownloader } from "./differentialDownloader/FileWithEmbeddedBlockMapDifferentialDownloader"
+import { DOWNLOAD_PROGRESS } from "./main"
 import { findFile } from "./providers/Provider"
 
 export class AppImageUpdater extends BaseUpdater {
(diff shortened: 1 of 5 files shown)


mastergberry:
Just as an update, I finally got a bit of time, and swapped over to Linux to test the AppImageUpdater system. It's working perfectly fine as well.

Editing the original PR to also include that this has been tested.

develar:
Awesome!

mastergberry:
Hey @develar can you throw a pre-release (or full release) with this included?

Thx :)

carterbs:
+1 @develar. Would love to have this in a release. Thank you @mastergberry for fixing it!

carterbs:
@develar let me know if there's anything I can do to help facilitate this. It would be a huge boon for my app, which will be releasing on windows very soon.

mmaietta:
@carterbs please update to v22.10.5 and that'll have this included

carterbs:
@mmaietta Thanks! Really appreciate this getting merged. Have a wonderful week.

PR #5524, opened 2021-01-05
fix(mac): Allow arm64 macs to update to x64 version if no arm64 version available

davej:
If an arm64 mac is running an intel version of an app (through Rosetta) and there are no arm64 versions of the app to update to then we should update to the intel version anyway.

Without this code change then an arm64 mac would be stuck on the current version until an arm64 version is released.


mmaietta (approved):
Can you add a comment to the new lines as to what is being done? Something simple like
```
// Allow arm64 macs to install universal or rosetta2(x64) - https://github.com/electron-userland/electron-builder/pull/5524
```

davej:
Thanks @mmaietta. Made suggested change.

PR #5557, opened 2021-01-18
fix: support Yarn 2/PNP projects

mitchchn:
Some build steps (e.g. `app-builder icon`) do not work with Yarn 2/PNP, causing electron-builder to fail. Unplugging the affected packages allows build to complete without rewriting all dependencies to be PNP-aware.

Some app-builder dev dependencies also needed to be declared as full dependencies since Yarn 2 sees that they are required at runtime and fails with an ambiguity error in strict mode.

### DIFF
--- a/packages/app-builder-lib/package.json
+++ b/packages/app-builder-lib/package.json
@@ -98,5 +98,6 @@
     "electron-builder-squirrel-windows": "workspace:*"
   },
   "//": "electron-builder-squirrel-windows and dmg-builder added as dev dep for tests (as otherwise `require` doesn't work using Yarn 2)",
-  "typings": "./out/index.d.ts"
+  "typings": "./out/index.d.ts",
+  "preferUnplugged": true
 }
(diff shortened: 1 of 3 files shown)


packages/app-builder-lib/package.json
```diff
@@ -93,10 +95,9 @@
...
     "@types/js-yaml": "^3.12.6",
     "@types/normalize-package-data": "^2.4.0",
-    "@types/semver": "^7.3.4",
-    "dmg-builder": "workspace:*",
-    "electron-builder-squirrel-windows": "workspace:*"
+    "@types/semver": "^7.3.4"
   },
-  "//": "electron-builder-squirrel-windows and dmg-builder added as dev dep for tests (as otherwise `require` doesn't work using Yarn 2)",
-  "typings": "./out/index.d.ts"
+  "//": "electron-builder-squirrel-windows and dmg-builder added as deps (as otherwise `require` doesn't work using Yarn 2)",
```
merceyz, line 100: This comment shouldn't exist as this is true for node_modules as well (Yarn 2 !== Yarn PnP btw), hoisting just happens to make it work if you're lucky.
https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

mmaietta:
Can you please rebase off latest master?

Issue #5565, opened 2021-01-20
Asar creation fails on Apple Silicon


elliottkember:
* **Version**: 22.10.4
* **Electron Version**: 11.1
* **Electron Type (current, beta, nightly)**: current
* **Electron-updater (current, beta, nightly)**: current
* **Target**: arm64, x64
* **Node**: v15.3.0

Apologies if you get déjà vu here. I have been working through this puzzle after writing this PR, and have edited it to update my findings.

This is my package.json configuration: 

```
"target": [
  {
    "target": "default",
    "arch": [
      "arm64",
      "x64"
    ]
  }
],
```

When building for both architectures, electron-builder seems to sometimes choke on a specific, particularly large 6.8MB file in my node_modules on M1 (`node_modules/react-icons/gi/index.js`, an icon set). 

The line that causes this is [here](https://github.com/electron-userland/electron-builder/blob/5203d7eb15726605e8987aeed7a5ccedc8152e04/packages/app-builder-lib/src/asar/asarUtil.ts#L209), which has a comment linking to [bulk file reads](https://github.com/yarnpkg/yarn/pull/3539) and changed in [this commit](https://github.com/electron-userland/electron-builder/commit/b061710f0126700ea8860d7db78441e8c7fcede2) three years ago.

```js
// https://github.com/yarnpkg/yarn/pull/3539
const stat = metadata.get(file)
if (stat != null && stat.size < (2 * 1024 * 1024)) {
  // This part works reliably, even on files that are > 2 * 1024 * 1024
  readFile(file)
    .then(it => {
      writeStream.write(it, () => w(index + 1))
    })
    .catch(e => reject(`Cannot read file ${file}: ${e.stack || e}`))
}
else {
  // This part fails, without emitting the "error" event below, ending the process
  const readStream = createReadStream(file)
  readStream.on("error", reject)
  readStream.once("end", () => w(index + 1))
  readStream.pipe(writeStream, {
    end: false
  })
}
```

This seems to work okay on Github Actions and on my x64 Mac Mini. But running it on the Apple Silicon mac, it quits before turning Electron.app into my MyApp.app bundle, every time. 

[...]

PR #5567, opened 2021-01-20
Add "open" guard for AsarUtil to fix concurrent builds

elliottkember:
Fixes https://github.com/electron-userland/electron-builder/issues/5565 by adding an `.on("open", () => {})` guard around using the file stream. 

From [the documentation](https://nodejs.org/en/knowledge/advanced/streams/how-to-use-fs-create-read-stream/):

```js
// This will wait until we know the readable stream is actually valid before piping
readStream.on('open', function () {
  // This just pipes the read stream to the response object (which goes to the client)
  readStream.pipe(res);
});
```

The original `createReadStream` call, without a guard, fails on my M1 mac when building multiple architectures with larger files (>7mb). This change makes packaging totally reliable and is best practice for a `readStream`.

### DIFF
--- a/packages/app-builder-lib/src/asar/asarUtil.ts
+++ b/packages/app-builder-lib/src/asar/asarUtil.ts
@@ -196,7 +196,6 @@ export class AsarPackager {
           return
         }
 
-        // https://github.com/yarnpkg/yarn/pull/3539
         readFile(file)
           .then(it => {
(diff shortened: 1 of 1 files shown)


mmaietta (approved):

elliottkember:
@develar this seems like a pretty simple PR, could you please merge?

PR #5596, opened 2021-01-29
fix(nsis): keep injected logger state

jupp0r:
This change avoids loosing state in the logger in case users inject a logger class instead of an object with functions.


jupp0r:
I don't think the CI failures are relevant to my changes, please advice on how to handle them.

mmaietta (approved):

PR #5645, opened 2021-02-19
set node-makensis charset to utf-8

fangjiapeng:
If fileAssociations were setting while there is Chinese character in electron-builder.yaml, an error will occur.


mmaietta:
Nice work! Can you include unit tests to cover this? (with + without Chinese character)
I think it'd be akin to the tests here: https://github.com/electron-userland/electron-builder/blob/master/test/src/configurationValidationTest.ts

PR #5664, opened 2021-03-01
feat(build-util): add param "logName" in log as log key and fix bad log display when win sign error 

codeudan:
log.ts use error.message as log display name, like below, when message is long, it will hard read

Xnip2021-03-01_13-12-37_mosaic

I just add param "logName" in log as log key and display 'sign error' as log name when winPackager.ts throw a error like below

Xnip2021-03-01_13-05-55_mosaic

change is  compatible

### DIFF
--- a/packages/app-builder-lib/src/winPackager.ts
+++ b/packages/app-builder-lib/src/winPackager.ts
@@ -250,6 +250,7 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
           log.warn({error: message, attempt: i + 1}, `cannot sign`)
           continue
         }
+        e.logName = 'sign error'
         throw e
       }
     }
--- a/packages/builder-util/src/log.ts
+++ b/packages/builder-util/src/log.ts
@@ -34,34 +34,34 @@ export class Logger {
     return debug.enabled
   }
 
-  info(messageOrFields: Fields | null | string, message?: string) {
-    this.doLog(message, messageOrFields, "info")
+  info(messageOrFields: Fields | null | string, message?: string, logName?: string) {
+    this.doLog(message, messageOrFields, "info", logName)
   }
 
-  error(messageOrFields: Fields | null | string, message?: string) {
-    this.doLog(message, messageOrFields, "error")
+  error(messageOrFields: Fields | null | string, message?: string, logName?: string) {
+    this.doLog(message, messageOrFields, "error", logName)
   }
 
-  warn(messageOrFields: Fields | null | string, message?: string): void {
-    this.doLog(message, messageOrFields, "warn")
+  warn(messageOrFields: Fields | null | string, message?: string, logName?: string) {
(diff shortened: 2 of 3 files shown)


mmaietta:
I appreciate you targetting this. I don't think this is the correct route to go though, as it's a lot of changes to only add a `logName` coming from an upstream win-only packager.

The reason being is that a couple of things stand out to me that we could fix as part of this:
- I see `Invalid Timestamp URL` is repeated twice so it must being wrapped somewhere on accident.
- The `error.message` and `error.stack` are almost identical (minus formatting shenanigans). I think this is due to how the console is piping output while recording it, only to have it all repeated in two vars when an error occurs? idk yet.

mmaietta:
Would you be willing to try this patch on your code locally? Should cleanup your output and it'd help in having another person verify that we're indeed solving for the same frustration 🙂 

```
diff --git a/packages/builder-util/src/util.ts b/packages/builder-util/src/util.ts
index 81441d99..3d86c371 100644
--- a/packages/builder-util/src/util.ts
+++ b/packages/builder-util/src/util.ts
@@ -240,7 +240,7 @@ function handleProcess(event: string, childProcess: ChildProcess, command: strin
       }
     }
     else {
-      reject(new ExecError(command, code, formatOut(out, "Output"), formatOut(errorOut, "Error output")))
+      reject(new ExecError(command, code, out, errorOut))
     }
   })
 }
diff --git a/packages/electron-builder/src/cli/cli.ts b/packages/electron-builder/src/cli/cli.ts
index bdc2e87d..3bf0700f 100644
--- a/packages/electron-builder/src/cli/cli.ts
+++ b/packages/electron-builder/src/cli/cli.ts
@@ -52,7 +52,7 @@ function wrap(task: (args: any) => Promise<any>) {
           log.error(null, error.message)
         }
         else if (!(error instanceof ExecError) || !error.alreadyLogged) {
-          log.error({stackTrace: error.stack}, error.message)
+          log.error({ culprit: task.name }, error.message)
         }
       })
   }
```

codeudan:
Yes, Clean output is a better way to resolve it, and this issue related to log.ts file also:

`
if (fieldValue != null && typeof fieldValue === "string" && fieldValue.includes("\n")) {
        valuePadding = " ".repeat(messagePadding + message.length + fieldPadding.length + 2)
        fieldValue = "\n" + valuePadding + fieldValue.replace(/\n/g, `\n${valuePadding}`)
    }
`

when fieldValue include '\n', log.ts will print too many blank string. What is you suggestion about ts?

@mmaietta

mmaietta:
Instead of `fieldValue.replace(/\n/g,"n${valuePadding})"`, howabout we do a regex to convert multi-line entries to a single new line. I think this would work? 
```
fieldValue.replace(/\n\s*\n/g, '\n${valuePadding}');
```
Can you give that a shot? 
I'm not able to repro any multiline errors after the patch I provided earlier.

PR #5669, opened 2021-03-03
feat: add option removePackageKeywords

krocheck:
This will add an option called removePackageKeywords, similar to removePackageScripts, to allow keywords to optionally not be removed from `package.json` files but leaving the current behavior to remove them as default.

The Companion project is intending to switch from including its control modules as git submodules to importing as dependencies in npm.  In this change we noted the keywords were being removed from the packages after building, which we use as an element of search filtering.  With over 250 modules, changing our inclusion method is long overdue, but we'd prefer to not have to change `keywords` to another key name in all those modules, instead continuing with this build option to allow keywords to be included in the package.json files.

This was tested through a local compile and build procedure for our app:
- Leaving the existing compile commands: passed ... no errors, package.json files did not include keywords
- Adding `"removePackageKeywords": false`: passed ... no errors, package.json files included the keywords

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -6211,6 +6211,11 @@
       "description": "Whether to remove `scripts` field from `package.json` files.",
       "type": "boolean"
     },
+	"removePackageKeywords": {
+	  "default": true,
+	  "description": "Whether to remove `keywords` field from `package.json` files.",
+	  "type": "boolean"
+	},
     "rpm": {
       "anyOf": [
         {
--- a/packages/app-builder-lib/src/configuration.ts
+++ b/packages/app-builder-lib/src/configuration.ts
@@ -232,6 +232,13 @@ export interface Configuration extends PlatformSpecificBuildOptions {
    * @default true
    */
   readonly removePackageScripts?: boolean
+  
+  /**
+   * Whether to remove `keywords` field from `package.json` files.
+   *
+   * @default true
+   */
+  readonly removePackageKeywords?: boolean
 }
 
 export interface AfterPackContext {
--- a/packages/app-builder-lib/src/fileTransformer.ts
+++ b/packages/app-builder-lib/src/fileTransformer.ts
@@ -28,10 +28,11 @@ export function hasDep(name: string, info: Packager) {
 export function createTransformer(srcDir: string, configuration: Configuration, extraMetadata: any, extraTransformer: FileTransformer | null): FileTransformer {
   const mainPackageJson = path.join(srcDir, "package.json")
   const isRemovePackageScripts = configuration.removePackageScripts !== false
+  const isRemovePackageKeywords = configuration.removePackageKeywords !== false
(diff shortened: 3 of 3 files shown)


mmaietta:
It seems that a test snapshot is still broken, but not sure if it should be since the default value of `false` shouldn't be requiring any tests to break.  Can you test again locally?
```
TEST_FILES=BuildTest yarn test-all
```

Can you also add a test for when `removePackageKeywords: true`?

mmaietta (changes requested):
Overall, this sounds like a great addition.
It seems that a test snapshot is broken though, but not sure if it should be since the default value of `false` shouldn't be requiring any tests to break.  Can you test again locally?
```
TEST_FILES=BuildTest yarn test-all
```

Can you also add a test for when `removePackageKeywords: true`?

krocheck:
Hi @mmaietta , so when I say I tested I wasn't using any of your internal tests, I was running "real-world" tests against our software.  To be honest, I'm a little lost when going into /test and trying to navigate fulfilling your request to add a test for this ... in part because it seems removePackageScripts doesn't have one and that was my guide for implementing this.

If I'm reading correctly in Circle

krocheck:
So I was able to to run the test suite as requested.  I ran both on my branch with the fix and the master branch as it currently sits:

```
Snapshot Summary
 › 7 snapshots failed from 4 test suites. Inspect your code changes or re-run jest with `-u` to update them.

Test Suites: 13 failed, 9 skipped, 17 passed, 30 of 39 total
Tests:       20 failed, 122 skipped, 106 passed, 248 total
Snapshots:   7 failed, 139 passed, 146 total
Time:        580.418 s
Ran all test suites.
```

krocheck:
So I was able to to run the test suite as requested.  Admittedly, my Windows environment runs into a lot of errors and is really only tuned for the projects I work on.  But, I ran the suite on both my branch with the fix and the master branch as it currently sits.  You'll notice that the results are the same:

With my fix:
```
Snapshot Summary
 › 7 snapshots failed from 4 test suites. Inspect your code changes or re-run jest with `-u` to update them.

Test Suites: 13 failed, 9 skipped, 17 passed, 30 of 39 total
Tests:       20 failed, 122 skipped, 106 passed, 248 total
Snapshots:   7 failed, 139 passed, 146 total
Time:        580.418 s
Ran all test suites.
```

Prior to my fix:
```
Snapshot Summary
 › 7 snapshots failed from 4 test suites. Inspect your code changes or re-run jest with `-u` to update them.

Test Suites: 13 failed, 9 skipped, 17 passed, 30 of 39 total
Tests:       20 failed, 122 skipped, 106 passed, 248 total
Snapshots:   7 failed, 139 passed, 146 total
Time:        621.542 s
Ran all test suites.
```

mmaietta:
Hi @krocheck can you please rebase off latest master?

mmaietta:
Thanks for your work. We've merged your feature through another PR that had resolved the merge conflict

PR #5680, opened 2021-03-05
chore: integrate prettier

codeudan:
This is a huge change, But I think it is worth for future development.

This codes of this repo is unformatted, So I integrate prettier to format

prettier is a format tool https://prettier.io/,  now I integrate it for this repo.

usage:
install deps
`yarn `

run prettier
`yarn run prettier ` 

Only format packages folder

It is compatible with eslint and work fine.

Notice: Because yarn2 has no node_modules, I do not  integrate with IDEs，if you want integrate with IDE or Editor, You need use pnpify,(https://yarnpkg.com/getting-started/editor-sdks) to link target folder, If you use vscode, use this extend (https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)


mmaietta:
I love the goal here and would like to get this integrated.
One request, can you use this yaml instead? The `printWidth` is the main kicker I'd like changed. The default is far too compact.
`.prettierrc.yaml`
```
singleQuote: false
semi: false
arrowParens: avoid
trailingComma: es5
printWidth: 250
quoteProps: as-needed
```
You'll probably need to run prettier from master with the different config as opposed to using the current commit. You can force push to condense all your commits as well. 🙂

codeudan:
@mmaietta done.

develar:
Ok, let's discuss code style and kill each other :)

First, thanks for pushing forward.

I am against some rules.

```
} else {
```

See https://github.com/electron-userland/electron-builder/blob/master/.editorconfig#L9

```
ij_any_else_on_new_line = true
```

`then` and `catch` — also should be on a new line because it is promise chaining.


I do not see any valid reason to change existing code style. Maybe it is possible to configure `prettier` to match existing code style?

mmaietta:
>Ok, let's discuss code style and kill each other :)

Hahaha love it. I'm flexible on styling, I just like enforcing consistency 😅 

> Maybe it is possible to configure prettier to match existing code style?

Fair point, let's try that first.
@codeudan would you be willing to take another go at this per develar's feedback?

codeudan:
@develar Prettier do not support else_on_new_line  config, I am Sorry for this🙂. 
And exist code has many inconsistent code style includes trailingComma、singleQuote and other. And prettier provide some style like this: 

Xnip2021-03-10_21-00-24

Xnip2021-03-10_21-16-23

And After local config, you can use editor extend to format code quickly and consistently depend prettier config
This is why I recommend use Prettier.

And If keep exist code, It fine.😀

develar:
Do I understand correctly, that most of web developers do not use IntelliJ IDEA, where IDE is responsible for code style and IDE forces you to format code consistently? And key point of Prettier is that every developer regardless of editor can use the same tool to format code? (and of course using this tool on CI).

codeudan:
@develar not force, Use IDE or editor extend to format is totally optional for every developer, I mean If someone want format use IDE or Editor to format, Prettier support it. In this PR,  I add prettier script `yarn run prettier` to help format quickly, and Eslint work fine with it, If code style do not match prettier config, Eslint will hint it, So someone can modify code to match style manually or  `yarn run prettier`

.prettierrc.yaml
```diff
@@ -0,0 +1,6 @@
+singleQuote: false
+semi: false
+arrowParens: avoid
+trailingComma: es5
```
develar, line 4: Please set to `all`.

develar (approved):

.prettierrc.yaml
codeudan: trailingComma: es5 not includes function params

develar:
Ok, burn it all :)

As @mmaietta  approved this, I will merge, thanks.

codeudan:
@develar  Oh I just add trailingComma: all and format it before you ignore comment  you can change to trailingComma: es5 and "yarn run prettier" .

mmaietta:
I appreciate it develar. 🙂 

Regarding IDE, I use VS Code. I've used IntelliJ in the past but for some reason, I just enjoy VS Code more (everything except the debugger json configurations haha)

develar:
> Regarding IDE, I use VS Code. I've used IntelliJ in the past but for some reason, I just enjoy VS Code more (everything except the debugger json configurations haha)

Thanks for feedback. BTW, every core contributor can request free license if needed (https://www.jetbrains.com/community/opensource/#support).

PR #5694, opened 2021-03-09
fix: windows async signing

mmaietta:
Allows promises and async arrow functions to be accepted for `win.sign`

Fixes issue I previously commented on and can no longer find via search ¯\_(ツ)_/¯


develar:
Pushed as https://github.com/electron-userland/electron-builder/commit/91627116fdffd5d5faec23e149392283b9979a22

PR #5710, opened 2021-03-13
chore: get rid of yarn 2 and migrate to pnpm, update deps, require no…

develar:
get rid of yarn 2 and migrate to pnpm, update deps, require node 12.0.0 instead of 8.0.0

IntelliJ IDEA don't fully support Yarn 2. See https://youtrack.jetbrains.com/issue/WEB-42178
Although this issue doesn't affect electron-builder, there are another issues like that, see for example https://github.com/electron-userland/electron-builder/pull/5680#issue-585517683

## Migrate from Yarn 2 to PNPM

> Because yarn2 has no node_modules, I do not integrate with IDEs，if you want integrate with IDE or Editor, You need use pnpify,

After migration from Yarn 2 to PNPM in other my projects, I find PNPM more suitable and more elegant than Yarn 2.

 * No need to install per project. Someone can argue, but which came first: the chicken or the egg?
 * No need to install some plugins, like interactive update for per project. PNPM has built-in good command for that. `pnpm update -i -r --latest`
 * Strict! Strict! Strict by default and without breaking all existing tools in the world. "pnpm creates a non-flat node_modules, so code has no access to arbitrary packages".
 * Fast. 

NPM is out of comparison. I don't want to waste my time — last time in 2018 when I tried NPM again to check is it completely unusable or not, package-lock was modified as part of install in another time package lock format was changed.  So, no go for NPM.

## Get rid of Babel for production sources in favour of raw tsc output

**Imports are not dynamic anymore and fully conforms to how JS module system works**. That's why there are changes in a source code (ParallelsVm and so on).

Get rid of Babel. No need in 2021. Still, ESM modules cannot be used because TS compiler cannot add `.js` extension, but Node requires it. Vicious circle. https://github.com/microsoft/TypeScript/issues/16577 As I am already tired of this JS hell and ecosystem and just want to a little bit refresh electron-builder build flow, I decided to set `module` to `CommonJS`. Maybe later will be some hero who will find a way to use ESM modules. Or just use esbuild to bundle project for Node, as definitely esbuild will be able to meet Node requirements in a more elegant and simple way. As TS always tries to be on line with standards,  but where is standard and JS ecosystem of various tools.

## Get rid of Babel for test sources in favour of esbuild

~IntelliJ IDEA doesn't support or I don't know how to configure running Jest tests from IDE using tsc output in `out`. That's why on CI `runTests` uses tsc output but from IDE `test/src` is used.~
Custom file transformer is used — when JEST request transformed source of TS file, corresponding file from `out` is returned.


zkochan:
I will try to help with this PR.

develar:
@zkochan How do I can force PNPM to ignore `os` and install package anyway? I cannot find such option. Project should be compiled regardless of is package available on some OS or not. Yarn 2 installs such dependency.

zkochan:
In that case, you should add it as a prod dependency, not an optional dependency.

develar:
@mmaietta Yes. Instead of pushing to master directly, I ask your approve to merge changes into master :)

mmaietta:
This is awesome! I can definitely take a look at the tests post-merge.

Quick Q, re: the failing tests.
I see this a few times in the test results (CI & locally). 
```
 FAIL  src/updater/differentialUpdateTest.ts
  ● Test suite failed to run

    Cannot find module '@babel/plugin-transform-modules-commonjs'
```
This popped up locally, but I haven't been able to find it in the CI logs.
```
  • packager.vm is used: Cannot read property 'readFileSync' of undefined
  • electron-builder  version=22.10.5 os=20.3.0
  • loaded configuration  file=package.json ("build" field)
```

Not sure what the first error is about because the dependency is indeed listed in the pnpm lock. ¯\\_(ツ)_/¯

develar:
> Just curious what it'd be improving.

I find that it is not good that even if runtime supports modules, still our code uses workarounds like `await Promise.resolve().then(() => require("./macPackager"))`
The issue —  https://github.com/microsoft/TypeScript/issues/16577 TS team doesn't want to support adding `.js` to produced imports, and Node.js team requires `.js` extension (https://nodejs.org/api/esm.html#esm_mandatory_file_extensions)

JS ecosystem how it is :( I didn't dig deeper — maybe there is some extra magic magic extra  option to marry raw TSC output ("module": "ES2020") and Node.js.

develar:
> FAIL  src/updater/differentialUpdateTest.ts

Fixed, thanks, it was regression.
This test is not executed on CI because too heavy. Indeed, will be great to enable it, but after fixing Windows CI failures.

mmaietta:
> await Promise.resolve().then(() => require("./macPackager"))

Hmm, I thought that was what `await import('./macPackager')` was meant to replace?

In any case, the normal CI tests are all passing from the previous (now legacy?) TravisCI, the usual signing failures and that pesky test with race condition in creating dmgs (I think I fixed in the windows-ci PR I was working on so I can follow up on that after this PR)

mmaietta (approved):

develar:
> Hmm, I thought that was what await import('./macPackager') was meant to replace?

Please see note **Imports are not dynamic anymore and fully conforms to how JS module system works.**
Previously special babel plugin was used to achieve normal and correct behaviour, like used in a good runtimes like JVM. Now, without such babel plugin, JS module is used directly — module is required on import time and not on use. That's why I was forced to fix code.

Why I decided to stop using special babel plugin? Because new and complex code should be written in a normal and correct language — Go in our case (app-builder). Not in TS.

Issue #512, opened 2016-06-16
Gnome Flatpak


ghost:
Since everyone seems to be listing Linux distribution tools, I'll add gnome flatpak to the list as well: 

http://flatpak.org/

develar:
Do you like flatpack or just report issue ;)?

kunkinkan:
AppImage, Flatpak and Snappy Apps are the 3 modern solutions at the moment. That's why I created the issue :-)

develar:
Moved to backlog to keep issue list clear.

develar:
AppImage #504 supported.

grafenhofer:
Flatpak is taking up in usage recently. It supports some features missing from appimage (support for automatic updates e.g. via proper integration in gnome software, KDE support seems to be coming,...). It could be time to reconsider the "backlog" tag.

prcastro:
I agree. Flatpak is becoming the standard not only in GNOME, but within KDE and probably ElementaryOS environments. Seems also to be a non-centralized way to keep applications reproducible and sandboxed, and seems way more secure than the alternatives. The number of contributors is also bigger than AppImage and Snap. GNOME Apps, Spotify, LibreOffice, Skype and Telegram already have Flatpak builds.

black-snow:
Anyone willing to tackle this? We have deb, rpm, AppImage and Snappy it would be a shame to not have flatpak as well.

prcastro:
Maybe these links are of interest:

https://github.com/endlessm/electron-flatpak-base-app
http://blog.manuq.com.ar/posts/building-electron-apps-offline-for-flathub/

develar:
Sorry, moved to backlog for now.

lbssousa:
Flatpak is the last missing piece for electron-builder to become a trully complete tool for packaging Electron apps. Is it too hard to port electron-forge Flatpak implementation to electron-builder?

laurent22:
A Flatpak version is also something that people frequently ask for in my application, so support for it would be very welcome.

develar:
As I am going to switch to vanilla gnome Linux from macOS, flatpack will be supported this/next month.

spiral2k:
any updates? thanks for the amazing work!

goncalossilva:
@develar If you have yet to begin working on it, this is likely of interest: http://docs.flatpak.org/en/latest/electron.html

julian-alarcon:
[...]

PR #5711, opened 2021-03-14
Add support for building Flatpak bundles

vially:
Add a new Linux target for exporting applications in Flatpak format. 

### Implementation details

The current implementation is delegating the actual build process to [@malept/flatpak-bundler](https://github.com/malept/flatpak-bundler), which is a Node wrapper over the [flatpak-builder](https://docs.flatpak.org/en/latest/flatpak-builder.html) tool.

Similar to the other Linux targets (e.g.: AppImage, snap, etc) the Flatpak target allows the user to customize some of the options. However, in order to make it as easy as possible to get started, some sensible defaults are also provided out-of-the-box:
- Runtime: `org.freedesktop.Platform`
- Runtime version: `20.08`
- SDK: `org.freedesktop.Sdk`
- Base: [org.electronjs.Electron2.BaseApp](https://github.com/flathub/org.electronjs.Electron2.BaseApp)
- Base version: `20.08`
- Finish Args:
```json5
[
    // Wayland/X11 Rendering
    "--socket=wayland",
    "--socket=x11",
    "--share=ipc",
    // Open GL
    "--device=dri",
    // Audio output
    "--socket=pulseaudio",
    // Read/write home directory access
    "--filesystem=home",
    // Allow communication with network
    "--share=network",
    // System notifications with libnotify
    "--talk-name=org.freedesktop.Notifications",
]
```

Also, an `electron-wrapper` script is automatically added to the bundle which uses [zypak](https://github.com/refi64/zypak) for sandboxing the applications.

When the electron version supports it (version 12 and above), the Wayland flags (`--enable-features=UseOzonePlatform --ozone-platform=wayland`) are also automatically added to this wrapper script in order to provide a better out-of-the-box experience for Wayland users.

#### Limitations and possible future work

**Note**: This pull-request only makes it possible to export `*.flatpak` bundles. 

In the future, it might make sense to also implement a Flatpak publisher (similar to the Snap Store publisher) that makes it possible to publish the application to a repository (e.g.: [Flathub](https://flathub.org/)).

However, it's not entirely clear how that would work so it is **not** part of this pull-request. But it is something that should be considered as a possible next step.

Related issues: #512


packages/app-builder-lib/src/targets/FlatpakTarget.ts
```diff
@@ -0,0 +1,169 @@
...
+    "--socket=x11",
+    "--share=ipc",
+    // Open GL
+    "--device=dri",
+    // Audio output
+    "--socket=pulseaudio",
+    // Read/write home directory access
+    "--filesystem=home",
+    // Chromium uses a socket in tmp for its singleton check
+    "--env=TMPDIR=/var/tmp",
```
refi64, line 147: Note that most Flatpaks instead set `TMPDIR=$XDG_RUNTIME_DIR/app/$FLATPAK_ID` instead.

packages/app-builder-lib/src/targets/FlatpakTarget.ts
```diff
@@ -0,0 +1,169 @@
...
+    "--talk-name=org.freedesktop.Notifications",
+  ],
+}
+
+function getElectronWrapperScript(executableName: string, isWaylandSupported: boolean): string {
+  return isWaylandSupported
+    ? `#!/bin/sh
+
+if [ "\${XDG_SESSION_TYPE}" == "wayland" ]; then
+    zypak-wrapper "${executableName}" --enable-features=UseOzonePlatform --ozone-platform=wayland "$@"
```
refi64, line 160: Is this a good default yet? I love the idea of Wayland support, but last time I tried it on Chromium itself it still had some rough edges (mostly in terms of font rendering & cursor scaling)...maybe make it opt-in by the packager to encourage use while still avoiding accidental breakage?

vially: That's a good point. I've changed it to an opt-in option which is `false` by default.

packages/app-builder-lib/src/targets/FlatpakTarget.ts
vially: Nice catch! Initially I've copied these values from [@malept/electron-installer-flatpak](https://github.com/malept/electron-installer-flatpak/blob/097e08654dfc8d0eeeee1ad3ed617e17e607c0b8/README.md#optionsfinishargs) without giving it too much thought.

But now that you've mentioned I've spent some time looking into it and I can't seem to find these chromium socket files anywhere (neither in `/tmp` nor in `$XDG_RUNTIME_DIR`). So now I'm wondering whether this config is still needed or is it just some artifact of the past where the documentation hasn't kept up with the latest changes.

So unless someone can confirm that this config is still needed I'm tempted to just remove it (0400b402da6174e4a872f4e3a11caa53baee2db1) and wait until there's a compelling reason to bring it back.

vially:
FYI I've rebased and force-pushed to fix some small conflicts introduced by the recent migration from Yarn to pnpm (#5710).

packages/app-builder-lib/src/targets/FlatpakTarget.ts
refi64: Try running with `ls -a`, the files are often hidden by default. At minimum, I want to say it's at least a pretty decent chunk of Electron apps that do this.

vially: You're right! It turns out the `TMPDIR` is actually used by Electron for creating the singleton socket.

However, this socket is only created when the application is using the [app.apprequestsingleinstancelock](https://www.electronjs.org/docs/api/app#apprequestsingleinstancelock) API. My previous tests were done using a sample application which wasn't using this API and that's why the socket was missing.

Therefore I've pushed another fix (https://github.com/electron-userland/electron-builder/pull/5711/commits/9ffc66f8da44c9ffc55006ccc0ae62e12e5a394c) that sets `TMPDIR=$XDG_RUNTIME_DIR/app/$FLATPAK_ID` in the wrapper script. I've done a small test and the socket file is indeed created in the right place (`$XDG_RUNTIME_DIR/app/$FLATPAK_ID`) and it gets cleaned up properly when the last instance of the app gets closed.

mmaietta:
Hi @vially! Thanks for the contribution!
We should have some tests to back this up, something like these: https://github.com/electron-userland/electron-builder/tree/master/test/src/linux

You might be able to literally just copy one of the test files and switch up the target to flatpak

vially:
@mmaietta Thanks for the feedback!

I added some tests (da167a1) and a fix for an issue uncovered by the tests (d7d2f78).

I think it's worth mentioning that these tests are **really** slow. On my machine it's taking around 5 minutes just for these Flatpak tests to run. For what it's worth it's pretty close to what I get when running the underlying [flatpak-builder](https://docs.flatpak.org/en/latest/flatpak-builder.html) commands directly, so I don't think this is due to some overhead introduced in `electron-builder`.

I'm not sure what would be considered a reasonable time frame for these tests, but it may be worth considering disabling them by default if you think this is unacceptable (similar to [these snap tests](https://github.com/electron-userland/electron-builder/blob/a879ba7f02ea704084b9672d6b9c9a47b7e5c27d/test/src/linux/snapHeavyTest.ts)).

mmaietta:
Thanks for adding the tests!

I'll need to figure out a way for us to get some more build node coverage for these heavy tests (similar as was done for mas and dmg tests)

@refi64 you're our resident expert! 😉  Could you run a final review please?

refi64:
😅 I'm admittedly not too familiar with electron-builder, but at a glance everything here does seem to check out.

mmaietta:
I joined the maintainer crew only a month or so ago :sweat_smile:

Thanks for reviewing the flatpak implementation! That helps a lot.
Might reach out to you in the future if any issues show up because I'm unfamiliar in this subject. If you wouldn't mind, can you officially hit the "approve" button on the PR?

refi64 (approved):

refi64:
Done!

mmaietta:
@vially would you be willing to also add documentation for this new feature to electron-builder `docs` branch?

vially:
> would you be willing to also add documentation for this new feature to electron-builder `docs` branch?

I think I may need some assistance in doing that. Just to make sure I'm on the right track here, the documentation is supposed to be auto-generated from the TypeScript comments and/or the [schema](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/scheme.json), right?

If that's the case then I must be doing something wrong because I didn't manage to get the scripts to run. I also get the feeling the [documentation](https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#documentation) may not be entirely up-to-date.

These are some of the things I tried that didn't work:

### mkdocs build

Out of the box it fails with an error complaining about the usage of IP address [0.0.0.0](https://github.com/electron-userland/electron-builder/blob/6b1be73e4510ac099d9e35f6339d398eb4a03f1a/mkdocs.yml#L16). Once I change that to `127.0.0.1` it get past that error but it now fails with Python stack-trace ending with: `AttributeError: 'dict' object has no attribute 'split'`. 

Maybe an older version of `mkdocs` and/or Python should be used (I've only tried with `mkdocs 1.1.2` and `Python 3.9.2`)?

### pnpm run docs

Initially it fails with some errors complaining about some missing node modules (`source-map-support` and `path-sort`). Once I get past those (`pnpm add -D path-sort source-map-support -w`) then the [jsdoc2md](https://github.com/electron-userland/electron-builder/blob/6b1be73e4510ac099d9e35f6339d398eb4a03f1a/package.json#L27) step fails with lots of errors (e.g.: `ERROR: Unable to parse a tag's type expression for source file...`) pointing to JSDoc comments that look like `Platform<>` (supposedly because this is invalid JSDoc code?).

For example, this is one such problematic comment: 

```js
/**
 * @param {Array<Platform<>>} platforms
 * @param {string | "undefined"} type
 * @param {string | "undefined"} arch
 * @returns {Map<Platform<> | Map<Arch | Array<string>>>}
 */
export function createTargets() {}
```

So, my main question is, are these actually the right commands for generating the documentation? And if they are, can anyone confirm that they have been able to run them *recently* on the `master`/`docs` branches?

razzeee:
It would be really handy, if this was documented

alariej:
Documentation of FlatpakOptions (https://www.electron.build/configuration/configuration.html#FlatpakOptions) would definitely be great. In the meantime, a simple example config file could be very useful. Has someone found one already?

alariej:
@mmaietta Thanks for coming back. I had seen this example, but it unfortunately doesn't contain any specific configuration for producing a flatpak build. And the link https://www.electron.build/configuration/configuration.html#FlatpakOptions seems to be pointing nowhere.

I've nevertheless tried to run `electron-builder --linux flatpak --config builder-flatpak.json` where builder-flatpak.json contains:

```
{
    "appId": "chat.quadrix",
    "productName": "quadrix",
    "linux": {
        "category": "Internet",
        "target": [
            "flatpak"
        ],
        "icon": "src/resources/logos/logo256.png"
    },
    "flatpak": {
        "description": "Minimal, simple, multi-platform chat app for the Matrix protocol"
    },
    "files": [
        "dist-web/**/*",
        "electron-main.js"
    ],
    "directories": {
        "output": "dist/"
    }
}
```
With this minimal config, I get the following output / error:

```
  • electron-builder  version=22.14.13 os=5.13.0-35-generic
  • loaded configuration  file=/home/.../quadrix/builder-flatpak.json
  • packaging       platform=linux arch=x64 electron=17.1.2 appOutDir=dist/linux-unpacked
  • building        target=flatpak arch=x64 file=dist/quadrix-0.6.4-x86_64.flatpak
  ⨯ flatpak failed with status code 1  failedTask=build stackTrace=Error: flatpak failed with status code 1
    at ChildProcess.<anonymous> (/home/.../quadrix/node_modules/@malept/flatpak-bundler/index.js:71:16)
    at ChildProcess.emit (node:events:390:28)
    at maybeClose (node:internal/child_process:1064:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:301:5)

```
`flatpak-builder` was installed on my Ubuntu system with `sudo apt install flatpak-builder` (which installed version 1.0.14)

As you can see, I'm new to flatpak, so I'm kind of stuck here at the moment. Any help or link to relevant info would be appreciated.

mmaietta:
Interesting that it's not linking anywhere on the website. Here's the interface/options for the flatpak config to get you in the right direction https://github.com/electron-userland/electron-builder/blob/a138a86fb7b59098f5dac0c0a6b59c034eb9b222/packages/app-builder-lib/src/options/linuxOptions.ts#L133-L232

cbartondock:
Just in case it helps anyone I was able to build flatpaks while getting essentially get all the benefits of using electron-builder (e.g. native module support) using an additional step in the build process with [@malept/electron-installer-flatpak](https://www.npmjs.com/package/@malept/electron-installer-flatpak). The relevant npm scripts are:
```json
  "scripts": {
    "postinstall": "electron-builder install-app-deps",
    "build:main": "cross-env NODE_ENV=production webpack --config ./webpack/main.js",
    "build:renderer": "cross-env NODE_ENV=production webpack --config ./webpack/renderer.js",
    "build:dist": "rimraf dist && npm run build:main && npm run build:renderer",
    "build:linux": "electron-builder --linux --ia32 --x64 --dir && cp LICENSE release/linux-unpacked/ && cp electron-version release/linux-unpacked/version",
    "build:flatpak": "electron-installer-flatpak --src release/linux-unpacked --dest release/ --arch x64",
  },
```
Then to make a flatpak package I simply do:
``` 
npm run build:dist; npm run build:linux; npm run build:flatpak
```

Essentially electron-builder is used to create an unpacked version of the build (`--dir`),  the files `LICENSE` and `version` that `@malept/electron-installer-flatpak` expects are copied into it, and then `@malept/electron-installer-flatpak` creates the flatpak (it is fully configurable - check their docs). The contents of `electron-version` is literally just the electron version number, or in my case `13.6.6`.

PR #5712, opened 2021-03-15
feat: add custom macro for checking uninstallation results

dlon:
Due to https://github.com/electron-userland/electron-builder/pull/5292, installers created using the default NSIS script ignore any errors returned by uninstallers, for example if files cannot be replaced, or if the uninstaller has been deleted. This behavior is often unwanted.

The changes here allow users to define `customUnInstallCheck` to handle such cases. In `customUnInstallCheck`, the error flag is set if the uninstaller cannot be executed at all (probably because it is missing). If the error flag is not set, `$R0` contains the return status/error level, which should be set to 0 if the uninstaller exits successfully.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/include/installUtil.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/installUtil.nsh
@@ -105,6 +105,36 @@ Var /GLOBAL isTryToKeepShortcuts
   ${endif}
 !macroend
 
+Function handleUninstallResult
+  Var /GLOBAL rootKey_uninstallResult
+  Exch $rootKey_uninstallResult
+
+  ${if} "$rootKey_uninstallResult" == "SHELL_CONTEXT"
+    !ifmacrodef customUnInstallCheck
+      !insertmacro customUnInstallCheck
+      Return
+    !endif
+  ${elseif} "$rootKey_uninstallResult" == "HKEY_CURRENT_USER"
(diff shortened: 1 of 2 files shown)


mmaietta:
Thanks for the PR @dlon! The changes seeem good to me. What's the best way for me to test this locally?

dlon:
Create an installer using something like this config:

```
nsis: {
    oneClick: false,
    perMachine: true,
    allowToChangeInstallationDirectory: false,
    include: 'install.nsh',
}
```

In `install.nsh`, add:

```
!macro customUnInstallCheck

  IfErrors 0 customUnInstallCheck_CheckCode

  MessageBox MB_OK "Uninstaller failed to run"
  SetErrorLevel 5
  Abort

  customUnInstallCheck_CheckCode:

  ${If} $R0 != 0
    MessageBox MB_OK "Uninstaller returned an error"
    SetErrorLevel 4
    Abort
  ${EndIf}

!macroend
```

To test the non-failing case, just run the installer twice (so that it will uninstall before re-installing). To test the case where the uninstaller fails to run, you can install it and delete uninstaller executable before running it again. To test the return code check, you can add the following to `install.nsh` before running the installer twice:

```
!macro customRemoveFiles

  SetErrorLevel 5
  Abort

!macroend
```

mmaietta:
@develar the changes LGTM atm and I'll run an example test later once I can get my Win env set up. Any thoughts/concerns from your side?

PR #5716, opened 2021-03-16
fix: fix husky not create hooks and improve lint when commit

codeudan:
1. ensure husky hooks install. husky5 changed behavior  see:https://github.com/typicode/husky/issues/866 
2. use "packages/**/*.ts": "eslint", this only lint staged file instead of run "pnpm lint", will lint more faster when commit


package.json
```diff
@@ -51,7 +51,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "fs-extra": "^9.1.0",
     "globby": "^11.0.2",
-    "husky": "5.1.3",
+    "husky": "^3.1.0",
```
develar, line 54: Why dependency was downgraded?

codeudan: huksy5 need run 1.husky install  2. husky add .husky/pre-commit   3.husky add .husky/pre-push to create pre-commit and pre-push  hooks.

In short, husky5 can not create hooks completely automatically, need more codes in postinstall. If you want add a new hook, you need modify script to create a hook like 'husky add .husky/pre-push' .

detail: typicode/husky#866

in my test, husky3 work nice and Ieasy to use

develar: I am not happy to use outdated version of husky. It has number of dependencies (see https://github.com/typicode/husky/blob/v3.0.6/package.json#L53) and over time we will have issues due to outdated dependencies.
Will be great if you can configure husky 5 properly or use another package.

codeudan:
@develar  The huksy version is 5 now😀

codeudan:
@mmaietta Would you like to help review it?😀

mmaietta (approved):
Just tested your branch out locally. LGTM 👍

PR #5727, opened 2021-03-22
feat: add Electron branding options

samuelmaddock:
The Electron project contains a [BRANDING.json](https://github.com/electron/electron/blob/665ac6f9c830edbf85596369b49844e1b910c59a/shell/app/BRANDING.json) file which will modify the resulting executable file names. A project I'm working on does exactly this in a fork of Electron.

Exposing these branding options allows Electron forks to build properly if branding has changed.

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -897,6 +897,18 @@
       },
       "type": "object"
     },
+    "ElectronBrandingOptions": {
+      "additionalProperties": false,
+      "properties": {
+        "projectName": {
+          "type": "string"
+        },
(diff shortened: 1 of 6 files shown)


mmaietta (approved):
I love this and had the exact same issue to work around, I also didn't realize that the BRANDING.json was available for changing exec names though. We had it manually patched in our fork. 😅 

Approved.

Will wait a little before merging in case anyone else would like to chime in with feedback

samuelmaddock:
@mmaietta sorry to bother, but is there any rough estimate on a schedule for publishing recent changes?

PR #5728, opened 2021-03-23
feat: add preInstall macro for customise NSIS script

altjz:
For now, Custom NSIS install Scripts include `customHeader`，`preInit`, `customInit`, `customInstall`.

Some tasks must be executed before install application files.

For example, delete some files those existed in the `$INSTDIR`. ( Sometime the uninstaller doesn't work if between Applications)

```nsis
!macro preInstall
  MessageBox MB_OK "preInstall"
  RMDir /r "$INSTDIR\resources\"
  RMDir /r "$INSTDIR\locales\"
  Delete "$INSTDIR\logo.ico"
  Delete "$INSTDIR\uninst.exe"
!macroend
```

`customInstall` is too late for the clean job. That would be nice if we could provide the `preInstall` macro.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/installSection.nsh
+++ b/packages/app-builder-lib/templates/nsis/installSection.nsh
@@ -57,6 +57,9 @@ SetOutPath $INSTDIR
   File /oname=uninstallerIcon.ico "${UNINSTALLER_ICON}"
 !endif
 
+!ifmacrodef preInstall
+  !insertmacro preInstall
+!endif
 !insertmacro installApplicationFiles
 !insertmacro registryAddInstallInfo
 !insertmacro addStartMenuLink $keepShortcuts


mmaietta:
Hi @altjz, thanks for the PR!
 
> For example, delete some files those existed in the $INSTDIR. ( Sometime the uninstaller doesn't work if between Applications)

Just curious, Is this the only use-case/bug we're solving for or are there additional cases that you will be using this for?

altjz:
> Hi @altjz, thanks for the PR!
> 
> > For example, delete some files those existed in the $INSTDIR. ( Sometime the uninstaller doesn't work if between Applications)
> 
> Just curious, Is this the only use-case/bug we're solving for or are there additional cases that you will be using this for?

Many thanks for your reply, I am glad to help.

Some cases I may use it in my application.

1. Check the files or subdirectory of $INSTDIR, (many primary users will install the application in desktop or some danger area)
2. Check the running processes, I know electron-builder integrate one macro something like this, but that macro is only for the same Application, If we have some background services, we still need to stop them before the installation.
3. Check the Registry before the installation, It's useful in some special cases, such as checking the if the program was not uninstalled completely last time. electron builder is able to do this but only for current Application, same GUID, same ProductName.

There are still many cases, The most important function of `preInstall` macro, I think is that **Ensure everything is working fine before the installation**, under `preInit`, `customInit` section, user still can cancel the installation(OneClick option is false). So I think the best time for these clean job is **User clicks the Next button, before copy application files**

mmaietta (approved):
Thanks for the details. Seems quite reasonable to me for this to be added.

Approved. Will leave this Open for a few days to see if any others have feedback to provide.

Thanks for your contribution!

mmaietta:
@altjz , just curious, is this PR also solving the same issue for you?
https://github.com/electron-userland/electron-builder/pull/5712
Seems to solve items 2 and 3 since any ongoing background services or previous files would have required an installation to already exist, potentially due to a failed uninstall/update.

I'm not sure if we need both these PRs if we're solving (one of) the same problems. What are your thoughts?

Issue #3012, opened 2018-06-14
Problem building multiple mac targets: TypeError: pattern.startsWith is not a function


paulingalls:
Electron Builder:
* **2.15.3**: 

Electron 2.0.2

Target:
* **dmg,mas**: 

I'm running into a problem when I try to build for both dmg,zip and mas.  

Basically
    "target": ["dmg", "zip"] // works
    "target": ["mas"] // works
    "target": ["mas", "dmg", "zip"] // shows the error below

Let me know if you need more info...

```
  TypeError: pattern.startsWith is not a function
    at FileMatcher.normalizePattern (/Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/fileMatcher.ts:42:17)
    at FileMatcher.patterns.map.it (/Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/fileMatcher.ts:37:54)
    at Array.map (<anonymous>)
    at new FileMatcher (/Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/fileMatcher.ts:37:39)
    at /Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/util/AppFileCopierHelper.ts:123:21
    at Generator.next (<anonymous>)
From previous event:
    at _bluebirdLst.default.mapSeries (/Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/util/AppFileCopierHelper.ts:120:16)
From previous event:
    at /Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/util/AppFileCopierHelper.ts:120:32
    at Generator.next (<anonymous>)
From previous event:
    at copyNodeModules (/Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/out/util/AppFileCopierHelper.js:319:18)
    at /Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/platformPackager.ts:261:42
    at Generator.next (<anonymous>)
From previous event:
    at then (/Users/paulingalls/src/fanzo/desktop/ripl-electron/node_modules/electron-builder-lib/src/platformPackager.ts:257:14)
From previous event:
[...]

PR #5739, opened 2021-03-26
Fix Problem building multiple mac targets:TypeError: pattern.startsWith is not a function

gaodeng:
more info https://github.com/electron-userland/electron-builder/issues/3012

### DIFF
--- a/packages/builder-util/src/DebugLogger.ts
+++ b/packages/builder-util/src/DebugLogger.ts
@@ -29,7 +29,7 @@ export class DebugLogger {
     }
 
     if (Array.isArray(o[lastName!])) {
-      o[lastName!].push(value)
+      o[lastName!] = [...o[lastName!], value]
     } else {
       o[lastName!] = value
     }


mmaietta:
This LGTM I think, but I'm wondering if we should cover this with some basic unit tests in fileTest.ts to validate the file pattern merges.  I was under the impression that `value` was an array.

Do you have a sample config for this?
I was trying to come up with some unit tests over the weekend, then my hackintosh crashed...so I'm working on fixing that

gaodeng:
Hi @mmaietta , sorry for not replying in time, I've been too busy lately

Actually fileTest.ts is fine. The root of the problem is DebugLogger.ts, which changes the input parameters.

I'm afraid I can't simply give a configuration file to reproduce the problem. This is because the problem is caused by compiling mac and mas in parallel, so the problem is related to the performance of your machine, and the same configuration that reproduces the problem on my machine may not reproduce it on your machine. 

Fortunately, we already know what the problem is, and it's easy to fix

If you still want to reproduce the problem, you can try @mifi 's https://github.com/mifi/lossless-cut/pull/709, which is an open source project, and see if you can reproduce it on your machine

mmaietta:
👍 Can you please rebase of latest master? Then we should be good to go

gaodeng:
> 👍 Can you please rebase of latest master? Then we should be good to go

done

mmaietta (approved):

PR #5752, opened 2021-03-30
Adding msi project created hook.

p120ph37:
Add a hook to allow modification of `project.wxs` before packaging.

Same idea as #4613 but for MSIs.

### DIFF
--- a/packages/app-builder-lib/src/configuration.ts
+++ b/packages/app-builder-lib/src/configuration.ts
@@ -199,11 +199,11 @@ export interface Configuration extends PlatformSpecificBuildOptions {
    */
   readonly afterAllArtifactBuild?: ((context: BuildResult) => Promise<Array<string>> | Array<string>) | string | null
   /**
-   * Appx manifest created on disk - not packed into .appx package yet.
+   * MSI project created on disk - not packed into .msi package yet.
    */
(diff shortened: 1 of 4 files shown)


packages/app-builder-lib/src/configuration.ts
```diff
@@ -201,6 +201,10 @@ export interface Configuration extends PlatformSpecificBuildOptions {
   /**
    * Appx manifest created on disk - not packed into .appx package yet.
    */
+  readonly msiProjectCreated?: ((path: string) => Promise<any> | any) | string | null
+  /**
+   * MSI project created on disk - not packed into .msi package yet.
+   */
   readonly appxManifestCreated?: ((path: string) => Promise<any> | any) | string | null
```
mmaietta, line 208: Looks like the documentation got flip-flopped here
Can you please swap the two lines `appxManifestCreated` and `msiProjectCreated` so they match your docs? 🙂

mmaietta (changes requested):

packages/app-builder-lib/src/configuration.ts
p120ph37: Good catch - sure...

mmaietta (approved):

PR #5811, opened 2021-04-19
fix after cancel download,the file haven't remove bug

liuestc:
when I cancel download,  the temp-update-download file is not close, the temp file is occupied . so when next time I try to download again , the last temp-file cant't be removed, when I cancel many times, It will generate many temp-file.

### DIFF
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -446,6 +446,7 @@ function configurePipes(options: DownloadCallOptions, response: IncomingMessage)
   let lastStream = response
   for (const stream of streams) {
     stream.on("error", (error: Error) => {
+      fileOut.close()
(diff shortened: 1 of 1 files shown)


packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -446,6 +446,7 @@ function configurePipes(options: DownloadCallOptions, response: IncomingMessage)
   let lastStream = response
   for (const stream of streams) {
     stream.on("error", (error: Error) => {
+      fileOut.close()
```
mmaietta, line 449: Why is no callback being supplied here? (as it is at line 457)
```
;(fileOut.close as any)(options.callback)
```
Shouldn't we be doing so?

Something like:
```
const closeFile = () => { ;(fileOut.close as any)(options.callback) }
```
Then use `closeFile()` both here and at L457.

PR #5813, opened 2021-04-19
Allow specifying additional WiX compiler options

p120ph37:
This is necessary to support certain customizations possible via #5752
For example, if you add a `<util:CloseApplication/>` tag to `project.wxs`, you must also specify some additional arguments to `candle.exe` and `light.exe` in order to include the appropriate extension DLLs.
This would also be useful if you have your own `CustomAction` DLLs that you need to link into your MSI.


mmaietta:
Thanks for the work!

Would you be willing to add testing for this? Probs in: `test/src/windows/msiTest.ts`

p120ph37:
New test shown passing here: https://github.com/p120ph37/electron-builder/runs/2451384511?check_suite_focus=true#step:9:202

bryantb2:
This would be an amazing feature to have. I know lots of folks, myself included, take issue with the lack of .MSI customization in Electron. ETA on how long before this is merged up, documented, and released?

p120ph37:
@bryantb2 , you can see a basic usage example in the `msiTest.ts` diff on this PR.  In that example, I use the `msiProjectCreated` hook to access the `project.wxs` file after it has been generated by `electron-builder`, but before it has been fed into WiX.  I then use `fs` to modify the file, inserting some custom XML just before the `</product>` tag.  This additional XML refers to a wix-util feature (I could have added the `xmlns` attribute to the root tag, but it was simpler to add it to the single tag I was adding), and I use the 'additionalWixArgs` property to add the needed arguments for WiX to compile this modified XML properly.  Hopefully the example there is useful enough for you to get started once this is released.  As far as when it will get merged, and when the next release will happen, I'm not sure.  @mmaietta seems to check and merge PRs once a week or so, and releases seem to happen every few months...

bryantb2:
@p120ph37 Fantastic work! I will check this out.

mmaietta (approved):

mmaietta:
Nice work! 

Would you be willing to open a PR for adding content to the docs? It's on the `docs` branch of this repo
https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#documentation

p120ph37:
I'm not sure how to correctly add to the docs.  It appears that a lot of the files in the docs branch contain autogenerated sections, which I assume are supposed to be produced via tsdoc or something - shouldn't the jsdoc/tsdoc comments I already created end up automatically pulled into the docs branch at some point?  I did try `pnpm run docs`, and it seems to be broken right now...  I guess that would explain why the docs site seems out-of-date compared to the comments in the source.

Issue #5841, opened 2021-04-29
getCertificateFromStoreInfo Should use -NoProfile


djbreen7:
* **Version**: 22.10.5

* **Electron Version**: 12.0.5
* **Electron Type (current, beta, nightly)**:

* **Target**: Windows 10

I have a custom powershell profile.ps1 that has the following line of code at the top:

```powershell
Write-Host "Loading Modules..."
```

This interferes with code signing locally and produces the error: `Unexpected token L in JSON at position 0`. The `L` is from "Loading Modules" in my powershell profile.

Removing the command from my profile above works as a workaround.

Recommended fix is to pass the `-NoProfile` flag.

**Full Error**

```
  ⨯ Unexpected token L in JSON at position 0  stackTrace=
    SyntaxError: Unexpected token L in JSON at position 0
    SyntaxError: Unexpected token L in JSON at position 0
        at JSON.parse (<anonymous>)
        at getCertificateFromStoreInfo (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\codeSign\windowsCodeSign.ts:113:73)
        at processTicksAndRejections (internal/process/task_queues.js:93:5)
        at WinPackager.signAndEditResources (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\winPackager.ts:292:106) 
    From previous event:
        at processImmediate (internal/timers.js:461:21)
    From previous event:
        at WinPackager.signApp (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\winPackager.ts:357:27)
        at WinPackager.doSignAfterPack (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\platformPackager.ts:270:16)  
        at WinPackager.doPack (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\platformPackager.ts:255:7)
        at WinPackager.pack (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\platformPackager.ts:115:5)
        at Packager.doBuild (C:\code\app\node_modules\app-builder-lib\src\packager.ts:434:9)
        at executeFinally (C:\code\App\app-desktop-v2\node_modules\builder-util\src\promise.ts:12:14)
        at Packager._build (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\packager.ts:369:31)
        at Packager.build (C:\code\App\app-desktop-v2\node_modules\app-builder-lib\src\packager.ts:333:12)
[...]

PR #5842, opened 2021-04-29
Fix: Use -NoProfile Flag on getCertFromStoreInfo

djbreen7:
#5841

### DIFF
--- a/packages/app-builder-lib/src/codeSign/windowsCodeSign.ts
+++ b/packages/app-builder-lib/src/codeSign/windowsCodeSign.ts
@@ -111,7 +111,7 @@ export async function getCertificateFromStoreInfo(options: WindowsConfiguration,
   // ExcludeProperty doesn't work, so, we cannot exclude RawData, it is ok
   // powershell can return object if the only item
   const rawResult = await vm.exec("powershell.exe", [
-    "Get-ChildItem -Recurse Cert: -CodeSigningCert | Select-Object -Property Subject,PSParentPath,Thumbprint | ConvertTo-Json -Compress",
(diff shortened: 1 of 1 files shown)


mmaietta:
Thanks for your contribution!

Just looking to understand. Can you explain a bit further what's occurring with the addition of `-NoProfile`? Is it specific now due to using `Write-Host`?

djbreen7:
@mmaietta No problem!

Sorry if I explain stuff you already know.

The `-NoProfile` flag prevents PowerShell from loading user profile. Profiles usually include things like aliases, quick functions, and some modules. If you have a Windows machine, you can peek at yours by entering `code $profile` into your terminal.

A common use case for a dev to modify their profile is to load `oh-my-posh` (a fork of `oh-my-zsh` that works in powershell). I'm sure a lot of profiles start out looking something like mine

```powershell
function Invoke-LoadModules() {
    Write-Host "Loading Modules..."
    Import-Module posh-git
    Import-Module oh-my-posh
    Set-Theme agnoster
    Write-Host "Loading Modules Complete!"
}

Invoke-LoadModules
```

So to answer your question, introducing `-NoProfle` could potentially improve performance since it will bypass loading this file. It should also prevent code signing from breaking if devs have imported any other modules that have output.

Note that it also seems to be the norm in this codebase everywhere else `powershell.exe` is used:
- https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/src/cli/create-self-signed-cert.ts#L31
- https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts#L34-L42
- https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts#L104

mmaietta (approved):

Issue #4676, opened 2020-02-14
Download block maps  Cannot download differentially, fallback to full download - Windows 10 - Github Private -Error: Maximum allowed size is 50 MB


hmpargi:
**electron-builder version: ^22.3.2"
electron version: ^8.0.0"
electron-updater version: 4.2.0"
Target: Windows 10
 Github Private repository**

``` 

`[2020-02-13 22:17:34.744] [info] Checking for update
[2020-02-13 22:17:37.465] [info] Found version 1.0.0 (url: psstracker-Setup-1.0.0.exe)
[2020-02-13 22:17:37.465] [info] Downloading update from psstracker-Setup-1.0.0.exe
[2020-02-13 22:17:37.469] [info] No cached update info available
[2020-02-13 22:17:37.471] [info] Download block maps (old: "https://api.github.com/repos/myusername/psstracker/releases/assets/18021931.blockmap", new: https://api.github.com/repos/myusername/psstracker/releases/assets/18021931.blockmap)
[2020-02-13 22:17:37.934] [error] Cannot download differentially, fallback to full download: Error: Maximum allowed size is 50 MB
    at Object.responseHandler (C:\Users\har\Desktop\Myapps\epssapp\psstracker\dist\win-unpacked\resources\app.asar\node_modules\electron-updater\node_modules\builder-util-runtime\out\httpExecutor.js:239:26)
    at ClientRequest.<anonymous> (C:\Users\har\Desktop\Myapps\epssapp\psstracker\dist\win-unpacked\resources\app.asar\node_modules\electron-updater\node_modules\builder-util-runtime\out\httpExecutor.js:298:17)
    at ClientRequest.emit (events.js:210:5)
    at SimpleURLLoaderWrapper.<anonymous> (electron/js2c/browser_init.js:2510:12)
    at SimpleURLLoaderWrapper.emit (events.js:210:5)`
```

Please help why I am having this issue. any idea or fix please. its GitHub Pvt repo. i am trying to make it work for windows 10.

hmpargi:
This is weird, so I build my program with the following dependencies by downgrading electron version and auto-updates working on windows.

"electron": "^5.0.13",
"electron-builder": "^22.3.2"


but whenever I compile with 8.0.0 it doesn't download

rancorfloydz:
I also got the same issue 
**[error] Cannot download differentially, fallback to full download: Error: Maximum allowed size is 50 MB**

[...]

Issue #5505, opened 2020-12-27
Auto-Update invalid header and maximum allowed size.


Daniele122898:
* **electron-builder**: 22.9.1

* **Electron Version**: 11.1.1
* **Electron Type (current, beta, nightly)**: current

* **electron-updater**: 4.3.5

* **Target**: Windows

So this is the code i use for updating: 
```ts
// Disable auto downloading of updates
autoUpdater.autoDownload = false;

const updateApp = (mainWindow: BrowserWindow) => {
  // Check for update
  autoUpdater.checkForUpdates()
    .catch(e => logger.error('Check for updates failed', e));

  autoUpdater.on('update-available', () => {
    // prompt user to start download
    dialog.showMessageBox({
      type: 'info',
      title: 'Update available',
      message: 'A new version of Media Dashboard is available. Do you want to update now?',
      buttons: ['Update', 'No']
    }).then(res => {
      const buttonIndex = res.response;

      if (buttonIndex === 0) {
        autoUpdater.downloadUpdate()
          .catch(e => logger.error('Update download failed', e));
      }
    });

  });

  // Listen for update download finished
  autoUpdater.on('update-downloaded', () => {
    mainWindow.setProgressBar(-1);
    dialog.showMessageBox({
      type: 'info',
      title: 'Update ready',
      message: 'Update downloaded. Install update & restart now?',
      buttons: ['Yes', 'Later']
    }).then(res => {
      const buttonIndex = res.response;

      // install and restart
      if (buttonIndex === 0) {
        autoUpdater.quitAndInstall(false, true);
      }
    });
  });

  // Listen for download progress
  autoUpdater.on('download-progress', (progress, bytesPerSecond, percent) => {
    mainWindow.setProgressBar(progress);
  });
};
```

When i start my electron app it correctly tells me that there is a new version of the app. Once i click on Update i receive this error message: 
![image](https://user-images.githubusercontent.com/20451770/103177429-c009a680-487a-11eb-8335-8c3545952785.png)

The release on github looks like this:
![image](https://user-images.githubusercontent.com/20451770/103177438-d152b300-487a-11eb-9ec7-7a77812eea80.png)

This is the electron-builder.json:
```json
{
  "productName": "Media Dashboard",
  "directories": {
    "output": "release/"
  },
    "files": [
        "**/*",
        "!**/*.ts",
        "!*.code-workspace",
        "!LICENSE.md",
        "!package.json",
        "!package-lock.json",
        "!src/",
[...]

PR #5843, opened 2021-04-29
Extend http file limit

mariusraht1:
Related to #4676 and propably some more

As an electron app takes up more than 50 MB pretty easily the updater stops working due to the very low chunk size limit of currently 50 MB.

### DIFF
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -218,8 +218,8 @@ Please double check that your authentication token is correct. Due to security r
             if (contentLength != null) {
               const size = parseInt(contentLength, 10)
               if (size > 0) {
-                if (size > 52428800) {
-                  callback(new Error("Maximum allowed size is 50 MB"))
+                if (size > 524288000) {
+                  callback(new Error("Maximum allowed size is 500 MB"))
                   return
                 }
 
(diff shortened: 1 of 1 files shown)


mmaietta:
Just to document here.
We have been unable to identify the reason for a 50mb limit. Upping the file size limit resolved the issue for many users, but not all. That being said, there are no other leads, so I'm merging this so that we may isolate any other/cascading issues.

mmaietta (approved):

LeonanCarvalho:
If accepted should close this issue: #5505  and all other related (there are more than 3)

LeonanCarvalho (commented):
I tried to understand why this limit was hardcoded to 50MB and don't have any comment about it from the author.

The limit was originally 5MB  when the method was created https://github.com/electron-userland/electron-builder/commit/c01b7c0b55d3466b826ea9cc9a11ad34118801c1#diff-f6445d9589029a2a1ce947c0940b7f7a715377570907a24a4e9087c582e07574


And this limit was increased from 5MB to 50MB By this PR:
https://github.com/electron-userland/electron-builder/commit/c5c2eeb80b9f357c18b486e94524580016009bd3#diff-f6445d9589029a2a1ce947c0940b7f7a715377570907a24a4e9087c582e07574

Maybe @develar  can help solve that?


Why do we need this limit? Why do we receive chunks bigger than 50mb? (network speed? server-side configuration?)

I guess the chunk size is a server-side configuration we can't control that size by application.

https://stackoverflow.com/questions/7058827/what-is-the-maximum-chunk-size-in-http-response-with-transfer-encoding-chunked


I tried to find when and why it was included  but no success :

https://github.com/electron-userland/electron-builder/commits/5d01ea6dc9daec4639ab54b38520b93833ac81d3/packages/builder-util-runtime/src/httpExecutor.ts

mmaietta:
Thanks for taking a look @LeonanCarvalho! 

Yeah, I'm not sure why the limit is present.
@develar could you please chime in when you have a chance? 🙂

PR #5852, opened 2021-05-03
chore: use pnpm publish instead of yarn 2 pack

develar:
pnpm is awesome — no need to pack and publish from tar.gz anymore.


mmaietta (approved):

mmaietta:
The breaking tests are unrelated to the changes. But I'm not sure how to fix these, we're now receiving `403 Forbidden` errors with `bintray`. 
```
× file url (bintray) (623 ms)
× downgrade (disallowed, bintray) (584 ms)
× downgrade (allowed) (574 ms)
```

Issue #5861, opened 2021-05-05
v22.11.1 intrododuces publishing bug


mixmix:
## Problem

**Node:** 14.16.1
**OS**: Linux, MacOS
 
have noticed the following error with publishing builds with `v22.11.1` :

```bash
> electron-builder --config build/electron-builder.config.js --publish always

  • cannot check updates  error=TypeError: update_notifier_1.default is not a function
  • electron-builder  version=22.10.5 os=5.4.0-72-generic
  • loaded configuration  file=/home/somename/projects/AHAU/whakapapa-ora/desktop/build/electron-builder.config.js
  • writing effective config  file=dist/installers/builder-effective-config.yaml
  • rebuilding native dependencies  dependencies=fd-lock@1.2.0, fsctl@1.0.0, leveldown@5.6.0, sodium-native@3.2.1, utp-native@2.4.0 platform=linux arch=x64
  • packaging       platform=linux arch=x64 electron=12.0.6 appOutDir=dist/installers/linux-unpacked
  • building        target=AppImage arch=x64 file=dist/installers/ahau-desktop-Linux-1.3.1-x86_64.AppImage
  • publishing      publisher=Github (owner: protozoa-nz, project: whakapapa-ora, version: 1.3.1)
  • uploading       file=ahau-desktop-Linux-1.3.1-x86_64.AppImage provider=GitHub
  ⨯ Cannot read property 'getType' of undefined  failedTask=build stackTrace=TypeError: Cannot read property 'getType' of undefined
    at GitHubPublisher.doUploadFile (/home/somename/projects/AHAU/whakapapa-ora/desktop/node_modules/electron-publish/src/gitHubPublisher.ts:190:36)
    at GitHubPublisher.doUpload (/home/somename/projects/AHAU/whakapapa-ora/desktop/node_modules/electron-publish/src/gitHubPublisher.ts:170:23)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at GitHubPublisher.upload (/home/somename/projects/AHAU/whakapapa-ora/desktop/node_modules/electron-publish/src/publisher.ts:87:5)
    at async Promise.all (index 1)
    at AsyncTaskManager.awaitTasks (/home/somename/projects/AHAU/whakapapa-ora/desktop/node_modules/builder-util/src/asyncTaskManager.ts:65:25)
[...]

PR #5870, opened 2021-05-08
Fix transpiled import syntax

bcomnes:
For whatever reason, importing the default value of mine started importing the .default object property off of module.exports in mime.   As a result, mime.getType could not be found on undefined. This fixes that.

### DIFF
--- a/packages/electron-publish/src/gitHubPublisher.ts
+++ b/packages/electron-publish/src/gitHubPublisher.ts
@@ -4,7 +4,7 @@ import { Fields } from "builder-util/out/log"
 import { httpExecutor } from "builder-util/out/nodeHttpExecutor"
 import { ClientRequest } from "http"
 import { Lazy } from "lazy-val"
-import mime from "mime"
+import * as mime from "mime"
 import { parse as parseUrl, UrlWithStringQuery } from "url"
 import { getCiTag, HttpPublisher, PublishContext, PublishOptions } from "./publisher"
 


bcomnes:
So here is the problem in the published asset: https://unpkg.com/browse/electron-publish@22.11.1/out/gitHubPublisher.js

```js
const mime_1 = require("mime");
//...

    doUploadFile(attemptNumber, parsedUrl, fileName, dataLength, requestProcessor, release) {
        return nodeHttpExecutor_1.httpExecutor
            .doApiRequest(builder_util_runtime_1.configureRequestOptions({
            hostname: parsedUrl.hostname,
            path: parsedUrl.path,
            method: "POST",
            headers: {
                accept: "application/vnd.github.v3+json",
                "Content-Type": mime_1.default.getType(fileName) || "application/octet-stream",
                "Content-Length": dataLength,
            },
//...
```

But of course, mime is cjs, and does not export a default property on the `module.exports` object.

https://unpkg.com/browse/mime@2.5.0/index.js

```js
'use strict';

let Mime = require('./Mime');
module.exports = new Mime(require('./types/standard'), require('./types/other'));
```

So it appears the issue that the `tsc --build` in the tests is producing a different output than whatatever was used to generate the published output. I know that there have been a lot of subtle config churns around how tsc treats cjs -> to default or * import transforms lately in various other modules.

Do you have any pointers or ideas why this may be the case?

### Environment

Node version 12

Electron builder versions:

```
electron-builder@22.11.1
  └─┬ app-builder-lib@22.11.1
    └── electron-publish@22.11.1 
```

bcomnes:
I just found the issue, but this appears to attempt to address this problem:

https://github.com/electron-userland/electron-builder/issues/5861

PR #5888, opened 2021-05-13
fix: quit on decompression errors in NSIS

indutny:
Unlike 7zip, zip decompression pushes a return value on stack. If it not
`success` installer should fail and display the error to the user
instead of assuming that installation was successful.

One of the causes of decompression errors might be a lingering child
process that still uses dlls and node.js bindings so that they cannot be
overwritten. In such case, prior to this patch nsis installer would
update some of the non-locked files and start the app as if it had been
fully updated.


indutny-signal:
Here's how the error message looks like:

image

indutny-signal:
Yay. Thank you!

PR #5900, opened 2021-05-17
feat(docker): update nodejs to 14.17.0

develar:


mmaietta:
Hi @develar, could you please perform a release as soon as possible?

It's proving hard to triage issues across 22.11.x depending on what compilation/runtime issue users are receiving when using electron-builder 😅

develar:
Do you think that it is safe to mark `22.11.5` as `latest` (and not as `next`)?

mmaietta:
Hmmm I would mark it as `next` still, as I'm not sure if we resolved all ts compilation/bundling issues just yet.

mmaietta:
Also, many thanks for the super quick response! 🙂

Issue #5672, opened 2021-03-03
Idea: if mac arm64 build becomes available, download this update on ARM macs


bartoszhernas:
macOS universal Electron build are twice as big (and more).
Because of that at FYM.fm we plan to provide separate builds for x64 and ARM64 architecture for macOS.

Currently this is problematic for our current users that are on x64 version of the app, running via Rosetta.
Even if we release the new ARM64 build, the users will be stuck on Intel version unless they download app from scratch from our website.

I checked the code and the fix seems easy, but I wanted to discuss first if this is something that electron-updater would like to do and if I should make a pull request for it.

Basically if I am on ARM Mac, using Intel version, I would like the updater to download ARM version for the update.

mmaietta:
Would using `mac: { defaultArch: 'arm64' }` in your config to get the right filename for it to default updating to arm64 if process.arch === arm64?

bartoszhernas:
Ah, you  are right, the comment mislead me. Indeed, thank you for pointing it out.

praneetrohida:
> Actually, this functionality already exists without any electron-builder config changes.
> ```
> if (files.some(isArm64)) {
>   files = files.filter(file => (process.arch === "arm64") === isArm64(file))
> }
> ```

@mmaietta I'm struggling to see how this will work since apps running in Rosetta will have the emulated process.arch value (x64), so they will get updated to the x64 version and not arm64. Correct me if I'm wrong.

[Source](https://stackoverflow.com/questions/65346260/get-real-architecture-of-m1-mac-regardless-of-rosetta)

mmaietta:
Hmm, great callout.

Hey @davej , you worked on https://github.com/electron-userland/electron-builder/pull/5524. Is your solution working for you? (Allow arm64 macs to update to x64 version if no arm64 version available)

praneetrohida:
Can we reopen this issue?

mmaietta:
@praneetrohida would you be willing to tackle this? 
It's simple to get a local dev env setup with your local project(s):
https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#to-setup-a-local-dev-environment

praneetrohida:
@mmaietta I can open a PR but I don't have access to an arm64 mac machine to test it

bartoszhernas:
I will make a PR today (Tallinn time) and I have an ARM Mac :)

davej:
[...]

PR #5901, opened 2021-05-17
Allow updating mac version running inside Rosetta to native ARM

bartoszhernas:
Fix for https://github.com/electron-userland/electron-builder/issues/5672

### DIFF
--- a/packages/electron-updater/src/MacUpdater.ts
+++ b/packages/electron-updater/src/MacUpdater.ts
@@ -1,7 +1,6 @@
 import { AllPublishOptions, newError, safeStringifyJson } from "builder-util-runtime"
 import { stat } from "fs-extra"
 import { createReadStream } from "fs"
-import { execSync } from "child_process"
 import { createServer, IncomingMessage, ServerResponse } from "http"
 import { AddressInfo } from "net"
 import { AppAdapter } from "./AppAdapter"
@@ -9,6 +8,7 @@ import { AppUpdater, DownloadUpdateOptions } from "./AppUpdater"
 import { ResolvedUpdateFileInfo } from "./main"
 import { findFile } from "./providers/Provider"
 import AutoUpdater = Electron.AutoUpdater
+import { execShellCommand } from "./util"
 
 export class MacUpdater extends AppUpdater {
   private readonly nativeUpdater: AutoUpdater = require("electron").autoUpdater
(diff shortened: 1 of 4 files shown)


bartoszhernas:
A lot of tests had been failing for me locally probably due to ARM Node environment I am running

packages/electron-updater/src/MacUpdater.ts
```diff
@@ -29,10 +30,13 @@ export class MacUpdater extends AppUpdater {
   protected doDownloadUpdate(downloadUpdateOptions: DownloadUpdateOptions): Promise<Array<string>> {
     let files = downloadUpdateOptions.updateInfoAndProvider.provider.resolveFiles(downloadUpdateOptions.updateInfoAndProvider.info)
 
+    // Detect if we are running inside Rosetta emulation
+    const isRosetta = execSync("sysctl sysctl.proc_translated").toString().includes("sysctl.proc_translated: 1")
```
mmaietta, line 34: Any chance we can convert this to async exec and use `await`?

mmaietta (approved):

packages/electron-updater/src/util.ts
```diff
@@ -35,3 +36,15 @@ export function blockmapFiles(baseUrl: URL, oldVersion: string, newVersion: stri
   const oldBlockMapUrl = newUrlFromBase(`${baseUrl.pathname.replace(new RegExp(escapeRegExp(newVersion), "g"), oldVersion)}.blockmap`, baseUrl)
   return [oldBlockMapUrl, newBlockMapUrl]
 }
+
+export function execShellCommand(cmd: string): Promise<{ error: ExecException | null; stdout: string | Buffer; stderr: string | Buffer }> {
+  return new Promise((resolve, reject) => {
+    try {
+      exec(cmd, (error, stdout, stderr) => {
+        resolve({ error, stdout, stderr })
```
davej, line 44: `exec` can error without outputting a `stderr`. So, you probably want to check for `error` object in MacUpdater.ts, or simply throw which should reject the promise.
```suggestion
        if (error) throw error
        resolve({ stdout, stderr })
```

mmaietta (approved):

packages/electron-updater/src/util.ts
```diff
@@ -35,3 +36,16 @@ export function blockmapFiles(baseUrl: URL, oldVersion: string, newVersion: stri
   const oldBlockMapUrl = newUrlFromBase(`${baseUrl.pathname.replace(new RegExp(escapeRegExp(newVersion), "g"), oldVersion)}.blockmap`, baseUrl)
   return [oldBlockMapUrl, newBlockMapUrl]
 }
+
+export function execShellCommand(cmd: string): Promise<{ stdout: string | Buffer; stderr: string | Buffer }> {
```
mmaietta, line 40: I was looking further into this util function and I think it may be duplicating 
https://github.com/electron-userland/electron-builder/blob/d307c21b34abee1cd2b8dcd3e705995914dda70f/packages/builder-util/src/util.ts#L208-L212

Is it possible to use `util.spawn(...)` for this operation instead of creating a new `execShellCommand`

bartoszhernas: Do I have access to electron-builder utils from electron-updater package?

bartoszhernas: Ahh,

```
import { spawn } from "builder-util"
```

I see it now :)

mmaietta (approved):

praneetrohida:
Which electron-updater version is this released in? I'm on the latest stable (4.3.9) and this doesn't seem to work

mmaietta:
It was released as part of 22.11.8 deployment. So ^4.4.0 should work

mmaietta:
Can you try latest 4.4.x?

praneetrohida:
Thanks for the response @mmaietta. Is 22.11.8 and 4.4.0 stable enough for production, or should I wait for a stable deployment? Asking because it is marked as a pre-release

davej:
@praneetrohida  see here if you’re using an older version of Electron. https://github.com/electron-userland/electron-builder/issues/6000

AFAIK this is why it was marked as pre-release. If you’re using a newer version of Electron then it might be fine.

PR #5902, opened 2021-05-17
fix(nsis): retry closing the app until success

indutny:
Do not continue installation while the app is running. The running
executable and dlls are locked until the app is closed and the 7z
extraction will result in a partial update. If app cannot be quit
forcibly - ask user to close it manually and retry the app check in
the loop until success.

This situation might happen when the app runs in elevated permissions
mode while the installer is started normally.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
@@ -62,16 +62,34 @@
       # to ensure that files are not "in-use"
       Sleep 300
 
-      ${nsProcess::FindProcess} "${APP_EXECUTABLE_FILENAME}" $R0
-      ${if} $R0 == 0
-        # wait to give a chance to exit gracefully
-        Sleep 1000
-        nsExec::Exec `taskkill /f /im "${APP_EXECUTABLE_FILENAME}" /fi "PID ne $pid"` $R0
-        ${If} $R0 != 0
-          DetailPrint `Waiting for "${PRODUCT_NAME}" to close (taskkill exit code $R0).`
-          Sleep 2000
+      # Retry counter
+      StrCpy $R1 0
+
+      loop:
+        IntOp $R1 $R1 + 1
+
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
```diff
@@ -62,16 +62,34 @@
...
+      # Retry counter
+      StrCpy $R1 0
+
+      loop:
+        IntOp $R1 $R1 + 1
+
+        ${nsProcess::FindProcess} "${APP_EXECUTABLE_FILENAME}" $R0
+        ${if} $R0 == 0
+          # wait to give a chance to exit gracefully
+          Sleep 1000
```
mmaietta, line 74: Was this 1s an arbitrary choice? I'm worried it's not enough time for low-performance hardware.

Is there an event we can listen for? Or at least have a retry-loop for several seconds to check if the process exists?

indutny: The logic in question was there before my changes, and the only difference between original code and new one is that it will ask for user confirmation and attempt to retry indefinitely (each time with a confirmation) until success. So the scope of this PR in short is to not let installation succeed while the executable is still locked.

As a follow up PR, I can try using `nsProcess::CloseProcess`, which closes all process windows waits for up to 3 seconds for the process to actually exit until it kills it. I'm not exactly sure if this is what we want though, since it is not unlikely that the electron process might ask user if they want to save some data before closing and 3 seconds is not enough for this to finish.

indutny:
Is there any way I can help with the review of this Pull Request?

mmaietta:
@indutny, I generally leave PRs open for a few days in case I miss something and catch it in a secondary review or in hopes that someone else will also jump in on a review 🙂 
For instance, I'm a bit rusty when it comes to NSIS scripts.

mmaietta (approved):

indutny:
Yay, thank you so much!

Issue #5857, opened 2021-05-05
${arch} in nsis artifactName causes build to fail


vespasianvs:
Hi all,

I *think* I've found a bug here, so thought I'd report it:

**Versions installed:**

"@capacitor-community/electron": "^1.4.2",
"electron-log": "^4.3.5",
"electron-notarize": "^1.0.0",
"electron-updater": "^4.3.8"
"electron": "^12.0.6",
"electron-builder": "^22.11.1"

npm@6.14.12
node@v14.16.1

**Build command:**
npm run build && ARCH=win64 NOTARIZE=disable electron-builder build --windows --x64

**Build config (this one *WORKS*):**

    "win": {
        "target": "nsis"
    },
    "nsis": {
        "artifactName": "Setup ${version} ${env.ARCH}.${ext}",
        "oneClick": false,
        "perMachine": false,
        "allowElevation": true,
        "allowToChangeInstallationDirectory": true,
        "deleteAppDataOnUninstall": true,
        "include": "./build/installer.nsh"
    },

**Build config (this one *FAILS*):**

    "win": {
        "target": "nsis"
    },
    "nsis": {
        "artifactName": "Setup ${version} ${arch}.${ext}",
        "oneClick": false,
        "perMachine": false,
        "allowElevation": true,
        "allowToChangeInstallationDirectory": true,
        "deleteAppDataOnUninstall": true,
        "include": "./build/installer.nsh"
    },

**Error shown:**

  • electron-builder  version=22.10.5 os=10.0.19042
  • cannot check updates  error=TypeError: update_notifier_1.default is not a function
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist\builder-effective-config.yaml
  • packaging       platform=win32 arch=x64 electron=12.0.6 appOutDir=dist\win-unpacked
  • building        target=nsis file=dist\Setup 2.2.7 x64.exe archs=x64 oneClick=false perMachine=false
  • building        target=nsis file=dist\Setup 2.2.7 x64.exe archs=x64 oneClick=false perMachine=false
  • output file is locked for writing (maybe by virus scanner) => waiting for unlock...  
  • building block map  blockMapFile=dist\Setup 2.2.7 x64.exe.blockmap
  ⨯ C:\Users\-----\AppData\Local\electron-builder\Cache\nsis\nsis-3.0.4.1\Bin\makensis.exe exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE

[...]

PR #5905, opened 2021-05-18
Issue 5857 ${arch} causes a build fail

vespasianvs:
There are two problems here as reported in Issue #5857 

If you are only using one architecture and you have a ${arch} in the pattern you get the file twice, once from this.archs and once from inside the loop added on line 124. This duplication meant that the builder was trying to output to the same file with two builds and failed. I have added a check to only add individual architectures, if the size of this.archs is greater than 1.

I've also changed each arch to build sequentially. The reason for this is the build was failing because it couldn't access the 7zip files or uninstaller for the build process because it was locked, causing the build to fail.

This *seems* to fix things, though someone else might have a better way of doing it?

### DIFF
--- a/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
+++ b/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
@@ -119,8 +119,15 @@ export class NsisTarget extends Target {
 
   async finishBuild(): Promise<any> {
     try {
-      const builds = new Set([this.archs])
-      await Promise.all([...builds].map(archs => this.buildInstaller(archs)))
+      const { pattern } = this.packager.artifactPatternConfig(this.options, this.installerFilenamePattern);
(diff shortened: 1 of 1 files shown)


packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
```diff
@@ -357,7 +360,7 @@ export class NsisTarget extends Target {
     // http://forums.winamp.com/showthread.php?p=3078545
     if (isMacOsCatalina()) {
       try {
-        await UninstallerReader.exec(installerPath, uninstallerPath)
+        UninstallerReader.exec(installerPath, uninstallerPath)
```
mmaietta, line 363: Just curious. Why is this `await` no longer needed?

mmaietta (approved):

Issue #3185, opened 2018-07-28
Nested node_modules disappears after building


bradydowling:
* **Version**: 20.26.0

* **Target**: macOS (.dmg)

Similar to https://github.com/electron-userland/electron-builder/issues/1539, I'm working with an app where I've cloned another git repo int and I have several `node_modules` folders. When I build, the `node_modules` folder in the nested git repo disappears. This happens whether asar is enabled or disabled. It also happens regardless of whether or not the inner git repo is officially designated as a git submodule or if I've just manually cloned it in.

Here's my folder structure:

```
main.js (starts app and forks process from the express app's index.js)
node_modules
package.json
cloned-express-app
   |
    --node_modules
      index.js
      package.json
```

Here's a [Stack Overflow issue](https://stackoverflow.com/questions/51573965/nested-node-modules-folder-disappears-after-packaging) I just created that may explain it better (I didn't want to copy/paste but let me know if that's preferred) and [here's the repo](https://github.com/bradydowling/electron-server-hello-world) I'm working with.

It's also worth noting that I've tried building this for Windows and my friend said it works fine on there. I haven't tested that recently though so I can't give much info on that immediately. I also haven't inspected the Resources contents of the packaged app (or whatever the .exe counterpart would be) so I can't provide info on that but I do know that when I look in the mac app's Resources directory there is no `node_modules` folder within the nested git repo.

fpsqdb:
Same issue here for me.

- **version:** 20.26.0

- **target:** windows

```
 "files": [
            "!**/**.map",
            "!**/**.ts",
            "!node_modules",
            "package.json",
            "node_modules/semver/**",
            "node_modules/fs-extra/**",
            "node_modules/graceful-fs/**",
            "node_modules/jsonfile/**",
            "node_modules/universalify/**"
        ],
```
When I build, there is no `node_modules` folder in the package. That's not my expected, it should have those folders and files in the package.
```
node_modules/semver,
node_modules/fs-extra,
node_modules/graceful-fs,
node_modules/jsonfile,
node_modules/universalify
```

But if I downgrade to 20.15.0, it works fine.

bradydowling:
Interesting, I'll need to give this a try, thanks for the note.

hn3000:
In version 20.16 the way node_modules includes work was changed: explicit includes are now ignored, instead all production dependencies are used. Unfortunately, excludes are still allowed to affect the node_modules folders and that means your !node_modules takes effect.

[...]

PR #5911, opened 2021-05-21
fix: optionally package nested node_modules (#3185)

bedney:
This patch adds the ability to package either all or a select set of *submodules* node_modules directories in an Electron app.


packages/app-builder-lib/src/util/AppFileWalker.ts
```diff
@@ -62,17 +76,30 @@ function createAppFilter(matcher: FileMatcher, packager: Packager): Filter | nul
...
 
 /** @internal */
 export class AppFileWalker extends FileCopyHelper implements FileConsumer {
+  readonly includeSubNodeModules: boolean
+  readonly matcherFilter
+
   constructor(matcher: FileMatcher, packager: Packager) {
     super(addAllPatternIfNeed(matcher), createAppFilter(matcher, packager), packager)
+    //  @ts-ignore
+    this.includeSubNodeModules = (packager._configuration || {}).includeSubNodeModules;
```
b-zurg, line 101: Another instance of a ts-ignore that could be fixed by better typing.

At the very least a type cast would be better.

packages/app-builder-lib/src/util/AppFileWalker.ts
```diff
@@ -83,7 +110,21 @@ export class AppFileWalker extends FileCopyHelper implements FileConsumer {
       // but do not filter if we inside node_modules dir
       // update: solution disabled, node module resolver should support such setup
       if (file.endsWith(nodeModulesSystemDependentSuffix)) {
-        return false
+        //  it's a node_modules directory
+
+        //  if includeSubNodeModules is true, then we just do nothing - we
+        //  definitely want it. But if its false (the default), then we match
+        //  and if the match is false, we return false (but if the match is
+        //  true, we do nothing).
```
b-zurg, line 118: Honestly this comment did not help me understand the code. If you feel the code needs this much explaining then it's better to represent the logic through more variable names or clearer code paths. Comments often get out of date with time.

packages/app-builder-lib/src/util/AppFileWalker.ts
```diff
@@ -48,6 +48,20 @@ export abstract class FileCopyHelper {
 }
 
 function createAppFilter(matcher: FileMatcher, packager: Packager): Filter | null {
+    //  @ts-ignore
```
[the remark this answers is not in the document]
bedney: In general, please see commentary that I've left in the bug that this references (https://github.com/electron-userland/electron-builder/issues/3185). I detail the compromises I made there. I didn't want to change APIs for these calls, hence the 'escaping' into regular JavaScript.

packages/app-builder-lib/src/util/AppFileWalker.ts
bedney: See above.

packages/app-builder-lib/src/util/AppFileWalker.ts
bedney: A stylistic choice. I've been on projects that are 50/50 code/comment and code reviews are employed to make sure comments stay up to date. I can easily remove.

mmaietta:
Please take a look at one of the failing tests
```
  ✕ ignore node_modules dev dep (1710 ms)
  ● ignore node_modules dev dep

    Path /tmp/et-260ab5d536c83164edf011b1145bf65e/t-gpDb7Z/test-project-3/dist/linux-unpacked/resources/app/node_modules/electron-osx-sign must not exist
```

packages/app-builder-lib/src/util/AppFileWalker.ts
mmaietta: I'm okay with comments, but I also agree with @b-zurg comment. To me, they read both convoluted to me in what their purpose is other than what an if-statement is already explaining. Happy to help out and contribute to this though.

We'll need some new test cases. Such as for when `includeSubNodeModules` is `true`, `false`, and unspecified in config.

bedney: I'm currently updating this patch to take into account the feedback from you and @b-zurg. In order to make this 'TypeScript friendly', I'll have to alter the API to this call so let me know if there is a big objection to that. I'll resolve the test failures and add some new test cases as well.

bedney:
This PR has been updated to solve several issues:

1. It is now TypeScript friendly (and a bit less chatty in comments ;-) ).
2. It fixes a bug when trying to copy just select submodule 'node_modules' directories.
3. It includes tests for this functionality
4. It should fix the failing test @mmaietta noted above.

Now, about tests: I am running my tests locally on an M1 (Apple Silicon) Mac. It looks like the directory creation call creates a 'tmp' directory that *includes* my chip architecture (`amd64`) as part of the path, but the `getResources` call does not do that unless you supply the 2nd parameter that matches an architecture. I do that in my tests (that `archFromString(process.arch` call). One or the other of these 2 calls need to be fixed to be consistent. I'm not sure if these tests will pass on other OS/architecture combinations because of this. This may also be causing other problems in the test suite and in the CI for this project.

Please advise @mmaietta or @b-zurg. Overall, if this patch looks good, it would be great to get it pulled in. If there are other things that need to be done, let me know.

bedney:
Still getting a test failure on the test that I thought I'd fixed. It was passing for me on my Mac. @mmaietta, please advise.

nathanielks:
Thank you so much for picking up the mantle on this, @bedney !

mmaietta:
Sorry for delay here, got swamped from work. Will try and take a look be end of this weekend

mmaietta:
I'm still trying to resolve the broken tests, one important one is `"ignore node_modules dev dep"` which is failing when `includeSubNodeModules: false` and for `undefined`, so that will be blocking the merge until resolved.

@bedney , most of the CI snapshots assume a Linux x64 environment, so you'll need to verify tests locally using the docker image. The `package.json` `pnpm run test-linux` needs to have the command updated to utilize command: `/bin/bash -c \"pnpm i && pnpm run compile && pnpm run test\"`

Then to verify the test I mentioned, run: `TEST_FILES=ignoreTest pnpm run test-linux`

I'll continue giving it a shot from my end, just wanted to give a quick status update and see if you'd be able to continue investigating the test as well.

bedney:
@mmaietta -

I can continue to work the problem from my end. I'll try to get the Linux 64 environment running here. I just took delivery of a sweet Apple M1 Mini and I have Parallels which will run x86 Linux images here (wow, is this machine fast! :-))

Thanks for the heads up about the test command. As an aside, it would be good to get things like that into the release. Also, I have some notes here about running the tests individually that I can contribute to the docs, as some things are not really that clear and/or are actually outdated. I can contribute those in a separate commit.

Cheers,

- Bill

bedney:
@mmaietta -

So I configured a native Linux environment (Ubuntu 20.04) rather than use the Docker image, which I had problems with since that Docker image is built for amd64 and this is an arm64 box (and an arm64 Ubuntu release).

I then execute just the `ignore` tests by executing:

`TEST_FILES=ignoreTest pnpm run test`

The weird thing is that I'm getting 2 failing tests that the CI build is passing:

`"ignore build resources"`
`"ignore known ignored files"`

but I'm not seeing `"ignore node modules dev dep"` fail on my local machine like we do in the CI builds here. I set `includeSubNodeModules` to `false` and to `undefined` and I can't get it to fail.

As an aside, I was able to 'fix' the `"ignore build resources"` test by adjusting the path it's looking for. As I stated above, I think these paths are being miscomputed, but since I don't know the tests that well, it's only a guess. The fact that it is one of the few tests that tests for file existence rather than non-existence like most of the other tests makes me a bit suspicious.

If you have any insight here, I'd appreciate it. Thanks for your continued effort on this project!

bedney:
@mmaietta -

So after tweaking my code a bit more, it looks like I got the tests to pass. I also added a further check to one of the tests that I added to the suite.

On further reflection, I guess the Docker container sets up an environment that the suite relies on. It would be good to make the test suite run cleanly outside of it though. One other thing is to maybe take advantage of Docker's newish multi-platform capability to produce builds for arm64 systems... there's going to be more of them moving ahead :-).

In any case, sorry for all of the back and forth. My feeling is that this patch is now ready to go, unless you have anything else that you need for me to do. Now that the code has settled out, I can do a commit for some docs. Not sure if I should include that in this same patch. Please advise.

bedney:
@mmaietta - This is now passing tests. If you have no objections, I'd like to get it merged. If there are further issues, let me know and I'll work through them as best I can.

Thanks!

mmaietta (approved):

mmaietta:
Fantastic work! Thanks for the contribution

Issue #5907, opened 2021-05-19
Missing 'libXtst.so.6' in wine-chrome container


MCMicS:
* **Version**:  electronuserland/builder:wine-chrome
Digest:sha256:7a41c8b334b60590fbd09727c5c1e29a00134e901fff474df74387aa5de7ef2c

```
electron-mocha test/main & electron-mocha --renderer test/renderer
warning Cannot find a suitable global folder. Tried these:
"/usr/local, /jenkins/.yarn"
/var/lib/jenkins/workspace/project/node_modules/electron/dist/electron:error while loading shared libraries: libXtst.so.6: cannot open shared object file: No such file or directory
/var/lib/jenkins/workspace/project/node_modules/electron/dist/electron: error while loading shared libraries: libXtst.so.6: cannot pen shared object file: No such file or directory
```
My older Images works. It seems happen after Update to Ubuntu 20.04

Workaround: Install `libxtst6` e.g. via `RUN apt-get update && apt-get install -y libxtst6` in own Dockerfile

There are official Prerequisites: https://www.electronjs.org/docs/development/build-instructions-linux#prerequisites

mmaietta:
I'm curious if this issue was introduced by this change
https://github.com/electron-userland/electron-builder/pull/5855/files#diff-a1a97c1c8efa82c884caec1f1e963f897544efb30119aee46f8e5e7ba470db52L3

Seems like the solution is to just add `libxtst6` to the Dockerfile though, no?

MCMicS:
yes
add `libxtst6 `in DOckerfile will fix it

I guess that the change come from ubuntu base container. There should the libxtst6  not included per default anymore

mmaietta:
@MCMicS I'm curious if I can create a simple test case for this, seems reasonable to add at least a github action for testing dockerfile changes.
Do you happen to have a simple/sample setup on what's causing the issue?
Would a basic electron-quick-start app with an empty electron-mocha test file work?

MCMicS:
Could provide some sample the days

MCMicS:
@mmaietta I've tried to reproduce it with simple test but this seems to runs
I lookt how to reproduce it in: https://github.com/MCMicS/electron-quick-start-electron-mocha/tree/issue/electron-builder-5907

MCMicS:
thanks for fast fixing 👍🏻

mmaietta:
Np! Took a hot second to figure out what was failing in the CI.

[...]

PR #5913, opened 2021-05-23
fix/dockerfile-libxtst6 (#5907)

mmaietta:
The dockerfile upgrade to ubuntu 20 removed libxtst6 from the image. Adding it explicitly to install 
Fixes: #5907


mmaietta:
@develar, I'm not sure why the build is failing for only a dockerfile change. After this is merged though, I think we should do a full release so that HEAD is 22.11.6 instead of 22.10.5. What are your thoughts?

mmaietta:
@develar all issues are resolved. Please perform a full release 🙂

Issue #5933, opened 2021-05-31
Debian post-inst script broken with linux.executableName


kris7t:
* **Version**: 22.11.3
* **Electron Version**: 13.0.1
* **Electron Type (current, beta, nightly)**: current
* **Target**: Linux .deb

Debian .deb packages built when `linux.executableName` is not the same as the `"productName"` from `package.json` don't run properly, because the `post-inst` script doesn't refer to the binaries at the appropriate path.

The problems seems to be that pacakge files are installed to `/opt/{sanitizedProductName}`,

https://github.com/electron-userland/electron-builder/blob/c623279a31d3d203ad1f150d83905a7e1f58a731/packages/app-builder-lib/src/targets/fpm.ts#L187

while the post-inst script uses `/opt{productFilename}`,

https://github.com/electron-userland/electron-builder/blob/28cb86bdcb6dd0b10e75a69ccd34ece6cca1d204/packages/app-builder-lib/templates/linux/after-install.tpl#L4

which is not the same as `/opt/{sanitizedProductName}` if the `executableName` is set:

https://github.com/electron-userland/electron-builder/blob/ea315aa11528a44189c860d22ce04e59b9ed6c64/packages/app-builder-lib/src/appInfo.ts#L66

See also the issue https://github.com/getferdi/ferdi/issues/1469. As a workaround, I set `linux.executableName` to the same as the `"productName"`  (https://github.com/getferdi/ferdi/pull/1472), but a correctly working package out of the box would be preferable.

mmaietta:
Fantastic find and investigative work, thank you!

Would you be willing to submit a PR for this? Getting a local dev env set up is fairly simple 🙂 
https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#to-setup-a-local-dev-environment

If not, I'll try and get around to this in the next week or so.

kris7t:
I'd be willing to create a PR, but I'm not sure what behavior is wanted here:

1. Should the `{sanitizedProductName}` be exposed to the post-inst script template, thus installing the binaries to `/opt/{sanitizedProductName}/{executableName}`,
2. or should rather everything be installed to `/opt/{productFilename}`?

[...]

PR #5941, opened 2021-06-05
fix(linux): executable path in Debian postinst script

kris7t:
Debian .deb packages built when linux.executableName is not the same as the "productName" from package.json don't run properly, because the post-inst script doesn't refer to the binaries at the appropriate path.

Add a test for the correct behavior.

Expose the correct sanitizedProductName value for the after-install.tpl postinst script template.

Still expose the productFilename, too, because users supplying their own templates might still be using it.

Make the associated "executable path in postinst script" test pass.

Closes: #5933


mmaietta (approved):

mmaietta:
Thank you @kris7t !

PR #5987, opened 2021-06-25
Escaping special characters in DMG entries

pedro-pedrosa:
Fixes an issue when the `productName` contains apostrophes. this is needed because in the python script it creates a literal string but the string is not escaped so an apostrophe terminates the string.

I have also added the backslash to the regex for completeness sake even though I don't think it's needed because it's sanitized out.


mmaietta:
Thanks for your contribution!

PR #6001, opened 2021-06-28
chore: update deps

develar:
Update app-builder-bin dependency. Required to perform a new release.


mmaietta:
All tests seem to be failing with `ReferenceError: jasmine is not defined` which I think we should consider a blocker prior to merging since deps are also being updated in this PR.

Btw, why are updating deps required for releasing?
Related note: I'm working on getting a auto-PR-labeler + dependabot set up for this repo.

packages/app-builder-lib/src/version.ts
```diff
@@ -1 +1 @@
-export const PACKAGE_VERSION = "22.11.7"
+export const PACKAGE_VERSION = "22.11.8"
```
mmaietta, line 1: What if we automated the manner in which this version is pulled?
```
import { version } from '../../package.json`
export const PACKAGE_VERSION = version
```

packages/dmg-builder/package.json
```diff
@@ -13,9 +13,9 @@
 		"vendor"
 	],
 	"dependencies": {
-		"app-builder-lib": "22.11.7",
-		"builder-util": "22.11.7",
-		"builder-util-runtime": "8.7.6",
+		"app-builder-lib": "22.11.8",
+		"builder-util": "22.11.8",
```
mmaietta, line 17: Should these also be changed to `"workspace:*"` just like `builder-util-runtime` is doing so?

packages/app-builder-lib/src/version.ts
develar: > Btw, why are updating deps required for releasing?

You asked me to do a release (thanks!), but to apply https://github.com/develar/app-builder/pull/60 we need to update app-builder-lib dependency.

mmaietta: Ahhh, I see now. Thanks for the quick reply 🙂

packages/dmg-builder/package.json
develar: Nice catch. Indeed.

mmaietta:
3 remaining tests to resolve.

Linux now has an additional entry, in Desktop file?
`"Recommends": "libappindicator3-1",`

And the other test seems to just have different whitespace for a line in the snapshot, which is weird.

develar:
> Linux now has an additional entry, in Desktop file?

Due to https://github.com/develar/app-builder/pull/53/files

mmaietta (approved):

mmaietta:
Hell to the Yes. Thanks for getting this release out @develar! I'm stoked 

Quick note, I'd like to shorten our release intervals to make it easier to triage issues/tickets that come in based on EB version. What are your thoughts?

mmaietta:
@develar may we please proceed on this? LGTM 🙂

PR #6020, opened 2021-07-03
chore: update deps

develar:
Update app-builder-bin — a new env `DISABLE_MULTIPART_DOWNLOADING` is added.


packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
```diff
@@ -101,6 +101,7 @@ function handleError(logger: Logger, error: Error | null, stderr: string | null)
   }
 
   try {
+    // @ts-ignore
```
develar, line 104: Some issue in a new typings, suppress incorrect error for now.

mmaietta (approved):

Emnmer1 (commented):
[pages-build-deployment](https://github.com/Lyttell/electron-builder/actions/workflows/pages/pages-build-deployment)

PR #6032, opened 2021-07-07
fix code error

zzxxiinn:
there is no callback for showMessageBox method, it return a promise.

https://github.com/electron-userland/electron-builder/blob/8a6c48116cbe1c1a2ad49db7648acf3080978bcc/encapsulated%20manual%20update%20via%20menu.js#L46-L53


mmaietta:
Thank you!

PR #6056, opened 2021-07-14
docs: fix misused command in README.md

ciffelia:
`yarn pack` is a built-in command of yarn: https://classic.yarnpkg.com/en/docs/cli/pack
We should run `yarn run pack` to execute the npm script.


mmaietta:
`yarn` doesn't require `run` to execute build `scripts` from the package.json

mmaietta (approved):

PR #6073, opened 2021-07-21
fix: Windows update fails for custom paths that require admin rights

hvuntokrul:
Proposed fix to the situation on Windows where an Electron is installed for all users of the machine e.g. into Program Files, but fails to update (installed app version is deleted but the update is never installed) due to failure to request admin rights.

node v12.16.2
npm v6.14.4

Current build.nsis settings in package.json: 
{
	"oneClick": false,
	"allowToChangeInstallationDirectory": true,
	"allowElevation": true,
	...
}

### DIFF
--- a/packages/electron-updater/src/BaseUpdater.ts
+++ b/packages/electron-updater/src/BaseUpdater.ts
@@ -57,19 +57,18 @@ export abstract class BaseUpdater extends AppUpdater {
     this.quitAndInstallCalled = true
 
     try {
-      this._logger.info(`Install: isSilent: ${isSilent}, isForceRunAfter: ${isForceRunAfter}`)
-      
       let installPathRequiresElevation = false
-      if (process.platform === 'win32') {
+      if (process.platform === "win32") {
         try {
-          var accessTestPath = path.join(path.dirname(process.execPath), 'access')
-          fs.writeFileSync(accessTestPath, ' ')
+          const accessTestPath = path.join(path.dirname(process.execPath), "access")
+          fs.writeFileSync(accessTestPath, " ")
         } catch(err) {
           // Require admin rights if needed
           installPathRequiresElevation = true
         }
       }
-      
+
+      this._logger.info(`Install: isSilent: ${isSilent}, isForceRunAfter: ${isForceRunAfter}, installPathRequiresElevation: ${installPathRequiresElevation}`)
       return this.doInstall({
         installerPath,
         isSilent,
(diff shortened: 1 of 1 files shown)


packages/electron-updater/src/BaseUpdater.ts
```diff
@@ -56,11 +58,23 @@ export abstract class BaseUpdater extends AppUpdater {
 
     try {
       this._logger.info(`Install: isSilent: ${isSilent}, isForceRunAfter: ${isForceRunAfter}`)
```
mmaietta, line 60: Would you mind moving this below the try-catch and add to the logging sentence `installPathRequiresElevation`, It'd help with future debugging

hvuntokrul:
@mmaietta Thank you very much for such a quick reply and for your suggestion, already done:) sorry for the delay

packages/electron-updater/src/BaseUpdater.ts
hvuntokrul: @mmaietta Thank you very much for such a quick reply and for your suggestion, already done:) sorry for the delay

packages/electron-updater/src/BaseUpdater.ts
```diff
@@ -55,12 +57,23 @@ export abstract class BaseUpdater extends AppUpdater {
     this.quitAndInstallCalled = true
 
     try {
-      this._logger.info(`Install: isSilent: ${isSilent}, isForceRunAfter: ${isForceRunAfter}`)
+      let installPathRequiresElevation = false
+      if (process.platform === "win32") {
+        try {
+          const accessTestPath = path.join(path.dirname(process.execPath), "access")
+          fs.writeFileSync(accessTestPath, " ")
+        } catch(err) {
```
mmaietta, line 65: Couple of notes here.
To avoid odd edge cases for the project that coincidentally includes a file named 'access', could you utilize a random id for this `.tmp` file? Something like this:
```access-${Math.floor(Math.random() * 100)}.tmp```
Also, if the write is successful, I think we should then delete the file afterward

**Review by mmaietta — APPROVED** [review]

**Comment by jmeinke** [comment]
I don't know for sure but I suppose there might be an error related to this PR:
After upgrading electron-builder to v22.11.11 and updater to the version 4.4.3 the elevated rights are being requested even if the app has been installed per-user into the standard path on Windows 10.

We'll downgrade electron-updater to version 4.3.10 in order to check, if this PR has introduced this unexpected behaviour. I'll add another post when we have the results.

**Comment by sandeep1995** [comment]
I can confirm this issue. It is happening with a lot of our users. 
```
    "electron-builder": "22.11.11",
    "electron-updater": "4.4.3",
```

PR #6074, opened 2021-07-21
feat(app-builder-lib): Add timestamp for electron-osx-sign options

ffflorian:
I noticed that the `timestamp` option from electron-osx-sign was missing.

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -241,7 +241,7 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     let binaries = options.binaries || undefined
     if (binaries) {
       // Accept absolute paths for external binaries, else resolve relative paths from the artifact's app Contents path.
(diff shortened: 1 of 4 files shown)


packages/app-builder-lib/src/macPackager.ts
```diff
@@ -284,6 +284,7 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       app: appPath,
       keychain: keychainFile || undefined,
       binaries,
+      timestamp: isMas ? masOptions && masOptions.timestamp : options.timestamp,
```
mmaietta, line 287: Minor: We can use optional chaining here
```
isMas ? masOptions?.timestamp : options.timestamp,
```

mmaietta:
Nice catch! Thanks for the add.
Can you confirm that passing `timestamp: undefined` does not impact current functionality?

ffflorian:
@mmaietta How can I confirm that? Write a test?

mmaietta (approved):

Issue #5764, opened 2021-04-03
Change extract folder for each instance for portable app or preventing delete extracted resources when closing second instance?


kobakou:
* **Version**: 22.9.1

* **Electron Version**: 9.4.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: portable

Backgrounds:
- build as portable apps for windows
- we want to run multiple instances of this app at the same time

Issues:
- From v20.40, the folder that is extracted when the app is launched is unique for each Build.
- As a result, when two or more apps are running, closing the second one will delete the extracted folder and some files will be deleted
- It may result in the first app not working properly.

How can I resolve this issue, such as changing extract folder for each launch instance or preventing delete extracted resources when closing the second instance?

I found a similar post in the issues log, but I can not find a resolution.
https://github.com/electron-userland/electron-builder/issues/5382#issue-733302775

Thanks,
Kou

kobakou:
関連？ウォッチかも。

https://github.com/electron-userland/electron-builder/issues/5759

kobakou:
I found related reports. I think this use-case is to be considered.
https://github.com/electron-userland/electron-builder/issues/4105

mmaietta:
Here's a quick patch I whipped up. Please apply it to your project and report back on how it works 🙂 

Use `unpackDirName: false` in the config, which internally does not define nsi var `$UNPACK_DIR_NAME`.  The portable app will default back to `$PLUGINSDIR\app` when set false (pre-v20.40 behavior).
Note: passing `true` will break the logic because this was scrappy.

```
diff --git a/node_modules/app-builder-lib/out/targets/nsis/NsisTarget.js b/node_modules/app-builder-lib/out/targets/nsis/NsisTarget.js
index a64e26f..8f5c29d 100644
--- a/node_modules/app-builder-lib/out/targets/nsis/NsisTarget.js
+++ b/node_modules/app-builder-lib/out/targets/nsis/NsisTarget.js
@@ -428,7 +428,10 @@ class NsisTarget extends _core().Target {
     if (isPortable) {
       const portableOptions = options;
       defines.REQUEST_EXECUTION_LEVEL = portableOptions.requestExecutionLevel || "user";
-      defines.UNPACK_DIR_NAME = portableOptions.unpackDirName || (await (0, _builderUtil().executeAppBuilder)(["ksuid"]));
+
+      if (portableOptions.unpackDirName !== false) {
[...]

PR #6093, opened 2021-07-24
feat(portable): Adding support for unique dir on each portable app launch

mmaietta:
Enabling config for extract folder to be unique for each instance of a portable app to prevent deletion of extracted resources when closing other app instance(s).
- Retains default of `ksuid` for unique extract dir **per-build** for empty strings and `false`
- `true` sets to a unique temp dir `$PLUGINSDIR/app` **per-app-launch**

Implements: #5764


kobakou (approved):
LTGM, and I have confirmed your new options work well as expected in my environment.
Thanks!

PR #6100, opened 2021-07-26
fix(nsis): should close app when `Silent` and `ONE_CLICK`

lwintermelon:
Currently, if uninstall When `Silent is true` and `ONE_CLICK is defined`, the uninstaller will try to unisntall the app without closing it, which cases problems if uninstall it sliently when the app is running, the uninstaller can't remove all files, so left a broken app dir.

How to reproduce:
create an app, set `"oneClick": true` in config, make a `NSIS` installer, install it and try `Uninstall {appName}.exe /S` to silently uninstall it when the app is **running**.
Actual behavior: see above description.
Expected behavior: close the app then do the uninstall.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/uninstaller.nsh
+++ b/packages/app-builder-lib/templates/nsis/uninstaller.nsh
@@ -1,13 +1,19 @@
+Function un.checkAppRunning
+  !insertmacro CHECK_APP_RUNNING
+FunctionEnd
+
 Function un.onInit
   !insertmacro check64BitAndSetRegView
 
-  ${IfNot} ${Silent}
+  ${If} ${Silent}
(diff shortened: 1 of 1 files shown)


mmaietta:
Just to confirm, did you test locally this fix with the various installer configurations? 🙂 A change like this wouldn't get caught in any unit tests (not sure if we can even create tests for nsis (un)installer scripts)

lwintermelon:
As seen in the code, it only affects whether u enable `oneClick` in config and whether u uninstall the app sliently or not, so i have tested the four posssible combined cases locally.

As for the test, it's not easy, as (un)installer may need user Interaction，so may need UI automation tools.

mmaietta (approved):

PR #6133, opened 2021-08-09
fix(electron-updater): arm64 updating issue with Github private repos as update server

nikcheerla:
When using a github private repo as update server, the `file.url.pathname` for an arm64 version does not actually include the `arm64` string, and instead it just looks like a github release artifact ID. Looking at `file.info.url`, which contains the actual name of the dmg, seems more robust.


mmaietta (approved):
Thanks for contributing!

nikcheerla:
@mmaietta Appreciate the quick merge! If you don't mind me asking, what's the fastest way for me to get this fix working in a production app? We urgently want to ship correct electron updates to our users on M1 macs. Tried forking the repo but there's no obvious way I could figure out to install just electron-updater from the monorepo.

PR #6149, opened 2021-08-12
fix(electron-updater): `null` object error when MacUpdater logs server port before it is listening

paulrosania:
This PR addresses an issue where macOS updates fail due to a `null` object reference.

Node's HTTP server `address()` method returns `null` until the server is listening for connections. `electron-updater` 4.4.4 and 4.4.5 contain a bug where `MacUpdater` attempts to log the server port before it is listening, which causes an error.

To fix this while maintaining the completeness of the added logs, I removed the server address from the existing log event but otherwise kept it in place. When the server is listening, a new event is recorded that includes the server address.

### DIFF
--- a/.changeset/stale-carrots-listen.md
+++ b/.changeset/stale-carrots-listen.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix(electron-updater): `null` object error when MacUpdater attempts to log the server port before it is listening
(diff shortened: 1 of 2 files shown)


aguynamedben:
This was a painful one for us. Luckily we caught it right after a deploy, but it would have broken autoupdates for all of our users.

amacneil:
Unfortunately we shipped this to production and broke auto updates for all our mac users.

PR #6160, opened 2021-08-16
fix: workaround vite replacing process.env in updater

pzeinlinger:
Vite replaces process.env.* during builds. 
This leads to problems with the updater. 

A known workaround is to replace process.env.* with process.env["*"].

https://github.com/vitejs/vite/issues/3176

### DIFF
--- a/.changeset/five-worms-work.md
+++ b/.changeset/five-worms-work.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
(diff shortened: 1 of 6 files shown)


mmaietta:
@pzeinlinger my main concern with this is that it'll still break in the future. I have no way to enforce or persist the knowledge for future contributors that `process.env.ANY_VAR` won't work for `vite`. I'm doubtful I'll consistently remember it tbh.

Issue #6130, opened 2021-08-08
Unknown option 'cache-min' when running electron-builder with pnpm


mhatvan:
* **Electron-Builder Version**: 22.11.11

* **Node Version**: 14.17.2 (also occurs with v16)

* **Electron Version**: 13.1.8
* **Electron Type (current, beta, nightly)**: current

* **Target**: linux x64

I am in the process of switching over a project from npm 6 to pnpm. There are no config changes, so I am wondering where this error is coming from.

## Screenshot

![Screenshot from 2021-08-08 17-23-23](https://user-images.githubusercontent.com/16797721/128637201-853f22e0-c88a-4061-a046-5b764226c54e.png)

mmaietta:
So that error is probably coming from here:
https://github.com/electron-userland/electron-builder/blob/ae363e51957d0abfc7d848f51aa23c7e5faf5f33/packages/app-builder-lib/src/util/yarn.ts#L102-L104

Seems like we need a way to discern if running `npm`, instead of determining on whether it is `yarn`?

mhatvan:
Ah, glad that you found the location of the issue. It might even be the case that npm, yarn and pnpm need to be handled differently, but unfamiliar with the code base.

mmaietta:
Just need a way to discern npm from pnpm. The code part will be easy after that, just need to create `isRunningNPM`

mhatvan:
I guess you could check for the existence of node_modules/.pnpm, related article: https://pnpm.io/symlinked-node-modules-structure

mmaietta:
Would you be willing to contribute? Setting up a dev env is fairly simple :)
https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#to-setup-a-local-dev-environment

mhatvan:
If you give me proper instructions on what to do, okay.

Why is the cache-min line there in the first place? 
Should we conditionally check for not yarn and not pnpm or rather just check for npm?

mhatvan:
I think I found an easy way to fix this:

According to https://docs.npmjs.com/cli/v7/using-npm/config#cache-min, the `--cache-min` option is deprecated anyways and it is recommend to use `--prefer-offline` instead.

pnpm doesn't support `--cache-min`, but both npm and pnpm support the usage of `--prefer-offline`.

PR #6165, opened 2021-08-18
fix: replace deprecated --cache-min option

mhatvan:
Closes https://github.com/electron-userland/electron-builder/issues/6130

I found out on https://docs.npmjs.com/cli/v7/using-npm/config#cache-min that we can replace deprecated `--cache-min` with `--prefer-offline` to avoid pnpm install failure.

This is the commit where pnpm added support for `--prefer-offline`: https://github.com/pnpm/pnpm/commit/b553c5cfe07d2fe9c0f6f3f8003950b64ec6b5bb

### DIFF
--- a/.changeset/honest-pandas-sneeze.md
+++ b/.changeset/honest-pandas-sneeze.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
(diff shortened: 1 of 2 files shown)


mhatvan:
Great to see this merged!

Can we get expect a new release soonish?

mmaietta:
Released 🙂

PR #6167, opened 2021-08-19
feat: Adding Keygen as an official publisher/updater for electron-builder

mmaietta:
CC @ezekg


mmaietta:
@ezekg few quick notes 🙂 

> If you want to ignore 5xx errors, I'd suggest also checking for other common 500s: 500, 503 and 504.

I actually pulled this logic from Bintray, but I'm more than happy to scan for the other errors as well. Can just check for any status code that starts with `50`?
https://github.com/electron-userland/electron-builder/blob/331406a0dc7643c894636c70abacee9dc7ba408b/packages/app-builder-lib/src/publish/BintrayPublisher.ts#L86-L88

> The data and errors properties will never be returned at the same time. So on API error, this will be accessing `undefined.id`, which will throw.

I actually pulled this from the documentation that both would be returned. Would you be willing to update it? I found it confusing.
https://keygen.sh/docs/api/#releases-upsert
```const { data, errors } = await response.json()```

I'll update the logic accordingly in my PR

**Comment by ezekg** [comment]
> I actually pulled this logic from Bintray, but I'm more than happy to scan for the other errors as well. Can just check for any status code that starts with `50`?

Yep, generally, you can use `statusCode >= 500 && statusCode <= 599` to test for ranges of HTTP status code.

I will update the API docs to make return values clearer when an error occurs, thanks for the feedback.

File: test/src/updater/nsisUpdaterTest.ts [hunk 1]
```diff
@@ -46,6 +46,17 @@ test("file url generic", async () => {
   await validateDownload(updater)
 })
 
+test.ifEnv(process.env.KEYGEN_TOKEN)("file url keygen", async () => {
+  const updater = await createNsisUpdater()
+  updater.addAuthHeader(`Bearer ${process.env.KEYGEN_TOKEN}`)
```
mmaietta, line 51: Hey @ezekg! Quick note for documentation: The end-user will need to know this detail of adding the "Bearer " prefix on their own. I think the function's name makes it self-explanatory that it's the `Authorization` header, but might still be worth noting in the docs.

develar (approved):

packages/app-builder-lib/src/publish/KeygenPublisher.ts
```diff
@@ -0,0 +1,116 @@
...
+    requestProcessor: (request: ClientRequest, reject: (error: Error) => void) => void,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _file?: string
+  ): Promise<any> {
+    return HttpExecutor.retryOnServerError(async () => {
+      const { data, errors } = await this.upsertRelease(fileName, dataLength)
+      if (errors) {
+        throw new Error(`Keygen - Upserting release returned errors: ${JSON.stringify(errors)}`)
+      }
+      const releaseId = data?.id
```
ezekg, line 44: Note: the release ID will always be present, given there are no `errors` (which is already handled above), so checking if it exists here isn't strictly needed.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -60,22 +60,40 @@ export class HttpError extends Error {
     this.name = "HttpError"
     ;(this as NodeJS.ErrnoException).code = `HTTP_ERROR_${statusCode}`
   }
+
+  // Check if 500 error
+  isServerError() {
+    return this.statusCode % 500 <= 99
```
ezekg, line 66: Nit: this is hard to parse. I think `this.statusCode >= 500 && this.statusCode <= 599` would be easier to read. 🙂

packages/builder-util-runtime/src/publishOptions.ts
```diff
@@ -146,6 +146,39 @@ export interface GenericServerOptions extends PublishConfiguration {
...
+  /**
+   * Keygen product's UUID
+   */
+  readonly product: string
+
+  /**
+   * The channel.
+   * @default stable
+   */
+  readonly channel?: string | null
```
ezekg, line 174: Should this match what the API allows?

```ts
readonly channel?: 'stable' | 'rc' | 'beta' | 'alpha' | 'dev' | null
```

Fine either way. The API will reject channels that do not match one of the above.

ezekg (approved):
Overall, looks great. A couple notes/nits which can be safely ignored. Excited for this! 🙂

packages/builder-util-runtime/src/httpExecutor.ts
mmaietta: Valid. Updating that

packages/app-builder-lib/src/publish/KeygenPublisher.ts
mmaietta: More so personal preference, I just strongly dislike force casting or force unwrapping a variable

packages/builder-util-runtime/src/publishOptions.ts
mmaietta: Great point! I wanted to leave it dynamic so that any updates from your side wouldn't necessarily require updating electron-builder as well.
That being said, it's much more handy/efficient to have the channel info all provided up front. Will update.

PR #6176, opened 2021-08-23
feat: add `beforePack` hook

gaodeng:


packages/app-builder-lib/src/configuration.ts
```diff
@@ -262,7 +264,7 @@ export interface Configuration extends PlatformSpecificBuildOptions {
   readonly removePackageKeywords?: boolean
 }
 
-export interface AfterPackContext {
+export interface PackContext {
```
mmaietta, line 267: As much as I like this simplification, it does result in a breaking change of the electron-builder API since this interface is exposed externally for users integrated with our hooks. To me, that feels like an unnecessary `major` semver version jump.

My advice would be to create a new interface that simply copies with explicit purpose.
```
interface PackContext {
...
}
export interface AfterPackContext extends PackContext {}
export interface BeforePackContext extends PackContext {}
```
That would retain backward compatibility.

gaodeng: Good point. Thanks for your guidance, I have made changes to fix the problem

PR #6212, opened 2021-08-30
fix: (mac) Fix intel mac upgrade flow when both x64 and arm64 published

johnnyopao:
On Intel macs, the error `No files provided` is thrown on the latest version of electron-updater (4.5.1) when the latest files include both arm64 and x64 builds.

The current filtering done when there are arm64 files available will return an empty array since there are arm64 files but we are not on a arm64 machine
https://github.com/electron-userland/electron-builder/blob/64fd696527e6f07cea7efc494d667276a9001430/packages/electron-updater/src/MacUpdater.ts#L55-L57

This change adjusts MacUpdater to filter the files to this criteria: 

- On an Intel Mac: Filter files to only include non-arm64 builds
- On an arm64 Mac: Filter files to only include arm64 builds
- If no arm64 builds are available: This will fallback to the condition to filter files to only include non-arm64 builds

**Steps to reproduce the bug:**
1. Install ` "electron-updater": "^4.5.1"`
2. Publish two versions of your app with both `x64` and `arm64` builds
```sh
electron-builder -p always --mac --x64 --arm64 
```
3. On an intel mac try to upgrade from the previous version to another

**Expected Behavior**
The files list is filtered down and the x64/intel build is selected for download

**Actual behavior**
No files are returned after filtering. The ` Error: No files provided` error is thrown

### DIFF
--- a/packages/electron-updater/src/MacUpdater.ts
+++ b/packages/electron-updater/src/MacUpdater.ts
@@ -40,7 +40,7 @@ export class MacUpdater extends AppUpdater {
 
     // detect if we are running inside Rosetta emulation
     const sysctlRosettaInfoKey = "sysctl.proc_translated"
-    let isRosetta: boolean
+    let isRosetta = false
     try {
       this.debug("Checking for macOS Rosetta environment")
       const result = execFileSync("sysctl", [sysctlRosettaInfoKey], { encoding: "utf8" })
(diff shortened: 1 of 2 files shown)


mmaietta (approved):

packages/electron-updater/src/MacUpdater.ts
```diff
@@ -50,13 +50,18 @@ export class MacUpdater extends AppUpdater {
       log.warn(`sysctl shell command to check for macOS Rosetta environment failed: ${e}`)
     }
 
+    const isArm64Mac = process.arch === "arm64" || isRosetta
+
     // allow arm64 macs to install universal or rosetta2(x64) - https://github.com/electron-userland/electron-builder/pull/5524
     const isArm64 = (file: ResolvedUpdateFileInfo) => file.url.pathname.includes("arm64") || file.info.url?.includes("arm64")
-    if (files.some(isArm64)) {
-      files = files.filter(file => (process.arch === "arm64" || isRosetta) === isArm64(file))
```
quanglam2807, line 56: Would you mind explaining the change you made here, @johnnyopao? From my understanding, `files = files.filter(file => (process.arch === "arm64" || isRosetta) === isArm64(file))` already handles both Intel & ARM64 cases. It would be equivalent to `files = files.filter(file => true === isArm64(file))` on ARM64 Macs and `files = files.filter(file => false === isArm64(file))` on Intel Macs.

packages/electron-updater/src/MacUpdater.ts
```diff
@@ -50,13 +50,18 @@ export class MacUpdater extends AppUpdater {
...
+    if (isArm64Mac && files.some(isArm64)) {
+      files = files.filter(file => isArm64Mac === isArm64(file))
+    } else {
+      files = files.filter(file => !isArm64(file))
```
quanglam2807, line 60: This is actually unnecessary as if there are no ARM64 files, we don't need to filter the files at all?

packages/electron-updater/src/MacUpdater.ts
quanglam2807: Ah I see, one of my friend pointed out `isRosetta === undefined` causes the problem previously.

quanglam2807: Ah I see, `isRosetta` could be `undefined`, causing the problem.

PR #6224, opened 2021-09-03
fix: update-desktop-database command not found

alx696:
```
/var/lib/dpkg/info/polong.postinst:行10: update-desktop-database：未找到命令
```

Debian 11 KDE can not execute `update-desktop-database`, check if a command exists. 

reference vscode /var/lib/dpkg/info/code.postinst
```
# Update mimetype database to pickup workspace mimetype
if hash update-mime-database 2>/dev/null; then
        update-mime-database /usr/share/mime
fi
```

Test under:
```
Operating System: Debian GNU/Linux 11
KDE Plasma Version: 5.20.5
KDE Frameworks Version: 5.78.0
Qt Version: 5.15.2
Kernel Version: 5.10.0-8-amd64
OS Type: 64-bit
```

### DIFF
--- a/packages/app-builder-lib/templates/linux/after-install.tpl
+++ b/packages/app-builder-lib/templates/linux/after-install.tpl
@@ -6,5 +6,10 @@ ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
 # SUID chrome-sandbox for Electron 5+
 chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
 
-update-mime-database /usr/share/mime || true
-update-desktop-database /usr/share/applications || true
+if hash update-mime-database 2>/dev/null; then
+    update-mime-database /usr/share/mime
+fi
+
+if hash update-desktop-database 2>/dev/null; then
+    update-desktop-database /usr/share/applications
+fi
--- a/packages/app-builder-lib/templates/linux/after-install.tpl
+++ b/packages/app-builder-lib/templates/linux/after-install.tpl
@@ -6,9 +6,7 @@ ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
 # SUID chrome-sandbox for Electron 5+
 chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
 
-if hash update-mime-database 2>/dev/null; then
-    update-mime-database /usr/share/mime
-fi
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/templates/linux/after-install.tpl
```diff
@@ -6,5 +6,10 @@ ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
 # SUID chrome-sandbox for Electron 5+
 chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
 
-update-mime-database /usr/share/mime || true
-update-desktop-database /usr/share/applications || true
+if hash update-mime-database 2>/dev/null; then
+    update-mime-database /usr/share/mime
```
mmaietta, line 10: Is there ever a chance this could fail? Would it be worth keeping the `|| true` for both commands to play it safe?

alx696: `update-mime-database` can execute correctly. `|| true` [used to ignore error](https://unix.stackexchange.com/questions/325705/why-is-pattern-command-true-useful/325727) , but error message is still displayed. `if hash update-mime-database 2>/dev/null` check the command before execution, no error message is displayed. I have installed vscode under Ubuntu 20.04(GNOME), works fine. VScode has a large number of users, so there should be no problem.

`|| true` and `if hash update-mime-database 2>/dev/null` both used to ignore error, keep bash script continue execution.  Since it doesn't matter if there is an error, there is no need to display an error message.

Of course, the safest thing is to keep `update-mime-database /usr/share/mime || true` and only change `update-desktop-database`.

packages/app-builder-lib/templates/linux/after-install.tpl
```diff
@@ -7,4 +7,7 @@ ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
 chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
 
 update-mime-database /usr/share/mime || true
-update-desktop-database /usr/share/applications || true
+
+if hash update-desktop-database 2>/dev/null; then
+    update-desktop-database /usr/share/applications
```
mmaietta, line 12: Can you please add `|| true` to this line in case the cmd were to fail?

mmaietta:
Also, please update the test snapshot that is failing
```
Snapshot name: `executable path in postinst script 5`
```

PR #6234, opened 2021-09-06
Add support for multiple extensions to Keygen publisher

### LINKED ISSUE #6229: Keygen Publisher fails with VERSION_TAKEN for blockmap artifacts
* **Electron-Builder Version**: 22.13.0
* **Node Version**: 16.6.1
* **Electron Version**: 14.0.0
* **Target**: mac

Hi! 👋 I'm really excited to see an official Keygen publisher. Zeke shared the news with me last week. As part of a new major release, my team is working on migrating our home-grown publisher/provider over to the official one, but we're encountering an issue when publishing on macOS.

When publishing a new release using the new Keygen publisher, the publish is failing with a `VERSION_TAKEN` error code. After digging around a bit, this seems to only happen for `blockmap` artifacts. I think this is because electron-builder [is setting the release filetype to `.blockmap`](https://github.com/electron-userland/electron-builder/blob/f45110cbf66572d5748d21fc24dc26cabd06f35f/packages/app-builder-lib/src/publish/KeygenPublisher.ts#L81) for both the `.zip` and `.dmg` artifacts.

Checking out my Keygen request logs, I get visibility into the following API requests from electron-builder:

The first release request for the `.dmg` artifact is successful,

```json
{
  "data": {
    "type": "release",
    "attributes": {
      "filename": "my-app-3.0.0.dmg.blockmap",
      "filetype": ".blockmap",
      "filesize": 88502,
      "version": "3.0.0",
      "platform": "mac",
      "channel": "stable"
    },
    "relationships": {
      "product": {
        "data": {
          "type": "product",
          "id": "15a7e0f1-c7da-4c05-984e-47ef1b664e70"
        }
      }
    }
  }
}
```

But a subsequent request for the `.zip` artifact fails,

```json
{
  "data": {
    "type": "release",
    "attributes": {
      "filename": "my-app-3.0.0-mac.zip.blockmap",
      "filetype": ".blockmap",
      "filesize": 91526,
      "version": "3.0.0",
      "platform": "mac",
      "channel": "stable"
    },
    "relationships": {
      "product": {
        "data": {
          "type": "product",
          "id": "15a7e0f1-c7da-4c05-984e-47ef1b664e70"
        }
      }
    }
  }
}
```

It fails with the following API error response,

```json
{
  "meta": {
    "id": "a71ea375-aa9c-408d-bc47-b1cce5428202"
  },
  "errors": [
    {
      "title": "Unprocessable resource",
      "detail": "already exists for 'mac' platform with 'blockmap' filetype on 'stable' channel",
      "source": {
        "pointer": "/data/attributes/version"
      },
      "code": "VERSION_TAKEN"
    }
  ]
}
```

As you can see, this makes Keygen reject the releases as duplicates, [since the `version` property has to be unique for each `product + filetype + filetype + channel` combination](https://keygen.sh/docs/api/#releases). Meaning, `.zip.blockmap` and `.dmg.blockmap` are both being released as `.blockmap`, but they're conflicting with each other due to the "shared" filetype across multiple valid releases. Their docs mention using the "complete" filetype in cases where the artifact has multiple "valid" extensions.

To resolve this, I think electron-builder actually needs to set the filetype to `.zip.blockmap` and `.dmg.blockmap`, which is what our home-grown publisher currently does. This would also become an issue for `.gz` artifacts (for example `.tar.gz`), and any other multi-extension artifacts that are built by electron-builder.

If it helps, our home-grown publisher uses this package to get the "complete" extname: https://www.npmjs.com/package/path-complete-extname. I'm not sure if that'd work for everybody, but worth a mention. 🙂

ezekg:
Matching multiple file extensions is tricky, but with a set heuristic, we can do it. I couldn't get the test env running (probably because I'm missing required env vars), but here's [a repl I used to test](https://replit.com/@ezekg/BusyInternationalPcboard#index.js).

image

Closes #6229.


packages/app-builder-lib/src/publish/KeygenPublisher.ts
```diff
@@ -113,4 +113,35 @@ export class KeygenPublisher extends HttpPublisher {
     const { account, product, platform } = this.info
     return `Keygen (account: ${account}, product: ${product}, platform: ${platform}, version: ${this.version})`
   }
+
+  // Get the filetype from a filename. Returns a string of one or more file extensions,
+  // e.g. `zip`, `dmg`, `tar.gz`, `tar.bz2`, `exe.blockmap`. We'd use `path.extname()`,
+  // but it doesn't support multiple extensions, e.g. `dmg.blockmap`.
+  private getFiletype(filename: string): string {
```
mmaietta, line 120: What are your thoughts on extracting this to: https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/util/sanitizeFileName.ts
Renaming that file to just be `util/filename.ts`?

And then we need some unit tests on this. Particularly because I'm horrible at working with regexes, but also I foresee some bug reports due to custom naming/macro expansions that might occur with artifact names.

ezekg: Ah, I didn't know custom naming was possible. Maybe this isn't the best solution then, to avoid any future issues. Do you think it'd better to use a whitelist then?

```ts
function getFiletype(filename: string): string {
  let extname = path.extname(filename)

  // Append leading extension for known compressed extensions
  switch (extname) {
    case '.blockmap':
    case '.bz2':
    case '.gz':
    case '.lz':
      extname = path.extname(filename.replace(extname, '')) + extname

      break
  }

  return extname
}
```

Probably much clearer, since it doesn't use regex. 🙂

ezekg: Moved to a util function and added unit tests, but still unsure of how to run them. I'm not super familiar with Jest but anytime I try to run tests, Jest doesn't run anything. Still trying to figure that one out. 😅

packages/app-builder-lib/src/util/filename.ts
```diff
@@ -0,0 +1,39 @@
+// @ts-ignore
+import * as _sanitizeFileName from "sanitize-filename"
+import * as path from "path"
+
+export function sanitizeFileName(s: string): string {
```
ezekg, line 5: Worth noting that renaming this file may be a breaking change if anybody is relying on `util/sanitizeFileName`.

mmaietta (approved):

mmaietta:
Many thanks, @ezekg!

Issue #5965, opened 2021-06-17
Shortcut icon is broken when installing with MSI


jfily:
* **Version**: 22.11.7
* **Electron Version**: 11.4.8
* **Electron Type (current, beta, nightly)**: current
* **Target**: Windows (MSI)

When installing the app with MSI package, the shortcut icon created on the desktop is "broken".
The icon is good when starting the app in the task bar and in the list of installed programs though.

![image](https://user-images.githubusercontent.com/10220170/122354767-9ab84400-cf51-11eb-8f37-070a07261933.png)

The shortcut icon is good when installing from NSIS package.

Here is package.json (used create-react-app with template cra-template-electron-builder-typescript)

```JSON
{
    "name": "xxx",
    "version": "1.0.1",
    "private": true,
    "author": "xxx",
    "description": "xxx",
    "dependencies": {
        "@material-ui/core": "^4.11.4",
        "@material-ui/icons": "^4.11.2",
        "@testing-library/jest-dom": "^5.13.0",
        "@testing-library/react": "^11.2.7",
        "@testing-library/user-event": "^12.8.3",
        "@types/jest": "^26.0.23",
        "@types/node": "^14.17.3",
        "@types/react": "^17.0.11",
        "@types/react-dom": "^17.0.7",
        "cross-env": "^7.0.3",
        "electron-is-dev": "^1.2.0",
        "exceljs": "^4.2.1",
        "ldap-filters": "^2.2.2",
        "ldapts": "^3.0.1",
        "node-cmd": "^4.0.0",
        "react": "^17.0.2",
        "react-dom": "^17.0.2",
        "react-scripts": "4.0.3",
        "typescript": "^4.3.2",
        "web-vitals": "^1.1.2"
    },
    "scripts": {
        "start": "concurrently -k \"cross-env BROWSER=none npm run react-start\" \"wait-on http://localhost:3000 && electronmon .\"",
        "build": "react-scripts build",
        "test": "react-scripts test",
        "eject": "react-scripts eject",
        "dist:win": "npm run build && electron-builder -w -c.extraMetadata.main=build/electron.js",
        "dist:mac": "npm run build && electron-builder -m -c.extraMetadata.main=build/electron.js",
        "dist:linux": "npm run build && electron-builder -l -c.extraMetadata.main=build/electron.js",
        "react-start": "react-scripts start"
    },
    "eslintConfig": {
        "extends": [
            "react-app",
            "react-app/jest"
        ]
    },
[...]

PR #6247, opened 2021-09-09
fix(msi): fix broken shortcut icon when using msi target

erikrz:
fix (msi): broken shortcut icon (#5965)

Add MSI option "iconId", replaces old static "icon.ico" icon identifier.

### DIFF
--- a/.changeset/tiny-dolls-smoke.md
+++ b/.changeset/tiny-dolls-smoke.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix (msi): broken shortcut icon for desktop and startup entry (#5965)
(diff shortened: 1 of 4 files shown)


mmaietta:
Is this backward compatible with previous builds? Just want to make sure the icon doesn't break when an update is installed.

erikrz:
Yes, it does not impact app icons in taskbar, installer and uninstaller. 
It only fixes start menu and desktop shortcut icons, which have been broken for a while because we did not comply with how the data is required by Wix.

mmaietta:
Can you try rebasing this off latest `master`? I'm not sure why the doc generator is failing.

hasanatmango:
@erikrz Where we need to add `iconId` in package.json file ?

PR #6276, opened 2021-09-23
Use fs-extra remove instead of fs.rm

pgherveou:
fs.rm breaks for me when the builder is trying to remove empty language folder.
fs-extra remove method seems to be a better choice here (see doc: https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/remove.md#removepath-callback)


mmaietta:
Please see failing tests. Looks like there's unused vars.

Also, is this swap from fs.rm to fs-extra needed/useful anywhere else?

jimmywarting:
I hope fs-extra could one day be removed entirely by the built in fs to reduce dependencies. maybe a quick check/try-catch is good enough?

mmaietta:
Closing due to inactivity. Feel free to reopen if unpausing work :)

Issue #6292, opened 2021-09-27
BitBucket Publish doesn't work if user and owner are different


sr258:
The way BitbucketPublisher is written, you can't use it to publish to repositories if you are not the owner of them.

It constructs the API endpoint like this:
https://github.com/electron-userland/electron-builder/blob/d017cd7de3fbee17176d00daea13151706450090/packages/app-builder-lib/src/publish/BitbucketPublisher.ts#L26

And the HTTP Basic auth header is created like this:
https://github.com/electron-userland/electron-builder/blob/d017cd7de3fbee17176d00daea13151706450090/packages/app-builder-lib/src/publish/BitbucketPublisher.ts#L63-L66

As you can see, owner and acting user must be identical. There can be situations in which you want to push to an organisation's repository and the username you're using is not identical to the organisation's name.

PR #6293, opened 2021-09-27
fix(publish): Bitbucket publish can have username different from owner

sr258:
The way BitbucketPublisher is written, you can't use it to publish to repositories if you are not the owner of them. The PR introduces a new environment variable that allows you to change the override the username if necessary.

Closes https://github.com/electron-userland/electron-builder/issues/6292

### DIFF
--- a/packages/app-builder-lib/src/publish/BitbucketPublisher.ts
+++ b/packages/app-builder-lib/src/publish/BitbucketPublisher.ts
@@ -1,4 +1,4 @@
-import { Arch, InvalidConfigurationError, isEmptyOrSpaces } from "builder-util"
+import { Arch, InvalidConfigurationError, isEmptyOrSpaces, log } from "builder-util"
 import { httpExecutor } from "builder-util/out/nodeHttpExecutor"
 import { ClientRequest, RequestOptions } from "http"
 import { HttpPublisher, PublishContext } from "electron-publish"
@@ -6,6 +6,7 @@ import { BitbucketOptions } from "builder-util-runtime/out/publishOptions"
 import { configureRequestOptions, HttpExecutor } from "builder-util-runtime"
 import * as FormData from "form-data"
 import { readFile } from "fs/promises"
+
 export class BitbucketPublisher extends HttpPublisher {
   readonly providerName = "bitbucket"
   readonly hostname = "api.bitbucket.org"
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/publish/BitbucketPublisher.ts
```diff
@@ -18,11 +18,13 @@ export class BitbucketPublisher extends HttpPublisher {
     super(context)
 
     const token = process.env.BITBUCKET_TOKEN
+    const username = process.env.BITBUCKET_USERNAME
+
```
mmaietta, line 22: Can you add some warning logging so that this usage is evident that we're defaulting to the user repo?
```
    if (isEmptyOrSpaces(username)) {
      log.warn('No Bitbucket username provided via "BITBUCKET_USERNAME". Defaulting to use repo owner.')
    }
```

packages/app-builder-lib/src/publish/BitbucketPublisher.ts
```diff
@@ -18,11 +18,13 @@ export class BitbucketPublisher extends HttpPublisher {
...
     if (isEmptyOrSpaces(token)) {
       throw new InvalidConfigurationError(`Bitbucket token is not set using env "BITBUCKET_TOKEN" (see https://www.electron.build/configuration/publish#BitbucketOptions)`)
     }
     this.info = info
-    this.auth = BitbucketPublisher.convertAppPassword(this.info.owner, token)
+    this.auth = BitbucketPublisher.convertAppPassword(this.info.owner, token, username)
```
mmaietta, line 27: I would recommend keeping the same function signature since it's already explicit and instead do:
```
this.auth = BitbucketPublisher.convertAppPassword(username ?? this.info.owner, token)
```

sr258:
@mmaietta Thank you for your feedback. I've changed the requested things. 

While working with the BitBucket publisher I've noticed that auto-updates don't work if the productName contains whitespaces as BitBucket converts whitespaces into underscores (_) when you upload files. The latest.yml file references the files with whitespaces, so the download doesn't work. I'd like to submit a PR to fix this. Should I include this in this PR (one big BitBucket PR) or create a separate one?

mmaietta:
Oomph. Great detection! Please open as new PR so that we can keep each all explicit.

mmaietta (approved):

PR #6307, opened 2021-09-30
fix(electron-updater): auto-updates work with filename with whitespaces

sr258:
There was a bug that prevented the auto updater from working if you used Bitbucket and had a whitespace in the productName. The reason was that Bitbucket converts whitespaces into underscores when files are uploaded into the download section. However, in electron-updater whitespaces are escaped with %20, which leads to a 404.

I've added the Bitbucket style conversion into the BitbucketProvider so that auto-updates also work when using whitespaces in productNames.

### DIFF
--- a/.changeset/breezy-humans-notice.md
+++ b/.changeset/breezy-humans-notice.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+BitbucketProvider converts whitespaces in filenames to underscores as does Bitbucket itself when uploading files
--- a/packages/electron-updater/src/providers/BitbucketProvider.ts
+++ b/packages/electron-updater/src/providers/BitbucketProvider.ts
@@ -33,7 +33,10 @@ export class BitbucketProvider extends Provider<UpdateInfo> {
   }
 
   resolveFiles(updateInfo: UpdateInfo): Array<ResolvedUpdateFileInfo> {
-    return resolveFiles(updateInfo, this.baseUrl)
+    return resolveFiles(
+      { ...updateInfo, files: updateInfo.files?.map(f => ({ ...f, url: f.url.replaceAll(" ", "_") })), path: updateInfo.path?.replaceAll(" ", "_") },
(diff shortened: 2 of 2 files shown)


mmaietta:
Can you try rebasing this PR off latest master?

One of the failing tests has nothing to do with your PR, but on principle, I shouldn't continue with merging.

mmaietta (approved):

sr258:
@mmaietta Looks like rebasing doesn't help.

mmaietta:
Can you please retry?
I've updated the test on master branch

sr258:
@mmaietta Before I do that: I've noticed that the GitHub publisher / provider has a similar challenge, as GitHub converts whitespaces to dots when you upload files. When I upload files through electron-builder I get dashes as separators. So the principle is the same as in Bitbucket but there's a different way of escaping the whitespace. When you use the GitHub publisher, the latest.yml files are changed accordingly, whereas my PR changes the update provider. I want to have a look at how you did this for GitHub and change the PR accordingly, for consistency.

robertpatrick:
> @mmaietta Before I do that: I've noticed that the GitHub publisher / provider has a similar challenge, as GitHub converts whitespaces to dots when you upload files. When I upload files through electron-builder I get dashes as separators. So the principle is the same as in Bitbucket but there's a different way of escaping the whitespace. When you use the GitHub publisher, the latest.yml files are changed accordingly, whereas my PR changes the update provider. I want to have a look at how you did this for GitHub and change the PR accordingly, for consistency.

Yes, the GitHub UI converts spaces to dots but electron-publisher converts them to dashes when it uploads them.  Since electron-publisher does not support proxy servers (yet?), I wrote my own code to do the publishing and I simply convert the spaces in file names to dashes and everything works fine.  Personally, I don't think it is necessary for electron-updater/electron-publisher to track the behavior of the providers' web application.

mmaietta:
Closing due to inactivity. Feel free to reopen if unpausing work :)

Issue #6282, opened 2021-09-24
Broken icon and dmg layout with non-ASCII characters in productName on macOS


sr258:
* **Electron-Builder Version**: 22.11.7
* **Node Version**: 16.5.0
* **Electron Version**: 15.0.0
* **Electron Type (current, beta, nightly)**: current
* **Target**: macOS dmg

When building an app that has non-ASCII characters like 'ü' in its product name, there are several issues:

the icon of the .app file is incorrect / empty
should be: 
<img width="148" alt="Bildschirmfoto 2021-09-24 um 21 48 56" src="https://user-images.githubusercontent.com/15268740/134732113-7de6a61f-1c82-48a7-970b-5793cd1f0802.png">
Is: 
<img width="148" alt="Bildschirmfoto 2021-09-24 um 21 47 13" src="https://user-images.githubusercontent.com/15268740/134732153-6f68f36b-76c7-4153-ae66-10165548998e.png">

Furthermore when opening the dmg file the layout of the contents is broken:

Should be:
<img width="538" alt="Bildschirmfoto 2021-09-24 um 21 48 50" src="https://user-images.githubusercontent.com/15268740/134732231-7b3ea049-a568-445f-abcb-0ca6a4d1f21f.png">

Is: 
<img width="540" alt="Bildschirmfoto 2021-09-24 um 21 47 24" src="https://user-images.githubusercontent.com/15268740/134732251-5ea9c6ac-c6ac-4d09-b542-5fd0173152ca.png">

I've created a minimal sample that reproduces the error here: https://github.com/sr258/electron-builder-umlaut-sample

Steps to reproduce:

- clone the repo
- yarn install
- CSC_IDENTITY_AUTO_DISCOVERY=false yarn dist

I would be willing to contribute to a fix, but I'm new to electron-builder and unsure where to look. How are macOS icons set for .app files and where is the dmg layout created?

sr258:
I've identified two causes for this:

- macOS doesn't like the non-ASCII character in the app's icon name. 
- the dmg-builder Python tool is broken with non-ASCII characters and doesn't set the correct icon Y position

I've found a fix and a workaround. Execute this script:

```sh
# Needed to create fully funtioning DMG and zip files for macOS.
# You may need to allow the execution of the downloaded 7zip program in the security settings of system control.

[...]

PR #6308, opened 2021-09-30
fix(app-builder-lib): macOS packager uses static icon name

sr258:
The filename of the app icon in macOS is now always 'icon.icns' instead of a derivate of the product name. The reason for this change is that macOS doesn't display icons with non-ASCII characters in their names, which is quite possible in languages other than English.

This fixes one of the issues described here https://github.com/electron-userland/electron-builder/issues/6282


mmaietta:
Can you try rebasing this off latest master? Not sure why one of the tests are failing

sr258:
@mmaietta I rebased this and fixed the test snapshots as far as possible. Unfortunately, I can't really run the tests on my dev machine, as I use an M1 processor, which causes lots of tests to fail even though they shouldn't (see #6310). I hope the CI pipeline runs through this time.

PR #6381, opened 2021-10-31
fix: get arch wrong with 'process.arch' in mac silicon

zhanjinfeng:


packages/electron-updater/src/MacUpdater.ts
```diff
@@ -50,7 +50,21 @@ export class MacUpdater extends AppUpdater {
...
+          if(err) {
+            resolve(process.arch === 'arm64')
+            return
+          }
+          resolve(stdout.includes("ARM"))
+        })
+      })
+    }
+
+    const isArm64Mac = await getArch() || isRosetta;
```
mmaietta, line 67: We need a try-catch here, and I'm not sure why we can't still use `execFileSync` to mirror the other code. I think we should replace the entire section with:
```

let isArm64Mac = false

// detect if we are running inside Rosetta emulation
const sysctlRosettaInfoKey = "sysctl.proc_translated"
try {
  this.debug("Checking for macOS Rosetta environment")
  const result = execFileSync("sysctl", [sysctlRosettaInfoKey], { encoding: "utf8" })
  const isRosetta = result.includes(`${sysctlRosettaInfoKey}: 1`)
  log.info(`Checked for macOS Rosetta environment (isRosetta=${isRosetta})`)
  isArm64Mac = isRosetta
} catch (e) {
  log.warn(`sysctl shell command to check for macOS Rosetta environment failed: ${e}`)
}

try {
  this.debug("Checking for arm64 in uname")
  const result = execFileSync("uname", ['-a'], { encoding: "utf8" })
  const isArm = result.includes('ARM')
  log.info(`Checked 'uname -a': arm64=${isArm}`)
  isArm64Mac = isArm64Mac || isArm
} catch (e) {
  log.warn(`uname shell command to check for arm64 failed: ${e}`)
}

isArm64Mac = isArm64Mac || process.arch === 'arm64'
```

zhanjinfeng: appreciate！ I will improve my code with your suggestion and resolve it

packages/electron-updater/src/MacUpdater.ts
```diff
@@ -50,7 +50,18 @@ export class MacUpdater extends AppUpdater {
...
     }
 
-    const isArm64Mac = process.arch === "arm64" || isRosetta
+    let isArm64Mac = false
+    try {
+      this.debug("Checking for arm64 in uname")
+      const result = execFileSync("uname", ['-a'], { encoding: "utf8" })
+      const isArm = result.includes('ARM')
+      log.info(`Checked 'uname -a': arm64=${isArm}`)
+      isArm64Mac = isArm64Mac || isArm
```
kaatt, line 59: this is the equivalent of `isArm64Mac = false || isArm`

kaatt:
why is this needed? `sysctl.proc_translated` should do the job too

PR #6390, opened 2021-11-03
fix: quitAndInstall not working on macOS with autoInstallOnAppQuit=false

### LINKED ISSUE #6389: quitAndInstall on macOS doesn't do anything
* **Electron-Builder Version**: 22.13.1

* **Node Version**: 14.17

* **Electron Version**: 13.6.1

* **electron-updater version**: 4.6.1

* **Target**: macOS x64

My code requests that `electron-updater`'s `autoUpdater` (now referred to as just `autoUpdater`) should check for updates and then download that update. And then once the user has confirmed that they wish to proceed calls `autoUpdater.quitAndInstall()`. Then nothing happens.

I have attached a logger which does output that the update was installed. But when playing around with the code I see that the native event `"downloaded-update"` never gets emitted so `squirrelDownloadedUpdate` is always `false`.

Nokel81:
- Would never call nativeUpdater.checkForUpdates() if not installing on
  quit. Thus neither squirrelDownloadedUpdate was true nor the
  nativeUpdater's event "update-downloaded" be emitted. Leading to
  nothing happening.

- This commit fixes it by calling the checkForUpdates() in
  MacUpdater.quitAndInstall() when necessary

fixes #6389

### DIFF
--- a/.changeset/olive-weeks-look.md
+++ b/.changeset/olive-weeks-look.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+Fix updating only on demand not working on macOS
--- a/packages/electron-updater/src/MacUpdater.ts
+++ b/packages/electron-updater/src/MacUpdater.ts
@@ -176,18 +176,18 @@ export class MacUpdater extends AppUpdater {
       // update already fetched by Squirrel, it's ready to install
       this.nativeUpdater.quitAndInstall()
     } else {
+      // Quit and install as soon as Squirrel get the update
+      this.nativeUpdater.on("update-downloaded", () => {
+        this.nativeUpdater.quitAndInstall()
+      })
+
       if (!this.autoInstallOnAppQuit) {
         /**
          * If this was not `true` previously then MacUpdater.doDownloadUpdate()
          * would not actually initiate the downloading by electron's autoUpdater
          */
-        this.nativeUpdater.checkForUpdates();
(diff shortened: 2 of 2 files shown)


PSNRigner:
Hey,

We also just found out a fix for that issue, which seems actually better.
Here's the full function:

```typescript
quitAndInstall(): void {
    if (this.squirrelDownloadedUpdate) {
      // Update already fetched by Squirrel, it's ready to install
      this.nativeUpdater.quitAndInstall()
    } else {
      // Quit and install as soon as Squirrel get the update
      this.nativeUpdater.on("update-downloaded", () => {
        this.nativeUpdater.quitAndInstall()
      })

      // And trigger the update
      if (!this.autoInstallOnAppQuit) {
          this.nativeUpdater.checkForUpdates();
      }
    } 
```

This is a very urgent issue @mmaietta and should be put in pre-release as soon as possible, It's affecting all our userbase.

Nokel81:
Yeah I was really surprised that there was no reported issue about this after I figured out what was going on.

I agree that yours is probably slightly more resilient, will update.

mmaietta:
Oh man, I'm so behind on my notifications. Will get this in ASAP

Nokel81:
@mmaietta Done

mmaietta:
Released **next**:
electron-builder@22.14.6
electron-updater@4.6.2

PR #6395, opened 2021-11-04
fix: Emit 'before-quit-for-update' when updating via novel means

### LINKED ISSUE #6394: No "before-quit-for-update" event when updating an AppImage
* **Electron-Builder Version**: 22.13.1

* **Node Version**: 14.17

* **Electron Version**: 13.6.1

* **electron-updater version**: 4.6.1

* **Target**: AppImage

I know that AppImage is only supported via this package and not natively via electron. But I think it would be nice if the event flow was more similar.

Nokel81:

fixes #6394

### DIFF
--- a/.changeset/sixty-queens-tell.md
+++ b/.changeset/sixty-queens-tell.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+Emulate electron.autoUpdater's event lifecycle for AppImageUpdater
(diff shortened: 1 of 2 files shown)


mmaietta:
Please rebase off latest master when you have a chance.

Also, is this a feature (minor) or a fix (patch)? It seems like we could call it a patch because it's "fixing" the updates for appimages. But I'm also not sure on the approach being taken.

Nokel81:
Sure I can rebase. I think I specified it as a patch in the changelog because I think that it should act like the other updates.

Nokel81:
@mmaietta Done

PR #6402, opened 2021-11-08
fix: latest node-gyp with old Electron versions

zcbenz:
Since node-gyp >= 8.4.0, building modules for old versions of Electron requires passing `--force-process-config` due to them lacking a valid config.gypi in their headers.

See also https://github.com/nodejs/node-gyp/pull/2497.


mmaietta (approved):

mmaietta:
Thank you @zcbenz !

PR #6429, opened 2021-11-17
feat: make `--no-sandbox` optional for building with AppImage

devinbinnie:
I've noticed that `--no-sandbox` is hardcoded into the arguments list when building an `AppImage`. This seems to be a good default for many implementations, but doesn't work for all of them. Example being the Mattermost Desktop App (see this issue: https://github.com/mattermost/desktop/issues/1804)

I've created this PR in order to potentially add an override option for users to specify their own arguments when launching the AppImage, using the `executableArgs` field which doesn't appear to be used by the AppImage at this time.

Let me know if this change makes sense and if I can make any further edits to support this.

### DIFF
--- a/packages/app-builder-lib/src/targets/AppImageTarget.ts
+++ b/packages/app-builder-lib/src/targets/AppImageTarget.ts
@@ -20,7 +20,7 @@ export default class AppImageTarget extends Target {
     super("appImage")
 
     this.desktopEntry = new Lazy<string>(() =>
-      helper.computeDesktopEntry(this.options, "AppRun --no-sandbox %U", {
+      helper.computeDesktopEntry(this.options, `AppRun ${[...(this.options.executableArgs ?? ["--no-sandbox"]), "%U"].join(" ")}`, {
         "X-AppImage-Version": `${packager.appInfo.buildVersion}`,
       })
     )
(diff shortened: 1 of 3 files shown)


packages/app-builder-lib/src/targets/AppImageTarget.ts
```diff
@@ -20,7 +20,7 @@ export default class AppImageTarget extends Target {
     super("appImage")
 
     this.desktopEntry = new Lazy<string>(() =>
-      helper.computeDesktopEntry(this.options, "AppRun --no-sandbox %U", {
+      helper.computeDesktopEntry(this.options, `AppRun ${[...(this.options.executableArgs ?? ["--no-sandbox"]), "%U"].join(" ")}`, {
```
mmaietta, line 23: Can you please split in to its own const?
```
const args = this.options.executableArgs?.join(" ") || "--no-sandbox"
helper.computeDesktopEntry(this.options, `AppRun ${args} %U`, {
```

mmaietta (approved):

mmaietta:
Can you please rebase off latest master? That should resolve the test failures

devinbinnie:
@mmaietta Done, I think I accidentally wiped out one of the `.md` files you added though. Might need to be re-added, let me know if I can help with that, apologies.

Issue #5468, opened 2020-12-11
Win64 system wide checkForUpdatesAndNotify() uninstalls but doesn't install


timvdalen:
* **Version**: 22.9.1

* **Electron Version**: 10.1.5
* **Electron Type (current, beta, nightly)**: current
* **electron-updater Version**: 4.3.5

* **Target**: win64

I'm running a generic server auto-update on Windows using NSIS.
Auto update works correctly if the application is installed per-user, but when the application is installed system wide:

1. The user launches an old version of the app
2. The user receives a notification that the app is ready to update on close
3. The user closes the app and received a UAC popup for `old-uninstaller.exe` (from a temporary folder)
4. The user gives permission
5. The old version of application is deleted
6. Nothing else happens

**Expected behaviour**
The new version of the application is installed

timvdalen:
I unfortunately don't have any interesting logs for this behaviour, the app simply logs that it has fout an update and is going to execute it.

timvdalen:
Still relevant. I've donated to the project to see if this can get priority.

randomgithubuser1:
I'm also having this same issue.  It seems there is an issue with silent update and machine wide installation.

Another thread about it here: https://github.com/electron-userland/electron-builder/issues/4474

There is a solution in that thread that checks the registry and sets isSilent to false if the app is installed for all users.

I'm not sure how to set isSilent when using checkForUpdatesAndNotify() though since it doesn't look like it accepts any arguments.

timvdalen:
@yorannn can you link to the workaround that ended up working for us?

Yorannn:
> @Yorannn can you link to the workaround that ended up working for us?

This was the workaround I used 
https://github.com/electron-userland/electron-builder/issues/2363

timvdalen:
Should be fixed by #6450

Issue #6425, opened 2021-11-16
electron-updater@next breaks NSIS "per machine" quitAndInstall() when not using silent mode


robertpatrick:
* **Electron-Builder Version**: 22.14.7

* **Node Version**: 16.13.0

* **Electron Version**: 15.3.1
* **Electron Type (current, beta, nightly)**: current

* **Electron-Updater Version**: 4.6.2

* **Target**:  Windows NSIS

When we move from electron-updater `latest` to `next` to pick up the `autoInstallOnAppQuit` fix for MacOS, it breaks the previously working auto-updater for NSIS when installing for "All Users" (works fine if installed using the "Only for Me" option).

What I am seeing is a little strange in that the very first time I run the installer (unsigned and built locally), the quitAndInstall() call works as expected.  However, if I uninstall the application and then install again, the quitAndInstall() call gets to the point of causing Windows to pop up the UAC dialog asking for permission but then nothing else happens (that is, the Installer dialog does not appear).  If I run the app again and call quitOnInstall(), it works as expected.  After that first time, this behavior is repeatable every single time.

In looking at the log file (produced by our logger that we pass onto the autoUpdater), the log entries are always the same:

```
info: Checking for update
info: Found version 1.0.0 (url: WebLogic-Kubernetes-Toolkit-UI-Setup-1.0.0.exe)
info: Downloading update from WebLogic-Kubernetes-Toolkit-UI-Setup-1.0.0.exe
debug: updater cache dir: C:\Users\rhpat\AppData\Local\wktui-updater
info: Update has already been downloaded to C:\Users\rhpat\AppData\Local\wktui-updater\pending\WebLogic-Kubernetes-Toolkit-UI-Setup-1.0.0.exe).
info: Download complete, install type: now
info: Install on explicit quitAndInstall
info: Install: isSilent: false, isForceRunAfter: true, installPathRequiresElevation: true
info: isAdminRightsRequired is set to true, run installer using elevate.exe
```

If I don't change anything other than to drop back to electron-updater@latest, everything works as expected the first time, every time.  If it will help, I can pass you an installer built with electron-updater@next that reproduces the bad behavior.  Please let me know what I can do to help move this forward.

robertpatrick:
@mmaietta OK, so digging in and adding some extra logging, I figured out why `electron-updater@4.3.9` is working and `electron-updater@4.6.2` is not when the old version of the app is installed for all users.

[...]

PR #6438, opened 2021-11-20
fix(nsis): silent upgrade of per-machine install not elevated (#5468)

krisdages:
Without elevation, app is uninstalled, but new files can't be extracted
 and no error message is displayed.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/installer.nsi
+++ b/packages/app-builder-lib/templates/nsis/installer.nsi
@@ -84,6 +84,28 @@ FunctionEnd
 
 Section "install"
   !ifndef BUILD_UNINSTALLER
+    # If we're running a silent upgrade of a per-machine installation, elevate so extracting the new app will succeed.
+    # For a non-silent install, the elevation will be triggered when the install mode is selected in the UI,
+    # but that won't be executed when silent.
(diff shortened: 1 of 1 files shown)


mmaietta:
Can you please rebase off latest master? That should resolve the test failures

robertpatrick (commented):
While I think this change is probably ok, I am concerned that the change added in PR #6073 already is elevating all Windows "per machine" installs.  In debugging my issue #6425, it seems that we need to limit the change made by PR #6073 to only use `elevate.exe` when the install is running in silent mode (maybe by rolling back #6073 and merging this PR?).  

In `electron-updater@4.6.2` (which includes #6073), the `autoInstallOnAppQuit` feature (which run the installer in silent mode) is already working using the NSIS installer so it's not clear to me that this change will do something not already being done by #6073.  As I mentioned above, maybe the PR #6073 change should have been accomplished using this PR instead so that #6425 is resolved by not trying to elevate installation when `isSilent` is set to `false`?  Am I missing something?

mmaietta:
Closing PR in favor of https://github.com/electron-userland/electron-builder/pull/6450
Thank you @krisdages for your contribution, we've utilized it as part of https://github.com/electron-userland/electron-builder/pull/6450 for the full solution and credit to you has been given as well :)

PR #6449, opened 2021-11-28
fix(app-builder-lib): channel alternation for github is not working

saadshahd:
After upgrading from v22.11.11 to v22.14.8 I can't alternate the channel file name anymore due to schema validation.

### DIFF
--- a/.changeset/channel-alternation-github.md
+++ b/.changeset/channel-alternation-github.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+Enable channel alternation for github publishing provider.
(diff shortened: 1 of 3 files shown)


mmaietta:
Editing the schema.json directly isn't a long-term approach unfortunately, it's actually automatically generated via `pnpm schema`

My best guess is the interface that declares `channel` needs to expose it as public.
I think it needs to be added to this interface: https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util-runtime/src/publishOptions.ts#L74
And probably should look like this: https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util-runtime/src/publishOptions.ts#L147-L151
Then just run `pnpm schema` to regenerate the schema.json file (which I may rename to be `schema-generated.json` to make it more understandable in the future)

saadshahd:
@mmaietta Thank you for your comment I used `pnpm schema` as you advised.

mmaietta (approved):

PR #6450, opened 2021-11-28
fix(nsis): fix per-machine installs

### LINKED ISSUE #6425: electron-updater@next breaks NSIS "per machine" quitAndInstall() when not using silent mode

robertpatrick:
Add @krisdages changes from PR #6438 to elevate silent per-machine installs.
Remove PR #6073 that incorrectly elevates all per-machine installs, breaking interactive per-machine installs.

Closes #6425, #5468


robertpatrick:
@mmaietta Not sure what to do about these tests since I am unable to get the tests to execute even without these changes.

robertpatrick:
@mmaietta I am far from an NSIS scripting expert but the errors seem to indicate that the variable `hasPerMachineInstallation` is not always defined when the install script tries to access it.  

In looking at the place referenced by the comment in PR #6438 that I copied over, the variable is defined like this:

```
!ifndef INSTALL_MODE_PER_ALL_USERS
  !ifndef ONE_CLICK
    Var hasPerUserInstallation
    Var hasPerMachineInstallation
  !endif
  Var PerUserInstallationFolder
  ...
```

but in the changes in the `installer.nsi` file, it is protected only by:

```
  !ifndef BUILD_UNINSTALLER
    ${if} $hasPerMachineInstallation == "1" # set in onInit by initMultiUser
    ...
```

I don't feel like I understand the exact implications of these various ifndef variables but I am experimenting with changes that will both allow those 15 failing tests (which thankfully, I am able to run one-by-one from IntelliJ even though the full `pnpm test` is still failing with dozens of errors even when run directly on master) and my installer to work properly.  I will update the PR if/when I figure out the magic combination that fixes the failing unit tests and allows my installer to work properly...

robertpatrick:
@mmaietta OK, this looks like a winner.  All unit tests seem to be passing and the 4 installer scenarios are working with my application:

1. per-machine, quitAndInstall()
2. per-machine, autoInstallOnAppQuit
3. per-user, quitAndInstall()
4. per-user, autoInstallOnAppQuit

mmaietta:
This is awesome.

Thank you @robertpatrick  and @krisdages for your great work here!

mmaietta (approved):

PR #6472, opened 2021-12-04
fix(nsis): Ignore other users processes

### LINKED ISSUE #6104: NSIS install fails on terminal servers with multiple users already running the program
22.11.7

 Which version of node are you using? 
14.17.1

 Which version of electron are you using? 
8.5.5

Which version of electron-updater are you using (if applicable)? 
4.3.9

For which target are you building for? 
``` json
"win": {
            "icon": "resources/icons/win/icon.ico",
            "target": {
                "target": "nsis-web",
                "arch": [
                    "ia32",
                    "x64"
                ]
            },
            "artifactName": "${name}_setup_${version}.${ext}"
        },
        "nsisWeb": {
            "createDesktopShortcut": false,
            "createStartMenuShortcut": true,
            "menuCategory": true,
            "guid": "com.update-test",
            "perMachine": false,
            "allowElevation": true,
            "differentialPackage": true,
            "uninstallDisplayName": "${productName}",
            "runAfterFinish": true,
            "oneClick": true,
            "allowToChangeInstallationDirectory": false
        },
```

The issue I am having is the NSIS installer incorrectly determines the application is still running on terminal servers if any user has it open, even when using per user.  This blocks the install most of the time, or worse yet, auto kills all instances on the terminal server for all users when one user wants to update their individual install.  Both of these are serious bugs.  The installer should not check processes for other users when it is installed to one user folder.  And it absolutely should not be able to kill the process for other users.

To reproduce.  Build an NSIS-web installer.
Connect to a terminal server with 2 different users.  Install and run the application one one user.  Then try to install on the other, it will fail.
image
On the right is the application running on a terminal server user.  On the left is the installer trying to run on the same server but a different user

Is there a way to disable to is running check with a custom nsis script or anything?  We encountered this bug during a large deployment which we have now paused.

I-Otsuki:
Only find and kill processes owned by current user on per-user install.

Closes #6104

### DIFF
--- a/.changeset/few-elephants-rhyme.md
+++ b/.changeset/few-elephants-rhyme.md
@@ -1,0 +1,6 @@
+---
+"app-builder-lib": patch
+---
+
+fix(nsis): Ignore other users processes when installing for only current user
+Closes #6104
--- a/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
@@ -35,6 +35,16 @@
     !endif
 !macroend
 
+!macro FIND_PROCESS _FILE _ERR
+  !ifdef INSTALL_MODE_PER_ALL_USERS
+    ${nsProcess::FindProcess} "${_FILE}" ${_ERR}
+  !else
+    # find process owned by current user
+    nsExec::Exec `cmd /c tasklist /FI "USERNAME eq %USERNAME%" /FI "IMAGENAME eq ${_FILE}" | find "${_FILE}"`
+    Pop ${_ERR}
+  !endif
+!macroend
+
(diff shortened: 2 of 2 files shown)


mmaietta (approved):

indutny-signal:
Just a quick FYI, we are investigating a [report](https://github.com/signalapp/Signal-Desktop/issues/5854) from a person on Windows 10 Education being unable to update. So far everything indicates that this commit might be the culprit.

I have a feeling that the Education version of Windows may come without some of the command-line tools, but I'm yet to confirm this.

indutny-signal:
@I-Otsuki I think the code is correct and I honestly updated dozens of times myself on both Windows 10 and Windows 11 before we shipped it to users. We have a number of reports from people running into this and so far it looks like it is a small percentage of users, but it is far from a singular event as far as I can tell.

Unfortunately, we had another unrelated issue with yellow banner so things are bit confusing in the bug tracker now.

Issue #4988, opened 2020-05-19
Clarification on allowPrerelease, pre-releases, and auto-updates with channels


eengoron:
**Electron Builder Version**:
22.4.0

**Electron Updater Version**:
4.2.0

Hey Electron Builder Team, 

I was hoping I could get some clarification on the expected behavior of using `allowPrerelease` with update channels  in `electron-builder` because I'm noticing a weird issue on my end. Here's the situation:

I use a public Github repository as my provider for releases, and I have 3 update channels in my application `alpha`, `beta`, and `latest`. 

I noticed that if I have `autoUpdater.channel = 'beta'` and `autoUpdater.allowPrerelease = 'true'` set in my app, and I pre-release a new version on the `alpha` channel, the autoUpdater will see the new `alpha` pre-release while on the `beta` channel and upgrade.

For instance, if I'm currently on `v4.0.5` of my application with my `autoUpdater.channel = 'beta'` and `autoUpdater.allowPrerelease = 'true'` and then I pre-release version `4.0.6-alpha` in my releases repo, when I check for updates again, even though I'm on the `beta` channel, the `autoUpdater` will start downloading `4.0.6-alpha`. 

Is this to be expected? **Note**: This does not happen if I **release** version `4.0.6-alpha`. This only happens when `4.0.6-alpha` is in the pre-release state on GH. 

Thanks!

eengoron:
Turns out, channels are not supported with Github as the provider. See: https://github.com/electron-userland/electron-builder/issues/1722#issuecomment-310468372

Issue #1722, opened 2017-06-21
Support channels for Github provider


MaKleSoft:
* **Version**: 19.6.3
* **Target**: win, mac

When using the github provider, electron builder always generates `latest.yml` and `latest-mac.yml`, even for alpha/beta releases.

I dug into the source code and the problem seems to be this part: https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/src/publish/PublishManager.ts#L476

The channel is correctly extracted from the file name but not assigned to the options object for any provider other than `generic`. However, looking at the source code of `electron-updater` it looks like the Github provider should support channels as well?

MaKleSoft:
Ok so it seems that as of now, `GithubProvider` is always looking for `latest.yml` regardless of the actual channel: https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/GitHubProvider.ts#L62

Does this mean channels are simply not supported for the Github provider at this point? Is this an oversight or intentional? Seems like it would be trivial to implement this?

The whole channel implementation seems kind of broken imo. Why do we need separate channel files in the first place? Or why separate files for all platforms for that matter? Wouldn't it be easier to simply have one `latest.yml` and have all the information in there? Something like:

```yaml
version: 2.0.0-alpha.1
releaseDate: '2017-06-21T20:25:04.622Z'
win:
  githubArtifactName: padlock-setup-2.0.0-alpha.1.exe
  path: Padlock Setup 2.0.0-alpha.1.exe
  sha512: ...
  sha2: ...
mac:
  githubArtifactName: padlock-2.0.0-alpha.1-mac.zip
  path: Padlock-2.0.0-alpha.1-mac.zip
  sha512: ...
```

MaKleSoft:
Another problem is that for the generic provider, information about previous releases is lost when `latest.yml` is updated which is a problem if you want to roll back a release for example. Imo it would be more appropriate to have a `releases.yml`, like so:

```yaml
- version: 2.0.0-alpha.1
  releaseDate: ...
  win:
    githubArtifactName: padlock-setup-2.0.0-alpha.1.exe
    path: Padlock Setup 2.0.0-alpha.1.exe
    sha512: ...
    sha2: ...
  mac: ...
- version: 1.5.0
  releaseDate: ...
  win:
    githubArtifactName: padlock-setup-1.5.0.exe
    path: Padlock Setup 1.5.0.exe
    sha512: ...
    sha2: ...
  mac: ...
[...]

PR #6505, opened 2021-12-17
feat(updater): Add Channel Support for Github with PreRelease

KenCorma:
Allows for apps to run updates using provided channel track.
This only works if `options.allowPreRelease` is true.

If Channel is set in the updater, then latest will return the latest release that satisfies the logic in the [builder docs](https://www.electron.build/tutorials/release-using-channels.html#your-application)

If Channel is NOT set but prerelease is set, then latest will return the latest release.

technically this also supports custom channels, but if you run one of those you will never be able to return to the Latest track. 

I have marked this as a PR Draft as I would like to get discussion to make sure my logic is correct as well as this being my first contribution to this repo.

Tag:
#1722
#4988

### DIFF
--- a/.changeset/github-channel-support.md
+++ b/.changeset/github-channel-support.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+feat(updater): Add Channel Support for Github with PreRelease
--- a/.changeset/github-channel-support.md
+++ b/.changeset/github-channel-support.md
@@ -1,5 +1,5 @@
 ---
-"electron-updater": patch
+"electron-updater": minor
 ---
 
 feat(updater): Add Channel Support for Github with PreRelease
--- a/packages/electron-updater/src/providers/GitHubProvider.ts
+++ b/packages/electron-updater/src/providers/GitHubProvider.ts
@@ -109,7 +109,7 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
 
     let rawData: string
     let channelFile: string = ''
-    let channelFileUrl: string = ''
+    let channelFileUrl: any = ''
     const fetchData = async (channelName: string) => {
       channelFile = getChannelFilename(channelName)
       channelFileUrl = newUrlFromBase(this.getBaseDownloadPath(String(tag), channelFile), this.baseUrl)
--- a/packages/electron-updater/src/providers/GitHubProvider.ts
+++ b/packages/electron-updater/src/providers/GitHubProvider.ts
@@ -106,18 +106,36 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
       throw newError(`No published versions on GitHub`, "ERR_UPDATER_NO_PUBLISHED_VERSIONS")
     }
 
-    const channelFile = getChannelFilename(this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0] || 'latest')) : this.getDefaultChannelName())
-    const channelFileUrl = newUrlFromBase(this.getBaseDownloadPath(tag, channelFile), this.baseUrl)
-    const requestOptions = this.createRequestOptions(channelFileUrl)
+
     let rawData: string
-    try {
-      rawData = (await this.executor.request(requestOptions, cancellationToken))!
-    } catch (e) {
-      if (!this.updater.allowPrerelease && e instanceof HttpError && e.statusCode === 404) {
-        throw newError(`Cannot find ${channelFile} in the latest release artifacts (${channelFileUrl}): ${e.stack || e.message}`, "ERR_UPDATER_CHANNEL_FILE_NOT_FOUND")
+    let channelFile: string = ''
+    let channelFileUrl: string = ''
+    const fetchData = async (channelName: string) => {
+      channelFile = getChannelFilename(channelName)
+      channelFileUrl = newUrlFromBase(this.getBaseDownloadPath(String(tag), channelFile), this.baseUrl)
+      const requestOptions = this.createRequestOptions(channelFileUrl)
+      try {
+          return (await this.executor.request(requestOptions, cancellationToken))!
+      } catch (e) {
+          if (e instanceof HttpError && e.statusCode === 404) {
+              throw newError(`Cannot find ${channelFile} in the latest release artifacts (${channelFileUrl}): ${e.stack || e.message}`, "ERR_UPDATER_CHANNEL_FILE_NOT_FOUND")
+          }
+          throw e
       }
-      throw e
-    }
+  }
+  
+  try {
+      const channel = this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0] || 'latest')) : this.getDefaultChannelName()
+      rawData = await fetchData(channel)
+  } catch (e) {
+      if (this.updater.allowPrerelease) {
+          // Allow fallback to `latest.yml`
+          rawData = await fetchData(this.getDefaultChannelName())
+      } else {
+          throw e
+      }
+  }
+
 
     const result = parseUpdateInfo(rawData, channelFile, channelFileUrl)
     if (result.releaseName == null) {
--- a/packages/electron-updater/src/providers/GitHubProvider.ts
+++ b/packages/electron-updater/src/providers/GitHubProvider.ts
@@ -57,7 +57,7 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
     let tag: string | null
     try {
       if (this.updater.allowPrerelease) {
-        const currentChannel = this.updater?.channel || null;
+        const currentChannel = this.updater?.channel || this.updater.currentVersion.prerelease?.[0] ||  null;
         for (const element of feed.getElements("entry")) {
           // noinspection TypeScriptValidateJSTypes
           const hrefElement = hrefRegExp.exec(element.element("link").attribute("href"))!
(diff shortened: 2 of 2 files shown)


packages/electron-updater/src/providers/GitHubProvider.ts
```diff
@@ -77,7 +106,7 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
       throw newError(`No published versions on GitHub`, "ERR_UPDATER_NO_PUBLISHED_VERSIONS")
     }
 
-    const channelFile = getChannelFilename(this.getDefaultChannelName())
+    const channelFile = getChannelFilename(this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0]) || 'latest') : this.getDefaultChannelName())
```
mmaietta, line 109: `String(undefined) || 'latest'` will always return `undefined`
I think you meant to do:
```
String(semver.prerelease(tag)?.[0] || 'latest')
```

packages/electron-updater/src/providers/GitHubProvider.ts
```diff
@@ -54,11 +54,40 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
...
+          //If no channel is set by the current version, then grab latest (including prerelease)
+          if (currentChannel == null || currentChannel == 'alpha' || currentChannel == 'beta'){
+            // Skip any "custom" channels
+            if (hrefChannel != null && hrefChannel !== 'alpha' && hrefChannel !== 'beta') continue
+            // Skip alphas if in Beta Channel
+            if (currentChannel == 'beta' && hrefChannel == 'alpha') continue
+            // Get tag
+            tag = hrefChannel
+            break
+          }
```
mmaietta, line 80: Quick Q. Are we still allowing Beta channel updates when the user is in Alpha channel? I see this `// Skip alphas if in Beta Channel` and am curious if we need to enforce vice-versa as well. Also, just curious, why wouldn't we want to enable custom channels?

Let's use `===` instead of `==`
Also, let's flatten this a bit:
```
const shouldFetchVersion = !currentChannel || ['alpha', 'beta'].includes(currentChannel)
const isCustomChannel = !['alpha', 'beta'].includes(hrefChannel)
const channelMismatch = currentChannel === 'beta' && hrefChannel === 'alpha'

if (shouldFetchVersion && !isCustomChannel && !channelMismatch) {
   tag = hrefChannel
   break
}
```

packages/electron-updater/src/providers/GitHubProvider.ts
```diff
@@ -54,11 +54,40 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
...
       if (this.updater.allowPrerelease) {
-        // noinspection TypeScriptValidateJSTypes
-        tag = hrefRegExp.exec(latestRelease.element("link").attribute("href"))![1]
+        const currentChannel = this.updater?.channel || String(semver.prerelease(this.updater.currentVersion)?.[0]) ||  null;
+        for (const element of feed.getElements("entry")) {
+          // noinspection TypeScriptValidateJSTypes
+          const hrefElement = hrefRegExp.exec(element.element("link").attribute("href"))!
+
+          // If this is null then something is wrong and skip this release
+          if (hrefElement == null) continue
```
mmaietta, line 66: Please place `continue` on a new line.

.changeset/github-channel-support.md
```diff
@@ -0,0 +1,5 @@
+---
+"electron-updater": patch
```
mmaietta, line 2: This should be `minor` for a `feat` introduction :)

packages/electron-updater/src/providers/GitHubProvider.ts
KenCorma: Thanks, changed.

.changeset/github-channel-support.md
KenCorma: Thanks, Changed.

packages/electron-updater/src/providers/GitHubProvider.ts
KenCorma: No problem, changed.

packages/electron-updater/src/providers/GitHubProvider.ts
KenCorma: For the custom channels my rational was, if you are using custom channels then you wouldn't want someone to get on that track as they might not be able to auto update to production branch without reinstalling. 

I think that also ties into the alpha to beta detail you mentioned. As it stands you can only go up, alpha -> beta -> release.  I feel like most projects would want this as going from beta to alpha could have issues that would prevent reverting or going to production. But I DO see the the opposite and it could be beneficial. I thought about having a some options like 'Stay on channel only' , 'Allow step down ' etc etc but I didn't want to introduce options that weren't already supported by the generic provider. 

I am open to feedback and want to help make the GitHub provider better.

mmaietta: Makes sense.

Please slim down the current logic by using this code block below instead of the nested If's
```
const shouldFetchVersion = !currentChannel || ['alpha', 'beta'].includes(currentChannel)
const isCustomChannel = !['alpha', 'beta'].includes(hrefChannel)
const channelMismatch = currentChannel === 'beta' && hrefChannel === 'alpha'

if (shouldFetchVersion && !isCustomChannel && !channelMismatch) {
   tag = hrefChannel
   break
}
const isNextPreRelease = hrefChannel && hrefChannel === currentChannel
if (isNextPreRelease) {
   tag = hrefChannel
   break
}
```

packages/electron-updater/src/providers/GitHubProvider.ts
mmaietta: This still isn't valid:
`String(semver.prerelease(tag)?.[0]) || 'latest'`

It needs to be 
`String(semver.prerelease(tag)?.[0] || 'latest')`

mmaietta (approved):

mmaietta:
Whoops, seems there was an error in some code I suggested:
```
packages/electron-updater/src/providers/GitHubProvider.ts:76:63 - error TS2345: Argument of type 'string | number | null' is not assignable to parameter of type 'string'.
  Type 'null' is not assignable to type 'string'.

76           const isCustomChannel = !['alpha', 'beta'].includes(hrefChannel)
```
Since we don't care about number/null when comparing against alpha/beta, let's just wrap `hrefChannel` to be `String(hrefChannel)`

KenCorma:
@mmaietta No problem, added that guard.

packages/electron-updater/src/providers/GitHubProvider.ts
```diff
@@ -77,7 +106,7 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
       throw newError(`No published versions on GitHub`, "ERR_UPDATER_NO_PUBLISHED_VERSIONS")
     }
 
-    const channelFile = getChannelFilename(this.getDefaultChannelName())
+    const channelFile = getChannelFilename(this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0] || 'latest')) : this.getDefaultChannelName())
```
mmaietta, line 109: So, this change ends up breaking backward compatibility with previous prereleases that still use the `latest.yml` tag.
The unit test fails for a current repo: https://github.com/develar/__test_nsis_release/releases/tag/v1.5.2-beta.3
```
HttpError: 404 Not Found
    "method: GET url: https://github.com/develar/__test_nsis_release/releases/download/v1.5.2-beta.3/beta.yml
```
I'm trying to brainstorm ways we can properly handle this. One way is to have `latest` always be a fallback to any release, so `allowPrerelease` would allow updates out of the alpha/beta cells to prod. What are your thoughts?

mmaietta: I think something like this could work?
```
const fetchData = (channelName: string) => {
    const channelFile = getChannelFilename(channelName)
    const channelFileUrl = newUrlFromBase(this.getBaseDownloadPath(tag, channelFile), this.baseUrl)
    const requestOptions = this.createRequestOptions(channelFileUrl)
    try {
        return (await this.executor.request(requestOptions, cancellationToken))!
    } catch (e) {
        if (e instanceof HttpError && e.statusCode === 404) {
            throw newError(`Cannot find ${channelFile} in the latest release artifacts (${channelFileUrl}): ${e.stack || e.message}`, "ERR_UPDATER_CHANNEL_FILE_NOT_FOUND")
        }
        throw e
    }
}

try {
    const channel = this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0] || 'latest')) : this.getDefaultChannelName()
    rawData = fetchData(channel)
} catch (e) {
    if (this.updater.allowPrerelease) {
        // Allow fallback to `latest.yml`
        rawData = fetchData(this.getDefaultChannelName())
    } else {
        throw e
    }
}
```

mmaietta:
Looks like eslint is throwing an error:
> Error:   111:9    error    Type string trivially inferred from a string literal, remove type annotation          @typescript-eslint/no-inferrable-types

The rest of the warnings should be easily resolved through `pnpm prettier && pnpx lint --fix` and then I think we're good to merge 🙂

KenCorma:
My bad I didn't realize I had prettier off on this project.

packages/electron-updater/src/providers/GitHubProvider.ts
```diff
@@ -77,17 +104,33 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
       throw newError(`No published versions on GitHub`, "ERR_UPDATER_NO_PUBLISHED_VERSIONS")
     }
 
-    const channelFile = getChannelFilename(this.getDefaultChannelName())
-    const channelFileUrl = newUrlFromBase(this.getBaseDownloadPath(tag, channelFile), this.baseUrl)
-    const requestOptions = this.createRequestOptions(channelFileUrl)
     let rawData: string
+    let channelFile: string = ""
```
mmaietta, line 108: This is being flagged by eslint.
Please remove explicit type (since it can be assumed automatically)
`let channelFile = ""`

mmaietta: In fact, is it possible to just move these to `let` vars into the `fetchData` function?
```
const fetchData = async (channelName: string) => {
      const channelFile = getChannelFilename(channelName)
      const channelFileUrl = newUrlFromBase(this.getBaseDownloadPath(String(tag), channelFile), this.baseUrl)
```

Issue #6507, opened 2021-12-17
Incorrect asar integrity plist key


indutny-signal:
* **Electron-Builder Version**: 22.14.5

* **Node Version**: 16.9.1

* **Electron Version**: 16.0.4
* **Electron Type (current, beta, nightly)**: current

* **Target**: macOS

Electron uses different plist key (and format) for asar integrity:

```
<key>ElectronAsarIntegrity</key><dict><key>Resources/default_app.asar</key><dict><key>algorithm</key><string>SHA256</string><key>hash</key><string>857172e4d62de44b72638d5fa2f787dfd525c4bb666d729ed0e0c987af6b1f38</string></dict></dict>
```

compare to the one returned by electron-builder:

```
<key>AsarIntegrity</key><string>{&#34;checksums&#34;:{&#34;app.asar&#34;:&#34;bwX9ldnTOsgsdkaiaZqeLtNTkiMtfd9zFpvnOe113tCZWphBlQatLkRnqtFVQLSPVWoQ4wMFf218Iddg0EHFjQ==&#34;}}</string><
```

indutny-signal:
In addition to that the version of `@electron/universal` is fixed in `app-builder-lib` and so even that can't generate correct `ElectronAsarIntegrity` (even though `@electron/universal` would do it itself)

indutny-signal:
Also, electron supports only SHA256 integrity has while electron-builder provides SHA-512.

indutny-signal:
FWIW, I'm working on a PR for both #6506 and #6507

Issue #6506, opened 2021-12-17
asar file is missing integrity hashes for each entry


indutny-signal:
* **Electron-Builder Version**:  22.14.5

* **Node Version**:  16.9.1

* **Electron Version**: 16.0.4
* **Electron Type (current, beta, nightly)**: current

* **Target**:  macOS

There are no integrity hashes in asar file at the moment, see [electron/asar](https://github.com/electron/asar/commit/94cb8bdfea6257d7bc67f72ed80a790c2b5dae3a)

indutny-signal:
FWIW, I'm working on a PR for both #6506 and #6507

PR #6511, opened 2021-12-22
feat(mac): ElectronAsarIntegrity in electron@15

indutny-signal:
See: https://github.com/electron/electron/pull/30667
Fix: #6506
Fix: #6507

### DIFF
--- a/.changeset/nervous-pandas-film.md
+++ b/.changeset/nervous-pandas-film.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat(mac): ElectronAsarIntegrity in electron@15
--- a/packages/app-builder-lib/src/asar/asar.ts
+++ b/packages/app-builder-lib/src/asar/asar.ts
@@ -3,6 +3,20 @@ import { close, open, read, readFile, Stats } from "fs-extra"
 import * as path from "path"
 
 /** @internal */
+export interface ReadAsarHeader {
+  readonly header: string
+  readonly size: number
+}
+
+/** @internal */
+export interface NodeIntegrity {
+  algorithm: "SHA256"
+  hash: string
+  blockSize: number
+  blocks: Array<string>
+}
+
+/** @internal */
 export class Node {
   // we don't use Map because later it will be stringified
   files?: { [key: string]: Node }
@@ -16,6 +30,8 @@ export class Node {
   executable?: boolean
 
   link?: string
+
+  integrity?: NodeIntegrity
 }
 
 /** @internal */
@@ -66,13 +82,14 @@ export class AsarFilesystem {
     return result
   }
 
-  addFileNode(file: string, dirNode: Node, size: number, unpacked: boolean, stat: Stats): Node {
+  addFileNode(file: string, dirNode: Node, size: number, unpacked: boolean, stat: Stats, integrity?: NodeIntegrity): Node {
     if (size > 4294967295) {
       throw new Error(`${file}: file size cannot be larger than 4.2GB`)
     }
 
     const node = new Node()
     node.size = size
+    node.integrity = integrity
     if (unpacked) {
       node.unpacked = true
     } else {
@@ -114,7 +131,7 @@ export class AsarFilesystem {
   }
 }
 
-export async function readAsar(archive: string): Promise<AsarFilesystem> {
+export async function readAsarHeader(archive: string): Promise<ReadAsarHeader> {
   const fd = await open(archive, "r")
   let size: number
   let headerBuf
@@ -135,7 +152,11 @@ export async function readAsar(archive: string): Promise<AsarFilesystem> {
   }
 
   const headerPickle = createFromBuffer(headerBuf)
-  const header = headerPickle.createIterator().readString()
+  return { header: headerPickle.createIterator().readString(), size }
+}
+
+export async function readAsar(archive: string): Promise<AsarFilesystem> {
+  const { header, size } = await readAsarHeader(archive)
   return new AsarFilesystem(archive, JSON.parse(header), size)
 }
 
--- a/packages/app-builder-lib/src/asar/asar.ts
+++ b/packages/app-builder-lib/src/asar/asar.ts
@@ -89,7 +89,9 @@ export class AsarFilesystem {
 
     const node = new Node()
     node.size = size
-    node.integrity = integrity
+    if (integrity) {
+      node.integrity = integrity
+    }
     if (unpacked) {
       node.unpacked = true
     } else {
--- a/packages/app-builder-lib/src/asar/asarUtil.ts
+++ b/packages/app-builder-lib/src/asar/asarUtil.ts
@@ -8,6 +8,7 @@ import { Packager } from "../packager"
 import { PlatformPackager } from "../platformPackager"
 import { getDestinationPath, ResolvedFileSet } from "../util/appFileCopier"
 import { AsarFilesystem, Node } from "./asar"
+import { hashFile, hashFileContents } from "./integrity"
 import { detectUnpackedDirs } from "./unpackDetector"
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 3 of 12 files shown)


indutny-signal:
Sorry about the failing tests, I can't seem to be able to run them locally. Pushed an attempted fix.

.changeset/nervous-pandas-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"app-builder-lib": minor
```
indutny-signal, line 2: Not sure if I made a right decision here. The scheme change below is a breaking change...

packages/app-builder-lib/src/asar/asarUtil.ts
```diff
@@ -112,9 +113,10 @@ export class AsarPackager {
       }
 
       const dirNode = currentDirNode!
-      const newData = transformedFiles == null ? null : transformedFiles.get(i)
+      const newData = transformedFiles == null ? undefined : transformedFiles.get(i)
       const isUnpacked = dirNode.unpacked || (this.unpackPattern != null && this.unpackPattern(file, stat))
-      this.fs.addFileNode(file, dirNode, newData == null ? stat.size : Buffer.byteLength(newData), isUnpacked, stat)
+      const integrity = newData === undefined ? await hashFile(file) : hashFileContents(newData)
```
indutny-signal, line 118: This could potentially be optional since it is expensive to generate for the platforms that are not going to use it anyway.

packages/app-builder-lib/src/asar/asar.ts
```diff
@@ -66,13 +82,14 @@ export class AsarFilesystem {
...
 
-  addFileNode(file: string, dirNode: Node, size: number, unpacked: boolean, stat: Stats): Node {
+  addFileNode(file: string, dirNode: Node, size: number, unpacked: boolean, stat: Stats, integrity?: NodeIntegrity): Node {
     if (size > 4294967295) {
       throw new Error(`${file}: file size cannot be larger than 4.2GB`)
     }
 
     const node = new Node()
     node.size = size
+    node.integrity = integrity
```
mmaietta, line 92: Since `integrity?` is optional, should we first check it before setting it on `Node()`?
```
if (integrity) {
   node.integrity = integrity
}
```
Similar to how `node.unpacked` is beneath an if-statement (as opposed to being `node.unpacked = unpacked`)

packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -214,7 +214,7 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
   }
 
   if (asarIntegrity != null) {
-    appPlist.AsarIntegrity = JSON.stringify(asarIntegrity)
+    appPlist.ElectronAsarIntegrity = asarIntegrity
```
mmaietta, line 217: Will this break anything for electron pre-15?

packages/app-builder-lib/src/asar/asarUtil.ts
mmaietta: Can you elaborate on this? I personally am in favor of defaulting to more security, i.e. asar integrity, and this will only impact the build-time though. Is this integrity check supposed to be platform-specific?

indutny-signal: The integrity check so far only works on macOS. See [fuses](https://github.com/electron/fuses/commit/bd6ba567bd0ae7f51a37fb68988e2b903bd2e6f6#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R20)

packages/app-builder-lib/src/asar/asar.ts
indutny-signal: Ack.

packages/app-builder-lib/src/electron/electronMac.ts
indutny-signal: There is no mention of `AsarIntegrity` in electron `14-x-y`, `13-x-y`, `12-x-y`, and `11-x-y` so it is safe to say that it wasn't used at all.

indutny-signal:
Thanks for the feedback! I believe everything is addressed now.

mmaietta:
Can you please take a look at the failing unit tests?
```
 FAIL  src/globTest.ts (25.929 s)
  ✕ unpackDir one (2465 ms)
  ✕ unpackDir (2133 ms)
  ✕ asarUnpack and files ignore (2191 ms)
  ✓ link (2195 ms)
  ✕ outside link (2169 ms)
  ✓ failed peer dep (8396 ms)
  ✓ ignore node_modules (3003 ms)
  ✕ asarUnpack node_modules (2862 ms)
```

It looks like the snapshots just need to be updated to include the new integrity field, this should do the trick:
`TEST_FILES=globTest UPDATE_SNAPSHOT=true pnpm test`
Or perhaps mirror BuildTest where you use `delete value.integrity`? Either way should work.

Also, since this schema has a breaking change, I suggest we incorporate this into a v23.0.0-alpha branch that I'm currently working on. Can we retarget this PR to that branch?

indutny-signal:
@mmaietta sorry about that. I didn't get a chance to look into it yesterday, and thanks so much for the command to update the snapshots. Pushed the fix!

indutny-signal:
FWIW, I couldn't make `pnpm test` pass without failures on my macbook. Maybe the issue is that I have Windows 11 Parallels VM installed, but there are always failures that seem unrelated to my changes.

indutny-signal:
Looks like my fix didn't quite work and the better approach is to delete the `integrity` for these tests as well (as you said it).

indutny-signal:
Alright. Rebased over the alpha branch and hopefully fixed the tests. Thanks for your patience!

indutny-signal:
Awesome. Thanks for merging it!

mmaietta:
Published in `23.0.0-alpha.0`. Enjoy!

indutny-signal:
Awesome. Thanks!

packages/app-builder-lib/src/asar/asar.ts
```diff
@@ -2,6 +2,20 @@ import { createFromBuffer } from "chromium-pickle-js"
 import { close, open, read, readFile, Stats } from "fs-extra"
 import * as path from "path"
 
+/** @internal */
+export interface ReadAsarHeader {
```
indutny-signal, line 6: @mmaietta I don't think this, below, and `AsarFilesystem` should be internal. I get following TS errors when importing `electron-builder`:

```
node_modules/app-builder-lib/out/asar/asar.d.ts:1:66 - error TS2304: Cannot find name 'ReadAsarHeader'.

1 export declare function readAsarHeader(archive: string): Promise<ReadAsarHeader>;
                                                                   ~~~~~~~~~~~~~~

node_modules/app-builder-lib/out/asar/asar.d.ts:2:60 - error TS2304: Cannot find name 'AsarFilesystem'.

2 export declare function readAsar(archive: string): Promise<AsarFilesystem>;
                                                             ~~~~~~~~~~~~~~

node_modules/app-builder-lib/out/asar/integrity.d.ts:2:10 - error TS2305: Module '"./asar"' has no exported member 'NodeIntegrity'.

2 import { NodeIntegrity } from "./asar";
           ~~~~~~~~~~~~~

[10:10:15 AM] Found 3 errors. Watching for file changes.
```

mmaietta: I have no qualms with making them public. Can you open a PR for it?

indutny-signal: Sure thing. Done: https://github.com/electron-userland/electron-builder/pull/6692 Thank you!

indutny-signal: btw, I'm happy to report that we have tested 23.0.1 on our app and both asar integrity and merging ASARs appears to be working great! Thanks!

PR #6410, opened 2021-11-10
fix(builder-util): enable proxy handling in NodeHttpExecutor

baparham:
makes use of 'http_proxy' and 'https_proxy' environment variables
in order to allow for various tasks to be performed from behind a
corporate proxy. Requires third party dependencies because node's
standard built in http.Agent does not handle proxies natively.

closes #5906
fixes #6286

PR #6516, opened 2021-12-27
fix: specify protocol as https to complete proxy support fix

### LINKED ISSUE #6286: electron-publish fails behind proxy server
* **Electron-Builder Version**: 22.11.7

* **Node Version**: 14.17.3

* **Electron Version**: 14.0.1

* **Electron Updater Version**: 4.3.9

* **Target**: AppImage, NSIS, DMG/ZIP

I have forked the electron-updater-example and updated it to use current versions and include MacOS code-signing.  You can find the repo at [Electron Updater Example](https://github.com/robertpatrick/electron-updater-example/).

After publishing artifacts for MacOS and Linux (so far), the MacOS app seems to work fine from inside of our corporate network (which requires the use of a proxy server to reach the Internet).  

MacOS

The AppImage version of the app is failing while trying to check for updates inside our corporate network.

AppImage

What am I missing?  curl and firefox are able to find github.com without issue.  How do I get electron-updater to use our proxy server on Linux?

robertpatrick:
Explicitly set the protocol to https on the request objects to allow publishing to work from behind a proxy server when the https_proxy environment variable is set.

Closes #6286

### DIFF
--- a/.changeset/shiny-kangaroos-juggle.md
+++ b/.changeset/shiny-kangaroos-juggle.md
@@ -1,0 +1,5 @@
+---
+"electron-publish": patch
+---
+
+fix: Explicitly set the protocol to https on the request objects to allow publishing to work from behind a proxy server when the https_proxy environment variable is set.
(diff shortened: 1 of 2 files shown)


robertpatrick:
@mmaietta with this change and the previously merged change from PR #6410, I can confirm that electron-publisher is able to successfully publish releases to GitHub.  Let me know what you think.

Issue #6488, opened 2021-12-10
MSI building process fails when build.productName contains hyphen


kazuhitoyokoi:
While upgrading the version of the electron-builder for [our project](https://github.com/node-red-hitachi/node-red-desktop), I found that the MSI building process fails as follows.

```
c:\node-red-desktop>npm run build

> node-red-installer@2.1.4 build
> electron-builder

  • electron-builder  version=22.14.5 os=10.0.19042
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist\builder-effective-config.yaml
  • rebuilding native dependencies  dependencies=bcrypt@5.0.1 platform=win32 arch=x64
  • packaging       platform=win32 arch=x64 electron=16.0.4 appOutDir=dist\win-unpacked
  • building        target=MSI arch=x64 file=dist\Node-RED 2.1.4.msi
  ⨯ Exit code: 14. Command failed: C:\Users\<username>\AppData\Local\electron-builder\Cache\wix\wix-4.0.0.5512.2\candle.exe -arch x64 -dappDir=c:\node-red-desktop\dist\win-unpacked -pedantic -wx project.wxs
c:\node-red-desktop\dist\__msi-x64\project.wxs(23) : error CNDL0010 : The Icon/@Id attribute was not found; it is required.
c:\node-red-desktop\dist\__msi-x64\project.wxs(72) : error CNDL0014 : The Shortcut/@Icon attribute's value, 'Node-REDIcon.exe', is not a legal identifier. Identifiers may contain ASCII characters A-Z, a-z, digits, underscores (_), or periods (.). Every identifier must begin with either a letter or an underscore.
c:\node-red-desktop\dist\__msi-x64\project.wxs(73) : error CNDL0014 : The Shortcut/@Icon attribute's value, 'Node-REDIcon.exe', is not a legal identifier. Identifiers may contain ASCII characters A-Z, a-z, digits, underscores (_), or periods (.). Every identifier must begin with either a letter or an underscore.
```
The problem occurs from electron-builder v22.14.3 when `build.productName`  in package.json contains a hyphen, "-".
[...]

PR #6530, opened 2022-01-06
feat(msi): add fileAssociation support for MSI target

### LINKED ISSUE #3694: fileAssociations fails on msi target
* **Version**: 20.38.5

* **Target**: Windows

fileAssociations not working on msi targets. After installing the app, the files with extension .asdfgh are not recognized, works fine with nsis and dmg targets though.  Bellow my config:
```
	"build": {
		"icon": "build_resources/icon.icns",
		"directories": {
			"buildResources": "build_resources",
			"app": "myAppDir",
			"output": "dist"
		},
		"fileAssociations": [
			{
				"ext": "asdfgh",
				"name": "Alphabet File",
				"role": "Viewer"
			},
		],
		"mac": {
			"category": "public.app-category.finance",
			"target": [
				"dir",
				"dmg"
			],
			"type": "development",
			"asar": true
		},
		"win": {
			"target": [
				"dir",
				"msi",         
			],
			"icon": "build_resources/icon.ico",
			"asarUnpack": [
				"**node_modules/regedit/**"
			]
		},
}
```

aplum:
Also fix iconId sometimes containing invalid characters, and iconId config option being ignored (was added in #6247, but it doesn't seem to have been used).

Marking this as a draft and targeting [v23.0.0-alpha](https://github.com/electron-userland/electron-builder/tree/v23.0.0-alpha) because I think there's some more related changes that should be made:
- I'm not sure why the iconId config option was added, and I think it should be removed. I don't see why anyone would want to manually configure it when we should be able to generate it ourselves. Even if the product name contains all invalid-for-MSI-Id characters, we could still generate a valid unique Id somehow (base64, seeded RNG, whatever).
  - Even if we _do_ want to keep the option to manually configure it, it's now too specific – we also need a product-specific Id prefix for ProgId Ids with this PR, and possibly others in the future.
- I didn't support FileAssociation.icon yet, but I'd be willing to if someone wants to lend a hand. The WiX side should be easy (add more [Icon](https://github.com/aplum/electron-builder/blob/4090ab62f16f05c5aa2bf7b2491415efd4594158/packages/app-builder-lib/templates/msi/template.xml#L23) elements as needed, reference their Ids [here](https://github.com/aplum/electron-builder/blob/4090ab62f16f05c5aa2bf7b2491415efd4594158/packages/app-builder-lib/src/targets/MsiTarget.ts#L266)), but I can't justify the time to figure out getting/converting the icon file since I don't need this feature at my job.

Also, I didn't use the FileAssociation.name config. I'm not sure what the expected values look like here, but the given example ("PNG") doesn't match up with the recommended format where it's used in [NsisTarget](https://github.com/electron-userland/electron-builder/blob/82ce942069dd54094951f6db890d335366048d4f/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L685) – in the [docs](https://nsis.sourceforge.io/FileAssoc) the example value is "myapp.textfile", which matches the format used in WiX [ProgId Id examples](https://www.firegiant.com/wix/tutorial/getting-started/beyond-files/). So in NSIS it looks like we should generate that value ourselves, as done in MSI ProgId here. Furthermore, on [Mac targets](https://github.com/electron-userland/electron-builder/blob/96d7813362d319c3141d78e179c3f6d71c798fab/packages/app-builder-lib/src/electron/electronMac.ts#L203) `name` is used for CFBundleTypeName, which I'm not familiar with but [this example](https://developer.apple.com/library/archive/qa/qa1959/_index.html) uses "Adobe PDF" and iMovie.app on my Mac uses values like "Color Preset"… so I'm thinking maybe the `name` and `description` options should be merged?

Closes #3694, #6488

### DIFF
--- a/.changeset/giant-dryers-beg.md
+++ b/.changeset/giant-dryers-beg.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat(msi): add fileAssociation support for MSI target
--- a/.changeset/serious-peas-help.md
+++ b/.changeset/serious-peas-help.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": major
+---
+
+BREAKING CHANGE: remove MSI option `iconId`
--- a/packages/app-builder-lib/src/options/FileAssociation.ts
+++ b/packages/app-builder-lib/src/options/FileAssociation.ts
@@ -1,9 +1,9 @@
 /**
  * File associations.
  *
- * macOS (corresponds to [CFBundleDocumentTypes](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-101685)) and NSIS only.
+ * macOS (corresponds to [CFBundleDocumentTypes](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-101685)), NSIS, and MSI only.
  *
- * On Windows works only if [nsis.perMachine](https://electron.build/configuration/configuration#NsisOptions-perMachine) is set to `true`.
+ * On Windows (NSIS) works only if [nsis.perMachine](https://electron.build/configuration/configuration#NsisOptions-perMachine) is set to `true`.
  */
 export interface FileAssociation {
   /**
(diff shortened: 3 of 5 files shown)


mmaietta:
Since we're targeting the v23 alpha branch, perhaps we can remove the iconId field and generate it ourselves as you mentioned? That way we can avoid this:
>it's now too specific – we also need a product-specific Id prefix for ProgId Ids with this PR, and possibly others in the future.

Not sure regarding the other items, the subject kind of flew over my head. Never worked with File Associations in the past

aplum:
I've removed the iconId config option and added a better fallback value than the "ElectronApp" I was using.

Supporting FileAssociation.icon could be done in a minor release in the future, and my rambling paragraph about FileAssociation.name would be more appropriate as a separate PR anyway (if it holds any merit) (FYI don't expect a PR about it from me), so I'm marking this as ready for review now.

mmaietta:
Can you please merge in latest master? That should enable latest test configuration to be run

aplum:
Done, I think?

mmaietta:
Not sure what happened, but the merge triggered much more "file changes" than what was originally in the PR.

Could you try rebasing your feature on top of `v23.0.0-alpha` branch? I just want to clean up the PR to only have your changes.

aplum:
That seems to have fixed it.

mmaietta (approved):

mmaietta:
Published in `23.0.0-alpha.0`. Enjoy!

PR #6547, opened 2022-01-12
fix(NSIS): prevent partial overwrites

indutny-signal:
`Nsis7z::Extract` ignores the errors when copying the files and thus can
leave us with the app that has old asar and bindings, but new assets. In
addition to that, the app will firmly believe that it is still running
an old version and would attempt to repeatedly auto-update until fixed,
leading to excessive bandwidth use and very unhappy customers.

This change extracts the contents of 7z archive into a separate
directory before attempting to copy them with `CopyFiles` function that
(unlike `Nsis7z::Extract`) does detect and report failures. To make our
lives easier the `CopyFiles` will also erase all files on a failure so
after retrying a few times we will ultimately have to fallback to old 7z
extraction directly into output folder.

### Testing

The easiest way to trigger the failure is to:

1. Go to the application folder
2. Copy exe file to `tmp.exe`
3. Start `tmp.exe` (thus locking the asar, bindings, and dlls)
4. Try to install over with freshly built installer

### DIFF
--- a/packages/app-builder-lib/templates/nsis/include/extractAppPackage.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/extractAppPackage.nsh
@@ -90,5 +90,45 @@
 !macroend
 
 !macro extractUsing7za FILE
-    Nsis7z::Extract "${FILE}"
+  Push $OUTDIR
+  CreateDirectory "$PLUGINSDIR\7z-out"
+  ClearErrors
+  SetOutPath "$PLUGINSDIR\7z-out"
+  Nsis7z::Extract "${FILE}"
+  Pop $R0
+  SetOutPath $R0
+
+  # Retry counter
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/templates/nsis/include/extractAppPackage.nsh
```diff
@@ -90,5 +90,45 @@
 !macroend
 
 !macro extractUsing7za FILE
-    Nsis7z::Extract "${FILE}"
+  Push $OUTDIR
+  CreateDirectory "$PLUGINSDIR\7z-out"
```
indutny-signal, line 94: Note: NSIS promises that `$PLUGINSDIR` is a temporary folder that is removed after the installation.

indutny-signal:
Yay. Thanks for merging it so quickly!

Issue #1839, opened 2017-07-19
get request timeout error


chenyuncai:
* **18.5.1**: 

* **win32 x64**: 

When I use the electron-builder to build a exe file in unbuntu 16.04, I get this err below:

electron-builder 18.5.1
Building NSIS installer
  Packaging NSIS installer for arch x64
Error: Request timed out
    at TLSSocket.socket.setTimeout (/home/work/code/p2p/node_modules/._electron-builder-http@18.5.1@electron-builder-http/src/httpExecutor.ts:184:18)
    at Object.onceWrapper (events.js:293:19)
    at emitNone (events.js:86:13)
    at TLSSocket.emit (events.js:188:7)
    at TLSSocket.Socket._onTimeout (net.js:351:8)
    at ontimeout (timers.js:386:14)
    at tryOnTimeout (timers.js:250:5)
    at Timer.listOnTimeout (timers.js:214:5)
From previous event:
    at Promise.longStackTracesCaptureStackTrace [as _captureStackTrace] (/home/work/code/p2p/node_modules/._bluebird@3.5.0@bluebird/js/release/debuggability.js:397:19)
    at new PromiseSpawn (/home/work/code/p2p/node_modules/._bluebird@3.5.0@bluebird/js/release/generators.js:43:17)
    at /home/work/code/p2p/node_modules/._bluebird@3.5.0@bluebird/js/release/generators.js:197:21
    at build (/home/work/code/p2p/node_modules/._electron-builder@18.5.1@electron-builder/out/builder.js:70:21)
    at Object.<anonymous> (/home/work/code/p2p/node_modules/._electron-builder@18.5.1@electron-builder/out/cli/build-cli.js:71:41)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:423:7)
    at startup (bootstrap_node.js:147:9)
    at bootstrap_node.js:538:3

what can i do to fix this?

Thanks

develar:
1. Please use 19.16.0
2. Do you have proxy?

develar:
@chenyuncai Please answer to a question about proxy. Are you in China?

chenyuncai:
Yes, china. has no proxy

chenyuncai:
[...]

PR #6550, opened 2022-01-14
feat(msi): support assisted installer for MSI target

aplum:
Add basic support for assisted installer, with UI to choose between per-user and per-machine. Supported config options: runAfterFinish, perMachine, oneClick. Not supported: license (EULA), allowToChangeInstallationDirectory, etc.

The UI dialogs are WelcomeDlg -> InstallScopeDlg -> VerifyReadyDlg. For `perMachine=true`, InstallScopeDlg doesn't allow changing the scope. Starting with the useless WelcomeDlg instead of going straight to InstallScopeDlg allows us to reuse an existing dialog set (WixUI_InstallDir) to reduce the custom code here, and makes it easier to add/remove dialogs if more features are supported in the future (EULA, allowToChangeInstallationDirectory).

I've tried to match the behaviour of NSIS for the supported config options based on what I came across in docs/PRs/issues.

This also prevents oneClick's runAfterFinish from executing when installed silently, because that's probably never desired (see [Rocket.Chat.Electron #1839](https://github.com/RocketChat/Rocket.Chat.Electron/issues/1839)).

I've tested upgrading from the existing oneClick MSI to this PR's oneClick and assisted MSIs (for both per-user and per-machine), and it worked in all cases, including when the installation directory is different (due to [this line](https://github.com/aplum/electron-builder/blob/089c4e5ffcc9636d62e98307c23481d7ac79060a/packages/app-builder-lib/src/targets/MsiTarget.ts#L175) changing to match NSIS). GPO installs also work.


mmaietta:
Re:
>Supported config options: runAfterFinish, perMachine, oneClick. Not supported: license (EULA), allowToChangeInstallationDirectory, etc.

It seems like we need to add a new top-level entry `msi` to provide these supported config options (similar to the `nsis` top-level key @ https://www.electron.build/configuration/nsis) as they're not available under the default `win` entry. What are your thoughts?

>I've tried to match the behaviour of NSIS for the supported config options based on what I came across in docs/PRs/issues.

Also, thanks for pursuing this approach

aplum:
I think that top-level key already exists from [this line](https://github.com/electron-userland/electron-builder/blob/bea51d6a8bb828d9b34734908f13b667aa55b0e9/packages/app-builder-lib/src/configuration.ts#L74). I've been using it in my projects and testing this PR. Those three options already exist under that key – runAfterFinish and perMachine already worked, and oneClick existed even though the resulting MSI didn't function quite correctly. Unless there's somewhere else it needs to go too?

davej:
@aplum is it possible that these changes broke the ability to use a custom `APPLICATIONFOLDER` param? Recent versions of Electron Builder don't seem to work with `APPLICATIONFOLDER` anymore. See issue here: https://github.com/electron-userland/electron-builder/issues/7193

aplum:
@davej Seems possible. I'd have to investigate, but in the meantime let's discuss in the issue you linked.

PR #6551, opened 2022-01-15
fix(nsis): use atomic rmdir on update

indutny-signal:
Previously uninstaller would run `RMDir /r $INSTDIR` to remove the
currently installed version, but this operation is not atomic and if
some of the files will fail to delete it will leave them in the
directory while erasing the rest. If we are updating, however, this
leads us to a tricky situation where we cannot update these files, but
cannot also cancel the installation. Because of the erased files the app
won't be able to start if the installation won't completely at least
partially. However, the downside of that is that the app can have new
asar.unpacked files along with the old asar, executable, and bindings.

The approach of this change is to recursive use `Rename` instead of a
single `RMDir /r` in uninstaller (when `isUpdated` is true) to move the
whole app directory file by file to a temporary folder. If this
operation fails due to busy files - we use `CopyFiles` to restore all
files that we managed to move so far. Because the whole uninstallation
process becomes interrupted - the app shortcut and file associations
have to be removed only *after* the successful recursive `Rename`.

This error is caught by installer running in an update mode (see
`installUtil.nsh`) and presented to user in a dialog. If this erro
happen the installation does not proceed normally.

In addition to all of the above, this patch simplifies the last resort
measure in `extractAppPackage` which should now only run when old
uninstaller (that still uses `RMDir /r`) leaves busy files behind.

## Testing

Tested by doing normal updates and keeping tmp.exe open as in https://github.com/electron-userland/electron-builder/pull/6547

https://user-images.githubusercontent.com/79877362/149609389-65e1ecc2-82eb-4153-a80b-89a95cbbbacd.mp4

### DIFF
--- a/packages/app-builder-lib/templates/nsis/include/installUtil.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/installUtil.nsh
@@ -159,7 +159,7 @@ Function uninstallOldVersion
     !endif
     ${if} $uninstallString == ""
       ClearErrors
-      Goto Done
+      Return
     ${endif}
   ${endif}
 
@@ -178,7 +178,7 @@ Function uninstallOldVersion
   ${if} $installationDir == ""
   ${andIf} $uninstallerFileName == ""
     ClearErrors
-    Goto Done
+    Return
   ${endif}
 
   ${if} $installMode == "CurrentUser"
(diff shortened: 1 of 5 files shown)


packages/app-builder-lib/templates/nsis/include/installUtil.nsh
```diff
@@ -126,7 +126,10 @@ Function handleUninstallResult
   Return
 
   ${if} $R0 != 0
+    MessageBox MB_OK|MB_ICONEXCLAMATION "$(uninstallFailed): $R0"
     DetailPrint `Uninstall was not successful. Uninstaller error code: $R0.`
+    SetErrorLevel 2
+    Quit
```
indutny-signal, line 132: I think ideally we should retry running uninstaller (with user's confirmation) in such case instead of failing whole installation immediately. Let me know if you want me to add this to the current patch.

indutny-signal: Oh, and btw. `Abort` doesn't quite work here for some reason (the installer just hangs), so I had to resort to `Quit`.

packages/app-builder-lib/templates/nsis/uninstaller.nsh
```diff
@@ -28,6 +28,109 @@ Function un.onInit
   !endif
 FunctionEnd
 
+Function un.atomicRMDir
```
indutny-signal, line 31: I'm not entirely proud of this, but it implements `RMDir /r` that preserves backups of all files and directories that it removed in `$PLUGINSDIR\old-install`.

packages/app-builder-lib/templates/nsis/uninstaller.nsh
```diff
@@ -38,6 +141,34 @@ Section "un.install"
 
   !insertmacro setLinkVars
 
+  # delete the installed files
```
indutny-signal, line 144: The order of operations changed as stated in the commit message. We try to keep as much state as possible in case we will need to restore the files on failure. The users would be really frustrated to see their shortcuts and file associations go if `uninstall.exe --updated` fails while restoring all app's files.

indutny-signal: I just wanted ahead and did it. Makes for a much better user experience!

packages/app-builder-lib/templates/nsis/include/installUtil.nsh
mmaietta: @indutny-signal This sounds like a great idea. Is it complex to add?
>we should retry running uninstaller (with user's confirmation) in such case instead of failing whole installation immediately

indutny: I actually already pushed it :wink:

indutny:
Awesome! Thanks for merging it!

mmaietta:
Thanks for the contribution @indutny! It'll make it in our upcoming v23 alpha release :)

PR #6563, opened 2022-01-18
fix: socket hang up error 422 issues in github publish

baparham:
I split this up into two commits, one with the actual fix, then one refactor so you could actually understand what the if statement was looking for before calling `this.overwriteArtifact`

### DIFF
--- a/packages/electron-publish/src/gitHubPublisher.ts
+++ b/packages/electron-publish/src/gitHubPublisher.ts
@@ -198,7 +198,12 @@ export class GitHubPublisher extends HttpPublisher {
         requestProcessor
       )
       .catch(e => {
-        if (e.statusCode === 422 && e.description != null && e.description.errors != null && e.description.errors[0].code === "already_exists") {
+        if (
+          e.statusCode === 422 &&
+          e.description &&
+          ((e.description.includes("errors") && e.description.includes("already_exists")) ||
+            (e.description.errors && e.description.errors.length >= 1 && e.description.errors[0].code === "already_exists"))
+        ) {
           return this.overwriteArtifact(fileName, release).then(() => this.doUploadFile(attemptNumber, parsedUrl, fileName, dataLength, requestProcessor, release))
         }
 
--- a/packages/electron-publish/src/gitHubPublisher.ts
+++ b/packages/electron-publish/src/gitHubPublisher.ts
@@ -198,12 +198,7 @@ export class GitHubPublisher extends HttpPublisher {
         requestProcessor
       )
       .catch(e => {
-        if (
-          e.statusCode === 422 &&
-          e.description &&
-          ((e.description.includes("errors") && e.description.includes("already_exists")) ||
-            (e.description.errors && e.description.errors.length >= 1 && e.description.errors[0].code === "already_exists"))
-        ) {
+        if (this.doesErrorMeanAlreadyExists(e)) {
           return this.overwriteArtifact(fileName, release).then(() => this.doUploadFile(attemptNumber, parsedUrl, fileName, dataLength, requestProcessor, release))
         }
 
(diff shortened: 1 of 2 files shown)


baparham:
this should in theory close #4940

packages/electron-publish/src/gitHubPublisher.ts
```diff
@@ -215,6 +215,19 @@ export class GitHubPublisher extends HttpPublisher {
       })
   }
 
+  private doesErrorMeanAlreadyExists(e: any) {
+    let doesErrorMeanAlreadyExists = true
+    doesErrorMeanAlreadyExists &&= e.statusCode === 422
```
baparham, line 220: this ends up transpiling to `doesErrorMeanAlreadyExists && (doesErrorMeanAlreadyExists = e.statusCode === 422);` which seems wrong to me, perhaps I need not use `&&=` which now that I read the [moz spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND_assignment) is not meant for assignment ... for whatever strange reason.

baparham: pushed an update

packages/electron-publish/src/gitHubPublisher.ts
```diff
@@ -215,6 +215,19 @@ export class GitHubPublisher extends HttpPublisher {
       })
   }
 
+  private doesErrorMeanAlreadyExists(e: any) {
+    let doesErrorMeanAlreadyExists = true
+    doesErrorMeanAlreadyExists = doesErrorMeanAlreadyExists && e.statusCode === 422
+    if (e.description !== null) {
```
baparham, line 221: this actually fails if description is undefined, so I'll push another fixup that resolves that situation as well

baparham: fixed

packages/electron-publish/src/gitHubPublisher.ts
```diff
@@ -215,6 +215,20 @@ export class GitHubPublisher extends HttpPublisher {
...
   }
 
+  private doesErrorMeanAlreadyExists(e: any) {
+    let doesErrorMeanAlreadyExists = true
+    doesErrorMeanAlreadyExists = doesErrorMeanAlreadyExists && e.statusCode === 422
+    if (e.description) {
+      const desc = e.description
+      const descIncludesAlreadyExists =
+        (desc.includes("errors") && desc.includes("already_exists")) || (desc.errors && desc.errors.length >= 1 && desc.errors[0].code === "already_exists")
+      doesErrorMeanAlreadyExists = doesErrorMeanAlreadyExists && descIncludesAlreadyExists
```
mmaietta, line 225: We can just simplify the logic flow here to be less nested (instead of refactoring after)
```
private doesErrorMeanAlreadyExists(e: any) {
  if (!e.description) {
  	return false
	}
  const desc = e.description
  const descIncludesAlreadyExists = (desc.includes("errors") && desc.includes("already_exists")) || (desc.errors && desc.errors.length > 0 && desc.errors[0].code === "already_exists")
	return e.statusCode === 422 && descIncludesAlreadyExists
}
```

packages/electron-publish/src/gitHubPublisher.ts
```diff
@@ -215,6 +215,20 @@ export class GitHubPublisher extends HttpPublisher {
       })
   }
 
+  private doesErrorMeanAlreadyExists(e: any) {
+    let doesErrorMeanAlreadyExists = true
+    doesErrorMeanAlreadyExists = doesErrorMeanAlreadyExists && e.statusCode === 422
+    if (e.description) {
+      const desc = e.description
+      const descIncludesAlreadyExists =
+        (desc.includes("errors") && desc.includes("already_exists")) || (desc.errors && desc.errors.length >= 1 && desc.errors[0].code === "already_exists")
```
mmaietta, line 224: I'm not sure I understand this:
`desc.includes("errors") && desc.includes("already_exists")`
Are we expecting a string or array for `desc`?

baparham: the old code seemed to expect an object, (e.g. `description = { ..., errors: [..., ..., ]}`) but the response back from github is a string with some prose, and some JSON style notation that we really can't parse properly, so I support both ways, not really sure what to expect.

packages/electron-publish/src/gitHubPublisher.ts
baparham: yeah, that might make more sense that way, a bit easier to follow at least

baparham: I've fixed it up to be more like your suggestion

mmaietta (approved):

mmaietta:
Great work! Thank you for the contribution :)

PR #6575, opened 2022-01-24
fix: Allow disabling of webinstaller files to avoid confusion with actual installers

devinbinnie:
This PR allows the disabling of web installer files for `electron-updater` when it is only expected that a full installer should be downloaded.

This prevents a potential flaw in which an attacker using a webinstaller could serve a package not authorized by the developer to the end-user with malicious code inside.

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -52,6 +52,17 @@ export abstract class AppUpdater extends EventEmitter {
   allowDowngrade = false
 
   /**
+   * Web installer files might not have signature verification, this switch prevents to load them unless it is needed.
+   *
+   * Currently false to prevent breaking the current API, but it should be changed to default true at some point that
+   * breaking changes are allowed.
+   *
+   * @default false
+   */
+
+  disableWebInstaller = false
+
+  /**
    * The current application version.
    */
   readonly currentVersion: SemVer
@@ -443,6 +454,7 @@ export abstract class AppUpdater extends EventEmitter {
         updateInfoAndProvider,
         requestHeaders: this.computeRequestHeaders(updateInfoAndProvider.provider),
         cancellationToken,
+        disableWebInstaller: this.disableWebInstaller,
       }).catch(e => {
         throw errorHandler(e)
       })
@@ -644,6 +656,7 @@ export interface DownloadUpdateOptions {
   readonly updateInfoAndProvider: UpdateInfoAndProvider
   readonly requestHeaders: OutgoingHttpHeaders
   readonly cancellationToken: CancellationToken
+  readonly disableWebInstaller?: boolean
 }
 
 function hasPrereleaseComponents(version: SemVer) {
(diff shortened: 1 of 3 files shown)


mmaietta:
Please also add a `.changeset` file for this PR via `pnpm changeset`. Seems like it should be a `major` bump if we do default `disableWebInstaller` to true (completely fine since we're in a v23-alpha so we can include breaking changes)

packages/electron-updater/src/NsisUpdater.ts
```diff
@@ -46,6 +49,9 @@ export class NsisUpdater extends BaseUpdater {
         }
 
         if (isWebInstaller) {
+          if (downloadUpdateOptions.disableWebInstaller) {
+            throw new Error("WebInstaller files are not allowed")
```
mmaietta, line 53: Please use:
```
throw newError(
   `Unable to download new version ${downloadUpdateOptions.updateInfoAndProvider.info.version}. Web Installers are disabled`,
   "ERR_UPDATER_WEB_INSTALLER_DISABLED"
)
```

packages/electron-updater/src/NsisUpdater.ts
```diff
@@ -31,7 +31,10 @@ export class NsisUpdater extends BaseUpdater {
       task: async (destinationFile, downloadOptions, packageFile, removeTempDirIfAny) => {
         const packageInfo = fileInfo.packageInfo
         const isWebInstaller = packageInfo != null && packageFile != null
-        if (isWebInstaller || (await this.differentialDownloadInstaller(fileInfo, downloadUpdateOptions, destinationFile, provider))) {
+        if (
+          (isWebInstaller && !downloadUpdateOptions.disableWebInstaller) ||
+          (await this.differentialDownloadInstaller(fileInfo, downloadUpdateOptions, destinationFile, provider))
+        ) {
```
mmaietta, line 37: Thoughts on just checking if the web installer is supposed to be disabled first and throwing the error at that point? Then we don't need to throw an error any further down (lines 52-54)
```
if (isWebInstaller && downloadUpdateOptions.disableWebInstaller) {
  throw newError(
   `Unable to download new version ${downloadUpdateOptions.updateInfoAndProvider.info.version}. Web Installers are disabled`,
   "ERR_UPDATER_WEB_INSTALLER_DISABLED"
  )  
}
if (!isWebInstaller && !downloadUpdateOptions.disableWebInstaller) {
  this._logger.warn('disableWebInstaller is set to false, you should set it to true if you do not plan on using a web installer.')
}
if (isWebInstaller || (await this.differentialDownloadInstaller(fileInfo, downloadUpdateOptions, destinationFile, provider))) {
  ....
}
```

packages/electron-updater/src/AppUpdater.ts
```diff
@@ -51,6 +51,17 @@ export abstract class AppUpdater extends EventEmitter {
...
+  /**
+   * Web installer files might not have signature verification, this switch prevents to load them unless it is needed.
+   *
+   * Currently false to prevent breaking the current API, but it should be changed to default true at some point that
+   * breaking changes are allowed.
+   *
+   * @default false
+   */
+
+  disableWebInstaller = false
```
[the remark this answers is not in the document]
mmaietta: Sounds good, we can leave it as false. I think we should log a warning then for the user that they should set the flag true if they're not using web installers, along with a note that it will default true in a future version.

packages/electron-updater/src/NsisUpdater.ts
devinbinnie: Works for me!

mmaietta (approved):

mmaietta (approved):

PR #6576, opened 2022-01-24
fix: Update certificate validation on Windows to check full DN

devinbinnie:
This PR changes the certificate validation for NSIS installer to allow for validation of the entire DN instead of just the Common Name portion. This should allow for a more secure certificate validation.

### DIFF
--- a/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
+++ b/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
@@ -56,7 +56,7 @@ export function verifySignature(publisherNames: Array<string>, unescapedTempUpda
           if (data.Status === 0) {
             const subject = parseDn(data.SignerCertificate.Subject)
             let match = false
-            publisherNames.forEach(name => {
+            for (const name of publisherNames) {
               const dn = parseDn(name)
               if (dn.size) {
                 // if we have a full DN, compare all values
@@ -65,12 +65,13 @@ export function verifySignature(publisherNames: Array<string>, unescapedTempUpda
                   return dn.get(key) === subject.get(key)
                 })
               } else if (name === subject.get("CN")!) {
+                logger.warn(`Signature validated using only CN ${name}. Please add your full Distinguished Name (DN) to publisherNames configuration`)
                 match = true
               }
-            })
-            if (match) {
-              resolve(null)
-              return
+              if (match) {
+                resolve(null)
+                return
+              }
             }
           }
 
--- a/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
+++ b/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
@@ -54,8 +54,21 @@ export function verifySignature(publisherNames: Array<string>, unescapedTempUpda
 
           const data = parseOut(Buffer.from(stdout, "base64").toString("utf-8"))
           if (data.Status === 0) {
-            const name = parseDn(data.SignerCertificate.Subject).get("CN")!
-            if (publisherNames.includes(name)) {
+            const subject = parseDn(data.SignerCertificate.Subject)
+            let match = false
+            publisherNames.forEach(name => {
+              const dn = parseDn(name)
+              if (dn.size) {
+                // if we have a full DN, compare all values
+                const allKeys = Array.from(dn.keys())
+                match = allKeys.every(key => {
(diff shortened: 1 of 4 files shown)


devinbinnie:
@mmaietta Added as a draft for now because I've added a unit test for the new case, but the tests don't seem to pass on my machine (Windows 10 running Git Bash). I have my own certificate as well, and some of the snapshots don't match. Is there a test runner in CI I can use? Or is there some way to set up my dev environment such that I can get the tests to pass?

mmaietta:
> @mmaietta Added as a draft for now because I've added a unit test for the new case, but the tests don't seem to pass on my machine (Windows 10 running Git Bash). I have my own certificate as well, and some of the snapshots don't match. Is there a test runner in CI I can use? Or is there some way to set up my dev environment such that I can get the tests to pass?

All CI is now handled through GitHub Actions. The test snapshots are OS-specific and for some, such as publish/updating, they need to have an API key or token that are part of this repo's env secrets. All in all, it's really difficult to get a singular test environment. `pnpm test-linux` can cover the linux-based snapshots by running via docker. Other than that, the mapping of test files/snapshots to platform are:
Linux
https://github.com/electron-userland/electron-builder/blob/59b6becfff65e76e27b578e89a109ea65aa3b782/.github/workflows/test.yaml#L21-L22
MacOS
https://github.com/electron-userland/electron-builder/blob/59b6becfff65e76e27b578e89a109ea65aa3b782/.github/workflows/test.yaml#L67
MacOS (updater-specific due to env tokens)
https://github.com/electron-userland/electron-builder/blob/59b6becfff65e76e27b578e89a109ea65aa3b782/.github/workflows/test.yaml#L86
Windows
https://github.com/electron-userland/electron-builder/blob/59b6becfff65e76e27b578e89a109ea65aa3b782/.github/workflows/test.yaml#L97-L98
Depending on what OS I'm working on, I have to update snapshots by booting up my corresponding OS. For non-token-based tests, I think you can activate github actions on a fork of this repo?

mmaietta:
All tests pass 🙌

packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
```diff
@@ -54,8 +54,21 @@ export function verifySignature(publisherNames: Array<string>, unescapedTempUpda
...
+            publisherNames.forEach(name => {
+              const dn = parseDn(name)
+              if (dn.size) {
+                // if we have a full DN, compare all values
+                const allKeys = Array.from(dn.keys())
+                match = allKeys.every(key => {
+                  return dn.get(key) === subject.get(key)
+                })
+              } else if (name === subject.get("CN")!) {
+                match = true
```
mmaietta, line 68: Can we log a warning here that the developer should provide their full Distinguished Name? Something like this. (Feel free to adjust the message as you see fit)
```
logger.warn(`Signature validated using only CN ${name}. Please add your full Distinguished Name (DN) to publisherNames configuration`)
```

packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
```diff
@@ -54,8 +54,21 @@ export function verifySignature(publisherNames: Array<string>, unescapedTempUpda
...
-            const name = parseDn(data.SignerCertificate.Subject).get("CN")!
-            if (publisherNames.includes(name)) {
+            const subject = parseDn(data.SignerCertificate.Subject)
+            let match = false
+            publisherNames.forEach(name => {
+              const dn = parseDn(name)
+              if (dn.size) {
+                // if we have a full DN, compare all values
+                const allKeys = Array.from(dn.keys())
+                match = allKeys.every(key => {
```
mmaietta, line 64: I might be overlooking something, but I'm not sure if this will work when multiple `publisherNames` are provided since we don't return early after `match = true`. If we have two fully defined publisher DNs (`p1` and `p2`), and `p1` passes/matches but then `p2` fails, `match` would be false after this `forEach`
I think we need to do a traditional for-loop so that we can `break` out of it once `match = true`
What are your thoughts?

devinbinnie: Nope, you're right, that will definitely cause an issue. I think it definitely makes sense to just break the loop here once we're done to make sure the match is set correctly. Nice catch!

mmaietta (approved):

PR #6578, opened 2022-01-25
feat: use `mergeASARs` API by @electron/universal

indutny-signal:
@electron/universal@1.2.0 introduced `mergeASARs` and `singleArchFiles`
options that enable merging of x64 and arm64 ASARs into one. This change
adds appropriate properties to the MacConfiguration interface and passes
them along to @electron/universal.

## Testing

Tested by building our app with linked packages.

### DIFF
--- a/packages/app-builder-lib/src/options/macOptions.ts
+++ b/packages/app-builder-lib/src/options/macOptions.ts
@@ -175,6 +175,23 @@ export interface MacConfiguration extends PlatformSpecificBuildOptions {
    * Specify the URL of the timestamp authority server
    */
   readonly timestamp?: string | null
+
+  /**
+   * Whether to merge ASAR files for different architectures or not.
+   *
+   * This option has no effect unless building for "universal" arch.
+   * @default false
+   */
+  readonly mergeASARs?: boolean
+
+  /**
+   * Minimatch pattern of paths that are allowed to be present in one of the
+   * ASAR files, but not in the other.
+   *
+   * This option has no effect unless building for "universal" arch and applies
+   * only if `mergeASARs` is `true`.
+   */
+  readonly singleArchFiles?: string
 }
 
 export interface DmgOptions extends TargetSpecificOptions {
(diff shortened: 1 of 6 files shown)


indutny-signal:
Should I target this to alpha branch?

packages/app-builder-lib/src/options/macOptions.ts
```diff
@@ -175,6 +175,23 @@ export interface MacConfiguration extends PlatformSpecificBuildOptions {
...
    */
   readonly timestamp?: string | null
+
+  /**
+   * Whether to merge ASAR files for different architectures or not.
+   *
+   * This option has no effect unless building for "universal" arch.
+   * @default false
+   */
+  readonly mergeASARs?: boolean
```
mmaietta, line 185: What are your thoughts on just having this default true since it'll only be applicable when building for universal?

mmaietta:
> Should I target this to alpha branch?

`master` target branch works perfectly fine. It's in an 'alpha' version notation for the CI/CD.

packages/app-builder-lib/src/options/macOptions.ts
indutny-signal: I think this is alright, but there are two downsides:

- It will require a major version bump
- The build will fail for some users and they will need to populate `singleArchFiles` with something like `"singleArchFiles": "node_modules/{@signalapp/signal-client/prebuilds/**,ringrtc/build/**,sharp/**}"`

If we are fine with both - I can flip it for sure!

mmaietta: Sounds good. I'm okay with both since we're currently in a v23-alpha state for our next major version bump. It also forces individuals building universal binaries to leverage the latest features of @electron/universal, which seems like a practice we'd always want to follow too

indutny-signal: I went ahead and flipped it :joy:

PR #6589, opened 2022-01-28
fix: More small security fixes

devinbinnie:
2 more small fixes:
- Removed backtick escaping for Windows code signing as it is unnecessary for Powershell and can cause the script to attempt to access the wrong file
- Updated the proxy filename to be more secure (512-bit string)

Let me know if I need to split these up or if I can leave it as one commit.

### DIFF
--- a/.changeset/neat-socks-dress.md
+++ b/.changeset/neat-socks-dress.md
@@ -1,0 +1,6 @@
+---
+"electron-updater": patch
+---
+
+- Removed backtick escaping for Windows code signing as it is unnecessary for Powershell and can cause the script to attempt to access the wrong file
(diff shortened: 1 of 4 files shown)


mmaietta (approved):

mmaietta:
It's okay if it's within one PR. Please update the changeset file to be more specific though as that's what generates the changelogs. It can just be your ticket description
>- Removed backtick escaping for Windows code signing as it is unnecessary for Powershell and can cause the script to attempt to access the wrong file
- Updated the proxy filename to be more secure (512-bit string)

devinbinnie:
Thanks @mmaietta. Should be good to merge now. 👍

Issue #6000, opened 2021-06-28
Breaking change in electron-updater 4.3.10


davej:
* **Version**: 4.3.10

* **Electron Version**: Effects Electron 11 and before. Does not effect Electron 12 or later.

* **Target**: All

Electron-updater has introduced `fs/promises` which will break the build on Electron 11 and earlier. This is because `fs/promises` is not supported in Node 12  (which is used by Electron 11). The change was introduced in this PR: https://github.com/electron-userland/electron-builder/pull/5890

This is a critical issue for anybody on Electron 11 or earlier because it will break their application after an update and it is not possible to recover by pushing a fix in an update.

The best mitigation that I think of here is to publish a `4.3.11` that does not use `fs/promises`. If you want to use `fs/promises` then you can publish `5.0.0` that uses `fs/promises`.

![electron-updater](https://user-images.githubusercontent.com/26971/123619180-d6c89000-d800-11eb-964b-2fab0aa8eeb3.png)

shownoso:
I got the same problem

tub5:
This has halted all development on our apps

theolivenbaum:
Same here

sFireFrog:
。。。I solved this problem by specifying the version in package.json

tub5:
> 。。。I solved this problem by specifying the version in package.json

That would be a temporary fix.
However, we're using the [Electron.NET](https://github.com/ElectronNET/Electron.NET) package which uses a CLI to build our apps. Therefore we can't just change it.
With a breaking change such as this, it shouldn't have been placed as a minor version.

sFireFrog:
> > 。。。I solved this problem by specifying the version in package.json
> 
> That would be a temporary fix.
> However, we're using the [Electron.NET](https://github.com/ElectronNET/Electron.NET) package which uses a CLI to build our apps. Therefore we can't just change it.
> With a breaking change such as this, it shouldn't have been placed as a minor version.

Yes, I also use Electron.NET, this problem has troubled me for a long time, I directly modified the CLI

tub5:
> > > 。。。I solved this problem by specifying the version in package.json
> > 
> > 
> > That would be a temporary fix.
> > However, we're using the [Electron.NET](https://github.com/ElectronNET/Electron.NET) package which uses a CLI to build our apps. Therefore we can't just change it.
> > With a breaking change such as this, it shouldn't have been placed as a minor version.
> 
[...]

PR #6594, opened 2022-01-30
fix(updater): Support Electron 11 and below (Node < 14)

mmaietta:
Replacing fs/promises with fs-extra to support legacy versions of Electron that use node 12 and below.
Fixes: #6000

Regenerating docs addition from PR #6585 (docs needed to be added to source code, which autogenerates the .md file)


kaatt:
`fs.promises` works in older node/electron versions, `fs/promises` doesn't.

kaatt:
fs-extra is a good solution

Issue #6322, opened 2021-10-08
electron-builder not generating app-update.yml for Linux


mohsintahir:
* **Electron-Builder Version**:  22.9.1

* **Node Version**: 12.12.69

* **Electron Version**: 10.1.4
* **Electron Type (current, beta, nightly)**: current
 4.2.5

 Linux default target
* **Target**: Linux default target

here is my package.json file 
```
{
  "name": "",
  "version": "",
  "description": "",
  "main": "built/main.js",
  "scripts": {
    "build": "tsc && (cp -r src/assets built/assets || xcopy \"src/assets\" \"built/assets\" /y /i /s)",
    "watch": "tsc -w",
    "lint": "tslint -c tslint.json -p tsconfig.json",
    "start": "yarn run build && electron .",
    "pack": "electron-builder --dir",
    "dist": "electron-builder build",
    "postinstall": "electron-builder install-app-deps",
    "build-mac": "yarn build && electron-builder build --mac",
    "build-windows": "yarn build && electron-builder build --windows nsis --x64 --ia32 --arm64",
    "build-linux": "yarn build && electron-builder build --linux rpm deb AppImage",
    "publish-mac": "yarn build && electron-builder build --mac --publish always",
    "publish-linux": "yarn build && electron-builder build --linux rpm deb AppImage --publish always",
    "publish-windows": "yarn build && electron-builder build --windows nsis --x64 --ia32 --arm64 --publish always",
    "publish-win-mac": "yarn build && electron-builder build --mac --windows nsis --x64 --ia32 --arm64 --publish always"
  },
  "repository": "",
  "keywords": [],
  "author": {
    "name": "",
    "email": "",
    "url": ""
  },
  "license": "",
  "build": {
    "appId": "",
    "productName": "",
    "asar": true,
    "asarUnpack": [
      "**/dict/*"
    ],
    "artifactName": "app-name-${version}-${arch}.${ext}",
    "afterSign": "build/notarize.js",
    "mac": {
      "category": "public.app-category.utilities",
      "icon": "build/icon.png",
      "type": "distribution",
      "hardenedRuntime": true,
      "gatekeeperAssess": false,
      "entitlements": "build/entitlements.mac.plist",
      "entitlementsInherit": "build/entitlements.mac.plist"
    },
    "dmg": {
      "sign": false
    },
    "win": {
      "icon": "build/icon.png"
    },
    "linux": {
      "icon": "build",
      "category": "Utility"
    },
    "publish": [
[...]

PR #6615, opened 2022-02-05
Add app-update.yml to avoid errors

siauderman:
Suggested fix for https://github.com/electron-userland/electron-builder/issues/6322 (not tested yet).

### DIFF
--- a/packages/app-builder-lib/src/targets/snap.ts
+++ b/packages/app-builder-lib/src/targets/snap.ts
@@ -211,6 +211,11 @@ export default class SnapTarget extends Target {
     if (snap.compression != null) {
       args.push("--compression", snap.compression)
     }
+    
+    const publishConfig = getAppUpdatePublishConfiguration(packager, arch, false /* in any case validation will be done on publish */)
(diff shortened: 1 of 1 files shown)


packages/app-builder-lib/src/targets/snap.ts
```diff
@@ -211,6 +211,11 @@ export default class SnapTarget extends Target {
     if (snap.compression != null) {
       args.push("--compression", snap.compression)
     }
+    
+    const publishConfig = getAppUpdatePublishConfiguration(packager, arch, false /* in any case validation will be done on publish */)
+    if (publishConfig != null) {
+      await outputFile(path.join(packager.getResourcesDir(stageDir.dir), "app-update.yml"), serializeToYaml(publishConfig))
+    }
```
mmaietta, line 218: I don't understand how this change, being applied to only snap, resovles the issue for deb/rpm builds that also do not support auto-updates.

mmaietta:
Closing, should be resolved through https://github.com/electron-userland/electron-builder/pull/6616

PR #6625, opened 2022-02-09
fix(packager): wait for event before starting an upload

DanielMcAssey:
When the configuration has a artifactBuildCompleted file which needs to be waited, the upload starts anyway as it listens to the artifactCreated event. This resolves that by first waiting for any hooks, then dispatching the event.


mmaietta:

Issue #5236, opened 2020-08-19
Files in folder`swiftshader` are not signed


cholic0304:
* **Version**: 21.2.0

* **Electron Version**: 5.0.11
* **Electron Type (current, beta, nightly)**: current

* **Target**: windows

I am trying to build and sign an electron app on windows, evyrything works fine but two files are not signed.
The folder is `AppInstalledRootPath/swiftshader`, and the files are `libEGL.dll` and `libGLEv2.dll`.

├───locales
├───resources
│   └───app.asar.unpacked
│       ├───node_modules
│       └───resources
└───swiftshader

Here is the code I found in electron-builder. Only `appOutDir` and `app.asar.unpacked` will be signed, `swiftshader` is not included. So is there any way to make these two files to be signed automaticlly?

```
    await BluebirdPromise.map(readdir(packContext.appOutDir), (file: string): any => {
      if (file === exeFileName) {
        return this.signAndEditResources(path.join(packContext.appOutDir, exeFileName), packContext.arch, packContext.outDir, path.basename(exeFileName, ".exe"), this.platformSpecificBuildOptions.requestedExecutionLevel)
      }
      else if (file.endsWith(".exe") || (this.isSignDlls() && file.endsWith(".dll"))) {
        return this.sign(path.join(packContext.appOutDir, file))
      }
      return null
    })

    if (!isAsar) {
      return
    }

    const outResourcesDir = path.join(packContext.appOutDir, "resources", "app.asar.unpacked")
    // noinspection JSUnusedLocalSymbols
    const fileToSign = await walk(outResourcesDir, (file, stat) => stat.isDirectory() || file.endsWith(".exe") || file.endsWith(".dll"))
    await BluebirdPromise.map(fileToSign, file => this.sign(file), {concurrency: 4})
```

SuperKirik:
Hello, 
any update on this ?

PR #6682, opened 2022-03-01
fix(signing): Include swiftshader in signing directories for windows

karliatto:
This PR adds `swiftshader` to the directories that contain `.dll` file that are required to be sign for windows in order  to satisfy AppLocker policy.

Closes https://github.com/electron-userland/electron-builder/issues/5236

### DIFF
--- a/packages/app-builder-lib/src/winPackager.ts
+++ b/packages/app-builder-lib/src/winPackager.ts
@@ -392,9 +392,18 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
       return
     }
 
-    const outResourcesDir = path.join(packContext.appOutDir, "resources", "app.asar.unpacked")
-    // noinspection JSUnusedLocalSymbols
-    const fileToSign = await walk(outResourcesDir, (file, stat) => stat.isDirectory() || file.endsWith(".exe") || (this.isSignDlls() && file.endsWith(".dll")))
-    await BluebirdPromise.map(fileToSign, file => this.sign(file), { concurrency: 4 })
+    const directoriesToBeSigned = [["resources", "app.asar.unpacked"], ["swiftshader"]]
+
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/winPackager.ts
```diff
@@ -392,9 +392,18 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
...
+      return path.join.apply(this, [packContext.appOutDir, ...directoryPath])
+    })
+
+    const filesToSignPromise = outDirectories.map(async outDir => {
+      return await walk(outDir, (file, stat) => stat.isDirectory() || file.endsWith(".exe") || (this.isSignDlls() && file.endsWith(".dll")))
+    })
+
+    const filesToSign = await Promise.all(filesToSignPromise)
+    const filesToSignFlat = filesToSign.flat(1)
+    await BluebirdPromise.map(filesToSignFlat, file => this.sign(file), { concurrency: 4 })
```
mmaietta, line 407: Can we make this entire section more simple?
```
const signPromise = (filepath: string[]) => {
  const outDir = path.join(packContext.appOurDir, ...filepath)
  return walk(outDir, (file, stat) => stat.isDirectory() || file.endsWith(".exe") || (this.isSignDlls() && file.endsWith(".dll")))
}
const filesToSign = await Promise.all([
    signPromise(["resources", "app.asar.unpacked"]),
    signPromise(["swiftshader"])
])
await BluebirdPromise.map(filesToSign.flat(1), file => this.sign(file), { concurrency: 4 })
```

karliatto: Thanks for checking it @mmaietta, I updated it with a fixup commit [4611e1b](https://github.com/electron-userland/electron-builder/pull/6682/commits/4611e1be484da49124eeed232363eba72c3c5250) let me know if you think it is good and I can rebase it so there is only one commit for that.

PR #6692, opened 2022-03-03
fix(app-builder-lib): export missing TS types

indutny-signal:
See [discussion here](https://github.com/electron-userland/electron-builder/pull/6511#discussion_r818931948).

cc @mmaietta


indutny-signal:
Thank you!

PR #6711, opened 2022-03-09
docs: fix newlines for pkg

ntnbrtnkv:
Now it looks like 
image


mmaietta:
IIRC, that `pkg.md` file is autogenerated. The formatting is created using this script: https://github.com/electron-userland/electron-builder/blob/master/scripts/jsdoc2md2html.js

Issue #6747, opened 2022-03-28
nsis.deleteAppDataOnUninstall property not work if package.json's name property has scope


lowfront:
* **Electron-Builder Version**: 22.14.5, 22.14.13

* **Node Version**:  16.13.2

* **Electron Version**: 16.0.7, 17.2.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: NSIS
* **OS**: Windows 10

Hi. I noticed that the `AppData/Roaming/` data folder is not deleted when `nsis.deleteAppDataOnUninstall` is `true`. When I tested it, I found that the folder was not deleted if the package.json's name has scope. Regardless of whether have a scope, I think the property `nsis.deleteAppDataOnUninstall` should work.
This is a repository [link](https://github.com/lowfront/electron-builder-remove-test) that can reproduce the problem.

lowfront:
I found the cause of the problem. This [line](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/nsis/uninstaller.nsh#L225) specifies the directory to be deleted as `$APPDATA\${APP_PACKAGE_NAME}` but for package name with scope, `APP_PACKAGE_NAME` is like `@lowfront/electron-builder-remove-test`.
At this time, the entire path is `$APPDATA\@lowfront/electron-builder-remove-test`, and the path contains a slash(/) and is incompatible with the Windows OS path.

So I modified this [line](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L192) as follows and it worked.
```ts
APP_PACKAGE_NAME: appInfo.name.replace(/\//g, '\\'), // delete path to $APPDATA\@lowfront\electron-builder-remove-test
```
I think can fix the `nsisTarget.ts` file, or `uninstaller.nsh` file to fix the problem. Regardless of which method choose, NSIS is an installation program for Windows, so there will probably be no problem with other OS builds.
I wonder which method looks better.

mmaietta:
Closing per #6772

PR #6762, opened 2022-04-02
fix(app-builder-lib): change slash to backslash in buildInstaller'APP_PACKAGE_NAME

lowfront:
Added a function util to replace slash with backslash, and modified slash to backslash so that APP_PACKAGE_NAME is the same folder address in Windows NSIS target.
This task will solve the issue #6747.


mmaietta:
Can you please rebase this PR on latest master? It seems that the CI tests weren't triggered for some reason, hoping that a rebase will do so.

PR #6771, opened 2022-04-07
fix(nsis): specify full path to system's find

indutny-signal:
GIT-SCM installation can override window's native `find` command line
tool with the one from unix tools. This in turn changes the exit code of
the whole find process command and makes us detect the app as running
even though it is not.

Credit and original fix by @theAddict

See https://github.com/signalapp/Signal-Desktop/issues/5856


mmaietta:
Thanks for your contribution!

Issue #6786, opened 2022-04-12
Is any other reason for checking oneClick options for deciding isAdminRightsRequired for update-info.json?


HwangTaehyun:
* **Electron-Builder Version**: 
  * 23.0.3

* **Node Version**: 
  * 16.3.1

* **Electron Version**:
  * 14.2.4
* **Electron Type (current, beta, nightly)**:
   * current

* **Target**: 
  * window 10

  * Is any other reason for checking oneClick options for deciding isAdminRightsRequired for update-info.json? When I update electron app on window service, it required isAdminRigthsRequired. However, my electron app needs some mui page, so i setup custom nsis installer script and set oneClick to false. In this case, because of the below logic, my electron installer cannot get isAdminRightsRequired.
Thank you for the amazing library!
  * https://github.com/electron-userland/electron-builder/blob/fca50aea42fc054b87bbe943bbe5c1c235a186e1/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#:~:text=if%20(updateInfo%20!%3D%20null,%7D

mmaietta:
@develar would you be willing to provide context here? I don't know the history

HwangTaehyun:
Resolved by #6787

PR #6787, opened 2022-04-12
fix(nsis): If window service needs to run installer for update, the installer must have admin previlege. #6786

HwangTaehyun:
If the electron app needs a custom NSIS installer script (aka. oneClick option is false), the electron installer cannot get isAdminRightsRequired and failed to update when the electron app is running on window service because it requires installer that being admin rights.
It seems that any other reason does not exist when the electron builder NSIS oneClick option is false, update-info's isAdminRightsRequired is false.
Thank you!

### DIFF
--- a/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
+++ b/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
@@ -159,6 +159,7 @@ export class NsisTarget extends Target {
         .join(", "),
     }
     const isPerMachine = options.perMachine === true
+
     if (!this.isPortable) {
       logFields.oneClick = oneClick
       logFields.perMachine = isPerMachine
@@ -319,7 +320,7 @@ export class NsisTarget extends Target {
       updateInfo = await createBlockmap(installerPath, this, packager, safeArtifactName)
     }
 
-    if (updateInfo != null && isPerMachine && oneClick) {
+    if (updateInfo != null && isPerMachine && (oneClick || options.packElevateHelper)) {
       updateInfo.isAdminRightsRequired = true
     }
 
--- a/.changeset/poor-windows-heal.md
+++ b/.changeset/poor-windows-heal.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+If window service needs to run installer for update, the installer must have admin previlege. Electron-updater detects whether elevating or not using isAdminRightsRequired in update-info.json. And this isAdminRightsRequired true option should be added to latest.yml using nsis's packElevateHelper option
--- a/test/snapshots/windows/assistedInstallerTest.js.snap
+++ b/test/snapshots/windows/assistedInstallerTest.js.snap
@@ -128,3 +128,29 @@ Object {
   ],
 }
 `;
+
+exports[`assisted, only perMachine and elevated 1`] = `
+Object {
+  "win": Array [
+    Object {
+      "arch": "x64",
+      "file": "Test App ßW Setup 1.1.0.exe",
+      "safeArtifactName": "TestApp-Setup-1.1.0.exe",
+      "updateInfo": Object {
+        "isAdminRightsRequired": true,
+        "sha512": "@sha512",
+        "size": "@size",
+      },
+    },
+    Object {
+      "file": "Test App ßW Setup 1.1.0.exe.blockmap",
+      "safeArtifactName": "TestApp-Setup-1.1.0.exe.blockmap",
+      "updateInfo": Object {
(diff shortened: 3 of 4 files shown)


mmaietta:
Instead of removing `&& oneClick`, is it possible to detect for this edge case?
>when the electron app is running on window service

This change breaks a snapshot test and I'm unsure if all assisted installers should have `"isAdminRightsRequired": true,`
```
FAIL  src/windows/assistedInstallerTest.ts (149.916 s)
  √ assisted (45469 ms)
  √ allowElevation false, app requestedExecutionLevel admin (6587 ms)
  × assisted, only perMachine (39978 ms)

  ● assisted, only perMachine

    expect(received).toMatchSnapshot()

    Snapshot name: `assisted, only perMachine 1`

    - Snapshot  - 0
    + Received  + 2

    @@ -3,18 +3,20 @@
          Object {
            "arch": "x64",
            "file": "Test App ßW Setup 1.1.0.exe",
            "safeArtifactName": "TestApp-Setup-1.1.0.exe",
            "updateInfo": Object {
    +         "isAdminRightsRequired": true,
              "sha512": "@sha512",
              "size": "@size",
            },
          },
          Object {
            "file": "Test App ßW Setup 1.1.0.exe.blockmap",
            "safeArtifactName": "TestApp-Setup-1.1.0.exe.blockmap",
            "updateInfo": Object {
    +         "isAdminRightsRequired": true,
              "sha512": "@sha512",
              "size": "@size",
            },
          },
        ],
```

HwangTaehyun:
@mmaietta Thank you for review. I'll try to detect for this edge case and change my code!

HwangTaehyun:
I detected an edge case for using the allowServiceToInstall boolean option of nsis in electron builder config

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
```diff
@@ -319,7 +322,7 @@ export class NsisTarget extends Target {
       updateInfo = await createBlockmap(installerPath, this, packager, safeArtifactName)
     }
 
-    if (updateInfo != null && isPerMachine && oneClick) {
+    if ((updateInfo != null && isPerMachine && oneClick) || allowServiceToInstall) {
```
mmaietta, line 325: Are you already using `packElevateHelper = true` by chance? Would this work instead?
```
(updateInfo != null && isPerMachine && (oneClick || options.packElevateHelper))
```

HwangTaehyun: Yes, elevate.exe file is alreay included in my electron app becauseo of perMachine true option. 

electron updater decides to elevate or not using isAdminRightsRequired in update info file.
(https://github.com/electron-userland/electron-builder/blob/5e381c556d12ce185bb7ea720380509c1ddc5cf7/packages/electron-updater/src/NsisUpdater.ts#:~:text=if%20(options.isAdminRightsRequired,%7D) So, it's seamless using packElevateHelper option beacuse if elevate.exe packed, then electron updater must using this to elevate when update.

mmaietta:
Just to play it safe, can you please also add this test to the `assistedInstallerTest.ts`?
```
test.ifNotCiMac(
  "assisted, only perMachine and elevated",
  app({
    targets: nsisTarget,
    config: {
      nsis: {
        oneClick: false,
        perMachine: true,
        packElevateHelper: true
      },
    },
  })
)
```

You can update the snapshot via:
```
TEST_FILES="assistedInstallerTest" UPDATE_SNAPSHOT=true pnpm test 
```

After that, I think we're good to merge!

HwangTaehyun:
I added the test code to this commit!
Finally, I will test the update in the Window service and tell you again.

HwangTaehyun:
I checked it. 

below nsis option's generate the following latest.yml.
```
nsis:
  oneClick: false
  perMachine: true
  packElevateHelper: true
  ...
  include: './scripts/nsis/installer.nsh'
```
```
...
files:
  - url: xxx-Setup.exe
    sha512: NZvMh7iNTT9/Uz5WsEVR2YS6+KFI7dAHerTJH743tJkeS/jNJr3sUp6OT8f4y3MyANQ6FgUsb2qRd99HjEAhug==
    size: 152614338
    isAdminRightsRequired: true // generated by nsis's packElevateHelper true option despite oneClick false option
path: xxx-Setup.exe
sha512: NZvMh7iNTT9/Uz5WsEVR2YS6+KFI7dAHerTJH743tJkeS/jNJr3sUp6OT8f4y3MyANQ6FgUsb2qRd99HjEAhug==
releaseDate: '2022-04-19T02:54:40.490Z'
```

Issue #6767, opened 2022-04-06
Exit code: ENOENT. spawn /usr/bin/python ENOENT


Sovgut:
* **Electron-Builder Version**: 23.0.3

* **Node Version**: 17.8.0

* **Electron Version**: 18.0.2
* **Electron Type (current, beta, nightly)**: current

* **Target**: mac
* **Mac Version**: 12.3.1

```
Done in 3.64s.
  • electron-builder  version=23.0.3 os=21.4.0
  • loaded configuration  file=package.json ("build" field)
  • loaded parent configuration  preset=react-cra
  • writing effective config  file=dist/builder-effective-config.yaml
  • packaging       platform=darwin arch=arm64 electron=18.0.2 appOutDir=dist/mac-arm64
  • default Electron icon is used  reason=application icon is not set
  • skipped macOS application code signing  reason=cannot find valid "Developer ID Application" identity or custom non-Apple code signing certificate, it could cause some undefined behaviour, e.g. macOS localized description not visible, see https://electron.build/code-signing allIdentities=     0 identities found
                                                Valid identities only
     0 valid identities found
  • building        target=DMG arch=arm64 file=dist/Sekeer-0.1.0-arm64.dmg
  • Detected arm64 process, HFS+ is unavailable. Creating dmg with APFS - supports Mac OSX 10.12+
  ⨯ Exit code: ENOENT. spawn /usr/bin/python ENOENT  failedTask=build stackTrace=Error: Exit code: ENOENT. spawn /usr/bin/python ENOENT
    at /Users/sovgut/Repositories/sekeer/node_modules/builder-util/src/util.ts:133:18
    at exithandler (node:child_process:406:5)
    at ChildProcess.errorhandler (node:child_process:418:5)
    at ChildProcess.emit (node:events:527:28)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:289:12)
    at onErrorNT (node:internal/child_process:476:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
The terminal process "/bin/zsh '-c', 'npm run electron:package:mac'" terminated with exit code: 1.
```

mmaietta:
Do you have python installed?

bytemain:
i have the same problem, because macos monterey 12.3 remove python 2.7 https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes

codehzy:
[...]

PR #6789, opened 2022-04-12
fix(dmg-builder): exec python fail on Macos12.3

bytemain:
close: #6767

### DIFF
--- a/.changeset/rare-dolls-sin.md
+++ b/.changeset/rare-dolls-sin.md
@@ -1,0 +1,5 @@
+---
+"dmg-builder": patch
+---
+
+Fix cannot exec python in MacOS 12.3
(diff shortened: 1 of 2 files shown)


mmaietta:
This change is causing all mac dmg tests to fail as it's trying to use python @ `/usr/local/bin/python`
```
FAIL  src/mac/dmgTest.ts (585.119 s)
  ✕ dmg (13943 ms)

  ● dmg

    Exit code: ENOENT. spawn /usr/local/bin/python
     ENOENT

      131 |           }
      132 |
    > 133 |           reject(new Error(message))
          |                  ^
      134 |         }
      135 |       }
      136 |     )

      at ../packages/builder-util/src/util.ts:133:18
```

Unfortunately, github actions doesn't have a MacOS 12 build machine yet otherwise I'd be running our tests on that.

bytemain:
it seem that `which python` got a result: `/usr/local/bin/python`.

but why it does not exists

bytemain:
it seem that I have fixed it.

the return value of `exec` func has a `'\n'` suffix.

mmaietta:

Seems the Docs automation is frozen, but the CI tests all passed. Merging now

PR #6791, opened 2022-04-13
feat: Use tar instead of 7zip to preserve file permissions in tar.gz packages

devinbinnie:
`7zip` on UNIX-based systems does not support conserving permissions when creating tarballs (see Backup and limitations: https://linux.die.net/man/1/7za), so tarballs created by `7zip` were always world-writable (permissions on files were set to `777` for all files added to the archive).

This is generally considered bad practice since it can lead to privilege escalation. This PR replaces the use of `7zip` with `tar` when creating the initial tarball, which preserves the permissions from the unpackaged folder. It also allows the user to change the permissions (in `afterPack`, if they want to) before creating the archive.

I couldn't find an issue created for this, so I'm not sure if this approach is correct, but I'd like to offer it as a solution to the issue described above. Open to feedback/changes as always :)

### DIFF
--- a/.changeset/spotty-impalas-listen.md
+++ b/.changeset/spotty-impalas-listen.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+Use tar instead of 7zip to preserve permissions in tar.gz builds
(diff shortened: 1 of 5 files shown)


devinbinnie:
> I love the idea. I'd target this as a feature though. Could you please update the changeset?

Done!

PR #6793, opened 2022-04-14
fix(nsis): cleanup temporary 7z folder

indutny-signal:
Before attempting to extract files with non-discriminating
`Nsis7z::Extract` that ignores the errors remove the temporary folder so
that we don't require twice as much disk space for this last resort
measure.

Additionally, replace `Quit` with a `Goto DoneExtract7za` so that the
last resort measure actually works.

## Testing

image

### DIFF
--- a/packages/app-builder-lib/templates/nsis/include/extractAppPackage.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/extractAppPackage.nsh
@@ -113,13 +113,21 @@
       # Try copying a few times before asking for a user action.
       Goto RetryExtract7za
     ${else}
-      MessageBox MB_RETRYCANCEL|MB_ICONEXCLAMATION "$(appCannotBeClosed)" /SD IDCANCEL IDRETRY RetryExtract7za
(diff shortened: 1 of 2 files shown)


indutny-signal:
Sorry for ready->draft->ready transition. I've realized that I had an issue which I uncovered in testing (ABORT wasn't handled properly). Now the PR is ready for a review.

packages/app-builder-lib/templates/nsis/include/extractAppPackage.nsh
```diff
@@ -113,13 +113,21 @@
       # Try copying a few times before asking for a user action.
       Goto RetryExtract7za
     ${else}
-      MessageBox MB_RETRYCANCEL|MB_ICONEXCLAMATION "$(appCannotBeClosed)" /SD IDCANCEL IDRETRY RetryExtract7za
+      MessageBox MB_ABORTRETRYIGNORE|MB_ICONEXCLAMATION "$(appCannotBeClosed)" /SD IDABORT IDABORT AbortExtract7za IDRETRY RetryExtract7za
```
indutny-signal, line 116: Do you the silent default (`/SD`) should be ABORT? I think it is the safest option, but just wanted to make sure it doesn't go unnoticed.

indutny-signal:
Thanks for merging this!

PR #6841, opened 2022-05-05
fix: Merge arrays from same config key.

MrMYHuang:
With these changes, electron-builder can merge extraResources from parent config and current conifg. For example:
common.json:
```
{
  "extends": null,
  "extraResources": [
    "common/*.files",
    "common.file"
  ]
}
```
specific.json:
```
{
  "extends": "common.json",
  "extraResources": [
    {
      "from": "specific",
      "to": ".",
      "filter": ["file.txt"]
    },
  ]
}
```
The merged config - dist/builder-effective-config.yaml:
```
directories:
  output: dist
  buildResources: build
extends: common.json
extraResources:
  - filter:
      - common/*.files
      - common.file
  - from: specific
    filter:
      - file.txt
files: []
electronVersion: 16.2.3
```

### DIFF
--- a/.changeset/good-cooks-sniff.md
+++ b/.changeset/good-cooks-sniff.md
@@ -1,0 +1,5 @@
+---
+"builder-util": patch
+---
+
+fix: Merge arrays from same config key in cascading electron-builder configs, such as `files`
--- a/packages/builder-util/src/deepAssign.ts
+++ b/packages/builder-util/src/deepAssign.ts
@@ -16,7 +16,12 @@ function assignKey(target: any, from: any, key: string) {
 
   const prevValue = target[key]
   if (prevValue == null || value == null || !isObject(prevValue) || !isObject(value)) {
-    target[key] = value
+    // Merge arrays.
+    if (Array.isArray(prevValue) && Array.isArray(value)) {
+      target[key] = prevValue.concat(value)
+    } else {
+      target[key] = value
+    }
   } else {
     target[key] = assign(prevValue, value)
   }


mmaietta:
@MrMYHuang this PR seems to have caused a bug in the latest version, can you please take a look?
A config of:
```
extraFiles:
  - from: node_modules/node-mac-permissions/build/Release
    to: MacOS
    filter:
      - '*.node'
  - from: node_modules/electron-deeplink/build/Release
    to: MacOS
    filter:
      - '*.node'
```
The effective electron-builder config shows it duplicated,
```
extraFiles:
  - from: node_modules/node-mac-permissions/build/Release
    to: MacOS
    filter:
      - '*.node'
  - from: node_modules/electron-deeplink/build/Release
    to: MacOS
    filter:
      - '*.node'
  - from: node_modules/node-mac-permissions/build/Release
    to: MacOS
    filter:
      - '*.node'
  - from: node_modules/electron-deeplink/build/Release
    to: MacOS
    filter:
      - '*.node'
```
This doesn't seem to cause an issue with `files`, but it does cause it to try and double-copy when in `extraFiles` which then throws an error on my linux build machine (can't repro on my mac locally, didn't check windows.)

MrMYHuang:
> The effective electron-builder config shows it duplicated,

I can't reproduce it on Linux and macOS with electron-builder 23.0.9. I tested with extraFiles in both parent config and current config, and each extraFiles has two elements with different "from" values.

Please revert this PR as your decision.

semireg:
I wonder if this could be related to https://github.com/electron-userland/electron-builder/issues/7021.

PR #6825, opened 2022-04-28
refactor: Add better typing support on AppUpdater as an event emitter

Nokel81:
- Helps with not having to look up the specific types for the events
  that are emitted

Signed-off-by: Sebastian Malton <sebastian@malton.name>

PR #6858, opened 2022-05-11
refactor: Replace typed-emitter with manual implementation of typed event emitter for AppUpdater

Nokel81:

This PR fixes issues related to the building of docs from #6825

### DIFF
--- a/.changeset/flat-toys-guess.md
+++ b/.changeset/flat-toys-guess.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+Remove typed-emitter as a dev-dep for electron-updater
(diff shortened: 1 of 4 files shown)


mmaietta:
Nice approach!

I was able to resolve it though via an override that parses any node_module in the jsdoc and just outputs it as a basic class (no hyperlink)
https://github.com/electron-userland/electron-builder/blob/fa72861f6cd2de97d191f1b2bbfddc6edf48ab6d/scripts/jsdoc/helpers.js#L34-L36
It also would allow other node_module packages to be utilized and/or the TypedEventEmitter to be utilized elsewhere in the app

Nokel81:
That is true.

Issue #5552, opened 2021-01-18
Can't use extraFiles that contain Universal binaries


IanMDay:
* **Version**: 22.10.4

* **Electron Version**: 11.2.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: macOS Universal

I have universal binaries built from another project which I want to include via extraFiles, and have been doing so a while, but I've since updated builds to create Universal binaries.

I'm specifying the folder for the binaries as:

```
"extraFiles": [
  {
    "from": "binaries/${arch}",
    "to": "binaries"
  }
],
```

So the configuration only expects to find files in x64 and arm64 folders, so it only attributes these files to one of those architectures.

Ideally, it should either test the contents of the files given in either folder, or have a new configuration option for a 'universal' arch to indicate the files are already universal, or additionally search the universal arch as well as x64 and arm64.

I've resorted to splitting the binaries with `lipo -extract` for now so that builder is happy, but I don't think this is an ideal solution.

XRenSiu:
I have x64 and arm64 binary built from another project, x64 need support macOS 10.13 and it includes `Frameworks/libswift*` , but arm64 doesn't include `Frameworks`

use `@electron/universal`, i get some error `While trying to merge mach-o files across your apps we found a mismatch, the number of mach-o files is not the same between the arm64 and x64 builds`

As @IanMDay said:
> or have a new configuration option for a 'universal' arch to indicate the files are already universal

that is great.

Hinton:
I just encountered the same issue as @XRenSiu, we bundle a Safari extension inside the `extraFiles` which contains `Frameworks/libswift*` which are no longer used in `arm64`. Adding an option to inject `extraFiles` after combining it should solve it for us.

XRenSiu:
Workaround
1. prepare universal `extralFiles`
2. use `after-sign` hook 

     -  copy `extraFiles` into packaged app
     - resign code
     
```
module.exports = async fucntion(packContext){
  const {
    appOutDir,
    outDir,
    packager
  } = packContext

  const appFile = path.join(
    appOutDir,
    `${packager.appInfo.productFilename}.app`
  )
  copy('<extraFile path>',path.join(appFile, 'Contents/Library/xxxxx/xxx.qlgenerator'))

  // resign
  await packContext.packager.signApp(packContext, true)
}
```

IanMDay:
Yes, its still relevant.

alectrocute:
+1. Currently stuck here.

[...]

PR #6887, opened 2022-05-25
fix: add afterPack call after macOS universal package is created

schetle:
The universal target for macOS fires the afterPack hook after both x64 and arm64, but not after the final package has been created. This is a problem, as there may need to be some final opportunity to adjust the final package before signing in doSignAfterPack.

**Context**:
In my scenario, this comes up when I'm embedding a macOS appex extension (not a kernel extension) in my PlugIns folder. I can't pre-sign my extension, as electron-builder and by extension, @electron/universal seems to require that my extension is separated out into 2 distinct arch extensions (x64, and arm64) for combining them via makeUniversalApp.

I cannot sign these in the current afterPack hooks, because each arch-dependent package gets signed, and then upon the attempted package combination, fails as the packages are different (they'll contain the non-binary signatures, which will be different).

electron-builder will not sign extensions inside of Content/PlugIns when supplied to the "binaries" option, either as it is purposely excluded from signing (even if the assumption that anything here is a kernel extension is erroneous) (see: https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/macPackager.ts#L271). "binaries" cannot contain their own entitlements anyway, making the removal of this exclusion a bad solution.

doSignAfterPack will fail validation unless the embedded appex is properly signed beforehand. It appears the only place to perform this operation will be this PR's new afterPack call after both x64 and arm64 have both been combined into the final package.

### DIFF
--- a/.changeset/wet-ligers-heal.md
+++ b/.changeset/wet-ligers-heal.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+add afterPack call after macOS universal package is created
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -132,7 +132,7 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
         })
         await fs.rm(x64AppOutDir, { recursive: true, force: true })
         await fs.rm(arm64AppOutPath, { recursive: true, force: true })
-        
+
         // Give users a final opportunity to perform things on the combined universal package before signing
         const packContext: AfterPackContext = {
           appOutDir,
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -132,6 +132,18 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
         })
         await fs.rm(x64AppOutDir, { recursive: true, force: true })
         await fs.rm(arm64AppOutPath, { recursive: true, force: true })
+        
+        // Give users a final opportunity to perform things on the combined universal package before signing
+        const packContext: AfterPackContext = {
+          appOutDir,
+          outDir,
+          arch,
+          targets,
+          packager: this,
+          electronPlatformName: platformName,
+        }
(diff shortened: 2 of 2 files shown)


mmaietta:

jtbandes:
> I can't pre-sign my extension, as electron-builder and by extension, @electron/universal seems to require that my extension is separated out into 2 distinct arch extensions (x64, and arm64) for combining them via makeUniversalApp.

It sounds like this is being fixed in https://github.com/electron/universal/pull/47

> I cannot sign these in the current afterPack hooks, because each arch-dependent package gets signed, and then upon the attempted package combination, fails as the packages are different (they'll contain the non-binary signatures, which will be different).

It's strange you say this — I haven't had the same problem (I was using `lipo -extract` to separate the architectures during afterPack so they could be re-combined successfully, as described in https://github.com/electron-userland/electron-builder/issues/5552) and now I need to change my code to avoid running this when `arch == universal`: https://github.com/foxglove/studio/pull/3595

jtbandes:
> If there's an afterPack call after x64 and a separate call for arm64, then it feels natural for the new, unique universal package to have the same treatment.

Agreed! I was just curious to understand what your signing issue was, because I never ran into one with `lipo -extract`.

schetle:
I am using lipo -extract as well to separate out the .appex into the individual architectures.

The issue then became my signing needs. I don't take the .appex as it is signed on disk as-is or before lipo. If I let electron-builder (and osx-sign) sign my .appex, it will fail notarization. So instead, I opted for afterPack scripts that I can fully control the entitlements (that are different than any of the helper/inherit/login entitlements that electron-builder let's you specify) and signing to resign the embedded .appex before sending it off for notarization.

The issue then became, I can't resign the .appex in afterPack because the separate architectures would individually get signed and then become unique and fail the recombination checks with makeUniversalApp.

jtbandes:
I'm also re-signing the embedded appex in afterPack: https://github.com/foxglove/studio/blob/a28ffd16ae71fa34a5f2771c7131b86e8a761565/desktop/afterPack.ts#L114-L122

I never had issues with the recombination in makeUniversalApp. I suppose your entitlements are somehow different 🤷

PR #6893, opened 2022-05-28
docs: add Flatpak documentation

vially:
This should have been a part of #5711 but, as I've [mentioned](https://github.com/electron-userland/electron-builder/pull/5711#issuecomment-817012077) before, I had some issues getting the documentation system to work on my machine in the past.

After some trial and error I did manage to get it working (fwiw, part of the problem was that version 1.2.4 of `mkdocs` is not compatible with Python 3.10 but downgrading to Python 3.8 fixed the problem), so now the `FlatpakOptions` should be properly [documented](https://deploy-preview-6893--car-park-attendant-cleat-11576.netlify.app/configuration/flatpak).


mmaietta:

Issue #6818, opened 2022-04-26
Add installLocation config for NSIS updater


panther7:
dynamically, via nsis json conf (https://www.electron.build/configuration/nsis)

PR #6907, opened 2022-05-31
feat: Add installDir property for NsisUpdater

panther7:
Add installDir property for NsisUpdater. Now is it posible change install folder from AppUpdater.

#6818

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -87,12 +87,6 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
   disableWebInstaller = false
 
   /**
-   * Install dir for Nsis Updater
-   *
-   */
-  installDirectory: string | null = null
-
-  /**
    * The current application version.
    */
   readonly currentVersion: SemVer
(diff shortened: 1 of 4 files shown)


packages/electron-updater/src/AppUpdater.ts
```diff
@@ -86,6 +86,12 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
 
   disableWebInstaller = false
 
+  /**
+   * Install dir for Nsis Updater
+   *
+   */
+  installDir = ''
```
mmaietta, line 93: I'd recommend using `installDirectory: string | null = null` since that will be picked up during doc generation

packages/electron-updater/src/NsisUpdater.ts
```diff
@@ -109,6 +109,11 @@ export class NsisUpdater extends BaseUpdater {
       args.push("--force-run")
     }
 
+    if (this.installDir && typeof this.installDir === 'string') {
```
mmaietta, line 112: I think we shouldn't need to check for type = string here since that should be autovalidated by the schema validator

panther7: ok

mmaietta:
I'm still not in favor of this PR as mentioned here: https://github.com/electron-userland/electron-builder/pull/6907#issuecomment-1143879130

We shouldn't add os-specific properties to the base AppUpdater

panther7:
@mmaietta Moved to NsisUpdater.ts

panther7:
Hi Mike,
all changes applied, what now?
Thanks

@mmaietta

PR #6913, opened 2022-06-01
feat: electron/universal has a new minimatch option 'x64ArchFiles'

obra:
Some projects need to ship x64 binaries which will be run under
Rosetta2 on arm64 Macs.

This option allows a packager to specify that they're aware that
certain files in their distribution which are included in both x64
and arm64 builds are actually x64 binaries in both builds.

Without this option, electron/universal will attempt to lipo
the two binaries together, resulting in a build failure.

### DIFF
--- a/.changeset/famous-cheetahs-run.md
+++ b/.changeset/famous-cheetahs-run.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+Expose electron/universal's new "x64ArchFiles" option to allow building universal binaries with single-architecture dependencies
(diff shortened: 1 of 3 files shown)


mmaietta:
I think this should be fine for now, but I'll need to make a breaking change in the future, as we're adding too many items to the `macOptions` that are specific just to `universal` builds.
Basically, what I'd like is for us to have a `universal: MakeUniversalOpts` utilizing this export: https://github.com/electron/universal/blob/master/src/index.ts#L16
That way we can use Typescript's `Omit` w/ a dedicated interface that stays up to date with electron/universal.
Unfortunately a change like that would break the current API/config, and I don't think that is purely necessary at this time

PR #6941, opened 2022-06-13
feat: upgrade Keygen integration to v1.1

ezekg:
Follow up to #6909, which covers the API changes more in-depth. This should have no breaking changes for app's that are upgrading from v1.0 to v1.1 of the API, since it's only related to publishing (the auto-upgrade functionality still works as it did before from a consumer standpoint), so I think it's safe to include in a minor electron-builder update.

I don't think this should be merged until #6920 is out, to give an option to pin to v1.0 for whatever reason (via #6909).

### DIFF
--- a/packages/app-builder-lib/src/publish/KeygenPublisher.ts
+++ b/packages/app-builder-lib/src/publish/KeygenPublisher.ts
@@ -6,6 +6,77 @@ import { KeygenOptions } from "builder-util-runtime/out/publishOptions"
 import { configureRequestOptions, HttpExecutor, parseJson } from "builder-util-runtime"
 import { getCompleteExtname } from "../util/filename"
 
+type RecursivePartial<T> = {
+  [P in keyof T]?: RecursivePartial<T[P]>
+}
+
+export interface KeygenError {
+  title: string
+  detail: string
+  code: string
+}
+
+export interface KeygenRelease {
+  id: string
+  type: "releases"
+  attributes: {
+    name: string | null
+    description: string | null
+    channel: "stable" | "rc" | "beta" | "alpha" | "dev"
+    status: "DRAFT" | "PUBLISHED" | "YANKED"
+    tag: string
+    version: string
+    semver: {
+      major: number
+      minor: number
+      patch: number
+      prerelease: string | null
+      build: string | null
+    }
+    metadata: { [s: string]: any }
+    created: string
+    updated: string
+    yanked: string | null
+  }
+  relationships: {
+    account: {
+      data: { type: "accounts"; id: string }
+    }
+    product: {
+      data: { type: "products"; id: string }
+    }
+  }
+}
+
+export interface KeygenArtifact {
+  id: string
+  type: "artifacts"
+  attributes: {
+    filename: string
+    filetype: string | null
+    filesize: number | null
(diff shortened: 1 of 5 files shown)


packages/app-builder-lib/src/publish/KeygenPublisher.ts
```diff
@@ -36,78 +107,143 @@ export class KeygenPublisher extends HttpPublisher {
...
-              type: "product",
-              id: this.info.product,
-            },
+
+    const data: RecursivePartial<KeygenRelease> = {
+      type: "releases",
+      attributes: {
+        version: this.version,
+        channel: this.info.channel || "stable",
+        status: "PUBLISHED",
```
ezekg, line 223: I'd prefer to start with a `DRAFT` status for the release, and publish only after all artifacts are uploaded, but I couldn't find a hook to accomplish that (i.e. letting the publisher perform an additional request once all artifacts are uploaded). LMK if one exists and I just happened to miss it.

mmaietta: Hmmm, I don't see any in the code for that functionality, but it totally seems like a valid reason to have one added :) 
Seems like an `onArtifactUploaded` could be added here? Simply need to add a task (i.e. Promise) to the `taskManager` after the `scheduleUpload`
https://github.com/electron-userland/electron-builder/blob/03cc9b96e0be07ef3450e3992eafcb7fe903a853/packages/app-builder-lib/src/publish/PublishManager.ts#L123-L133

From my quick look, an `onAllArtifactsUploaded` hook could probably be added here:
https://github.com/electron-userland/electron-builder/blob/03cc9b96e0be07ef3450e3992eafcb7fe903a853/packages/app-builder-lib/src/publish/PublishManager.ts#L230-L241

mmaietta: If only adding `onAllArtifactsUploaded`, which is probably all we need for now, the best place might just be adding it here:
https://github.com/electron-userland/electron-builder/blob/03cc9b96e0be07ef3450e3992eafcb7fe903a853/packages/app-builder-lib/src/index.ts#L113

ezekg: I was thinking about doing that in a separate follow-up PR, and then updating the Keygen publisher while I'm there. Is that cool? As-is, I think the current publisher works even if it does skip the draft stage.

mmaietta: Works for me 😄

ezekg: Yeah, it's ready for review but I wanted to wait until #6920 is out (which includes #6909) before we consider merging it, so I'm keeping the PR in draft mode. Detailed more of the reasoning there in the PR description.

LMK if you want me to move it out of a draft state. 👍

mmaietta: Published 23.2.0 + electron-updater@5.0.6

ezekg: Awesome. Marked this as ready for review. 👍

PR #6961, opened 2022-06-24
fix: Remove DISABLE_WAYLAND for snaps

### LINKED ISSUE #6888: Please enable wayland for snaps
Currently, the snap builder is setting `DISABLE_WAYLAND: 1`, but I don't think that's necessary anymore. Electron apps run fine under Wayland since they default to Xwayland, but there are some commandline arguments available to run it under native Wayland: `--enable-features=WaylandWindowDecorations --ozone-platform=wayland`. The Chromium snap has also [recently re-enabled wayland support](https://git.launchpad.net/~chromium-team/chromium-browser/+git/snap-from-source/commit/?id=a4f41b5c065e14e954bc81d2d3e2b0c7fc26fc4d). Keeping `DISABLE_WAYLAND: 1` means that even the commandline options to force native Wayland don't work.

aripollak:
Fixes #6888

### DIFF
--- a/.changeset/stupid-kiwis-eat.md
+++ b/.changeset/stupid-kiwis-eat.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: Optionally allow removing DISABLE_WAYLAND flag for snaps
--- a/packages/app-builder-lib/src/options/SnapOptions.ts
+++ b/packages/app-builder-lib/src/options/SnapOptions.ts
@@ -50,7 +50,7 @@ export interface SnapOptions extends CommonLinuxOptions, TargetSpecificOptions {
 
   /**
    * The list of [plugs](https://snapcraft.io/docs/reference/interfaces).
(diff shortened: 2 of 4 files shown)


mmaietta:
This feels like a breaking change for electron-builder. I'm wondering if we can make this a feature instead by adding `allowNativeWayland` as a property in SnapOptions.
```
allowNativeWayland: boolean | null = null
```
Then `DISABLE_WAYLAND: Number(!allowNativeWayland)`

What are your thoughts?

aripollak:
Unfortunately I'm not sure if this would be a breaking change, but I'm happy to add an option to enable it in the meantime. Maybe in the future it could be enabled by default if a minimum known-working electron version is used.

aripollak:
I'll have to use some different logic for setting DISABLE_WAYLAND because [this](https://github.com/snapcore/snapcraft/blob/b81550376df7f2d0dfe65f7bfb006a3107252450/extensions/desktop/common/desktop-exports#L248) expects it to be unset/"" for false, or a non-empty value for true.

aripollak:
Okay, I added that option. I wasn't able to update the docs locally because I haven't been able to get my local environment set up to compile this properly (pnpm install is complaining about yarn exiting with status 1).

PR #6964, opened 2022-06-28
fix: nsis-web target set APP_PACKAGE_URL_IS_INCOMPLETE

geovie:
When specifying `appPackageUrl` in the `NsisWebOptions` config `APP_PACKAGE_URL_IS_INCOMPLETE` would not be set to `null` and thus the installer would not include the packages.

Now I'm not sure if `APP_PACKAGE_URL_IS_INCOMPLETE` is even needed or if the check in `webPackage.nsh` could even be removed.

### DIFF
--- a/.changeset/cuddly-books-approve.md
+++ b/.changeset/cuddly-books-approve.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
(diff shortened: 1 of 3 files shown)


grovesNL:
This seems to cause an extra path to be appended to `appPackageUrl`, see #7732

Is this expected or is there a known workaround?

PR #6970, opened 2022-06-29
fix(mas): Allow signing with "3rd Party Mac Developer Application"

csett86:
Despite the "Developer" in the name, this is only used for Mac App
Store Distribution.

This was by accident removed in #6105


suchal:
@csett86 any idea when it will be merged?

PR #6976, opened 2022-06-30
fix: Update snapcraft.yaml to use gnome-3-38-2004

eeickmeyer:
The core18 and plug gnome-3-28-1804 lose standard (free/community) support in less than a year in accordance with the Ubuntu LTS timeline as these both correspond with Ubuntu 18.04 LTS, which loses standard support in April 2023 and moves to paid-only support. To continue to have secure builds for Electron apps, new snap builds should move to core20 and use plug gnome-3-38-2004.

As an added bonus, this will also allow Ubuntu and official Ubuntu flavors to seed Electron snaps built with core20 and gnome-3-38-2004 because gnome-3-38-2004 provides the proper facility for this process.

### DIFF
--- a/.changeset/twelve-snakes-whisper.md
+++ b/.changeset/twelve-snakes-whisper.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: Update snapcraft.yaml to use gnome-3-38-2004
--- a/packages/app-builder-lib/templates/snap/snapcraft.yaml
+++ b/packages/app-builder-lib/templates/snap/snapcraft.yaml
@@ -24,7 +24,7 @@ parts:
   app:
     plugin: "nil"
       # cd ~ && rm -rf ~/squashfs-root && unsquashfs /media/psf/ramdisk/electron-builder-test/dist/__snap-x64/se-wo-template_1.1.0_amd64.snap
-      # comm -12 <(ls ~/squashfs-root/usr/lib/x86_64-linux-gnu/) <(ls /snap/gnome-3-28-1804/current/usr/lib/x86_64-linux-gnu/) > /media/psf/Home/f.txt
+      # comm -12 <(ls ~/squashfs-root/usr/lib/x86_64-linux-gnu/) <(ls /snap/gnome-3-38-2004/current/usr/lib/x86_64-linux-gnu/) > /media/psf/Home/f.txt
       # run snap-exclude-list.js
     stage:
       - '-usr/lib/python*'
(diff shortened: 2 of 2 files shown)


ItzSwirlz:
snaps also do other weird things; for example it opens a multipass on arm64 snap builds

eeickmeyer:
@ItzSwirlz Perhaps, but that's unrelated to this pull request.

eeickmeyer:
I actually now realize that this might not alone do the job. There are several 18.04-based dependencies that are being distributed as well that would need to be updated (as I tried a snap build with just updating the yaml file and the snap wouldn't run). Of course, I could be wrong there as well, but the long and short of it is that 18.04 dependencies should be moved to 20.04 ASAP.

mmaietta:
What are the 18.04-based dependencies?

eeickmeyer:
> What are the 18.04-based dependencies
That would be pretty much everything listed under app: stage:.

mmaietta:
Oh jeez, I have no idea how one would transition it to 20.04, I'm not familiar with snaps or linux tbh. Happy to review a PR though 🙂

MikeJerred:
FYI, I am patching this for my electron build, simply changing `base: core18` to `base: core20` in `app-builder-lib/templates/snap/snapcraft.yaml` and it seems to work correctly.

MikeJerred:
Changing the base alone works, but I also need to update the gnome extension to fix bugs e.g. this one: https://forum.snapcraft.io/t/snapped-app-not-loading-fonts-on-fedora-and-arch/12484/116

When I update to `gnome-3-38-2004` by doing this:

app-builder-lib/templates/snap/snapcraft.yaml
```diff
-base: core18
+base: core20
...
plugs:
- gnome-3-28-1804:
+ gnome-3-38-2004:
    interface: content
    target: $SNAP/gnome-platform
-   default-provider: gnome-3-28-1804
+   default-provider: gnome-3-38-2004
```

app-builder-lib/out/targets/snap.js
```diff
        const appDescriptor = {
+           extensions: ["gnome-3-38"],
            command: "command.sh",
            plugs: plugNames,
            adapter: "none",
        };
```

The snap will not run, giving errors like:
```
g_module_open() failed for /snap/.../usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so: librsvg-2.so.2: cannot open shared object file: No such file or directory
/snap/.../usr/lib/x86_64-linux-gnu/libgtk-3-0/gtk-query-immodules-3.0: error while loading shared libraries: libharfbuzz.so.0: cannot open shared object file: No such file or directory
```

PR #6989, opened 2022-07-06
fix: parallel release creation with Keygen publisher

ezekg:
Looks like the previous publisher changes contained a bug that slipped through the cracks, as far as automated tests. 🙈

Under a normal publish, electron-builder will upload multiple artifacts. The Keygen publisher has a find-or-create method for creating a new release bucket before publishing artifacts. What I didn't account for in #6941 was that electron-builder uploads these artifacts in parallel, so the find-or-create method isn't as robust as it needed to be. This resulted in `409 Conflict` errors when multiple create attempts were made in parallel for a new release.

This wasn't caught in tests because the test for the Keygen publisher only uploads a single artifact, so a conflict will never occur. That test has been modified to upload multiple artifacts, rather than a single artifact.

This PR resolves the issue by making an additional fetch if a conflict error is encountered.


mmaietta:
> this may be a good indication that before/after hooks are needed for publishers. It'd be much nicer to be able to tell electron-builder to create the release _before_ uploading any artifacts, then publish the release _after_ uploading all artifacts. This would alleviate any need for this find-or-create method.

I'm in full agreement of this approach as well.

Issue #1501, opened 2017-04-26
NSIS build - add welcome page


thavasukkannu:
Version: 17.0.3
Target: NSIS (windows)

Thanks for given NSIS features. 

I have taken the build, it's working good. When I install the .exe setup it starting with "choose install location" in the Windows Installer setup page. But my requirement is adding the "Welcome page" as the first page. 

![test](https://cloud.githubusercontent.com/assets/28038281/25420986/4a24e3cc-2a78-11e7-9936-2b64485ec358.png)

![welcome](https://cloud.githubusercontent.com/assets/28038281/25420994/53708256-2a78-11e7-848d-2201b613593d.png)

Is that possible with using Installer.nsh? or any other way to achieve this? please suggest.

develar:
`MUI_PAGE_WELCOME` is not added by intention. Only for uninstaller — https://github.com/electron-userland/electron-builder/blob/master/packages/electron-builder/templates/nsis/boringInstaller.nsh#L51

For what do you need welcome page?

thavasukkannu:
@develar Thanks for quick replay :) .

For greeting window page.

Actually I need to create a custom page before the installation page. Is it possible?
I would like to create same as below image.

![requirement](https://cloud.githubusercontent.com/assets/28038281/25424089/b3467b84-2a84-11e7-9b10-174d8b7f47bc.png)

develar:
Well, if https://github.com/electron-userland/electron-builder/wiki/NSIS#custom-nsis-script will be not enough, you can fork project, modify as you need and send PR.

thavasukkannu:
@develar  Ok thanks :) .

Sivasss94:
Actually i just used electron builder (nsis target) to create a setup file for windows application, but i need to add another one installer page in the setup, so i can get the input (serverIP) from the end user and i can modify the api_url in app.asar file or asar unpacked main.js file. Is there any another way? I checked the Custom Installer script, but it is little confusing. So Please help me in this. And Thanks in advance.

Issue #3809, opened 2019-04-05
NSIS - Custom pages appear twice


agpreynolds:
* **Version**:  20.39.0

* **Target**:  NSIS (With custom include)

Hello i've noticed some strange behaviour when attempting to create a custom installer page.

My config is as follows:
```
  "build": {
    "appId": "my-app",
    "directories": {
      "output": "dist"
    },
    "extraResources": {
      "from": "..\\..\\output\\test\\api",
      "to": "api/",
      "filter": [
        "**/*"
      ]
    },
    "nsis": {
      "oneClick": false
    },
    "win": {
      "target": [ "nsis" ]
    }
  }
```

With an additional script (build/installer.nsh):

```
!include MUI2.nsh

!macro customWelcomePage

	!insertMacro MUI_PAGE_WELCOME

!macroEnd
```

What appears to be happening is the pages appear in the following order:
- **Welcome**
- per-user / per-machine
- Administrator access confirmation
- **Welcome**
- Installation
- Completed

With the welcome page appearing twice.

This also happens with more sophisticated custom pages defined with the command:
`Page custom someFunction`

Is there something obvious which i'm doing wrong?

agpreynolds:
As an additional piece of information. This doesn't seem to happen with oneClick set to false and perMachine set to true.

I suspect it is related to https://github.com/electron-userland/electron-builder/blob/4ee4be63159472d621abc30d5fbe017d7a88981a/packages/app-builder-lib/templates/nsis/multiUserUi.nsh#L9

However i'm not sufficiently experiences in NSIS and MUI to suggest how to resolve it

EthannYakabuski:
Can't seem to get it to work with your suggestion of perMachine = true either. Any other suggestions?

PR #2384, opened 2017-12-13
add customWelcomePage macro

dengyaolong:
https://github.com/electron-userland/electron-builder/issues/2372
https://github.com/electron-userland/electron-builder/issues/1501

PR #7001, opened 2022-07-14
docs: Add customWelcomePage example to Custom NSIS script

hueyyeng:
The docs doesn't cover customWelcomePage for NSIS installer as it is not possible to configure the Welcome Page using configuration.

This is useful if one need to display Welcome Page for NSIS installer as the default behaviour is intentionally omitted. Refer to https://github.com/electron-userland/electron-builder/issues/1501#issuecomment-297274642

### DIFF
--- a/docs/configuration/nsis.md
+++ b/docs/configuration/nsis.md
@@ -37,30 +37,35 @@ Two options are available — [include](#NsisOptions-include) and [script](#Nsis
 Keep in mind — if you customize NSIS script, you should always state about it in the issue reports. And don't expect that your issue will be resolved.
 
 1. Add file `build/installer.nsh`.
-2. Define wanted macro to customise: `customHeader`, `preInit`, `customInit`, `customUnInit`, `customInstall`, `customUnInstall`, `customRemoveFiles`, `customInstallMode`.
-    
+2. Define wanted macro to customise: `customHeader`, `preInit`, `customInit`, `customUnInit`, `customInstall`, `customUnInstall`, `customRemoveFiles`, `customInstallMode`, `customWelcomePage`.
+
     !!! example
         ```nsis
         !macro customHeader
           !system "echo '' > ${BUILD_RESOURCES_DIR}/customHeader"
         !macroend
-        
+
         !macro preInit
           ; This macro is inserted at the beginning of the NSIS .OnInit callback
           !system "echo '' > ${BUILD_RESOURCES_DIR}/preInit"
         !macroend
-        
+
         !macro customInit
           !system "echo '' > ${BUILD_RESOURCES_DIR}/customInit"
         !macroend
-        
+
         !macro customInstall
           !system "echo '' > ${BUILD_RESOURCES_DIR}/customInstall"
         !macroend
-        
+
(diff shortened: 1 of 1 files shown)


docs/configuration/nsis.md
```diff
@@ -37,31 +37,36 @@ Two options are available — [include](#NsisOptions-include) and [script](#Nsis
...
         !macro customInstallMode
-          # set $isForceMachineInstall or $isForceCurrentInstall 
+          # set $isForceMachineInstall or $isForceCurrentInstall
           # to enforce one or the other modes.
         !macroend
+
+        !macro customWelcomePage
+          # Welcome Page is not added by default for installer.
+          !insertMacro MUI_PAGE_WELCOME
+        !macroend
```
mmaietta, line 69: Hiya! I'm not sure I understand this change.  What is `MUI_PAGE_WELCOME` and why does the developer have to custom specify it in a separate include.nsh file?

I only see `customWelcomePage` being used here:
https://github.com/electron-userland/electron-builder/blob/28cb86bdcb6dd0b10e75a69ccd34ece6cca1d204/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh#L7-L11
So I don't understand how adding this `MUI_PAGE_WELCOME` to the `customWelcomePage` will work for an installer. Could you clarify this change?

hueyyeng: Hi @mmaietta my understanding after going through Nullsoft Scriptable Install System docs is that one will need to specify the pages manually and the macro `customWelcomePage` introduced in this PR https://github.com/electron-userland/electron-builder/pull/2384 allows dev to display the Welcome Page when running the setup for installation.

Without modifying the logic of Electron Builder for generating NSIS build, the `build/installer.nsh` file as suggested by the docs didn't mentioned anything about  `customWelcomePage` macro. Specifying the `!insertmacro MUI_PAGE_WELCOME` in the `customWelcomePage` macro allows me to restore the Welcome Page.

As for what is `MUI_PAGE_WELCOME`? It is one of the available page macros (more specifically it refers to the installer page) as part of NSIS Modern User Interface https://nsis.sourceforge.io/Docs/Modern%20UI/Readme.html (scroll down to Pages section). If you see in the `assistedInstaller.nsh` file, there is `MUI_UNPAGE_WELCOME` and `MUI_UNPAGE_INSTFILES` for the uninstaller setup.

https://github.com/electron-userland/electron-builder/blob/28cb86bdcb6dd0b10e75a69ccd34ece6cca1d204/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh#L67-L74

Let me know if you need real world examples. I've been searching for how to insert the Welcome Page for the installer setup and there is not many results except for this issue https://github.com/electron-userland/electron-builder/issues/3809 although I found it after figuring out the `customWelcomePage` macro method. Ironically the uninstaller setup does have "Welcome Page" on line 68 in the code snippet above.

mmaietta: Thanks for the explanation! I'm not familiar with NSIS scripting so that helps a lot. I'm kicking off the final CI build/tests now

PR #7032, opened 2022-07-30
fix: inappropriate `electron-updater` cache directory on macOS

kidonng:
On macOS, `~/Library/Application Support` is where applications persist data, and caches are stored in `~/Library/Caches`.

Currently `electron-updater` mistakenly created its own `Caches` directory under `~/Library/Application Support`, which is bad practice and not align with e.g. the usage of `XDG_CACHE_HOME` below.

### DIFF
--- a/.changeset/smooth-hounds-peel.md
+++ b/.changeset/smooth-hounds-peel.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix: use appropriate `electron-updater` cache directory on macOS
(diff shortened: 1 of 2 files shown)


mmaietta:
Would this be a breaking change? Not sure what else relies on the Cache

kidonng:
> Would this be a breaking change?

This should only affects the updater so I don't think it could break anything.

> Not sure what else relies on the Cache

I'm not very familiar with electron-builder so I can't tell. But if this refers to the one under `Application Support` then anything else using it needs to be fixed as well.

PR #7060, opened 2022-08-12
feat: Introducing deb and rpm auto-updates

mmaietta:
[Draft]
Utilizes GUI-based `sudo` for installation which is pulled via `which gksudo || which kdesudo || which pkexec || which beesu` and executed using `/bin/bash -c` in order to coalesce all commands under a single `sudo` prompt
Deb is fairly straightforward and evals to this:
```
`which dnf || which yum` -i path-to-update.deb || apt-get install --fix-broken -y
```

Rpm requires one of these package managers available: `zypper`, `dnf`, `yum`
```
zypper -y remove AppName; zypper -y install path-to-update.rpm
```
js is optimized, but the shell cmd basically is:
```
`which dnf || which yum` remove -y AppName;
`which dnf || which yum` clean --all;
`which dnf || which yum` --no-refresh install --allow-unsigned-rpm -y -f path-to-update.rpm
```

Testing still in progress on Ubuntu 14/18/20/22 and Fedora 32. No clue what else to test on, suggestions would be appreciated

Things I could use help/verification on:
 - idk if `--allow-unsigned-rpm` is a good idea, but that's what I use in my home project.
 - are those enough package managers to cover?
 - 

Note: This would need to go through a significant `alpha` version stage for electron-builder


mmaietta:
@develar what are your thoughts on this?

Kilian:
I would love to configure this for my app but there's (seemingly) no documentation on how to set up autoupdating for .deb. What are the steps needed to support and test this?

mmaietta:
@Kilian it should be the exact same setup as when using AppImage, just need to add `deb` as a target in electron-builder config.

PR #7074, opened 2022-08-19
feat: In non-silent installation mode, allows the user to decide whether to run the installed application

shenglianlee:
Currently only in silent installation mode allows user to set "isForceRunAfter" value to decide whether to run the installed application. Actually sometimes in non-silent installation mode, the user does not want to run the installed application after the installation is complete.


mmaietta:
This feels like it's a breaking change in previous functionality, not sure if we can just have it as a `minor` semver bump. What are your thoughts?

shenglianlee:
> This feels like it's a breaking change in previous functionality, not sure if we can just have it as a `minor` semver bump. What are your thoughts?

Agree. I will submit a new PR.

PR #7075, opened 2022-08-19
fix: use `buildVersion` not `buildNumber` for fpm `--iteration` flag

### LINKED ISSUE #6945: Unable to install .deb packages due to embedding CI tag in Version number
* **Electron-Builder Version**: 23.1.0

* **Node Version**: 16.5.0

* **Electron Version**: 18.0.3

* **Target**: Linux .deb package

I've recently gotten electron-builder building and uploading my packaged app in CI (using Azure Pipelines) for Mac, Windows and Linux. It's mostly working great, and I was all set to do a release, when I discovered that the generated .deb file won't install (using `dpkg` on Ubuntu 20) because of the format of the `Version` field in the generated .deb file.
The issue appears to be that, under CI, electron-builder is automatically adding a tag (presumably from an env var set by the CI system) to the generated version number:

```sh
$ sudo dpkg -i tad_0.10.1_amd64.deb
[sudo] password for parallels: 
dpkg: error processing archive tad_0.10.1_amd64.deb (--install):
 parsing file '/var/lib/dpkg/tmp.ci/control' near line 2 package 'tad':
 'Version' field value '0.10.1-IndividualCI_refs_heads_master_25': invalid character in revision number
Errors were encountered while processing:
 tad_0.10.1_amd64.deb
```

Looking at the .deb package contents:
```sh
$ dpkg --info tad_0.10.1_amd64.deb
 new Debian package, version 2.0.
 size 138734940 bytes: control archive=9440 bytes.
     510 bytes,    14 lines      control              
   32336 bytes,   292 lines      md5sums              
     257 bytes,    10 lines   *  postinst             #!/bin/bash
      66 bytes,     4 lines   *  postrm               #!/bin/bash
 Package: tad
 Version: 0.10.1-IndividualCI_refs_heads_master_25
 License: MIT
 Vendor: Antony Courtney <antony@xxxxxx.com>
 Architecture: amd64
 Maintainer: Antony Courtney <antony@xxxxxx.com>
 Installed-Size: 525349
 Depends: libgtk-3-0, libnotify4, libnss3, libxss1, libxtst6, xdg-utils, libatspi2.0-0, libuuid1, libsecret-1-0
 Recommends: libappindicator3-1
 Section: default
 Priority: optional
 Homepage: https://www.tadviewer.com
 Description: 
   Tabular data viewer desktop app, built with Electron
```

Is there some way to change the behavior of electron-builder so that it just uses the `version` from `package.json` as the `Version` field in the generated .deb file, without attempting to append the build tag / build number?

(That said, I can see that it's quite helpful and convenient to have this build tag stamped into the generated package for debugging purposes. Would love to know how others are dealing with this, or if there is some other recommended solution for getting installable .deb files).

Thanks in advance for any help!

davej:
Fixes #6945.

`buildNumber` is not overridable in the config and is provided directly by the CI, this should be `buildVersion` instead.

Dashes are not supported for iteration in some versions of fpm, so I replaced dashes with underscores. https://github.com/jordansissel/fpm/issues/1833

Also, the linter complained, so I fixed up the type definition for the `use` function.

### DIFF
--- a/packages/app-builder-lib/src/configuration.ts
+++ b/packages/app-builder-lib/src/configuration.ts
@@ -131,13 +131,13 @@ export interface Configuration extends PlatformSpecificBuildOptions {
 
   /**
    * The build number. Maps to the `--iteration` flag for builds using FPM on Linux.
-   * If not defined then will fallback to `BUILD_NUMBER` or `TRAVIS_BUILD_NUMBER` or `APPVEYOR_BUILD_NUMBER` or `CIRCLE_BUILD_NUM` or `BUILD_BUILDNUMBER` or `CI_PIPELINE_IID` env.
+   * If not defined, then it will fallback to `BUILD_NUMBER` or `TRAVIS_BUILD_NUMBER` or `APPVEYOR_BUILD_NUMBER` or `CIRCLE_BUILD_NUM` or `BUILD_BUILDNUMBER` or `CI_PIPELINE_IID` env.
    */
   readonly buildNumber?: string | null
 
   /**
    * The build version. Maps to the `CFBundleVersion` on macOS, and `FileVersion` metadata property on Windows. Defaults to the `version`.
-   * If `buildVersion` is not defined and `buildNumber` (or one of the `buildNumber envs) is defined, it will be used as a build version (`version.buildNumber`).
+   * If `buildVersion` is not defined and `buildNumber` (or one of the `buildNumber` envs) is defined, it will be used as a build version (`version.buildNumber`).
    */
   readonly buildVersion?: string | null
 
--- a/packages/app-builder-lib/src/configuration.ts
+++ b/packages/app-builder-lib/src/configuration.ts
@@ -130,8 +130,14 @@ export interface Configuration extends PlatformSpecificBuildOptions {
   readonly npmRebuild?: boolean
 
   /**
+   * The build number. Maps to the `--iteration` flag for builds using FPM on Linux.
+   * If not defined then will fallback to `BUILD_NUMBER` or `TRAVIS_BUILD_NUMBER` or `APPVEYOR_BUILD_NUMBER` or `CIRCLE_BUILD_NUM` or `BUILD_BUILDNUMBER` or `CI_PIPELINE_IID` env.
+   */
+  readonly buildNumber?: string | null
+
+  /**
    * The build version. Maps to the `CFBundleVersion` on macOS, and `FileVersion` metadata property on Windows. Defaults to the `version`.
-   * If `TRAVIS_BUILD_NUMBER` or `APPVEYOR_BUILD_NUMBER` or `CIRCLE_BUILD_NUM` or `BUILD_NUMBER` or `bamboo.buildNumber` or `CI_PIPELINE_IID` env defined, it will be used as a build version (`version.build_number`).
+   * If `buildVersion` is not defined and `buildNumber` (or one of the `buildNumber envs) is defined, it will be used as a build version (`version.buildNumber`).
    */
   readonly buildVersion?: string | null
 
--- a/packages/app-builder-lib/src/targets/fpm.ts
+++ b/packages/app-builder-lib/src/targets/fpm.ts
@@ -181,7 +181,7 @@ export default class FpmTarget extends Target {
     }
 
     use(packager.info.metadata.license, it => args.push("--license", it))
-    use(appInfo.buildVersion, it =>
+    use(appInfo.buildNumber, it =>
       args.push(
         "--iteration",
         // dashes are not supported for iteration in older versions of fpm
--- a/packages/app-builder-lib/src/targets/fpm.ts
+++ b/packages/app-builder-lib/src/targets/fpm.ts
@@ -180,10 +180,17 @@ export default class FpmTarget extends Target {
       }
     }
 
-    use(packager.info.metadata.license, it => args.push("--license", it!))
-    use(appInfo.buildNumber, it => args.push("--iteration", it!))
+    use(packager.info.metadata.license, it => args.push("--license", it))
+    use(appInfo.buildVersion, it =>
+      args.push(
+        "--iteration",
+        // dashes are not supported for iteration in older versions of fpm
+        // https://github.com/jordansissel/fpm/issues/1833
+        it.split("-").join("_")
+      )
+    )
 
-    use(options.fpm, it => args.push(...(it as any)))
+    use(options.fpm, it => args.push(...it))
 
     args.push(`${appOutDir}/=${installPrefix}/${appInfo.sanitizedProductName}`)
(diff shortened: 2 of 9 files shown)


sandeep1995 (approved):

sandeep1995:
Tested. Works well.

davej:
Fixed a few more lint errors. All tests/lints should pass now.

mmaietta:
I'm pretty sure this is a breaking change. I rely on` buildNumber` to be used for `--iteration `as well. Or rather I set the env var manually. It comes from here:
https://github.com/electron-userland/electron-builder/blob/369f1fa793c28d8743e19ef07ce6eb091c191fb0/packages/app-builder-lib/src/appInfo.ts#L42-L55

Why can't we retain current behavior? Why not just allow overriding with `metadata.buildVersion` in the package.json or electron-builder config file? Similar to this but for `buildNumber`
https://github.com/electron-userland/electron-builder/blob/369f1fa793c28d8743e19ef07ce6eb091c191fb0/packages/app-builder-lib/src/appInfo.ts#L64

davej:
@mmaietta yes, technically it would be a breaking change. I'm not sure there are any real-world scenarios where this might cause real-world issues? I assumed this was a bug as `buildNumber` isn't used directly by any other distribution formats, only fpm. fpm doesn't seem to use `buildVersion` anywhere either (like other formats do).

> Why not just allow overriding with metadata.buildVersion in the package.json or electron-builder config file?

Isn't that pretty close to what we're doing in this PR? If `buildVersion` is explicitly defined then `--iteration` will be `${buildVersion}`. If `buildVersion` isn't defined then `--iteration` will be `${version}.${buildNumber}` (with `buildNumber` defined by the env var).

That's my thinking on it but I'm happy to proceed with whatever direction you think makes sense.

Unrelated to this PR but one final note about `buildNumber`: the envs it uses aren't always just build numbers. In my case (Azure Pipelines), it leaked the associated commit message and email address. It's unlikely that the env var will have sensitive information in it but it's just something to be aware of.

mmaietta:
>one final note about buildNumber: the envs it uses aren't always just build numbers. In my case (Azure Pipelines), it leaked the associated commit message and email address.

Hmmm That, by definition, is not a build **number**. I'm not sure how to solve for that edge case as it doesn't seem normal. We can add a way to override via metadata or config, but even that could be considered a breaking change

davej:
@mmaietta how do you want me to proceed? I'm happy to make whatever changes you think are reasonable. The main thing that I want to do is fix #6945, everything else is just a personal preference.

I guess the minimal changeset for that would be to just add `it.split("-").join("_")` to the `--iteration` param. Even this would technically be a breaking change but if the `--iteration` "number" had a dash then the build would have failed . We could go a step further and do more aggressive sanitization of the string and just allow alphanumerics, periods (`.`) and underscores (`_`) because I'm not actually sure if other characters will break the build.

If we want to do a non-breaking change then I guess we just add a `buildNumber` to the config that will take precedence over any env vars if it is explicitly defined. You mentioned that this is a breaking change above but I'm not sure what you mean? Do you mean that the user may have provided a `buildNumber` config option even though it isn't used by electron-builder?

mmaietta:
> If we want to do a non-breaking change then I guess we just add a buildNumber to the config that will take precedence over any env vars if it is explicitly defined.

This sounds perfect. Something like this below, but you may need to make other edits for the `buildNumber` property to be available. Not sure if it's accessible via `config` either
```
this.buildNumber = info.config.buildNumber || info.metadata.buildNumber ||
      process.env.BUILD_NUMBER ||
      ..... // the other env vars
```

davej:
@mmaietta Thanks! I've made the requested changes. Let me know if I'm missing anything.

mmaietta:
Looks good. Can you please add a changeset for this PR via `pnpm generate-changeset`? That way it gets auto-picked up by the release automation.

packages/app-builder-lib/src/configuration.ts
```diff
@@ -129,9 +129,15 @@ export interface Configuration extends PlatformSpecificBuildOptions {
    */
   readonly npmRebuild?: boolean
 
+  /**
+   * The build number. Maps to the `--iteration` flag for builds using FPM on Linux.
+   * If not defined then will fallback to `BUILD_NUMBER` or `TRAVIS_BUILD_NUMBER` or `APPVEYOR_BUILD_NUMBER` or `CIRCLE_BUILD_NUM` or `BUILD_BUILDNUMBER` or `CI_PIPELINE_IID` env.
```
mmaietta, line 134: Minor grammar change (comma + 'it'): "If not defined, then it will fallback"

packages/app-builder-lib/src/configuration.ts
```diff
@@ -129,9 +129,15 @@ export interface Configuration extends PlatformSpecificBuildOptions {
...
+   */
+  readonly buildNumber?: string | null
+
   /**
    * The build version. Maps to the `CFBundleVersion` on macOS, and `FileVersion` metadata property on Windows. Defaults to the `version`.
-   * If `TRAVIS_BUILD_NUMBER` or `APPVEYOR_BUILD_NUMBER` or `CIRCLE_BUILD_NUM` or `BUILD_NUMBER` or `bamboo.buildNumber` or `CI_PIPELINE_IID` env defined, it will be used as a build version (`version.build_number`).
+   * If `buildVersion` is not defined and `buildNumber` (or one of the `buildNumber envs) is defined, it will be used as a build version (`version.buildNumber`).
```
mmaietta, line 140: "`buildNumber envs" is missing a closing backtick around buildNumber

packages/app-builder-lib/src/targets/fpm.ts
```diff
@@ -180,10 +180,17 @@ export default class FpmTarget extends Target {
...
-    use(appInfo.buildNumber, it => args.push("--iteration", it!))
-
-    use(options.fpm, it => args.push(...(it as any)))
+    use(packager.info.metadata.license, it => args.push("--license", it))
+    use(appInfo.buildNumber, it =>
+      args.push(
+        "--iteration",
+        // dashes are not supported for iteration in older versions of fpm
+        // https://github.com/jordansissel/fpm/issues/1833
+        it.split("-").join("_")
```
mmaietta, line 189: Change to `it.replaceAll('-', '_')`

davej: `replaceAll` isn't supported in Node 14.

mmaietta:
Hmmm, are you using a different version of typescript (installed globally)? I just ran a checkout of your PR via `gh pr checkout 7075` and ran `pnpm install && pnpm generate-all` locally and successfully.

davej:
@mmaietta downgraded to node 14 (from node 16) and everything worked ok. See the latest commit based on your comments.

PR #7094, opened 2022-08-26
fix: close file stream when error

HppZ:
file stream should be closed when error.


packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -485,9 +485,8 @@ function configurePipes(options: DownloadCallOptions, response: IncomingMessage)
   let lastStream = response
   for (const stream of streams) {
     stream.on("error", (error: Error) => {
-      if (!options.options.cancellationToken.cancelled) {
-        options.callback(error)
-      }
+      fileOut.close()
+      options.callback(error)
```
mmaietta, line 489: Great callout on closing the stream

I think we still need: `if (!options.options.cancellationToken.cancelled) {` though

HppZ: callback should be invoked even if the downloading is cancelled by user.

HppZ: > Great callout on closing the stream
> 
> I think we still need this though:
> 
> ```
> if (!options.options.cancellationToken.cancelled) {
>        options.callback(error)
> }
> ```

callback should be invoked even if the downloading is cancelled by user.

mmaietta: Hmm, I disagree though, as user cancelation is an explicit action that I think should not resolve into an error event.

HppZ: agree!~

Issue #7104, opened 2022-08-31
Snap.yaml has invalid fields


MikeJerred:
* **Electron-Builder Version**: 23.3.3

* **Node Version**: 16.14.2

* **Electron Version**: 19.0.14
* **Electron Type (current, beta, nightly)**: current

* **Target**: snap

Currently app-builder-lib copies all the fields specified in `snapcraft.yaml` into `meta/snap.yaml`, but there are some fields that shouldn't be copied (see https://code.launchpad.net/~logicoversnacks/review-tools/+git/review-tools/+merge/429123). Including any of these invalid fields will cause the snap to fail validation and require manual review.

A list of fields that should be stripped out:

```
compression
contact
donation
issues
source-code
website
```

I would make a pull request but I am not sure what app-builder-lib is doing when `this.isUseTemplateApp` is false, it seems to make `snap/snapcraft.yaml` instead of `meta/snap.yaml`

mmaietta:
Hi @MikeJerred , I'm honestly not familiar with the implementation of snap within electron-builder. Would be happy to accept a PR for this change.

MikeJerred:
I opened a PR: https://github.com/electron-userland/electron-builder/pull/7110
It looks like when `isUseTemplateApp` is false then the snap cli tool is used which should create the `snap.yaml` correctly, so I have only stripped the fields out when `isUseTemplateApp == true`

PR #7110, opened 2022-09-01
fix: strip extra fields out before creating snap.yaml (#7104)

MikeJerred:
I have also removed the line `TMPDIR: "$XDG_RUNTIME_DIR"` as it causes problems with multiple instances (see the chromium bug: https://bugs.launchpad.net/ubuntu/+source/chromium-browser/+bug/1838508)


packages/app-builder-lib/src/targets/snap.ts
```diff
@@ -214,6 +213,16 @@ export default class SnapTarget extends Target {
...
 
+    if (this.isUseTemplateApp) {
+      // remove fields that are valid in snapcraft.yaml, but not snap.yaml
+      delete snap.compression;
+      delete snap.contact;
+      delete snap.donation;
+      delete snap.issues;
+      delete snap['source-code'];
+      delete snap.website;
+    }
```
mmaietta, line 224: Just so I understand this change further, what's the difference in usage between snapcraft.yaml and snap.yaml?

mmaietta: Also, to be more clear about intent, thoughts on adjusting the logic to:
```
// remove fields that are valid in snapcraft.yaml, but not snap.yaml
const forbiddenSnapFields = ['compression', 'donation', 'issues', 'source-code', 'website']
for (field of forbiddenSnapFields) {
   delete snap[field]
}
```
Could be `of` or `in`, I always forget the difference when reviewing PRs on-the-fly

MikeJerred: AFAIK the `meta/snap.yaml` file is present in the built apps directory before it gets run through squashfs, and so is present in the final snap package when it is released; `snapcraft.yaml` seems to be used to tell the snap cli tool how to build the snap. The two files are very similar, but don't have exactly the same schema. It looks like `electron-builder` was removing the `parts` field, but not some of the newer fields that were added more recently to snap.

mmaietta: Additional note, we have this at Line 178
```
if (this.isUseTemplateApp) {
      delete snap.parts
}
```
Should we incorporate your additional logic there or is it essential it be placed down here?

MikeJerred: For reference, here are the documented schemas:
- https://snapcraft.io/docs/snap-format
- https://snapcraft.io/docs/snapcraft-yaml-reference

MikeJerred: > Additional note, we have this at Line 178
> 
> ```
> if (this.isUseTemplateApp) {
>       delete snap.parts
> }
> ```
> 
> Should we incorporate your additional logic there or is it essential it be placed down here?

I have stripped right before we write the file: this is necessary for example with the `compression` field, as `--compression` is only pushed to `args` if `snap.compression` is truthy, so deleting the field early would cause incorrect behavior. I could move the delete of parts down into the if block I just added though.

mmaietta:
Looks like test snapshots need to be updated using `UPDATE_SNAPSHOT=true pnpm test-linux`

Any thoughts on why this needs to be added to the test snapshot now? I don't recognize the code. Maybe that's why `snap.parts` was earlier in the logic 🤷 
```
+       "override-build": "mkdir -p \"$SNAPCRAFT_PART_INSTALL/data-dir/themes\" mkdir -p \"$SNAPCRAFT_PART_INSTALL/data-dir/icons\" mkdir -p \"$SNAPCRAFT_PART_INSTALL/data-dir/sounds\" mkdir $SNAPCRAFT_PART_INSTALL/gnome-platform
    + ",
    +       "plugin": "nil",
    +     },
    +     "launch-scripts": Object {
    +       "plugin": "dump",
    +       "source": "scripts",
    +     },
    +   },
```

MikeJerred:
> Looks like test snapshots need to be updated using `UPDATE_SNAPSHOT=true pnpm test-linux`
> 
> Any thoughts on why this needs to be added to the test snapshot now? I don't recognize the code. Maybe that's why `snap.parts` was earlier in the logic 🤷
> 
> ```
> +       "override-build": "mkdir -p \"$SNAPCRAFT_PART_INSTALL/data-dir/themes\" mkdir -p \"$SNAPCRAFT_PART_INSTALL/data-dir/icons\" mkdir -p \"$SNAPCRAFT_PART_INSTALL/data-dir/sounds\" mkdir $SNAPCRAFT_PART_INSTALL/gnome-platform
>     + ",
>     +       "plugin": "nil",
>     +     },
>     +     "launch-scripts": Object {
>     +       "plugin": "dump",
>     +       "source": "scripts",
>     +     },
>     +   },
> ```

Looks like `effectiveOptionComputed` should be taking the finished `snap` file, so I need to move my code before it.

MikeJerred:
I tried to run this but I get an error:
```
. prepare: fatal: detected dubious ownership in repository at '/project'
. prepare: To add an exception for this directory, call:
. prepare:      git config --global --add safe.directory /project
```
(running `git config --global --add safe.directory /project` makes no difference)

mmaietta:
Never seen that error before. Anyways, I pulled your PR locally and made some edits to add a unit test for template apps. :) Please take a look when you have a chance.
https://github.com/MikeJerred/electron-builder/pull/1

MikeJerred:
@mmaietta I've updated this PR, but cannot generate the snapshots or check that the tests work

mmaietta:
> @mmaietta I've updated this PR, but cannot generate the snapshots or check that the tests work

It seems that not all the changes from my PR were incorporated into your PR. Namely, the snapshot updates and the reinsert of `delete snap.parts`
https://github.com/MikeJerred/electron-builder/pull/1/files#diff-0354454e32db9abdc255b34f4188224fd0ea698bfa76355aa99b43ce64c3beb0

MikeJerred:
> > @mmaietta I've updated this PR, but cannot generate the snapshots or check that the tests work
> 
> It seems that not all the changes from my PR were incorporated into your PR. Namely, the snapshot updates and the reinsert of `delete snap.parts` https://github.com/MikeJerred/electron-builder/pull/1/files#diff-0354454e32db9abdc255b34f4188224fd0ea698bfa76355aa99b43ce64c3beb0

I changed the test to check for removal of snap.compression so I think the snapshot will need updating for that? Also I'm pretty sure the removal of snap.parts doesn't need to be moved up.

mmaietta:
>Also I'm pretty sure the removal of snap.parts doesn't need to be moved up.

It does. Otherwise current snapshots break.

I'm able to run these tests with just `TEST_FILES=snapTest UPDATE_SNAPSHOT=true pnpm test` as opposed to having to use docker. Can you try that?

MikeJerred:
> > Also I'm pretty sure the removal of snap.parts doesn't need to be moved up.
> 
> It does. Otherwise current snapshots break.
> 
> I'm able to run these tests with just `TEST_FILES=snapTest UPDATE_SNAPSHOT=true pnpm test` as opposed to having to use docker. Can you try that?

Awesome! On Windows it skipped the test but using WSL this works :)
I've updated this PR with the new snapshots and fixed the existing compression test which was checking for the existence of `snap.compression` which has now been removed. Test suite is now passing.

test/src/linux/snapTest.ts
```diff
@@ -288,9 +288,35 @@ test.ifDevOrLinuxCi(
     },
     effectiveOptionComputed: async ({ snap, args }) => {
       expect(snap).toMatchSnapshot()
-      expect(snap.compression).toEqual("xz")
+      expect(snap.compression).toBeUndefined()
```
mmaietta, line 291: I don't think this test should be changing. This test "compression option" is making sure that compression is properly applied, so it shouldn't be undefined.

mmaietta: I think the easy fix to retain this logic is to add to the config, `useTemplateApp: false`

MikeJerred: 👍 I've made this change now

test/snapshots/linux/snapTest.js.snap
```diff
@@ -342,7 +340,144 @@ Object {
   "confinement": "strict",
   "description": "Test Application (test quite “ #378)",
   "grade": "stable",
+  "icon": "snap/gui/icon.png",
   "name": "sep",
+  "parts": Object {
```
mmaietta, line 345: This is why `delete snap.parts` needs to be higher up in the logic flow. This addition shouldn't be here.

MikeJerred: The reason I saw `snap.parts` deletion doesn't need to move up is because if you look at everywhere the `snap` variable is used it is only these lines:

```typescript
snap.icon = "snap/gui/icon.png"
...
const desktopFile = path.join(snapMetaDir, "gui", `${snap.name}.desktop`)
...
if (this.isElectronVersionGreaterOrEqualThan("5.0.0") && !isBrowserSandboxAllowed(snap)) {
...
if (snap.compression != null) {
  args.push("--compression", snap.compression)
}
...
if (packager.packagerOptions.effectiveOptionComputed != null && (await packager.packagerOptions.effectiveOptionComputed({ snap, desktopFile, args }))) {
  return
}
```

The code that strips out the snap fields is now above that bottom one (that is what was causing all the tests to fail previously), and none of the others look at `snap.parts` at all (`isBrowserSandboxAllowed` only looks at `snap.plugs`).

Nevertheless I just added that `snap.parts` removal code back in and re-ran `TEST_FILES=snapTest UPDATE_SNAPSHOT=true pnpm test` and the snapshots do not change.

mmaietta: Ah right, we explicitly set `useTemplateApp: false` to account for the no template flow w/ compression

PR #7113, opened 2022-09-02
docs: declare node-linker when using Yarn 3

ouliuquan:
Because of yarn 3 use PnP by default, but electron-builder still need node-modules.
So add configuration in the  `.yarnrc.yaml`:
```
nodeLinker: "node-modules"
```
will declare to use node-modules instead of PnP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,14 @@ public-hoist-pattern=*
 shamefully-hoist=true
 ```
 
+### Note for Yarn 3
+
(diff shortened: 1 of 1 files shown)


ouliuquan:
#https://github.com/yarnpkg/berry/issues/4804

mmaietta:
Thank you!

Can you please also add that info here so it is shown on the docs website?
https://github.com/electron-userland/electron-builder/blob/master/docs/index.md

ouliuquan:
> Thank you!
> 
> Can you please also add that info here so it is shown on the docs website? https://github.com/electron-userland/electron-builder/blob/master/docs/index.md

Yeah, I had been create a PR for website here. https://github.com/electron-userland/electron-builder/pull/7233

PR #7120, opened 2022-09-03
docs: fix typo

cjeonguk:
FIx typo of `docs/generated/NsisOptions.md`.

### DIFF
--- a/.changeset/smart-countries-decide.md
+++ b/.changeset/smart-countries-decide.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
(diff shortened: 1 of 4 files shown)


mmaietta:
This is an auto-generated file (in `generated` folder)
You'll need to update the jsdoc code here and then regen the docs via `pnpm generate-all`
https://github.com/electron-userland/electron-builder/blob/5eb002b806d5bf439992bb0a9c5d7504ccc6d7dc/packages/app-builder-lib/src/targets/nsis/nsisOptions.ts#L135

Issue #6987, opened 2022-07-06
default unwelcome page with electron builder nsis can't be changed or removed


moulinierf:
* **Electron-Builder Version**: 
22.14.5

* **Node Version**: 
16.15.1

* **Electron Version**:
18.3.0
* **Electron Type (current, beta, nightly)**:

current

* **Target**: 
win nsis ia32

based on this PR https://github.com/electron-userland/electron-builder/pull/1173

Hi, based on this feature, it seems there is an issue to interact with the unwelcome page.
by default when we use electron builder with nsis, we have now during the uninstallation , the unwelcome page and the unfinish page like you described bellow. And it's cool
but we didn't interact with them like this

!define MUI_PAGE_CUSTOMFUNCTION_PRE un.custom_pre_page
!define MUI_PAGE_CUSTOMFUNCTION_LEAVE un.custom_leave_page
!define MUI_WELCOMEPAGE_TITLE_3LINES
!define MUI_WELCOMEPAGE_TITLE 'unwelcome_title'
!define MUI_WELCOMEPAGE_TEXT 'unwelcome_text'
!insertmacro MUI_UNPAGE_WELCOME

so when we have are in a language where the default text is too long we don't have the 3 lines.
and we can't change the text. see the screenshot.

I've test many ways to solve it, manage anothe unpage welcome and it works for the othe page but the default page that you added is never removed.
I tried to manipulate the window item to hide it, try to click on the next button with a command to navigate 
 but it doesn't work. And I verified that handles are correct.
 
 so what it the solution in this context to interact with your unwelcome page or to remove it?
 thanks in advance

![image](https://user-images.githubusercontent.com/39991826/177599759-873125fd-bb63-409c-b8a2-7983dcf8a52f.png)

moulinierf:
nobody can help on this topic?

moulinierf:
after searched deeper, 
I can see in the electron-builder code (after cloning the repo) that the assistedInstaller.nsh
managed the uninstaller as is
![image](https://user-images.githubusercontent.com/39991826/189417425-42f7278b-c0f1-49b6-80a2-cc7238046360.png)
the !insertmacro MUI_UNPAGE_WELCOME is added without condition whereas the customWelcomePage is added with a condition.
I guess it's the root cause of my problem.
If I try to add a custom unwelcome page like this

`!define MUI_PAGE_CUSTOMFUNCTION_SHOW un.ModifyUnWelcome
[...]

PR #7141, opened 2022-09-16
feat: add nsis option to remove the default uninstall welcome page

moulinierf:
<subject>add nsis option to remove the default uninstall welcome page</subject> <BLANK LINE>
<Scope>nsis</scope>
<body>until now we can't modify the default unwelcome page, this feature adds the option removeDefaultUninstallWelcomePage to give the possibility to remove from the nsis option this default page in order to custom it in the installer.nsh </body> <BLANK LINE>
<footer>BREAKING CHANGE: the nsis option removeDefaultUninstallWelcomePage is used in the assistedInstaller.nsh to decide if we must add the MUI_UNPAGE_WELCOME or not. based on the issue based on this issue https://github.com/electron-userland/electron-builder/issues/6987 </footer>

### DIFF
--- a/.changeset/tasty-pianos-jump.md
+++ b/.changeset/tasty-pianos-jump.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: add nsis option to remove the default uninstall welcome page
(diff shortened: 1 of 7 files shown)


mmaietta:
LGTM. Can you clarify how this is a breaking change though? Seems with the default as false, the current functionality is being preserved

moulinierf:
Hello Mike,
You're right the value false preserves the current behavior.
I thought that the add of a new nsis option was big change, but it's my first contribution on the project.
if it's ok for you go ahead.
thanks

mmaietta:
Thanks for the contribution! :)

PR #7163, opened 2022-09-28
feat: update note for pnpm

romadryud:
It will more clear to developer to see that note in `electron-builder` docs. This line was taken from pnpm [docs](https://pnpm.io/npmrc#public-hoist-pattern)


romadryud:
@develar Could you please take a look on this tiny one?

romadryud:
@mmaietta Done ✅

PR #7174, opened 2022-10-04
fix: Allow non-semver versions in getVersionInWeirdWindowsForm

### LINKED ISSUE #7173: `extraMetadata: version` must be semver on Windows, but can be whatever on macOS and Linux
* **Electron-Builder Version**: 23.3.3

* **Node Version**: v16.14.2

* **Electron Version**: 19.0.13
* **Electron Type (current, beta, nightly)**: ???

* **Target**: Issue exists on Windows. We build for macOS and Linux also.

## TLDR:

The following way of injecting our app's version works fine on macOS and Linux, but fails with a semver error on Windows:

`distribution.js`:
```js
extraMetadata: {
  version: '2022.6'
},
```

On Windows this results in:
```
[12:04:15] 'builder-win' errored after 29 s
[12:04:15] TypeError: Invalid Version: 2022.6
    at new SemVer (C:\Users\build\mullvadvpn-app\gui\node_modules\semver\classes\semver.js:38:13)
    at AppInfo.getVersionInWeirdWindowsForm (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\appInfo.ts:78:27)
    at WinPackager.signAndEditResources (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\winPackager.ts:288:46)
    at C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\winPackager.ts:378:21
    at bound (node:domain:421:15)
    at runBound (node:domain:432:12)
From previous event:
    at bound (node:domain:421:15)
    at MappingPromiseArray.runBound (node:domain:432:12)
    at processImmediate (node:internal/timers:464:21)
    at process.topLevelDomainCallback (node:domain:152:15)
    at process.callbackTrampoline (node:internal/async_hooks:128:24)
From previous event:
    at WinPackager.signApp (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\winPackager.ts:376:27)
    at WinPackager.doSignAfterPack (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\platformPackager.ts:329:16)
    at WinPackager.doPack (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\platformPackager.ts:314:7)
    at WinPackager.pack (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\platformPackager.ts:136:5)
    at Packager.doBuild (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\packager.ts:441:9)
    at Object.executeFinally (C:\Users\build\mullvadvpn-app\gui\node_modules\builder-util\src\promise.ts:12:14)
    at Packager._build (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\packager.ts:376:31)
    at Packager.build (C:\Users\build\mullvadvpn-app\gui\node_modules\app-builder-lib\src\packager.ts:337:12)
    at Object.executeFinally (C:\Users\build\mullvadvpn-app\gui\node_modules\builder-util\src\promise.ts:12:14)
[12:04:15] 'pack-win' errored after 57 s
```
### Expected

I expect the version format requirements to 1) be the same everywhere and 2) not need to be semver formatted.

## Longer explanation

The version format of our app is `$year.$number`. For example `2022.6`. The `version` field in `package.json` requires this field to be semver. So we have always appended a `.0` to our version to make it semver compatible. We have a hacky build script that injects this into `package.json` when building. We now wanted to improve on this hack. We found a solution that works sooo nicely on Linux and macOS, but fails on Windows (see TLDR).

In our actual code, the version string in `distribution.js` is not hardcoded, it looks like:
```js
version: execSync('cargo run --bin mullvad-version', { encoding: 'utf-8' }).trim()
```
To use existing Rust code to compute the desired product version. This includes git hashes etc for development builds. Normal versions are `2022.6-beta1`, `2022.7-dev-afafaf` (`-dev-$git_hash[0..6]`), but that is besides the point here.

We don't see any reason why **applications** must have versions compatible with semver. Semver is mostly for libraries IMO, where there is API compatibility to think about.

The relevant code that crashes is here: https://github.com/electron-userland/electron-builder/blob/8166267d487cd26b154e28cf60d89102a487a353/packages/app-builder-lib/src/appInfo.ts#L79. We don't think this should require the app being semver at all. Can this maybe be replaced with opportunistically trying semver, and if it fails, it fills in some reasonable defaults? The code does not care about more than just `major`, `minor` and `patch`. So it could just split the version string at `.` and iterate, assigning the first part to `major`, the second to `minor` and so fourth, but default to version `0` when the string split iterator comes to an end.

## Why allowing non-semver?

So `package.json` requires the version to be correct semver. Why should `extraMetadata: version` not when what it does is override the `version` field in `package.json`, they should have the same requirements? I'd argue no! Because `package.json` is for both libraries and applications, but most commonly for libraries. And for libraries it makes a lot of sense to version with semver. npmjs.com would be very messy if packages did not adhere to one versioning scheme. However, `electron-builder` is for applications, and not libraries. Applications need not be semver, because they don't have a public API to provide stability guarantees for)

faern:
Fixes #7173

This new code allows the user to not the lower version parts, and have them default to zero instead. With this change the resulting weird-windows-version will be:

```
# this.version -> windows-weird-version (buildNumber = 0)
'1.2.3' => '1.2.3.0'
'1.2' => '1.2.0.0'
'9.2-beta' => '9.2.0.0'
'45' => '45.0.0.0'
'2022.6-beta2-dev-afafaf' => '2022.6.0.0'
```

### DIFF
--- a/.changeset/chilled-queens-think.md
+++ b/.changeset/chilled-queens-think.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+Allow non-semver version formats on Windows
--- a/packages/app-builder-lib/src/appInfo.ts
+++ b/packages/app-builder-lib/src/appInfo.ts
@@ -1,5 +1,5 @@
 import { isEmptyOrSpaces, log } from "builder-util"
-import { prerelease, SemVer } from "semver"
+import { prerelease } from "semver"
 import { PlatformSpecificBuildOptions } from "./options/PlatformSpecificBuildOptions"
 import { Packager } from "./packager"
 import { expandMacro } from "./util/macroExpander"
@@ -76,13 +76,25 @@ export class AppInfo {
   }
 
   getVersionInWeirdWindowsForm(isSetBuildNumber = true): string {
-    const parsedVersion = new SemVer(this.version)
+    const [major, maybe_minor, maybe_patch] = this.version.split(".").map(versionPart => parseInt(versionPart))
+    // The major component must be present. Here it can be either NaN or undefined, which
+    // both returns true from isNaN.
+    if (isNaN(major)) {
+      throw new Error(`Invalid major number in: ${this.version}`)
+    }
+    // Allow missing version parts. Minor and patch can be left out and default to zero
+    const minor = maybe_minor ?? 0
+    const patch = maybe_patch ?? 0
+    // ... but reject non-integer version parts. '1.a' is not going to fly
+    if (isNaN(minor) || isNaN(patch)) {
+      throw new Error(`Invalid minor or patch number in: ${this.version}`)
+    }
     // https://github.com/electron-userland/electron-builder/issues/2635#issuecomment-371792272
     let buildNumber = isSetBuildNumber ? this.buildNumber : null
     if (buildNumber == null || !/^\d+$/.test(buildNumber)) {
       buildNumber = "0"
     }
-    return `${parsedVersion.major}.${parsedVersion.minor}.${parsedVersion.patch}.${buildNumber}`
+    return `${major}.${minor}.${patch}.${buildNumber}`
   }
 
   private get notNullDevMetadata() {


mmaietta:
Hi there! I'm not sure I understand. Why would something like `3.9.2-beta` not want to include the `-beta` in the metadata version? Or is it already parsing out the `-beta` with the current `SemVer` logic?

Side note, this might be a breaking change, right? (not opposed, just would need to coordinate with other "alpha" release submissions)

faern:
> Hi there! I'm not sure I understand. Why would something like 3.9.2-beta not want to include the -beta in the metadata version? Or is it already parsing out the -beta with the current SemVer logic?

Yes, semver already removes this. Or well, of course semver understands it's a beta. But the old code here just used the `major`, `minor` and `patch` fields. So it was already stripping out any `-beta`, `-dev` `-pre` etc tags. The "windows weird format" is just four integers. It does not allow for other data than those four integers.

If you set `extraMetadata: verion` to `3.9.2-beta` the result would be the same before and after this PR: Windows version metadata field set to `3.9.2.0`. The difference is just that now it's also allowed to give the version as `3.9-beta` and have that gracefully translate to `3.9.0.0`

And this will not affects the version in most parts of the app. Only what's filled into the "weird version field" on Windows, not other parts.

faern:
> Side note, this might be a breaking change, right?

I don't think so. I can't come up with any version that was previously allowed that is now forbidden, or that translates to something different than before. The version `x.y.z[whatever]` will still translate to `x.y.z.buildNumber`. The only difference is that you are now allowed to omit `z`, `z`+`y` or `z`+`y`+`x` and have those default to zero instead of throwing an exception.

If implemented as I intended it, this should just allow for more versions to be valid, not change behavior any currently allowed version.

Sidenote: We might want to still require the `major` and not have that default to zero. Making it a tiny bit stricter, but so we at least have some number to give to Windows... I'll look into that.

mmaietta:
Please also create a changeset file for this change (`pnpm changeset`)

faern:
I managed to generate a changeset, and I added it to the same commit and force-pushed. I hope this is OK now :+1:

mmaietta:
Can you please rebase/merge latest `master`?

Unrelated to your PR, unit tests broke on master due to an npm package changing file contents (we snapshot the unpacked node_modules dir in some tests). I just pushed the changes to fix it

faern:
Speaking of tests. I could add tests to `getVersionInWeirdWindowsForm` or somewhere higher to ensure that the non-semver versions are now accepted on all platforms, if you want me to. But I'm not sure where such a test would go, so some directions would be nice.

mmaietta:
Oooo I like your idea.

My best route would be to extract it to a function outside of the AppInfo class (just an exported function), we'd then have a `utilsTest.ts` that would explicitly test for various entries of semver.
Similar to this `smarten` being an exported function:
https://github.com/electron-userland/electron-builder/blob/8166267d487cd26b154e28cf60d89102a487a353/packages/app-builder-lib/src/appInfo.ts#L10-L20

We can rename `urlUtilsTest.ts` to just be `utilsTest.ts` and then you can write very similar tests for handling `getVersionInWeirdWindowsForm`

Thoughts?

faern:
To not miss the release, maybe we should merge as is and consider tests later? Because I will probably not have the time to do that right now 🙈

mmaietta:
https://github.com/electron-userland/electron-builder/releases/tag/v24.0.0-alpha.1
🙂

PR #7188, opened 2022-10-12
docs: Warn users not to disable zip for macos

taratatach:
This warning was added in #6496 but removed by the very next release. It is essential to warn users about this as disabling the `zip` target on macos will prevent apps to auto-update.


tinytb (approved):

mmaietta:
Thanks for the contribution! That file is actually auto-generated from 
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/options/macOptions.ts#L16

Can you please update the comment directly, then regenerate the docs and schema via `pnpm generate-all`?

taratatach:
Done. Good thing this is auto-generated as I'd removed a closing `<li>` tag :sweat_smile:

tinytb (approved):

Issue #7175, opened 2022-10-04
Rebuilding native dependencies don't work for Electron 20 and up


oalfroukh:
* **Electron-Builder Version**:  23.6.0

* **Node Version**: 14.18.1 32-bit

* **Electron Version**:20.3.0 and up
* **Electron Type (current, beta, nightly)**:current

* **Target**: nsis

- Rebuilding native dependencies don't work for Electron 20 and up, I added `sqlite3 v4.2.0` as dependencies to prove the issue.
- The issue doesn't exist in Electron `v19.x`.
- Here is my package.json file:

```js
{
  "name": "electron-quick-start",
  "productName": "electron-quick-start",
  "description": "My Electron application description",
  "keywords": [],
  "main": "./main.js",
  "version": "1.0.0",
  "author": "oalfroukh",
  "scripts": {
    "start": "electron .",
    "dist": "electron-builder"
  },
  "build": {
    "buildVersion": "1.0.0.0",
    "compression": "normal",
    "win": {
      "target": "nsis",
      "verifyUpdateCodeSignature": false
    },
    "nsis": {
      "oneClick": false,
      "perMachine": false,
      "allowToChangeInstallationDirectory": true,
      "differentialPackage": false
    },
    "squirrelWindows": {
      "msi": "true"
    },
    "msi": {
      "warningsAsErrors": false
    },
    "appx": {
      "publisherDisplayName": "test"
    },
    "electronVersion": "20.3.0",
    "buildDependenciesFromSource": true,
    "nodeGypRebuild": false,
    "npmRebuild": true,
    "extends": null
  },
  "dependencies": {
    "sqlite3": "4.2.0"
  },
  "devDependencies": {
    "electron": "20.3.0",
    "electron-builder": "23.6.0"
  }
}
```

- Here is the error for the issue:
```cmd
$ npm run dist

> electron-quick-start@1.0.0 dist C:\Users\oalfroukh\Desktop\Test projects\electron-quick-start
> electron-builder

  • electron-builder  version=23.6.0 os=10.0.19044
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist\builder-effective-config.yaml
  • rebuilding native dependencies  dependencies=sqlite3@4.2.0 platform=win32 arch=ia32
  ⨯ cannot execute  cause=exit status 1
                    out=
[...]

PR #7196, opened 2022-10-19
fix: Migrate to electron-rebuild for handling native dependencies

mmaietta:
Fixes: #7175


ludvigintel:
This fix doesn't work.. do you have any updates?
I still get the /std++17 compiler flag error

mmaietta:
Please don't copy/cross-post across from https://github.com/electron-userland/electron-builder/issues/7175#issuecomment-1296431796

Issue #7210, opened 2022-10-24
Question: Is there value putting together a PR for non-proprietary FFmpeg download?


baparham:
Looking for some feedback if it's worth putting together a PR to support downloading the non-proprietary FFmpeg binary that the Electron org releases with each version of Electron.

This can be accomplished with some basic `afterPack` scripting to find the asset, and replace the current FFmpeg library in the unpacked folder, but I'm curious if this would be valuable as a direct option that just handles this in the standard flow.

Thanks for the feedback!

mmaietta:
I not familiar with FFmpeg binary or its usage (or why it needs to be replaced). Could you elaborate on what that provides?

baparham:
Electron ships by default with an ffmpeg.dll (on windows at least...libffmpeg.so on linux) to support [audio/video in the browser window](https://www.chromium.org/audio-video/), and from what I can tell, you can't just remove it if you don't need it (not without compiling electron yourself).

The default dll contains proprietary h264 codecs, which may cause issues for most commercial apps shipping electron, so the Electron team releases another version of ffmpeg.dll that does not contain proprietary codecs. These are available alongside the normal assets for each release and can serve as drop in replacements for the default dll that ships with normal electron release assets.

Some related commentary on the subject: https://github.com/electron/electron/issues/13445

AFAIK the h.264 codec is really the only one in the default ffmpeg library that could cause issues, and of course I'm not a lawyer, but it seems like using the "non proprietary" ffmpeg build that the electron team releases would mitigate some of the risk for developers of electron apps.

Making it easier for electron-builder users to incorporate this into their flows seems like a helpful strategy, since helping the ecosystem comply with licensing and patent obligations likely helps us all in the long run.

mmaietta:
Very interesting! I did not know of that.

Re: the FFmpeg, is it already being downloaded with the current electron-builder implementation? IIRC, it's currently being downloaded using an upstream GoLang library app-builder-bin.

baparham:
I am actually not sure if the current `electron-builder`/`app-builder-lib` implementation downloads the target electron asset (I believe it does) or just reuses what has already been downloaded/installed into node_modules. Since you can cross-"compile" with electron-builder, I suspect it downloads the electron dist asset directly itself.

[...]

PR #7247, opened 2022-11-08
feat: add downloadAlternateFFmpeg option

baparham:
This is a possible solution to the question proposed in #7210 addressing the need/desire to have an automatic FFmpeg download as part of downloading the standard electron distributable.

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -6416,6 +6416,11 @@
       ],
       "description": "macOS DMG options."
     },
+    "downloadAlternateFFmpeg": {
+      "default": false,
+      "description": "Whether to download the alternate FFmpeg library from Electron's release assets and replace the default FFmpeg library prior to signing",
+      "type": "boolean"
+    },
     "electronBranding": {
       "$ref": "#/definitions/ElectronBrandingOptions",
       "description": "The branding used by Electron's distributables. This is needed if a fork has modified Electron's BRANDING.json file."
--- a/packages/app-builder-lib/src/configuration.ts
+++ b/packages/app-builder-lib/src/configuration.ts
@@ -142,6 +142,11 @@ export interface Configuration extends PlatformSpecificBuildOptions {
   readonly buildVersion?: string | null
 
   /**
+   * Whether to download the alternate FFmpeg library from Electron's release assets and replace the default FFmpeg library prior to signing
+   */
+  readonly downloadAlternateFFmpeg?: boolean
+
+  /**
    * Whether to use [electron-compile](http://github.com/electron/electron-compile) to compile app. Defaults to `true` if `electron-compile` in the dependencies. And `false` if in the `devDependencies` or doesn't specified.
    */
   readonly electronCompile?: boolean
--- a/packages/app-builder-lib/src/electron/ElectronFramework.ts
+++ b/packages/app-builder-lib/src/electron/ElectronFramework.ts
@@ -1,5 +1,5 @@
 import BluebirdPromise from "bluebird-lst"
-import { asArray, executeAppBuilder, log } from "builder-util"
+import { asArray, executeAppBuilder, isEmptyOrSpaces, log } from "builder-util"
 import { CONCURRENCY, copyDir, DO_NOT_USE_HARD_LINKS, statOrNull, unlinkIfExists } from "builder-util/out/fs"
 import { emptyDir, readdir, rename } from "fs-extra"
 import { Lazy } from "lazy-val"
@@ -181,6 +181,18 @@ async function unpack(prepareOptions: PrepareApplicationStageDirectoryOptions, o
       return
     }
     await executeAppBuilder(["unpack-electron", "--configuration", JSON.stringify([options]), "--output", appOutDir, "--distMacOsAppName", distMacOsAppName])
+    if (packager.config.downloadAlternateFFmpeg) {
+      log.info("downloading and unpacking alternate FFmpeg")
+      const ffmpegOptions = Object.assign({}, options)
+      ffmpegOptions.customFilename = `ffmpeg-v${options.version}-${options.platform}-${options.arch}.zip`
+      let ffmpegLocation = path.join(appOutDir, "..")
+      if (!isEmptyOrSpaces(process.env.ELECTRON_BUILDER_CACHE)) {
+        ffmpegLocation = process.env.ELECTRON_BUILDER_CACHE
+      }
+      ffmpegOptions.cache = ffmpegLocation
+      await executeAppBuilder(["download-electron", "--configuration", JSON.stringify([ffmpegOptions])])
+      await executeAppBuilder(["unzip", "--input", path.join(ffmpegLocation, ffmpegOptions.customFilename), "--output", appOutDir])
+    }
   } else {
     isFullCleanup = true
     const source = packager.getElectronSrcDir(dist)
--- a/.changeset/silly-carpets-attend.md
+++ b/.changeset/silly-carpets-attend.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: Adding new downloadAlternateFFmpeg option to download non-proprietary ffmpeg library


packages/app-builder-lib/src/configuration.ts
```diff
@@ -141,6 +141,11 @@ export interface Configuration extends PlatformSpecificBuildOptions {
    */
   readonly buildVersion?: string | null
 
+  /**
+   * Whether to download the alternate FFmpeg library from Electron's release assets and replace the default FFmpeg library prior to signing
+   */
+  readonly downloadAlternateFFmpeg?: boolean
```
mmaietta, line 147: What are your thoughts on renaming this to: `injectElectronFFmpeg` or something similar? `downloadAlternate` seems odd to me because it can also use `ELECTRON_BUILDER_CACHE` and this seems to be more Electron-specific than just a generic alternate library.

packages/app-builder-lib/src/electron/ElectronFramework.ts
```diff
@@ -181,6 +181,18 @@ async function unpack(prepareOptions: PrepareApplicationStageDirectoryOptions, o
       return
     }
     await executeAppBuilder(["unpack-electron", "--configuration", JSON.stringify([options]), "--output", appOutDir, "--distMacOsAppName", distMacOsAppName])
+    if (packager.config.downloadAlternateFFmpeg) {
+      log.info("downloading and unpacking alternate FFmpeg")
+      const ffmpegOptions = Object.assign({}, options)
+      ffmpegOptions.customFilename = `ffmpeg-v${options.version}-${options.platform}-${options.arch}.zip`
```
mmaietta, line 187: Looking at the logic above for `const zipFile = `electron-v${options.version}-${platformName}-${options.arch}.zip``, I think we can use `platformName` instead of `options.platform`? That way we're labeling for sure the same download file?

packages/app-builder-lib/src/electron/ElectronFramework.ts
```diff
@@ -181,6 +181,18 @@ async function unpack(prepareOptions: PrepareApplicationStageDirectoryOptions, o
...
+      let ffmpegLocation = path.join(appOutDir, "..")
```
mmaietta, line 188: Would `path.dirname(appOutDir)` work better here? Either that or `path.resolve`?

packages/app-builder-lib/scheme.json
```diff
@@ -6416,6 +6416,11 @@
       ],
       "description": "macOS DMG options."
     },
+    "downloadAlternateFFmpeg": {
```
mmaietta, line 6419: For the schema, we'll also need to run `pnpm generate-all` to update the docs+schema together

baparham: aha, ok, thanks for the tip. I'll run that and submit another commit with the updates.

packages/app-builder-lib/src/configuration.ts
baparham: yeah, a name for this is kind of hard. It's also not just `downloading` it is really `using` the alternate ffmpeg library. I was also hesitant to call it `nonProprietary` since I don't really have a guarantee from electron that it is non proprietary...we as consumers are supposed to be the ones checking this sort of thing. Granted, it's built with the `proprietary_codecs=false` flag from chromium.

Anyway, all Electron's come with this ffmpeg library, its just that we want to _use_ an alternative one that supposedly has only non proprietary codecs in it. With that, I'd say the best name maybe is `useNonProprietaryFFmpeg` ... that seems to be as descriptive and accurate as it gets.

packages/app-builder-lib/src/electron/ElectronFramework.ts
baparham: the naming logic here mirrors the `unpack-electron` options (since the full `options` object is just passed to the app-builder-bin call), since this case is only handled when we are downloading the default electron distributable (i.e. `dist === null`). The zipFile reference above is referencing something the user can manually override and specify custom platforms for, and we won't run this ffmpeg option in such a case. The user is expected to have the dlls in that electron distributable the way they want them.

packages/app-builder-lib/src/electron/ElectronFramework.ts
baparham: yeah, probably best to just stick to `path.dirname(appOutDir)` to avoid the `..`. I'll fix.

packages/app-builder-lib/src/electron/ElectronFramework.ts
```diff
@@ -181,6 +181,18 @@ async function unpack(prepareOptions: PrepareApplicationStageDirectoryOptions, o
...
+      if (!isEmptyOrSpaces(process.env.ELECTRON_BUILDER_CACHE)) {
+        ffmpegLocation = process.env.ELECTRON_BUILDER_CACHE
+      }
+      ffmpegOptions.cache = ffmpegLocation
```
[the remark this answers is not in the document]
baparham: Hmm, I was basing this off of the app-builder arguments, where it expects the `cache` property in the config to be a directory.

https://github.com/develar/app-builder/blob/4e2aa6a12e2bc3d31ec0d01d661fb3a4d65248ff/pkg/electron/electronDownloader.go#L20

and

https://github.com/develar/app-builder/blob/4e2aa6a12e2bc3d31ec0d01d661fb3a4d65248ff/pkg/electron/electronDownloader.go#L62

However, I mistakenly read the `ELECTRON_CACHE` as `ELECTRON_BUILDER_CACHE` and defaulted to that if it were explicitly specified other than the default to take advantage of caching this ffmpeg download between runs if users had specified the ELECTRON_CACHE folder.

I was really hoping to use a `getCacheFolder` method that I could call and not have to specify any specific directory, since we a) probably want to keep this download cached like we do electron distributables and b) we need to know where it ends up after downloading so that we can unzip it.

Perhaps the caching is a premature optimization and it's easiest and cleanest to just keep it up in `path.dirname(appOutDir)` for now without the caching logic, since this zip is ever only around 1mb.

baparham (commented):
Thanks for the feedback! I've answered your questions though it seems like there's still some more discussion needed before I push anything back up to the branch, so I'll hold off until I hear your thoughts.

packages/app-builder-lib/src/electron/ElectronFramework.ts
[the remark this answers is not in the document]
mmaietta: Gotcha gotcha. Yes, great callout as well, I think the caching is an unnecessary optimization for now when it's that is only 1mb.

Just to double check, is this working in your personal/test project? 
Have you given any thought on how we could integration test this? I'll continue thinking from my end. Is there a way for us to detect whether a new `FFmpeg` was downloaded or not? As we do have a post-packaging hook for verifying app contents.

baparham:
@mmaietta are you waiting on me to do anything else?

baparham:
> @mmaietta are you waiting on me to do anything else?

ah, sorry, I see some things I said I would fix. My mistake.

calisven:
> > @mmaietta are you waiting on me to do anything else?
> 
> ah, sorry, I see some things I said I would fix. My mistake.

@baparham Hello, is there any update on when this will be released? Appreciate your and the team's great work!

baparham:
Hey, been in holiday mode and then sick. Hopefully I can get back on it and get this merged in within a few weeks

baparham:
> Hey, been in holiday mode and then sick. Hopefully I can get back on it and get this merged in within a few weeks

Still on my to-do list, but I haven't forgotten! Thanks for the patience

mmaietta:
Hey @baparham ! During some other research, I came across this npm package for handling non-proprietary ffmpeg (https://www.npmjs.com/package/electron-packager-plugin-non-proprietary-codecs-ffmpeg) 

I gave it a shot. Would you be willing to test it after merge? I don't know how to set up a test for this PR https://github.com/electron-userland/electron-builder/pull/7477

baparham:
The only test I've come up with is doing like an `objdump` or `dumpbin.exe` on windows to look for some specific strings like `x264` or `aac` but even that is subjective, since they are still defined, there are just many fewer occurrences of that string when proprietary_codecs=false, it makes for a somewhat fragile test, but it's something.

mmaietta:
Closing. Added functionality via https://github.com/electron-userland/electron-builder/pull/7477 :)

PR #7251, opened 2022-11-11
feat(nsis): add ShutdownBlockReasonCreate for blocking Win Shutdown prompt

ptol:
I want a user to undestand why they cannot shutdown a computer, because the installation process have not been completed yet.
On windows we can use ShutdownBlockReasonCreate to show a better message when shutdown is blocked instaed of default one.
Before

After

### DIFF
--- a/.changeset/olive-planets-grin.md
+++ b/.changeset/olive-planets-grin.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat(nsis): add ShutdownBlockReasonCreate for blocking Windowns Shutdown alert/prompt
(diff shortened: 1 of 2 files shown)


ptol:
I don't know much about nsis scripting so I am not sure if it's a correct way to do it. But it works.

mmaietta:
Thanks for opening this PR! I'm not sure I understand it though. What business requirements/flow is this implementing for? Could you add more details to the PR description?

ptol:
I have added more details. I hope it's more clear now.

mmaietta:
Thanks for the contribution!

Issue #7296, opened 2022-12-01
app-builder-lib - `TS2304: Cannot find name 'Node'` since v23


t3chguy:
* **Electron-Builder Version**: v23.6.0 and v24.0.0-alpha.4

* **Node Version**: 16

* **Electron Version**: Unrelated
* **Electron Type (current, beta, nightly)**: Unrelated

* **Target**: Unrelated

Errors can be seen at https://github.com/vector-im/element-desktop/actions/runs/3586425963/jobs/6035600059

Similar to that which got fixed by https://github.com/electron-userland/electron-builder/pull/6692

PR #7297, opened 2022-12-01
fix(app-builder-lib): export missing TS types

t3chguy:
Fixes https://github.com/electron-userland/electron-builder/issues/7296
Based on https://github.com/electron-userland/electron-builder/pull/6692

### DIFF
--- a/.changeset/twenty-dragons-carry.md
+++ b/.changeset/twenty-dragons-carry.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(app-builder-lib): export missing TS types
--- a/packages/app-builder-lib/src/asar/asar.ts
+++ b/packages/app-builder-lib/src/asar/asar.ts
@@ -14,7 +14,6 @@ export interface NodeIntegrity {
   blocks: Array<string>
 }
 
-/** @internal */
 export class Node {
   // we don't use Map because later it will be stringified
   files?: { [key: string]: Node }


t3chguy:
@mmaietta any chance to get a 23.x patch release containing this fix? Would rather avoid having to upgrade to alpha for it and can't stay on 22.x if I want `options.mac.notarize` to be in the schema.json

Thanks

mmaietta:
Sorry, unfortunately, the way the CI/CD automation works with `@changesets` package, I have no idea how to roll out a patch for a previous version. It's stuck in `pre` (alpha) until we promote it to a proper `v24` release. Are you not able to use `24.0.0-alpha.5`?

t3chguy:
We've pinned app-builder-lib back to 22.x as we only use it for `getElectronVersion` whilst having the rest of `electron-builder` at `23.6.0` - https://github.com/vector-im/element-desktop/pull/558 moves us wholly to `24.x-alpha` but will be blocked until alpha stabilises. 

If it is a faff then don't worry about it, initially I thought `mac.notarize` support was present in 23.6.0 and us pinning back `app-builder-lib` caused schema validation issues but in fact it was added to scheme.json in https://github.com/electron-userland/electron-builder/commit/00d0dbc2d74fbac3e9ce7a046427c1e1d9a11301 (alpha13) and the docs are at fault for  including it without stating what version it is available from (whilst listing v23.6.0 in the top right github links) https://www.electron.build/configuration/mac

PR #7314, opened 2022-12-13
feat: Add the accelerate option for s3 buckets

lbestftr:
Currently it appears there's no support for the s3 accelerated bucket option.
Attempting to supply an endpoint causes the bucket to be appended on the URL, pushing the artifacts into a subfolder.

This PR adds the option to 'accelerate' the s3 bucket, where the s3 URL will be formatted as expected by s3.

For more information on s3 bucket acceleration and the expected URL formats, see:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/transfer-acceleration.html

Let me know if I missed anything, I'm happy to throw this PR away if I am off the mark, or there's an easier way to achieve this.

### DIFF
--- a/.changeset/good-balloons-admire.md
+++ b/.changeset/good-balloons-admire.md
@@ -1,0 +1,6 @@
+---
+"app-builder-lib": minor
+"builder-util-runtime": minor
+---
+
+added the accelerate option to handle accelerated s3 buckets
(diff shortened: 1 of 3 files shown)


mmaietta:
Curious. Why not set the `endpoint` to the full s3 url and then set `bucket: ''`? Then it should render:
```
url = https://${yourActualBucket}.s3-accelerate.amazonaws.com/
```

lbestftr:
> Curious. Why not set the `endpoint` to the full s3 url and then set `bucket: ''`? Then it should render:
> 
> ```
> url = https://yourActualBucket.s3-accelerate.amazonaws.com/
> ```

Sadly I tried that and it throws `- minimum field size of 1, PutObjectInput.Bucket.`

mmaietta:
Ahh, I see. It gets pulled elsewhere via `getBucketName()` and that in turn calls app-builder-bin directly. (not sure why aws SDK wasn't used, maybe it didn't exist yet 🤷 )

Okay, this seems fine to me. I can't see another way to decrease the wonky logic going on here without breaking backward compatibility.

PR #7324, opened 2022-12-18
docs: fix publish with BitbucketOptions

onucsecu2:
Removed ambiguous parameter names and added  clear description about token key of BitbucketOptions.


mmaietta:
Hi @onucsecu2 , thanks for the PR!
That .md file is actually generated from https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util-runtime/src/publishOptions.ts
You'll need to update `publishOptions.ts` and then run `pnpm generate-all` to regen schema+docs

PR #7331, opened 2022-12-22
feat: relpace powershell to native verify signature module

### LINKED ISSUE #7105: Planning to remove the use of PowerShell to verify updates? Execution of powershell.exe fails due to Anti-virus software, Software Restriction Policies (SRP), and general misconfiguration of machines
* **Electron-Builder Version**:  23.3.3
* **Electron-Updater Version**:  5.2.2

* **Node Version**: 16.5.0

* **Electron Version**: 20.1.0

* **Target**: windows

Hi,

electron-updater has been using powershell.exe to verify the signature of the downloaded update for a long time. The app I am working on is being used by multiple users and companies for a long time now and we have telemetry for updater failures. Failures to start powershell.exe have always been present to some extent, even though it works for most users. The failure rate has been small enough to ignore for a long time, but it is always there. Occasionally some enterprise customers would complain from their anti-virus blocking the execution of powershell.exe, or their SRP policies blocking it, or our telemetry would show that powershell failed to execute for various reasons.

I am filing this ticket not to get a recommendation to do manual rollouts, or to disable certificate validation, or otherwise to avoid the issue. This ticket is to ask if there are plans and request if possible to replace the execution of powershell.exe with calling Win32 APIs directly. This should make the update process friendlier to large companies and should increase the success rate of the updates.

There is the CryptQueryObject() API (example https://docs.microsoft.com/en-US/troubleshoot/windows/win32/get-information-authenticode-signed-executables), which is deprecated, but should still be a better option than launching powershell.exe.

The replacement API is Cryptography API: Next Generation, but I don't find an example of checking the code signing certificate of a file.

Thanks,
Ivan

beyondkmp:
Resolve #7105 

**Descrptions**

Because of the security concern, lots of enterprise users disable the PowerShell.  Once the PowerShell is disabled,  the code signature will fail and then the app will never be able to update.  It totally breaks the auto update feature in electron-updater.

**My solution**
Replace the PowerShell with a native verify signature module(https://github.com/beyondkmp/win-verify-trust)
It turns out work on many of our Windows PCs. 

This module(https://github.com/beyondkmp/win-verify-trust) has been used in my company production environment, no problems found so far.


gujinku:
We need this solution. It helps us. Thank you!

mmaietta:
This is awesome, thanks for the PR.

Quick Q, was any additional setup required in order for this native module to be properly bundled in with electron-updater? I'm worried that default electron-builder configs won't pick up the new `.node` addon

Also, why are we using the forked package instead of https://github.com/xan105/node-win-verify-trust?

beyondkmp:
> This is awesome, thanks for the PR.
> 
> Quick Q, was any additional setup required in order for this native module to be properly bundled in with electron-updater? I'm worried that default electron-builder configs won't pick up the new `.node` addon
I have published the pre build binary into it. And If we want to rebuild in local window computer, we just need to install the vs community.
> Also, why are we using the forked package instead of https://github.com/xan105/node-win-verify-trust?
This native module forks https://github.com/xan105/node-win-verify-trust, but I found some issues in it:
1. There are too many dependencies that he wrote himself
2. It uses CryptQueryObject to get sign subject. But it is too slow for large file and deprecated(https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptqueryobject).
So I have rewrite this module: Removed all unnecessary dependencies and use CertGetNameStringW to get the sign subject(refer to the microsoft example: https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/Win7Samples/security/cryptoapi/VerifyNameTrust/VerifyNameTrust/VerifyNameTrust.cpp)

mmaietta:
> There is on additional setup. If we want to rebuild in local window computer, we just need to install the vs community with c++ module

That's one of my worries in that not all users will have access to a Windows build machine if their project does not currently require a native module.  For instance, compiling with Wine on linux/docker would no longer work, right?

Is there a way for us to conditionally require the native module?

Alternatively, we could provide an electron-updater hook that would allow an app to apply a custom "verify" function and then add to the electron-builder Windows documentation the newer approach of verifying signature natively.

beyondkmp:
> Alternatively, we could provide an electron-updater hook that would allow an app to apply a custom "verify" function and then add to the electron-builder Windows documentation the newer approach of verifying signature natively.

We can public the verify function like the logger in electron-updater, if user don't pass the verify function, we use the powershell to verify. The user can pass their own native method to electron-updater, So the electron-updater doesn't need to import the native module. What do you think of it?

beyondkmp:
There is another way. We don't need that conditionally require the native module. We can provide the prebuild binary in the win-verify-trust module. So the users don't need to compile locally anymore.

mmaietta:
I'm in favor of this route unless pre-built binaries can be provided? I think that should still allow bundling with `wine` right?
>There is another way, we public the verifySignature function , the user can set their custom "verify" function. I created a new PR(https://github.com/electron-userland/electron-builder/pull/7337) for this way.
The user need to learn that how to use native verify module and how to fix the native verify module bugs.

https://github.com/electron-userland/electron-builder/pull/7337 looks solid to me but I do acknowledge it increases complexity of users leveraging native verification. Maybe we need a separate note in that PR's Markdown Docs to call out this awesome enhancement?

beyondkmp:
Agree with you. https://github.com/electron-userland/electron-builder/pull/7337 doesn't change the old verify logic, So many users don't need to make any changes.  
If they want to use the native verify signature module, they just need to integrate the native module and know how to use it.

mmaietta:
Merged https://github.com/electron-userland/electron-builder/pull/7337 to `master`, so I'm closing this PR. Many thanks for your contribution! Should be able to deploy to npm later today.

Issue #3754, opened 2019-03-14
[mac] Autoupdate fails if 'LSFileQuarantineEnabled' flag is enabled in Info.plist


paulbennet:
* **Version**: 20.38.5
Note: It should be the same case in 20.39.0, since I didn't find any issues related to this reported earlier, and no fixes pushed for this.

* **Updater Version**: 4.0.5

* **Target**: Mac [DMG]

Adding and enabling 'LSFileQuarantineEnabled' flag in the Info.plist file, makes our app mark all 
 new downloaded/created files in user's machine via NodeJS as "quarantined" (downloaded from the internet). We require this feature in our app, since it downloads external files to user's machine.

But, it seems when the auto-updater downloads an update file ( zip or dmg ), it is also marked as "quarantined" and hence the update fails throwing error as in below.

"Can’t be updated when it’s running from a read-only volume .."

Initially we thought this would be an issue for users who didn't move the app into Applications folder, but it was happening for almost all mac users. So I guess the issue is with electron-builder marking files are quarantined.

Solution:
- Unmark update files by removing the quarantine attributes, entries once downloaded

gch1p:
I have the same issue in my Electron app. I want quarantine bit to be set on files downloaded by user, but `LSFileQuarantineEnabled` in `Info.plist` breaks autoUpdater... I guess I'll have to set quarantine bit manually by calling xattr or by some other way.

abarsov:
the issue is still actual

PR #7333, opened 2022-12-22
fix: remove `com.apple.quarantine` attribute from downloaded update files on mac (#3754)

abarsov:
This is a potentioal fix of https://github.com/electron-userland/electron-builder/issues/3754

### DIFF
--- a/packages/electron-updater/src/MacUpdater.ts
+++ b/packages/electron-updater/src/MacUpdater.ts
@@ -95,6 +95,15 @@ export class MacUpdater extends AppUpdater {
     const updateFileSize = zipFileInfo.info.size ?? (await stat(downloadedFile)).size
 
     const log = this._logger
+
+    try {
+      this.debug("Clearing quarantine bit")
+      const result = execFileSync("xattr", ["-r", "-d", "com.apple.quarantine", downloadedFile])
+      log.info(`Quarantine bit is cleared: ${result}`)
+    } catch (e) {
+      log.warn(`xattr shell command for clearing quarantine bit failed: ${e}`)
+    }
+
     const logContext = `fileToProxy=${zipFileInfo.url.href}`
     this.debug(`Creating proxy server for native Squirrel.Mac (${logContext})`)
     this.server?.close()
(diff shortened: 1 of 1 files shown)


packages/electron-updater/src/MacUpdater.ts
```diff
@@ -95,6 +95,15 @@ export class MacUpdater extends AppUpdater {
     const updateFileSize = zipFileInfo.info.size ?? (await stat(downloadedFile)).size
 
     const log = this._logger
+
+    try {
+      this.debug("Clearing quarantine bit")
+      const result = execFileSync("xattr", ["-r", "-d", "com.apple.quarantine", downloadedFile])
+      log.info(`Quarantine bit is cleared: ${result}`)
```
mmaietta, line 102: What's this `result` being printed out going to look like? Perhaps we could also append `?.trim()` in case there's extra whitespace

packages/electron-updater/src/MacUpdater.ts
```diff
@@ -95,6 +95,15 @@ export class MacUpdater extends AppUpdater {
...
+    } catch (e) {
+      log.warn(`xattr shell command for clearing quarantine bit failed: ${e}`)
```
mmaietta, line 104: Please use `e.message` as I don't think we'll need additional data from the `e` object.

mmaietta:
Just to confirm, have you tested this approach?

packages/electron-updater/src/MacUpdater.ts
abarsov: in my case result was an empty string, perhaps it is always the case, need to check it.

abarsov:
> Just to confirm, have you tested this approach?

I tried it locally, setting quarantine was executed sussessfully, 
but auto-update failed later due to the issue with my test package signature, 
so I can't say I have managed to perform end-to-end test - just checked that added code was executed succesfully.

mmaietta:
@abarsov do you happen to have any updates on this PR? Specifically looking for more insight on:
- message `result` when `xattr` fails to verify `e.message` response content
- testing the full flow with the auto-update

PR #7337, opened 2022-12-24
feat: Provide a custom verify function interface in NsisUpdater

beyondkmp:
Replace the PowerShell with [a native verify signature module](https://github.com/beyondkmp/win-verify-trust)， but not all users will have access to a Windows build machine if their project does not currently require a native module. For instance, compiling with Wine on linux/docker would no longer work. 

Providing a custom verify function interface, if the user want to use native verify module, they can pass it and they need to have an access to a windows build machine. If they don't pass it, will use the default verify signature function(https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts).

**Interface**
```js
/**
   * The verifyUpdateCodeSignature. You can pass [win-verify-signature](https://github.com/beyondkmp/win-verify-trust) or another custom verify function: ` (publisherName: string[], path: string) => Promise<string | null>`.
   * The default verify function uses [windowsExecutableCodeSignatureVerifier](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts)
   */
// return null if verify signature succeed
// return error message if verify signature failed
export type verifyUpdateCodeSignature = (publisherName: string[], path: string) => Promise<string | null>
```

**Example**

```js
import { NsisUpdater } from "electron-updater"
import { verifySignatureByPublishName } from "beyondkmp/win-verify-signature"
// Or MacUpdater, AppImageUpdater

export default class AppUpdater {
    constructor() {
        const options = {
            requestHeaders: {
                // Any request headers to include here
            },
            provider: 'generic',
            url: 'https://example.com/auto-updates'
        }

        const autoUpdater = new NsisUpdater(options)
        autoUpdater.verifyUpdateCodeSignature = (publisherName: string[], path: string) => {
            const result = verifySignatureByPublishName(path, publisherName);
            if(result.signed) return Promise.resolve(null);
            return Promise.resolve(result.message);
        }
        autoUpdater.addAuthHeader(`Bearer ${token}`)
        autoUpdater.checkForUpdatesAndNotify()
    }
}
```

### DIFF
--- a/.changeset/slow-avocados-carry.md
+++ b/.changeset/slow-avocados-carry.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": minor
+---
+
+feat: Provide a custom verify function interface to enable nsis signature verification alternatives instead of powershell
--- a/docs/configuration/win.md
+++ b/docs/configuration/win.md
@@ -60,9 +60,7 @@ exports.default = async function(configuration) {
 
 #### How do use a custom verify function to enable nsis signature verification alternatives instead of powershell?
 
-You need to pass a custom verify function to the nsis updater. For example, if you want to use a native verify function, you can use [win-verify-signature](https://github.com/beyondkmp/win-verify-trust).
-
-The `verifyUpdateCodeSignature` interface
+Use the `verifyUpdateCodeSignature` interface:
 
 ```js
 /**
@@ -72,7 +70,8 @@ The `verifyUpdateCodeSignature` interface
 export type verifyUpdateCodeSignature = (publisherName: string[], path: string) => Promise<string | null>
 ```
 
-pass a custom verify function to the nsis updater
+Pass a custom verify function to the nsis updater. For example, if you want to use a native verify function, you can use [win-verify-signature](https://github.com/beyondkmp/win-verify-trust).
+
 
 ```js
 import { NsisUpdater } from "electron-updater"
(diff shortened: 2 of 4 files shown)


beyondkmp:
>Maybe we need a separate note in that PR's Markdown Docs to call out this awesome enhancement?
Could I add the example to [slow-avocados-carry.md](https://github.com/electron-userland/electron-builder/pull/7337/commits/f7681c9b852fd8e66d072f852cea53f95e5bf23a) ?

mmaietta:
`slow-avocados-carry.md` is specifically for the release automation to pick up a PR for auto-publishing to npm. So let's keep that file as-is.

Your example provided looks great though. Perhaps we could add it to the Common Questions for Windows configuration?
https://github.com/electron-userland/electron-builder/blob/master/docs/configuration/win.md

I can't seem to find any other docs specifically on auto-update configuration, just for publishers but not the AppUpdater itself? (Maybe I need to create a separate Page just for that in the future?)

beyondkmp:
I have added the example to win.md. Please help check it.

mmaietta:
@rasgo-cc I think this should help you 
https://www.electron.build/configuration/win#how-do-use-a-custom-verify-function-to-enable-nsis-signature-verification-alternatives-instead-of-powershell

Issue #7334, opened 2022-12-23
NSIS Multi Language  License File support does not work


zanzara:
* **Electron-Builder Version**: 23.6.0 

* **Node Version**: 16.16.0 

* **Electron Version**: 21.1.0
* **Electron Type (current, beta, nightly)**: current

* **Target**:  Windows10 64bit

I have multiple license files as `license_en.html`  and `license_de.html`  i.e.
but it doesn't matter in which folder I put them, the installer `setup.exe` will never show them.

If i put the `license` prop directly to one license  file location as ie.
`license: license_en.html`
it works.

So I assume, that this feature does not work as described in the electron-builder docs under
[https://www.electron.build/configuration/nsis](https://www.electron.build/configuration/nsis)

ielektronick:
I do understand that the issue is closed more than a year ago, but still.

According to `app-builder-lib` package source code, the license files could be placed in root on build resources (https://www.electron.build/nsis#license), meaning if you put the license files (`license_de.txt`, `license_en.txt` etc) to the root and run electron-builder to make nsis installer, the files should be handled accordingly. 

To check it, after the installer script review the `builder-debug.yml` file. There should be something like:

```
!macro licensePage
  LicenseLangString MUILicense 1033 "path\build\license_en.txt"
  LicenseLangString MUILicense 1031 "path\build\license_de.txt"
  !insertmacro MUI_PAGE_LICENSE "$(MUILicense)"
!macroend
```

PR #7339, opened 2022-12-25
fix: missing html extension for multi language license files in nsis target

zanzara:
Regarding issue #7334  
[https://github.com/electron-userland/electron-builder/issues/7334]( https://github.com/electron-userland/electron-builder/issues/7334)
there was missing the `.html` extension.

### DIFF
--- a/.changeset/plenty-spoons-prove.md
+++ b/.changeset/plenty-spoons-prove.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: add missing html extension for multi language license files in nsis target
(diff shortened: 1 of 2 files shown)


vvillaluna:
Hey guys from which version it will be added ?

mmaietta:
@vvillaluna 24.0.0-alpha.9

vvillaluna:
What's the difference in the process of reading the license file when using the `multiLanguage` and adding them directly on `license` nsis config ?

Seems like it has differences in the output, 

Using `license: "license_ja.html"` works well.
image


But using multilanguage like putting licenses under build resources:
image


@mmaietta can you guide me where to check that functionality would like to contribute

PR #7351, opened 2023-01-05
fix: Update MacOS signOptions on macPackager #7317

### LINKED ISSUE #7317: Mac code signing compatibility with @electron/osx-sign 1.0.4
* **Electron-Builder Version**: 24.0.0-alpha.7

* **Node Version**: 16

* **Electron Version**: 22.0.0

* **Target**: mac

I'm in the process of updating to electron-builder 24 and having issues with code signing.  It looks like EB is using the legacy "identity-validation" property https://github.com/electron-userland/electron-builder/blob/ffe99677610e933a5a22ceb70ba0cbb7edf55695/packages/app-builder-lib/src/macPackager.ts#L273 which should be camelCase now.  

More importantly, while tracking this down, it became clear there is some conflated use of the `options.identity` string and the return value of the `findIdentity` function that returns an osx-sign `Identity` object.  In EB versions before 24 the legacy `identity-validation=false` working properly would mask this for my particular use case.

At this line, I think the identity value should be `qualifer` instead of the `Identity` instance.

https://github.com/electron-userland/electron-builder/blob/ffe99677610e933a5a22ceb70ba0cbb7edf55695/packages/app-builder-lib/src/macPackager.ts#L297

filfreire:
cc @mmaietta 

Trying (almost) the same fix as suggested by @mayfield in this comment https://github.com/electron-userland/electron-builder/issues/7317#issuecomment-1351168459. I'm leaving the old legacy options in for now, I reckon they will be ignored when using the new https://github.com/electron/osx-sign ?

Not sure if in this case we should leave the legacy stuff in. There's a migration guide that can be checked for reference: https://github.com/electron/osx-sign/blob/a9cfce77512358420a850abfc102f784e02f8138/MIGRATION.md

### DIFF
--- a/.changeset/chilled-shoes-sort.md
+++ b/.changeset/chilled-shoes-sort.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+Update MacOS signOptions on macPackager
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -270,7 +270,6 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     }
 
     const signOptions: any = {
-      "identity-validation": false,
       identityValidation: false,
       // https://github.com/electron-userland/electron-builder/issues/1699
       // kext are signed by the chipset manufacturers. You need a special certificate (only available on request) from Apple to be able to sign kext.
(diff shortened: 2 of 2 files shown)


mmaietta:
>I'm leaving the old legacy options in for now, I reckon they will be ignored when using the new

Let's just remove the legacy options as otherwise I bet someone (incl. myself) may forget why we have both keys present. If something breaks, we should know fairly quickly 😅

filfreire:
@mmaietta I removed the legacy options following your feedback.

Issue #7357, opened 2023-01-09
spctl process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE on 24.0.0-alpha.9


filfreire:
* **Electron-Builder Version**: 24.0.0-alpha.9

* **Node Version**: 16.17

* **Electron Version**: 22.0.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: MacOS

We're currently facing an issue when signing a Mac App where we get an error  like:
```
spctl process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
```

It seems to happen on [this line of code](https://github.com/electron-userland/electron-builder/blob/1e8dad8bc58f53780c9fac3b0c48e248a8b5467c/packages/app-builder-lib/src/macPackager.ts#L492).

This is [the config](https://github.com/Kong/insomnia/blob/1cb78093c22a2a99b4512d9f91840c77b6a7cf7f/packages/insomnia/electron-builder.config.js).

Original [debug log](https://github.com/Kong/insomnia/actions/runs/3875812475/jobs/6608865197#step:5:796): 
```
• signing         file=dist/mac-universal/Insomnia.app identityName=Developer ID Application: Kong Inc. (FX44YY62GV) identityHash=8A3A2A7038D62E18916CF6F5752BB292F289B213 provisioningProfile=none
  • spawning        command=spctl -a -t open --context context:primary-signature -v "/Users/runner/work/***/***/packages/***/dist/mac-universal/Insomnia.app"
"/Users/runner/work/***/***/packages/***/dist/mac-universal/Insomnia.app": No such file or directory
  • exited          command=spctl code=1 pid=27881
  • executing       file=security args=delete-keychain /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/d586d1f05b6aa072b4f4079123b8263389cb4317b0003db560c539082840b86b.keychain
  • executed        file=security
  ⨯ spctl process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
Exit code:
1  failedTask=build stackTrace=Error: spctl process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
Exit code:
1
    at ChildProcess.<anonymous> (/Users/runner/work/***/***/packages/***/node_modules/builder-util/src/util.ts:250:14)
    at Object.onceWrapper (node:events:628:26)
    at ChildProcess.emit (node:events:513:28)
[...]

PR #7361, opened 2023-01-10
fix: Remove spctl check from Mac's notarization

filfreire:
Workaround fix for issue #7357 

We found that the `spctl` check doesn't work well on Github Actions (and maybe other CI providers). This PR ~~adds a workaround to skip that check when notarizing builds on Mac.~~ (UPDATE) removes the `spctl` check after notarizing the Mac builds.

cc @mmaietta

### DIFF
--- a/.changeset/thin-bags-tap.md
+++ b/.changeset/thin-bags-tap.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": major
+---
+
+Add skipVerification flag in Mac's notarize options
--- a/.changeset/thin-bags-tap.md
+++ b/.changeset/thin-bags-tap.md
@@ -2,4 +2,4 @@
 "app-builder-lib": major
 ---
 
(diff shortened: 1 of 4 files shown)


mmaietta:
I'm wondering if it'd be best to just remove that line of code entirely. What are your thoughts?

I used `spctl` locally and it's worked historically quite well, but it may be unnecessary since @electron/notarize seems to handle the verification as part of its Promise (and spitting out an error logs too).
https://github.com/electron/notarize/blob/cce06044e40049074f64ce0f58dd9e2be5406683/src/notarytool.ts#L80-L98

filfreire:
@mmaietta it was one of my temptations as well - remove it - because regularly - users who want to - do it themselves.

I'll change the PR to remove the step entirely.

filfreire:
@mmaietta done - I removed the line entirely and updated the PR

PR #7378, opened 2023-01-16
fix: Remove adapter if core22 is set as base on snapcraft

### LINKED ISSUE #7372: Extra field 'adapter' not permitted when using core22 as base for Snapcraft configs
* **Electron-Builder Version**: 24.0.0-alpha.10

* **Node Version**: 16.17

* **Electron Version**: 22.0.0

* **Target**: Snapcraft (snap)

If users define `core22` as `base` and define some stage packages, like:

```
stagePackages: ['default', 'libdbus-1-dev', 'libnss3-dev']
```

when packaging their apps they might get an error like:
```
Bad snapcraft.yaml content:
- extra field 'adapter' not permitted in 'apps.insomnia' configuration
Full execution log: '/home/parallels/.cache/snapcraft/log/snapcraft-20230113-002320.808750.log'
  ⨯ exit status 1
github.com/develar/app-builder/pkg/util.ExecuteAndPipeStdOutAndStdErr
        /Volumes/data/Documents/app-builder/pkg/util/exec.go:23
github.com/develar/app-builder/pkg/package-format/snap.buildWithoutTemplate
        /Volumes/data/Documents/app-builder/pkg/package-format/snap/snap.go:352
github.com/develar/app-builder/pkg/package-format/snap.Snap
(...)
```

this is due to `adapter` being legacy. See [this doc for more info](https://forum.snapcraft.io/t/micro-howto-migrate-from-core20-to-core22/30188) in the `Snapcraft provided environment` section.

It ends up being this way due to `isUseTemplateApp`ending up as `false` when folks define `stagePackages`, hitting this codepath in snap.ts:
```
this.isUseTemplateApp =
      this.options.useTemplateApp !== false &&
      (arch === Arch.x64 || arch === Arch.armv7l) &&
      buildPackages.length === 0 &&
      isArrayEqualRegardlessOfSort(stagePackages, defaultStagePackages)
```

one possible workaround might be adding a check for `core22`
```
if (snap.base === "core22") {
  delete appDescriptor.adapter
}
```

filfreire:
Closes #7372

cc @mmaietta

### DIFF
--- a/packages/app-builder-lib/src/targets/snap.ts
+++ b/packages/app-builder-lib/src/targets/snap.ts
@@ -72,7 +72,7 @@ export default class SnapTarget extends Target {
     if (options.base != null) {
       snap.base = options.base
       // from core22 onwards adapter is legacy
-      if (snap.base === "core22") {
+      if (Number(snap.base.split("core")[1]) >= 22) {
         delete appDescriptor.adapter
       }
     }
--- a/packages/app-builder-lib/src/targets/snap.ts
+++ b/packages/app-builder-lib/src/targets/snap.ts
@@ -71,6 +71,10 @@ export default class SnapTarget extends Target {
     }
     if (options.base != null) {
       snap.base = options.base
+      // from core22 onwards adapter is legacy
+      if (snap.base === "core22") {
+        delete appDescriptor.adapter
+      }
     }
     if (options.grade != null) {
       snap.grade = options.grade
--- a/.changeset/thin-terms-change.md
+++ b/.changeset/thin-terms-change.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": major
+---
+
+Remove extra adapter field if core22 is set as base for snapcraft


packages/app-builder-lib/src/targets/snap.ts
```diff
@@ -71,6 +71,10 @@ export default class SnapTarget extends Target {
     }
     if (options.base != null) {
       snap.base = options.base
+      // from core22 onwards adapter is legacy
+      if (snap.base === "core22") {
+        delete appDescriptor.adapter
```
mmaietta, line 76: Do we need to future-proof this for core22+? This feels too hardcoded to me. Perhaps we can extract the number from `core__` and compare `>= 22`, what are your thoughts?

filfreire: Agreed. We can change it to something like `Number(snap.base.split("core")[1]) >= 22` instead?

mmaietta: That sounds perfect.

Issue #7393, opened 2023-01-25
electron-updater 6.0.0-alpha.8 does not work on Ubuntu 22.04.1 LTS with Error: EPIPE


ganthern:
* **Electron-Builder Version**: 23.6.0

* **Node Version**: 16.16.0

* **Electron Version**: 22.0.3
* **Electron Type (current, beta, nightly)**: current
* **Which version of electron-updater are you using** 6.0.0-alpha.8

* **Target**: linux

Currently, the updater will start the updated app with `child_process.spawn()` with `stdio: "pipe"`. This leads to the following error on Ubuntu LTS:

![image](https://user-images.githubusercontent.com/29049728/214585766-eebb768b-90f2-4536-af6a-9616ffab1e98.png)

Reproducible with https://github.com/ganthern/electron-updater-epipe

mmaietta:
Thanks @ganthern for calling this out! Deploying electron-updater@6.0.0-alpha.9 now, please give it a shot when you have a chance :)

PR #7394, opened 2023-01-25
fix: inherit stdio for updated processes (#7393)

ganthern:
https://github.com/electron-userland/electron-builder/issues/7393


mmaietta:
Thanks for the contribution!

PR #7407, opened 2023-02-02
feat: Allow for NSIS windows installer to be wrapped in an MSI

ghost1face:
Hey all,

I ran into a similar issue as others where we build an .exe and .msi installer only to find out they are completely different.  I looked into a few potential tools and solutions and have come to this one.  Ultimately, the goal here is that we install a single application using the same Product Code regardless of installer.

The way I'm accomplishing this here is by building an msi installer that wraps and invokes an exe installer behind the scenes.  
This pull request adds a new target: `msiWrapped` which is a "dependent target" that depends on the `nsis` target to also be configured.  Order of the configuration doesn't matter here as I use the `isAsyncSupported` flag for this target to ensure the `nsis` build goes down before the `msiWrapped`.

```json
{
  "build": {
    ...other stuff...
    "win": {
      "target":  [
        "nsis",
        "msiWrapped"
      ]
  }
}
```

I did the following:
- [X] Added tests
- [X] Updated the schema
- [ ] Updated documentation (I was getting the exact same error as https://github.com/electron-userland/electron-builder/pull/7136#issuecomment-1263029896)

Take a look and let me know what you guys think.  I kept the implementation basic so it definitely does have room to develop over time.


mmaietta:
This is awesome! Thanks for the contribution. 

Super odd that the doc generation isn't working, but I can take care of that separately if need be.

ghost1face:
Ok, let me know if there's anything you need from me whether it be screenshot samples or anything else of the sort.  I think I covered all the pre-requisites: schema, tests, naming conventions, etc.  But if there's something I missed I'd be happy to make updates

package.json
```diff
@@ -57,6 +57,7 @@
     "eslint": "8.26.0",
     "eslint-config-prettier": "8.5.0",
     "eslint-plugin-prettier": "4.2.1",
+    "fast-xml-parser": "^4.0.15",
```
mmaietta, line 60: I think this needs to be a prod `dependencies` in `app-builder-lib` specifically.

packages/app-builder-lib/src/targets/MsiWrappedTarget.ts
```diff
@@ -0,0 +1,224 @@
...
+const projectTemplate = new Lazy<(data: any) => string>(async () => {
+  const template = (await readFile(path.join(getTemplatePath("msiWrapped"), "template.xml"), "utf8"))
+    .replace(/{{/g, "<%")
+    .replace(/}}/g, "%>")
+    .replace(/\${([^}]+)}/g, "<%=$1%>")
+  return ejs.compile(template)
+})
+
+// WiX doesn't support Mono, so, dontnet462 is required to be installed for wine (preinstalled in our bundled wine)
+export default class MsiWrappedTarget extends Target {
```
mmaietta, line 29: I'm noticing a **lot** of copied code from `MsiTarget`, perhaps we should extend `MsiTarget` instead of `Target`?
This would mirror the current setup of 
```
interface MsiWrappedOptions extends CommonWindowsInstallerConfiguration, TargetSpecificOptions { ... }
```

package.json
ghost1face: `fast-xml-parser` is only referenced in the unit tests.  So my thought is that it's only a devDependency here

packages/app-builder-lib/src/targets/MsiWrappedTarget.ts
ghost1face: Good point, I copied it at the start not knowing how far it may stray, let me try to clean that up

package.json
mmaietta: ahhh fair enough. Does it still need to be at the electron-builder workspace level or can it be installed within the `/test/package.json` then?

ghost1face: I went ahead and moved the dependency to the appropriate location, under test

ghost1face:
I did some cleanup between the targets so there's less duplication

mmaietta:
The failing tests are unrelated to this PR and seem to be an environment issue.

The PR LGTM, thanks for the contribution! I'll work on getting the docs regenerated

mmaietta (commented):
LGTM

PR #7425, opened 2023-02-17
fix(linux): sanitize apostrophe in postinst script

baiyuanneko:
Currently if the product name includes apostrophes, there will be an error when installing ```.deb``` file that electron-builder generates.

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -47,6 +47,7 @@
     "smooth-hounds-peel",
     "sour-chicken-study",
     "spotty-vans-pay",
+    "stale-dots-build",
     "swift-ties-love",
     "thin-bags-tap",
     "thin-terms-change",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# [](https://github.com/electron-userland/electron-builder/compare/v24.0.0-alpha.12...v) (2023-02-21)
+
+
+### Bug Fixes
+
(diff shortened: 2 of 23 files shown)


packages/app-builder-lib/templates/linux/after-install.tpl
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 # Link to the binary
-ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
+ln -sf $'/opt/${sanitizedProductNameInBashScript}/${executable}' '/usr/bin/${executable}'
```
mmaietta, line 4: What's the additional `$` operating as here?
What if we changed the command to use `"` instead of `'` so that apostrophes are retained?
```
ln -sf "/opt/${sanitizedProductName}/${executable}" "/usr/bin/${executable}"
chmod 4755 "/opt/${sanitizedProductName}/chrome-sandbox" || true
```

baiyuanneko: You can refer to [this link](https://wiki.bash-hackers.org/syntax/quoting#ansi_c_like_strings) to learn more about ```$'string'```. It can allow escape sequences to be decoded.

Using ```"``` instead of ```'``` can also solve this problem. Do you want me to push another commit that uses this way?

mmaietta: Let's use the `"` route for simplicity, that way we keep the template to a minimal number of vars.

PR #7431, opened 2023-02-20
fix(packager): return success status from doSign function calls

nsrCodes:
Recent changes (https://github.com/electron-userland/electron-builder/pull/7311) in platform packager `afterSign` hook show that the correct success status was not being returned after signing the app on macOS

This results in the app being signed but the terminal has the following message

>  skipping "afterSign" hook as no signing occurred, perhaps you intended "afterPack"?

found this on electron-builder v24.0.0-alpha.12

_Couldn't get a local development environment properly setup, so hoping someone could help test if these help solve the issue_

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -454,10 +454,6 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       })
       return true
     }
-
-    const appFileName = `${this.appInfo.productFilename}.app`
-    await readDirectoryAndSign(packContext.appOutDir, await readdir(packContext.appOutDir), file => file === appFileName)
-
     if (!isAsar) {
       return true
     }
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -376,7 +376,7 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
   }
 
   //noinspection JSMethodCanBeStatic
-  protected async doSign(opts: SignOptions): Promise<any> {
+  protected doSign(opts: SignOptions): Promise<any> {
     return signAsync(opts)
   }
 
@@ -454,6 +454,9 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       })
       return true
     }
+
+    const appFileName = `${this.appInfo.productFilename}.app`
+    await readDirectoryAndSign(packContext.appOutDir, await readdir(packContext.appOutDir), file => file === appFileName)
     if (!isAsar) {
       return true
     }
--- a/packages/app-builder-lib/src/winPackager.ts
+++ b/packages/app-builder-lib/src/winPackager.ts
@@ -255,13 +255,11 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
   }
 
   private async doSign(options: WindowsSignOptions) {
-    let didSign = true
     for (let i = 0; i < 3; i++) {
       try {
         await sign(options, this)
         return true
       } catch (e: any) {
-        didSign = false
         // https://github.com/electron-userland/electron-builder/issues/1414
         const message = e.message
         if (message != null && message.includes("Couldn't resolve host name")) {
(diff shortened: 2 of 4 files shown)


packages/app-builder-lib/src/winPackager.ts
```diff
@@ -269,6 +271,7 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
         throw e
       }
     }
+    return didSign
```
mmaietta, line 274: This `didSign` flag logic needs to be flipped in order to handle the `continue` within the `catch`
```
let didSign = false
for (let i = 0; i < 3; i++) {
  try {
    await sign(options, this)
    didSign = true
    break
  } catch (e: any) {
    // https://github.com/electron-userland/electron-builder/issues/1414
    const message = e.message
    if (message != null && message.includes("Couldn't resolve host name")) {
      log.warn({ error: message, attempt: i + 1 }, `cannot sign`)
      continue
    }
    throw e
  }
}
return didSign
```

packages/app-builder-lib/src/winPackager.ts
```diff
@@ -366,7 +369,9 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
       if (this.shouldSignFile(file)) {
         const parentDir = path.dirname(file)
         if (parentDir !== packContext.appOutDir) {
-          return new CopyFileTransformer(file => this.sign(file))
+          return new CopyFileTransformer(async file => {
+            await this.sign(file)
```
mmaietta, line 373: Is this change intentional?
`file => this.sign(file)` returns `Promise<boolean>`
`async file => { await this.sign(file) }` returns `Promise<void>` due to the brackets

packages/app-builder-lib/src/macPackager.ts
```diff
@@ -450,7 +450,7 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     await BluebirdPromise.map(readdir(packContext.appOutDir), async (file: string): Promise<any> => {
```
mmaietta, line 450: If you change the return of this larger function to `Promise<boolean>`, intellisense will still flag that this function isn't returning the correct values. I'll take a look on refactoring this function entirely

mmaietta (commented):
Thank you so much for calling out this issue!

I found that the `BluebirdPromise.map` doesn't require `return` as `this.sign` throws an error if it fails. We can just return true after signing completes.

I'm going to pick up this work and file a new PR. I'll take a larger stab at refactoring this code (and investigate some other `Promise<any>` that ended up letting this issue sneak by intellisense.

packages/app-builder-lib/src/winPackager.ts
nsrCodes: I was getting linting issues that this callback expected a return type of `Promise<void>`, which is why I made the change, but these do not seem important, I will remove this if needed.

nsrCodes:
@mmaietta I have rebased this with master which contains your larger rewrite. Doing this just for completeness, and a few small improvements I thought were relevant. 
In case this PR is now just redundant please close it.
If not, probably a better idea to squash and merge because the commit history isn't great 😓

Thanks for quickly releasing the fix!

packages/app-builder-lib/src/winPackager.ts
```diff
@@ -210,7 +210,7 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
     const cscInfo = await this.cscInfo.value
     if (cscInfo == null) {
       if (this.platformSpecificBuildOptions.sign != null) {
-        await sign(signOptions, this)
+        return await sign(signOptions, this)
```
mmaietta, line 213: I don't think we want to return here, right?
We still have this at the bottom of this function 
```
return this.doSign({ ...
```

nsrCodes: @mmaietta  We won't reach there if we enter this scope, because of the `return true` on L#219
Added this here for cases when `sign` might return a falsy value, but looking at its definition, seems like that's never the case.

So I assume the fix here would be to 
1. remove this return, await `sign`
2. remove the `return true` in `cscInfo == null` scope so that we do execute `doSign`
3. add some sort of null check for `cscInfo` on L#226 and beyond

But that (step#2) seems wrong since if`cscInfo == null` we have already tried to sign and wouldn't want to try anymore

I am not familiar with what the code does here so I am hoping you can help frame the right logic. To me, the current state looks correct.

PR #7436, opened 2023-02-22
fix(linux): use double quotes in postinst script

baiyuanneko:
This pull request solves the problem mentioned [here](https://github.com/electron-userland/electron-builder/pull/7425).

> Currently if the product name includes apostrophes, there will be an error when installing ```.deb``` file that electron-builder generates.


mmaietta:
Thank you for your contribution! Kicking off the CI builds

mmaietta:
This test snapshot is failing, looks like the snapshot just needs to be updated to match the quotation marks
```
 FAIL  src/linux/debTest.ts (103.627 s)
  ✕ executable path in postinst script (14419 ms)
```

mmaietta:

mmaietta:
Not too sure why the test is still failing but I guess you'll need to run the full test snapshot update flow
```
TEST_FILES=debTest UPDATE_SNAPSHOT=true pnpm run test-linux
```
I'd offer to do so, but the docker image doesn't work on arm64 macs.

Issue #7441, opened 2023-02-22
afterSign does not support mjs files


Nokel81:
* **Electron-Builder Version**: 23.6.0

* **Node Version**: v16.16.0

* **Electron Version**: v19.1.9
* **Electron Type (current, beta, nightly)**: current

* **Target**: any

Specifying a file `"./build/notarize.mjs"` as the value of `.build.afterSign` in our `package.json` fails with 
```
Error [ERR_REQUIRE_ESM]: require() of ES Module <repo>/build/notarize.mjs not supported.
```

PR #7442, opened 2023-02-22
feat: Support mjs files for lifecycle operations

Nokel81:
- beforePack
- afterSign
- artifactBuildStarted
- artifactBuildCompleted
- appxManifestCreated
- msiProjectCreated
- beforeBuild
- afterPack
- afterAllArtifactBuild
- sign
- onNodeModuleFile

fixes https://github.com/electron-userland/electron-builder/issues/7441

### DIFF
--- a/packages/app-builder-lib/src/platformPackager.ts
+++ b/packages/app-builder-lib/src/platformPackager.ts
@@ -752,6 +752,22 @@ export function normalizeExt(ext: string) {
   return ext.startsWith(".") ? ext.substring(1) : ext
 }
 
+/**
+ * NOTE: this is required because tsc converts the following:
+ *
+ * ```ts
+ * await import("path/to/some/module");
+ * ```
+ *
+ * to
+ *
+ * ```ts
+ * Promise.resolve().then(() => require("path/to/some/module"))
+ * ```
+ */
+/* eslint-disable @typescript-eslint/no-implied-eval */
+const importDynamic = new Function("modulePath", "return import(modulePath)")
+
 export async function importFunction<T>(executor: T | string, name: string): Promise<T> {
   if (executor == null || typeof executor !== "string") {
     return executor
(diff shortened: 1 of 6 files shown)


packages/app-builder-lib/src/platformPackager.ts
```diff
@@ -769,8 +769,7 @@ export function resolveFunction<T>(executor: T | string, name: string): T {
     p = path.resolve(p)
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const m = require(p)
+  const m = await import(p)
```
mmaietta, line 772: Is this a backward-compatible change that will still allow cjs?

Nokel81: Yes it should be.

mmaietta: Can you please double-check with a local test project? Can just use a boilerplate repo/project from here: https://www.electron.build/#boilerplates

It's either that route or we need to add explicit `beforeBuild.js` and `beforeBuild.mjs` in the `test/fixtures` + `BuildTest.ts` tests, which is fine, but just much longer of a route.

Nokel81:
@mmaietta Tested locally using the `electron-react-boilerplate` repo and both a `"buildBuild": "./scripts/beforeBuild.mjs"` and a `"buildBuild": "./scripts/beforeBuild.js"` run. They work. Though I had to put a fix in.

mmaietta:

Nokel81:
@mmaietta done

mmaietta:
Need to revert this PR until further testing is done.
https://github.com/electron-userland/electron-builder/pull/7461

Basically, Windows unit tests broke (that somehow I overlooked earlier?). It hangs and receives an Access Violation consistently - `3221225477` -> `0xC0000005`

Nokel81:
Oh okay. I'll see if I can reproduce on a Windows machine.

Nokel81:
Oh okay. I'll see if I can reproduce on a Windows machine.

PR #7449, opened 2023-02-27
docs: fix missing opening curly bracket in publish.md

ms10596:
fix missing opening curly bracket and format snippet as json


mmaietta:
I believe that rendered/generated doc is coming from here:
https://github.com/electron-userland/electron-builder/blob/93930cf0b04b60896835e1d9feeab20722cd1b98/packages/builder-util-runtime/src/publishOptions.ts#L292-L309

The downside is that the entire jsdoc2html needs to be rewritten to properly support formatting. IIRC Markdown in Netlify doesn't properly render lists, so the API docs were a complete mess.

I don't think we can merge this PR, but thanks for calling out the issue. I'll need to re-investigate the documentation generation pipeline again 😭

PR #7450, opened 2023-02-27
chore: Adding node 18 docker images for electron-builder

xmireczech:
Adding node 18 docker images for electron-builder (Electron 23 uses node 18).


mmaietta:

PR #7465, opened 2023-03-07
fix: cross-platform build for native modules using prebuild

cocktailpeanut:
The `rebuild()` method has not been passing the platform to the electron-rebuild.

This means electron-builder has been using the `process.platform` no matter what platform you specify, which means no cross-platform build if an app includes native modules that include prebuilt binaries.

Combined with the fix for `@electron/rebuild` at https://github.com/electron/rebuild/pull/1072 I have confirmed that now I can successfully build using electron-builder for ALL platforms using the correct native built binaries.

### DIFF
--- a/packages/app-builder-lib/src/util/yarn.ts
+++ b/packages/app-builder-lib/src/util/yarn.ts
@@ -26,7 +26,7 @@ export async function installOrRebuild(config: Configuration, appDir: string, op
     }
     await installDependencies(appDir, effectiveOptions)
   } else {
-    await rebuild(appDir, config.buildDependenciesFromSource === true, options.arch)
+    await rebuild(appDir, config.buildDependenciesFromSource === true, options)
   }
 }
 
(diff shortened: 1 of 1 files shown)


mmaietta:
Attempting to reimplement this via https://github.com/electron-userland/electron-builder/pull/7629

mmaietta:
@cocktailpeanut can you try `next` 24.5.1 with the new implementation I added?

willemdjong:
@mmaietta I tried using 24.5.2 and set npmRebuild at true but it ain't working as previous versions of electron builder. Previously I used electron-builder  version=23.6.0. Rebuilding native dependencies and install prebuilt binary are the missing lines in my electron-builder output.

mmaietta:
@willemdjong please create a Issue ticket, tag me, and we can consolidate our convo there. I'll need a minimum reproducible repo of the project to test with as well

PR #7466, opened 2023-03-07
fix(arm64): RPM architecture name

jkroepke:
It feels like just having `aarch64` in the filename is not sufficient.

```
[root@7b42929061f0 /]# uname -a
Linux 7b42929061f0 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux

yum install https://github.com/MuhammedKalkan/OpenLens/releases/download/v6.4.4/OpenLens-6.4.4.aarch64.rpm
Error:
 Problem: conflicting requests
  - package open-lens-6.4.4-1.arm64 does not have a compatible architecture
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

There is also an fix available on [fpm](https://github.com/jordansissel/fpm/commit/cc0a4e736941401ed6f3a5b20269d1415c53f45c), but this requires fpm 1.13.0 or higher

### DIFF
--- a/.changeset/selfish-games-sniff.md
+++ b/.changeset/selfish-games-sniff.md
@@ -1,0 +1,5 @@
+---
+"builder-util": patch
+---
+
+fix(arm64): use RPM architecture `aarch64` in name
--- a/packages/builder-util/src/arch.ts
+++ b/packages/builder-util/src/arch.ts
@@ -17,7 +17,7 @@ export function toLinuxArchString(arch: Arch, targetName: string): string {
     case Arch.armv7l:
       return targetName === "snap" || targetName === "deb" ? "armhf" : targetName === "flatpak" ? "arm" : "armv7l"
     case Arch.arm64:
-      return targetName === "pacman" || targetName === "flatpak" ? "aarch64" : "arm64"
+      return targetName === "pacman" || targetName === "rpm" || targetName === "flatpak" ? "aarch64" : "arm64"
 
     default:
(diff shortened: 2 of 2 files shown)


mmaietta:
>It feels like just having aarch64 in the filename is not sufficient.

Can you elaborate on this? This seems like a breaking change

jkroepke:
> Can you elaborate on this? This seems like a breaking change

I post a console output. You can reproduce it by trying it out. On RedHat systems, the arch called aarch64.

In case you avoid this breaking change in the code, bump the embedded fpm version which does the same. See upstream https://github.com/jordansissel/fpm/commit/cc0a4e736941401ed6f3a5b20269d1415c53f45c#diff-a93d61ffd6fa3f2b9405387c603649469db300e9c5145c62020e47b44d71b16cR242-R243 .

mmaietta:
>Here is another example, where a users renamed the arch in the rpm to solve the issue

Sorry, can you elaborate? I'm not sure how the `aarch64` name suffix is needed. Does the install require that suffix in order to identify arch?

jkroepke:
At this
Moment, the RPM is build with arm64 arch, but there is no arm64 arch on RPM based systems. On RPM based systems, the arch is called aarch64. The arch Information is part of the RPM metadata which is currently wrong.

Please explain precisely what is still missing?

Issue #7489, opened 2023-03-22
Missing quote syntax error in nsis uninstaller


JamieRuderman:
* **Electron-Builder Version**: 1342f87

* **Node Version**: v14.18.3

* **Electron Version**: ^23.1.3"
* **Electron Type (current, beta, nightly)**: current

* **Target**: nsis

Looking through installer.nsh file as I'm trying to debug an issue and noticed a missing quote `"`

PR #7490, opened 2023-03-22
fix: Missing quote syntax error in nsis uninstaller

JamieRuderman:
for issue https://github.com/electron-userland/electron-builder/issues/7489


mmaietta:
Nice catch!

Can you please add this changeset file?
`.changeset/big-rings-design.md`
```
---
"app-builder-lib": patch
---

fix: missing quote syntax error in nsis uninstaller
```

PR #7498, opened 2023-03-24
chore(docs): clarify yalc setup instructions

NoahAndrews:
The publish and link commands were grouped together, even though they needed to be run from entirely different locations.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ pnpm install
 popd
 ```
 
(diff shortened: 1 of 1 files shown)


NoahAndrews:
Hmm, the Powershell script to apply changes to the yalc script is also broken.

mmaietta:
@NoahAndrews I actually don't know how PowerShell works tbh. I believe that was a community contribution

NoahAndrews:
It was a simple error that's now fixed in this PR.

Issue #7500, opened 2023-03-25
Use `update-alternatives` in Linux to avoid direct symlinks


markizano:
* **Electron-Builder Version**: 24.1.1

* **Node Version**: Any (current=14.21.3)

* **Electron Version**: Any (current=23.2.0)
* **Electron Type (current, beta, nightly)**: Current

* **Electron-Updater version**: N/A

* **Target**: Linux (apt/DEB based)

### What is the summary of the issue?

Electron app builder library should produce an app that leverages `update-alternatives` when installing software to Linux.

### What is the expected results?

When compiling an app using `app-builder-lib`, it should produce a packaged result that uses `update-alternatives` so userland developers can assign a higher priority alternatives that middleware software may want to run/override later.

### What is the actual results?

This library creates a hardcoded symlink to the binary executable when installing software. As a result, if a userland software wants a different script to execute in place of the binary, it'll get overwritten on each update.

### What is an example of what you are seeking?

Example: signal-desktop

Each time I run `sudo apt update` to upgrade the software, I find myself needing to also `update-alternatives` immediately after to restore the symlinks I have created in `/usr/bin/signal-desktop`.

If this change is accepted, then I can simply set a higher priority script once and not need to worry about this maintenance on my system. This is also non-impactful enough to be accepted without much after-effects as a result of the change.

If `update-alternatives` is not available, then as a fallback, it will continue to use the symlink approach as before so there won't be any misses to existing desired functionality.

There are a few Electron-based apps that I use this would improve, so I figured I'd come to the source for a solution. Signal Desktop is just one of the few apps I would think this library update would help.

I have a Pull Request I will submit along with this ticket for a fix I hope is quickly reviewed and accepted without much of issues.

markizano:
I have opened PR #7501 for a fix to this issue and it's ready for review =)

markizano:
PR Merged into master.

Now Electron based apps will use `update-alternatives` if present and remove the old symlink method if it was previously set on future installs.

Thanks much to the team for helping get this resolved!

PR #7501, opened 2023-03-25
fix: Use `update-alternatives` instead of symlinks for #7500

markizano:
Where possible, use `update-alternatives` instead to avoid hardcoding links in Linux. This will allow for downstream users to specify paths to their own executables with higher priority if they wish.

Backwards compatibility is preserved by still using the symlinking route if the command `update-alternatives` is not available.


markizano:
okay, I think I have included all the requirements for this to be accepted.

Ready for review!

mmaietta:
Thank you!
Test is failing
```
  ● executable path in postinst script

    expect(received).toMatchSnapshot()

    Snapshot name: `executable path in postinst script 5`

    - Snapshot  - 1
    + Received  + 5

    @@ -1,9 +1,13 @@
      #!/bin/bash

      # Link to the binary
    - ln -sf '/opt/foo/Boo' '/usr/bin/Boo'
    + if type update-alternatives 2>/dev/null >&1; then
    +    update-alternatives --install '/usr/bin/Boo' 'Boo' '/opt/foo/Boo' 100
    + else
    + ln -sf '/opt/foo/Boo' '/usr/bin/Boo'
    + fi
```
You'll need to run the specific test file and update the snapshot
```
TESt_FILES= debTest UPDATE_SNAPSHOT=true pnpm test
```
(Might need to run within docker image, not sure off the top of my head

markizano:
How can I update snapshots?

mmaietta:
Good to note that the docs need to be updated.
You'll need to compile before running the tests
```
pnpm compile
TEST_FILES= debTest UPDATE_SNAPSHOT=true pnpm test-linux
```

pnpm-lock.yaml
```diff
@@ -37,7 +37,7 @@ importers:
       typescript: 4.8.4
       typescript-json-schema: 0.51.0
       v8-compile-cache: 2.3.0
-    dependencies:
+    optionalDependencies:
```
mmaietta, line 40: This breaks the build

markizano:
Apologies for the sporadic work here and there on this and thanks so much for the incredible responsiveness on this, @mmaietta !!!

.husky/pre-push
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-pnpm run pretest
+npx pnpm run pretest
```
mmaietta, line 4: Can you please revert this?
Not sure why we're adding `npx` here

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-pnpm run lint-staged
+npx pnpm run lint-staged
```
mmaietta, line 4: Can you please revert this?

pnpm-lock.yaml
```diff
@@ -5139,6 +5139,7 @@ packages:
     engines: {node: '>=8'}
     os: [darwin]
     hasBin: true
+    requiresBuild: true
```
mmaietta, line 5142: Can you please revert this? It doesn't seem relevant to the changes you're looking to add

mmaietta (changes requested):

.husky/pre-commit
markizano: Okay, all reverted.

packages/app-builder-lib/templates/linux/after-install.tpl
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
 # Link to the binary
-ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
+if type update-alternatives 2>/dev/null >&1; then
+   update-alternatives --install '/usr/bin/${executable}' '${executable}' '/opt/${sanitizedProductName}/${executable}' 100
```
mmaietta, line 5: I just read this in your changeset. (My apologies for pinging back n forth on this)
> If you are having issues on install or get errors about `/usr/bin/${executable}` already
exists, then simply remove the link and `update-alternatives` will take over from there.

Do we need to update this code as follows? We can't expect all linux users to know they need to remove the link on their own during the install process
```
previousLink = "/usr/bin/${executable}"
# remove previously existing symlink if applicable
[ -L "${previousLink}" ] && [ -e "${previousLink}" ] && rm -f "${previousLink}"

update-alternatives --install "${previousLink}" "${executable}" "/opt/${sanitizedProductName}/${executable}" 100
```

Note: I didn't test this logic, just pulled resources from online and changed `'` to `"`.  So you'd need to test this change

What are your thoughts?

packages/app-builder-lib/templates/linux/after-install.tpl
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
-# Link to the binary
-ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
+# Remove previous link if it doesn't use update-alternatives
+if [ `readlink "/usr/bin/${executable}"` != "/etc/alternatives/${executable}" ]; then
+  rm -f "/usr/bin/${executable}"
+fi
+
+update-alternatives --install '/usr/bin/${executable}' '${executable}' '/opt/${sanitizedProductName}/${executable}' 100
```
mmaietta, line 8: We still need this check, right?
```
if type update-alternatives 2>/dev/null >&1; then
   update-alternatives --install '/usr/bin/${executable}' '${executable}' '/opt/${sanitizedProductName}/${executable}' 100

else
    ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
fi
```

packages/app-builder-lib/templates/linux/after-install.tpl
markizano: No worries, it was an upstream change I wanted to propose into this project and grateful for the thoughtful discussion!

> Do we need to update this code as follows? We can't expect all linux users to know they need to remove the link on their own during the install process

I think I see what you mean in the updates here, so I updated the changeset accordingly. This is very valid and for a project as core as this to so many others, it would cause a lot of heartburn.

In this corrected way, the end-user nor middleware have to take any action in consuming this update.

With the above proposed change, it might produce output durring the install about having to correct a broken link as such:

    update-alternatives: warning: forcing reinstallation of alternative /opt/${sanitizedProductName}/${executable} because link group ${executable} is broken


on each update.

Single quotes are considered string literals. Double-quotes are considered interpolatable strings. Since the templates are passing through a template processor and not the bash interpreter itself, single quotes in this case would be fine since it's not in-script variables that are being used. Since those variables are "userland" variables outside of the script (interpolated by the template filter), single quotes would also be more secure in this case, so I think it might have been correct. In-script variable use might require double-quotes.
(sorry if this sounds wordy -- I am an instructor on my TikTok & YouTube channels :D )

packages/app-builder-lib/templates/linux/after-install.tpl
```diff
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-# Link to the binary
-ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
+if type update-alternatives 2>/dev/null >&1; then
+    # Remove previous link if it doesn't use update-alternatives
+    if [ -L '/usr/bin/${executable}' -a -e '/usr/bin/${executable}' -a "`readlink '/usr/bin/${executable}'`" != '/etc/alternatives/${executable}' ]; then
```
mmaietta, line 5: I don't know what this line means, but I trust 😅

mmaietta (approved):
Kicking off CI

packages/app-builder-lib/templates/linux/after-install.tpl
markizano: No worries :)

https://ss64.com/bash/test.html

`-L` - file is symlink.
`EXPR -a EXPR` - AND operation between expressions.
`-e` - check if file exists.

Should result in:
if ( isSymLink() && isExist() && readlink(file) != "/etc/alternatives/$exec" )

markizano:
@mmaietta -- Thanks so much for working with me to get this into the project!!

Issue #7494, opened 2023-03-23
Builds failing after upgrading to electron-builder 24.x


kylekirkby:
* **Electron-Builder Version**: 24.1.0

* **Node Version**: v18.14.2

* **Electron Version**: 23.2.0 

* **Target**:  Mac OSX / Windows / Linux

After upgrading to electron-builder `24.1.0`, our builds have started failing with:

```
OSX/node_modules/app-builder-lib/src/electron/electronVersion.ts:102
error	23-Mar-2023 07:49:46	    throw new InvalidConfigurationError(
error	23-Mar-2023 07:49:46	          ^
error	23-Mar-2023 07:49:46	InvalidConfigurationError: Cannot compute electron version from installed node modules - none of the possible electron modules are installed.
error	23-Mar-2023 07:49:46	See https://github.com/electron-userland/electron-builder/issues/3984#issuecomment-504968246
```

I've tried every suggestion mentioned on https://github.com/electron-userland/electron-builder/issues/3984#issuecomment-504968246 but have had no success.

Any idea what's happening here?

mmaietta:
Oomph, I can take a look at this asap. Surprised it wasn't caught in any of the alpha versions.

Would you be willing to put together a minimum reproducible repo for this? That'd help immensely

glennmartinez:
Getting the same here, latest electron with latest electron-builder.

mmaietta:
I need more info to work with here. Please post build logs with env var `DEBUG=electron-builder` or a minimum reproducible repo/gist (such as with electron-quick-start).

Are your projects using native modules?

zibra:
same on my end

1zilc:
[same problem](https://github.com/1zilc/fishing-funds/actions/runs/4543819080/jobs/8009011417)

glennmartinez:
@mmaietta I'm using this project https://github.com/electron-react-boilerplate/electron-react-boilerplate and to reproduce just try to bump electron-builder dependency. I don't have any probs with the latest 23.alpha versions, only from 24 onwards.

mmaietta:
Hmmm does this repro for you guys on m1 devices? Or is it only on intel-based macs?

@glennmartinez I just tried that project: Set electron-builder 24.1.1 and upped electron to 23.2.0, then ran `npm run package` successfully.

1zilc:
[...]

PR #7511, opened 2023-03-30
fix: two-package.json electron version query

mmaietta:
Fixes: #7494


davej:
@mmaietta I believe this PR has broken the `v24.1.2` release. I'm getting the following error:
```
    Error: Cannot compute electron version from installed node modules - none of the possible electron modules are installed.
    See https://github.com/electron-userland/electron-builder/issues/3984#issuecomment-504968246
```

Error is resolved by downgrading back to `v24.0.0`

davej:
> Hi @davej, this one explicitly fixed a related "cannot compute electron version" for 2 package.json projects (#7494)
> 
> Can you post logs with DEBUG=electron-builder env var or provide a sample gist/repo or public repo that I can fork?

Thanks @mmaietta, yeah, it's curious that this caused the same issue but in the exact opposite direction for me. I don't have an easy repro but I'll try and put something together.

Issue #7512, opened 2023-03-30
"Can't reconcile two non-macho files" after upgrading to 24.x.x


Slapbox:
* **Electron-Builder Version**:  24.1.1

* **Node Version**: 16.17.0

* **Electron Version**: 23.3.x
* **Electron Type (current, beta, nightly)**:
 Current

* **Target**: mac-universal

It's presumably related to this underlying issue: https://github.com/electron/universal/issues/41

I know there's already 3 other issues about this, but they all seem stale and different in their cause.

Slapbox:
We also see this error instead intermittently: `Detected unique file "node_modules/argon2/bin/darwin-arm64-113" in "/Users/me/Documents/app/release/prod/mac-universal--arm64/app.app/Contents/Resources/app.asar" not covered by allowList rule: "undefined"`

**But...** the file that is causing the problem always seems to be different, whichever error type we get.

Slapbox:
Copying my comment from electron/universal#41:

-----

In my case adding `'!node_modules/**/*.{mk,a,o,h}'` just results in it failing on some other file, and when I got to the end of those it failed in another way.

```
Detected unique file "node_modules/argon2/bin/darwin-arm64-113" in "/Users/me/Documents/app/release/prod/mac-universal--arm64/app.app/Contents/Resources/app.asar" not covered by allowList rule: "undefined"
```

and then in Electron-Builder if I add that file to `singleArchFiles` then I get:

```
 Command failed: lipo /private/var/folders/y5/pp8xq1qd1f78jth_11nsxndh0000gn/T/arm64-mw1MJx/node_modules/argon2/bin/darwin-arm64-113/argon2.node /private/var/folders/y5/pp8xq1qd1f78jth_11nsxndh0000gn/T/x64-BD57LD/node_modules/argon2/bin/darwin-arm64-113/argon2.node -create -output /private/var/folders/y5/pp8xq1qd1f78jth_11nsxndh0000gn/T/x64-BD57LD/node_modules/argon2/bin/darwin-arm64-113/argon2.node
[...]

PR #7519, opened 2023-04-04
fix: "Can't reconcile two non-macho files" due to new "PreGyp Copy" functionality in electron/rebuild integration

mmaietta:
Fixes: #7512 7512


packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -148,15 +146,15 @@ export interface RebuildOptions {
...
+export async function rebuild(appDir: string, buildFromSource: boolean, frameworkInfo: DesktopFrameworkInfo, arch: Arch) {
   log.info({ arch, version: frameworkInfo.version, appDir }, "executing @electron/rebuild")
   const rootPath = await searchModule.getProjectRootPath(appDir)
   const options: electronRebuild.RebuildOptions = {
     buildPath: appDir,
     electronVersion: frameworkInfo.version,
-    arch,
-    debug: log.isDebugEnabled,
+    arch: Arch[arch],
     projectRootPath: rootPath,
```
sandeep1995, line 156: can you also add a different `cachePath` argument for each arch? 

https://github.com/electron/rebuild/blob/main/src/types.ts#L10

mmaietta: What for? Why can't the default cache be used?

PR #7524, opened 2023-04-05
fix: Fix electron-updater error handling when spawning a process asynchronously

### LINKED ISSUE #7523: NSIS: When `isAdminRightsRequired` is incorrectly set to false in `update-info.json`, updating fails
* **electron-updater version**: 6.0.2

* **Node Version**: 16.19.1

* **Target**: NSIS

https://github.com/electron-userland/electron-builder/pull/7503 may have fixed the happy case, but it (re?) introduced a problem with error handling. PR incoming.

NoahAndrews:
Fixes #7523 

When deciding if a process that we attempted to spawn actually started, we need to look at the PID of the spawned process (which will be `undefined` if it failed to start), not the PID of the original Electron process. #7503 introduced this error.

### DIFF
--- a/.changeset/tricky-flies-bow.md
+++ b/.changeset/tricky-flies-bow.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+Fixed error handling when launching updater (fixes NSIS updates when isAdminRightsRequired is incorrectly set to false)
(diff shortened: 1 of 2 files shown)


mmaietta:
Thank you!

PR #7529, opened 2023-04-06
fix: msiWrapped: Normalize target names before checking for NSIS target

NoahAndrews:
Target names are not case sensitive, and are normalized by this function: https://github.com/electron-userland/electron-builder/blob/cd7b79f8273a32d3d6a4f4c4a139d847835df637/packages/app-builder-lib/src/targets/targetFactory.ts#L67-L78

This PR makes it so that the `msiWrapped` target doesn't complain when the `nsis` target has leading/trailing spaces or is not all-lowercase.

### DIFF
--- a/.changeset/shiny-jars-fetch.md
+++ b/.changeset/shiny-jars-fetch.md
@@ -1,0 +1,6 @@
+---
+"app-builder-lib": patch
+"electron-builder": patch
+---
+
+When using the msiWrapped target, allow the nsis target to be capitalized in the configuration file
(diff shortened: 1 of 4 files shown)


mmaietta:
Sorry, I'm confused. Why do we need to allow multi-case such as `NSIS` allowed?

NoahAndrews:
We don't really _need_ to, but we should be consistent. If you're just configuring NSIS, using all-caps `NSIS` works. If you want to also add `msiWrapped`, now you get an error unless you change `NSIS` to `nsis`.

NoahAndrews:
@mmaietta Do you plan to review this eventually? It's a pretty confusing DX right now if you have a working setup that has an `NSIS` configuration block (matching the capitalization of the tool), and then you want to build a wrapped MSI and it tells you it can't find the confusion block that you know for a fact is there and is being used.

Issue #7532, opened 2023-04-10
Silent application exit in case user canceled the installation on the UAC step.


vitto-moz:
"electron-updater": "5.3.0",
but the issue is relevant for the latest version as well 

* **Electron-Builder Version**:  "23.6.0",

* **Node Version**:  v16.16.0

* **Electron Version**:  "23.1.4",

* **Target**:  Windows

We have the long-running widget application, where we have the auto-update functionality implemented. 
The application uses silent mode for update by invoking `autoUpdater.quitAndInstall(true, true)`, which works perfectly fine in case of happy path (user agreed update during UAC step), but act quite confusing in case user has the installation canceled during UAC.

The concern we have: 
The application exits automatically (which is expected and intended) and the UAC Windows dialog appears on the screen. 
Then once the user clicks CANCEL the update process is interrupted, _but the user still thinks the application is running_ (remember it's a widget, and we don't have it exposed all the time).

Would be great to have the following option: 
Extract the functionality of the update process initiation, and provide the ability of using it without app quit.
Would be something like this
`autoUpdater.install(true, true)` 

We already made a patch for our needs (just commented 2 lines) and everything works fine: the application just exits during the installation process (after the UAC is confirmed by the user and installation process is proceeded)
```
-                require("electron").autoUpdater.emit("before-quit-for-update");
-                this.app.quit();
+                // require("electron").autoUpdater.emit("before-quit-for-update");
+                // this.app.quit();
```

PR #7533, opened 2023-04-10
feat: nsis install method - exposed as public to avoid quit the app for the install

vitto-moz:
Resolves the issue: 
https://github.com/electron-userland/electron-builder/issues/7532

### DIFF
--- a/.changeset/bright-numbers-provide.md
+++ b/.changeset/bright-numbers-provide.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": minor
+---
+
+feat: nsis install method - exposed as public to avoid quit the app for the install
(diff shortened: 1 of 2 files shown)


vitto-moz:
> By making it `abstract` specific to Windows, it breaks the build. I'd recommend going an alternative route Just make this change
> 
> ```
> -  protected install(isSilent: boolean, isForceRunAfter: boolean): boolean {
> +  install(isSilent: boolean = false, isForceRunAfter: boolean = false): boolean {
> ```
> 
> Then add an import, I think you'd be able to use: `import { install } from "electron-updater/out/BaseUpdater";` or `NsisUpdater`?

@mmaietta 
have the PR updated

Issue #6885, opened 2022-05-25
Electron-builder's installer "safe directory" issue


PerfectionVR:
We’ve chosen to use electron-builder to create the installer because it’s easy to integrate. However, there is a small, mostly-inconsequential buglet concerning how it chooses its install directory.

By default, it’ll use the productName as its leaf install directory. So you get something like \%PROGRAMFILES\%/$\{productName\}. 
This is fine, unless you want to install somewhere else, like %SOMEPLACE%/foo. If that directory doesn’t exist yet, everything works fine. However, on the second install, it’ll install into %SOMEPLACE%/foo/${productName}. This is because electron-builder has an [empty-directory check](https://github.com/electron-userland/electron-builder/blob/28cb86bdcb6dd0b10e75a69ccd34ece6cca1d204/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh), which will append productName to the requested directory if the latter isn’t empty. 

It shouldn’t just do an ‘empty’ check, but also see if this dir is the current install dir.

The reason I stumbled on this is that I’d prefer to have the directory be named “example-project”. This can be done by using productName="example-project", but then that also shows up in the text of the installer. I’d prefer that to be “Example Project”, but then the directories get weird.

For w use “example-project”, but we would like to know how this can be resolved.

Details:

```
node -v # v16.15.0
yarn -v # 1.22.15
````
Windows

**package.json**
```json
{
  "name": "example-project",
  "description": "Example Project",
  "version": "0.0.1",
  "license": "MIT",
  "scripts": {
    "build:start": "node ./scripts/run-script",
    "build:clean": "rimraf .dist",
    "build:prepare": "node ./build.prepare.js",
    "build:main": "webpack --config webpack.config.main.js --mode=production",
    "build:renderer": "webpack --config webpack.config.renderer.js --mode=development",
    "build:main:dev": "webpack --watch --mode=development --config webpack.config.main.js",
    "build:renderer:dev": "webpack serve --port 3000 --mode=development --config webpack.config.renderer.js",
    "build:compile": "yarn electron-builder",
    "build": "yarn build:clean && npm-run-all --parallel build:prepare build:main build:renderer && yarn build:compile",
    "dev": "yarn build:clean && npm-run-all --parallel build:main:dev build:renderer:dev",
    "dev:start": "nodemon --watch ./.dist/www/index.js --exec electron ./.dist/www/index.js",
  },
  "dependencies": {
    "axios": "^0.27.2",
[...]

PR #7552, opened 2023-04-20
fix(nsis): Ensure application name sub-folder on fresh installs.

p2004a:
During the first program install when the selected directory doesn't exist it's not enforced that path contains application name as sub-folder. On subsequent installs e.g. on update, that condition is enforces though, which creates a new installation in sub-folder.

This tries to address #6885

### DIFF
--- a/.changeset/two-goats-shop.md
+++ b/.changeset/two-goats-shop.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(nsis): Ensure application name sub-folder on fresh installs.
--- a/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh
+++ b/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh
@@ -31,11 +31,9 @@
 
     # sanitize the MUI_PAGE_DIRECTORY result to make sure it has a application name sub-folder
     Function instFilesPre
-      ${If} ${FileExists} "$INSTDIR\*"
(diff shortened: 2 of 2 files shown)


mmaietta:
@p2004a, thanks for the contribution! Just to confirm, how were you able to test this fix? Do you have a `patch-package` 
(or something similar) that you were applying locally to your builds to test with?

p2004a:
@mmaietta Sorry for slow response: I've tested it by applying the change locally in the `node_modules` of the electron package I'm building.

Here is the behavior tested in Windows Sandbox:
- before PR: https://user-images.githubusercontent.com/2212599/235661468-d8014524-3985-4620-9f67-1a97376de456.mp4
- with this PR: https://user-images.githubusercontent.com/2212599/235661592-f7dd61a5-51de-4bce-adb6-a4babc1dcd68.mp4

p2004a:
It's not reverting that change, but amending/fixing it. 958a7ae2bfe1b86cb7318e12c60bff25d65c9221 made sure that when user installs application in location of their choosing AND that location is non-empty, it's appending the `${APP_FILENAME}` to install path.

That resolved #1298 because without that check, user could pick non-empty e.g. `C:\games`, program would be installed directly there, and when when uninstalling *everything* in `C:\games` would be dropped, not only application's files.

My change is dropping the "AND that location is non-empty" from the 958a7ae2bfe1b86cb7318e12c60bff25d65c9221 because guess what? When update happens, the previous install directory is non empty, so installer creates a new subdirectory.

mmaietta:
Great explanation. LGTM

PR #7558, opened 2023-04-24
feat: allow specifying recommended dependencies for deb target

t3chguy:
Closes https://github.com/develar/app-builder/issues/87

`deb.fpm: ['--deb-recommend=blah']` is insufficient as https://github.com/develar/app-builder always forces a `libappindicator3-1` into the recommendeds.

### DIFF
--- a/packages/app-builder-lib/src/targets/FpmTarget.ts
+++ b/packages/app-builder-lib/src/targets/FpmTarget.ts
@@ -1,5 +1,5 @@
 import { path7za } from "7zip-bin"
-import { Arch, executeAppBuilder, getArchSuffix, log, TmpDir, toLinuxArchString, use, serializeToYaml } from "builder-util"
+import { Arch, executeAppBuilder, getArchSuffix, log, TmpDir, toLinuxArchString, use, serializeToYaml, asArray } from "builder-util"
 import { unlinkIfExists } from "builder-util/out/fs"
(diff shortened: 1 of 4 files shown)


packages/app-builder-lib/src/targets/FpmTarget.ts
```diff
@@ -196,6 +196,22 @@ export default class FpmTarget extends Target {
...
+          // noinspection SuspiciousTypeOfGuard
+          if (typeof recommends === "string") {
+            fpmConfiguration.customRecommends = [recommends as string]
+          } else {
+            throw new Error(`recommends must be Array or String, but specified as: ${recommends}`)
+          }
+        }
+      }
+    }
+
```
mmaietta, line 214: Since the schema validator will automatically flag incorrect types being passed in, we can simplify this to use `asArray`. I think we should also check a general falsey for `if (recommends) {...}`, as I don't think we should allow an empty string to be passed into customRecommends. What are your thoughts?
```
import { asArray } from "builder-util"

if (target === "deb") {
  const recommends = (options as DebOptions).recommends
  if (recommends) {
      fpmConfiguration.customRecommends = asArray(recommends)
  }
}
```

mmaietta (approved):

t3chguy:
@mmaietta thanks for landing the PRs

:shipit:

PR #7581, opened 2023-05-23
feat: add mips64el and loongarch64 architecture support

yujianchi:
add mips64el and loongarch64 architecture support


mmaietta:
These Arch's need to be added here and then the schema regenerated via `pnpm generate-all`
https://github.com/electron-userland/electron-builder/blob/80a3ae857b9305aff1774e1b2b932601af50231b/packages/builder-util/src/arch.ts#L4

mmaietta:
Closing due to inactivity

PR #7600, opened 2023-06-06
fix(mac): Wrap hditutil detach in retry w/ backoff

roryabraham:
We are [experiencing a lot of flakiness](https://github.com/Expensify/App/issues/20283) with our Electron deploys in a GitHub Actions macos-12 runner. This PR is an attempt to wrap the failing command in a retry, as has been done elsewhere in this repo.

I also added a simple backoff mechanism to the retry function.

### DIFF
--- a/.changeset/large-items-argue.md
+++ b/.changeset/large-items-argue.md
@@ -1,0 +1,6 @@
+---
+"builder-util": patch
+"dmg-builder": patch
+---
+
+fix(mac): wrap hdiutil detach in retry w/ backoff
--- a/packages/builder-util/src/util.ts
+++ b/packages/builder-util/src/util.ts
@@ -396,14 +396,14 @@ export function executeAppBuilder(
   }
 }
 
-export async function retry<T>(task: () => Promise<T>, retriesLeft: number, interval: number): Promise<T> {
+export async function retry<T>(task: () => Promise<T>, retriesLeft: number, interval: number, backoff = 0, attempt = 0): Promise<T> {
   try {
     return await task()
   } catch (error: any) {
     log.info(`Above command failed, retrying ${retriesLeft} more times`)
     if (retriesLeft > 0) {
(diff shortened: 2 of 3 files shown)


roryabraham:
First-time contributor here, so happy to do whatever is needed to get this relatively simple change ready for review.

AndrewGable:
FYI - We [tested the fix here](https://github.com/Expensify/App/actions/runs/5338582643/jobs/9676148081) and it worked as intended in fixing the error we were seeing earlier. Nice work @roryabraham!

mmaietta:
Pretty sure there's no way to detect this from javascript side. All dmg-builder is doing is calling `hdiutil` directly via exec/spawn. There's no callback support to observe the corresponding events.

We do try to use `detach --force` on failure, but even then, it appears the merged fix still does not resolve the issue for all developers.

In a previous project, I had Obj-C code (via a `.node` module) that tapped into the NSNotificationCenter to observe system-level events of dmg mount/unmounting for native support of dmg auto-updates. I've previously attempted to do the same thing with javascript and it doesn't work 😓

mmaietta:
@laurent22 can you create a new Github issue for this and ping me there/link it here? I'd rather contain discussions there as I forsee this to be a larger topic and not best suited as comments on a merged PR.

I'm whipping up a `patch-package` that you can test with for pulling more debug data. 🙂

roryabraham:
was another issue every created? this issue has started coming back to haunt us

mmaietta:
@roryabraham I don't think a new one exists. Are you on the same version of electron-builder and the issue came back? Or did you update electron-builder? I recall there's a new retry mechanism for hdiutil that was added about 4 months ago for wrapping the retry for all hdiutil commands: https://github.com/electron-userland/electron-builder/blob/master/packages/dmg-builder/src/hdiuil.ts

roryabraham:
We are currently on `24.13.2`, we'll try updating to `25.0.0`. Thanks 👍🏼

Issue #7570, opened 2023-05-09
Build dmg and MAS with different CFBundleIdentifier


GabrielNSD:
* **Electron-Builder Version**: 23.6.0

* **Node Version**: 14.21.3

* **Electron Version**: 20.3.0
* **Electron Type (current, beta, nightly)**:

* **Target**: dmg, zip and mas

Hi there! I have an app that is currently just distributed by .dmg files. I use the electron-update to provide updates for the users.  
Now we want to publish the app in the App Store, for doing so the CFBundleIdentifer must match with the one already registered in the App Store. The problem is that we have been releasing the app with a different appId.  
Just changing the appId to the new one will break the update for the installed .dmg.  
My idea is to keep keep building the dmg using the old id and build the MAS using the "new" id.  
I have tried doing that by change my build configuration, either by using `extendInfo` in the `mas` key and add the key `CFBundleIdentifier` with the new id, what makes the .pkg file being able to upload by transporter, but making dmg impossible to update since it also changes the Info.plist of the .zip. Or tried to put the `CFBundleIdentifier` key directly in the `entitlements.mas.plist` file, which had no effect :/.
Here is part of my json file with build configuration:

```json
{...
    "mac": {
      "icon": "src/assets/icons/icon.icns",
      "appId": "com.old-id.app",
      "category": "public.app-category.education",
      "target": [
        "dmg",
        "zip",
        "mas"
      ],
      "files": [
        "entitlements.mac.plist",
        "entitlements.mas.plist"
      ],
      "hardenedRuntime": true,
      "entitlements": "entitlements.mac.plist",
      "extendInfo": {
        "NSMicrophoneUsageDescription": "Please give us access to your microphone"
      }
    },
    "mas": {
      "entitlements": "entitlements.mas.plist"
    },
...}
```

Is there a way to achieve this? Keep the old CFBundleIdentifier in the .zip file and use the new id only for the MAS (.pkg)?

mmaietta:
I think you can set `bundleVersion ` uniquely on `mac` and `mas` targets, and that should override the CFBundleVersion to be different per platform. I'm just running off of memory here though, don't have the code handy to check the behavior.

GabrielNSD:
Hi @mmaietta , thank you for the answer. As far as I understood, the CFBundleIdentifier is the key used by electron-updater to locate the bundle that is eligible to update. The `bundleVersion` overrides the CFBundleVersion, which I am not having trouble

[...]

PR #7603, opened 2023-06-07
fix: Allow building MAS and dmg targets with different appId

GabrielNSD:
This changes allows packaging different macOS targets with different CFBundleIndetifiers.  
It makes the `createMacApp` to look for a CFBundleIdentifier in mas config, if the app is MAS.
  
Solving the issue related here: #7570 

Current implementation ignores any `appId` specified in `mas`, directly using `appInfo.macBundleIdentifier` when writing Plist file.

### DIFF
--- a/packages/app-builder-lib/src/electron/electronMac.ts
+++ b/packages/app-builder-lib/src/electron/electronMac.ts
@@ -118,10 +118,12 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
     log.warn("build.helper-bundle-id is deprecated, please set as build.mac.helperBundleId")
   }
 
-  const cfBundleIdentifier = filterCFBundleIdentifier((isMas ? packager.config.mas?.appId : packager.platformSpecificBuildOptions.appId) || appInfo.macBundleIdentifier)
+  const defaultAppId = packager.platformSpecificBuildOptions.appId
+  const cfBundleIdentifier = filterCFBundleIdentifier((isMas ? packager.config.mas?.appId : defaultAppId) || defaultAppId || appInfo.macBundleIdentifier)
 
+  const defaultHelperId = packager.platformSpecificBuildOptions.helperBundleId
   const helperBundleIdentifier = filterCFBundleIdentifier(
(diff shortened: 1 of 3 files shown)


packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -117,7 +117,14 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
   if (oldHelperBundleId != null) {
     log.warn("build.helper-bundle-id is deprecated, please set as build.mac.helperBundleId")
   }
-  const helperBundleIdentifier = filterCFBundleIdentifier(packager.platformSpecificBuildOptions.helperBundleId || oldHelperBundleId || `${appInfo.macBundleIdentifier}.helper`)
+
+  const cfBundleIdentifier = filterCFBundleIdentifier((isMas ? packager.config.mas?.appId : packager.platformSpecificBuildOptions.appId) || appInfo.macBundleIdentifier)
```
mmaietta, line 121: This will break previous logic if `isMas = true` & `mas?.appId = null`
```
(isMas ? packager.config.mas?.appId : packager.platformSpecificBuildOptions.appId)
```

I think it needs be
```
const defaultAppId = packager.platformSpecificBuildOptions.appId
const cfBundleIdentifier = filterCFBundleIdentifier((isMas ? packager.config.mas?.appId : defaultAppId) || defaultAppId || appInfo.macBundleIdentifier)

const defaultHelperId = packager.platformSpecificBuildOptions.helperBundleId
const helperBundleIdentifier = filterCFBundleIdentifier(
    (isMas ? packager.config.mas?.helperBundleId : defaultHelperId) || defaultHelperId || oldHelperBundleId || `${cfBundleIdentifier}.helper`
  )
```
I honestly don't like the way to write it, but `defaultAppId` is required twice, unfortunately.

GabrielNSD: Thanks for the heads up @mmaietta ! You are right, I have pushed a commit with your suggestion.

PR #7610, opened 2023-06-13
build: update Dockerfile -- Outdated Node version v14 --> v18

jbool24:
node v14 is no longer maintained. Update node to current LTS version.

### DIFF
--- a/.changeset/lemon-cows-thank.md
+++ b/.changeset/lemon-cows-thank.md
@@ -1,0 +1,5 @@
+---
+
+---
+
+chore(docker): update Dockerfile -- Update `latest` tag from Node version v14 --> v18
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -6,12 +6,14 @@ DATE=$(date +%m.%y)
 
 docker build -t electronuserland/builder:base -t "electronuserland/builder:base-$DATE" docker/base
 
-# Node 14
-docker build --build-arg NODE_VERSION=14.19.3 -t electronuserland/builder:14 -t "electronuserland/builder:14-$DATE" -t electronuserland/builder:latest docker/node
+## NOTE: Order the latest to oldest versions. The most recent node LTS should be tagged as the latest image
 
-docker build --build-arg IMAGE_VERSION=14 -t electronuserland/builder:14-wine -t "electronuserland/builder:14-wine-$DATE" -t electronuserland/builder:wine docker/wine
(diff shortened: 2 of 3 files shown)


mmaietta:
We already build a node 18 image.
https://github.com/electron-userland/electron-builder/blob/master/docker/build.sh#L24
Tag `electronuserland/builder:18` (-wine, -wine-mono, -wine-chrome)

Seems the docs need updating though 😅 
https://www.electron.build/multi-platform-build#docker

jbool24:
@mmaietta ok, thanks Mike. I'll check and update my scripts with that container.

jbool24:
@mmaietta should we not update the latest tag with the current Node LTS version? And in reverse make v14 or v16 available as tags

jbool24:
@mmaietta I'm confused what you mean. I see in the [build.sh](../blob/master/docker/build.sh) its doing multiple tags here but the one I'm concerned with is line 10. You can see that its double tagging v14 as also 'latest' which is the default that will pull if you just use `docker pull electronuserland/build` that line is the one I think should swap with v18. In essence this file will reverse its order such that the latest LTS release of node should be added to the top and tagged with 'latest'. I can add the proposal to this PR to make more sense what I'm saying. I don't think this causes backwards compat issues unless the end user did what I did and didn't get more specific on my `docker pull` command.

```Bash
# Node 14
docker build --build-arg NODE_VERSION=14.19.3 -t electronuserland/builder:14 -t "electronuserland/builder:14-$DATE" -t electronuserland/builder:latest docker/node
```

jbool24:
With these changes, I'm proposing that we always tag the `latest` and `wine` images with the LTS version of Node. End users can always be specific to make sure they are locally using the exact versions as it always was, such as `:14` `:14-wine` or `:14-wine-chrome` etc

mmaietta:
Makes sense, I'm in favor 🙂 

Linux tests are failing though with
```
+ docker build --build-arg NODE_VERSION=18.16.1 -t electronuserland/builder:18 -t electronuserland/builder:18-06.23 docker/node -t electronuserland/builder:latest docker/node
"docker build" requires exactly 1 argument.
```
`docker/node` is being provided twice.

jbool24:
> Makes sense, I'm in favor 🙂 
> 
> 
> 
> Linux tests are failing though with
> 
> ```
> 
> + docker build --build-arg NODE_VERSION=18.16.1 -t electronuserland/builder:18 -t electronuserland/builder:18-06.23 docker/node -t electronuserland/builder:latest docker/node
> 
> "docker build" requires exactly 1 argument.
> 
> ```
> 
> `docker/node` is being provided twice.

Thanks for the catching that typo 🙏

Issue #7620, opened 2023-06-17
mac: ambiguous (matches "Developer ID Application: X (X)" in


lutzroeder:
* **Electron-Builder Version**: 24.4.0

* **Target**: `npx electron-builder --mac --universal --publish never`

After updating to 24.4.0, publishing to macOS `zip` and `dmg` regressed with this error:

```
  ⨯ Command failed: codesign --sign Developer ID Application: AAAA (BBBB) --force --keychain /var/folders/rs/xxxxx/xxxx.keychain --timestamp --options runtime --entitlements ~/Projects/xxxx/node_modules/app-builder-lib/templates/entitlements.mac.plist ~/Projects/xxxx/dist/mac-universal/XXXX.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Resources/af.lproj/locale.pak
Developer ID Application: AAAA (BBBB): ambiguous (matches "Developer ID Application: AAAA (BBBB)" in /private/var/folders/rs/xxxx/xxxx.keychain and "Developer ID Application: AAAA (BBBB)" in /Users/lutzroeder/Library/Keychains/login.keychain-db)
``` 

@filfreire @mayfield @mmaietta [this](https://github.com/electron-userland/electron-builder/blob/00136aa6889ec81de8f0125c05cc3edfd4a5f3e1/packages/app-builder-lib/src/macPackager.ts#L300) change in #7317 seems to be the root cause.

`identity.name` being present in both `.keychain` files previously worked as `identity.hash` was passed to `codesign` instead of `identity.name`.

mmaietta:
@lutzroeder can you try this patch via patch-package and see if it resolves the problem for you?
```
diff --git a/node_modules/app-builder-lib/out/macPackager.js b/node_modules/app-builder-lib/out/macPackager.js
index cfa4b51..bd1da47 100644
--- a/node_modules/app-builder-lib/out/macPackager.js
+++ b/node_modules/app-builder-lib/out/macPackager.js
@@ -162,6 +162,7 @@ class MacPackager extends platformPackager_1.PlatformPackager {
         }
     }
     async sign(appPath, outDir, masOptions, arch) {
+        var _a;
         if (!(0, macCodeSign_1.isSignAllowed)()) {
             return false;
         }
@@ -246,7 +247,7 @@ class MacPackager extends platformPackager_1.PlatformPackager {
                   https://github.com/electron-userland/electron-builder/issues/5383
                   */
             },
[...]

PR #7622, opened 2023-06-20
fix(mac): use Identity `hash` instead of `name` if it exists

mmaietta:
Fixes: #7620


mmaietta:
Broken test is unrelated to this change. Merging

PR #7630, opened 2023-06-24
fix(linux): make semver pre-release versions valid for "pacman" and "rpm" target

### LINKED ISSUE #7601: invalid version for semver pre-releases when targeting `"pacman"`
* **Electron-Builder Version**: 23.6.0

* **Node Version**: v20.2.0

* **Target**: pacman

building for target `"pacman"` with a [semver compatible pre-release version](https://semver.org/#spec-item-9) (e.g. `0.0.0-alpha.0`, `3.0.0-preview.19`) results in an invalid version format, since [arch linux does not allow hyphens in package versions](https://wiki.archlinux.org/title/PKGBUILD#pkgver). installing the package then results not in installing the desired `package-name@0.0.0-alpha.0` but instead `package-name-0.0.0@alpha.0`.

changing the hyphen to an underscore (like the arch wiki recommends) or a period (also common, though not recommended) results in an `Invalid version: "0.0.0_alpha.0"` error.

i am unsure how exactly to deal with this, but i would probably recommend following the advice from the arch wiki and replacing the hyphen with an underscore when building for the `"pacman"` target.

m4rch3n1ng:
fix #7601

this replaces all hyphens (`-`) in versions when targeting `"pacman"` with underscores (`_`) to follow [the pacman version format](https://wiki.archlinux.org/title/PKGBUILD#pkgver).

#### notes:

* i am only replacing the version of the package, not of the `.pacman` file, since doing so would (i think) require overwriting the version in `appInfo.version`, which is marked as readonly,
https://github.com/electron-userland/electron-builder/blob/285aa766c2675448689f2e465b6fa2b2acacdbc6/packages/app-builder-lib/src/appInfo.ts#L24
or adding a seperate rule to the `macroExpander`, which, as far as i can tell, does not have access to the current build target (which could be solved fairly easily with a new optional parameter tho)
https://github.com/electron-userland/electron-builder/blob/285aa766c2675448689f2e465b6fa2b2acacdbc6/packages/app-builder-lib/src/util/macroExpander.ts#L4
neither of which seemed like a good idea to me, especially without prior feedback.
it also doesn't matter for `pacman` and looks more consistent with the other installer files, if unchanged.

* i am using `.replace(/-/g, "_")` over `.replaceAll("-", "_")` since the latter was only added in [node v15.0.0](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility) and this package still [supports node v14](https://github.com/electron-userland/electron-builder/blob/285aa766c2675448689f2e465b6fa2b2acacdbc6/package.json#L77)

* i have a very limited idea on how this codebase is structured so tell me if i should do something differently

### DIFF
--- a/.changeset/honest-comics-bake.md
+++ b/.changeset/honest-comics-bake.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(pacman): replace hyphens with underscores in version
--- a/.changeset/honest-comics-bake.md
+++ b/.changeset/honest-comics-bake.md
@@ -2,4 +2,4 @@
 "app-builder-lib": patch
 ---
 
-fix(pacman): replace hyphens with underscores in version
+fix(linux): make semver pre-release versions valid for `"pacman"` and `"rpm"` target
--- a/packages/app-builder-lib/src/targets/FpmTarget.ts
+++ b/packages/app-builder-lib/src/targets/FpmTarget.ts
@@ -152,7 +152,7 @@ export default class FpmTarget extends Target {
       "--description",
       smarten(target === "rpm" ? this.helper.getDescription(options)! : `${synopsis || ""}\n ${this.helper.getDescription(options)}`),
       "--version",
-      target == "pacman" ? appInfo.version.replace(/-/g, "_") : appInfo.version,
+      this.helper.getSanitizedVersion(target),
       "--package",
       artifactPath,
     ]
--- a/packages/app-builder-lib/src/targets/FpmTarget.ts
+++ b/packages/app-builder-lib/src/targets/FpmTarget.ts
@@ -152,7 +152,7 @@ export default class FpmTarget extends Target {
       "--description",
       smarten(target === "rpm" ? this.helper.getDescription(options)! : `${synopsis || ""}\n ${this.helper.getDescription(options)}`),
       "--version",
-      appInfo.version,
+      target == "pacman" ? appInfo.version.replace(/-/g, "_") : appInfo.version,
       "--package",
       artifactPath,
     ]
--- a/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
+++ b/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
@@ -76,6 +76,20 @@ export class LinuxTargetHelper {
     return options.description || this.packager.appInfo.description
   }
 
+  getSanitizedVersion(target: string) {
+    const {
+      appInfo: { version },
+    } = this.packager
+    switch (target) {
+      case "pacman":
+        return version.replace(/-/g, "_")
+      case "rpm":
+        return version.replace(/-/g, "~")
+      default:
+        return version
+    }
+  }
+
   async writeDesktopEntry(targetSpecificOptions: LinuxTargetSpecificOptions, exec?: string, destination?: string | null, extra?: { [key: string]: string }): Promise<string> {
     const data = await this.computeDesktopEntry(targetSpecificOptions, exec, extra)
     const file = destination || (await this.packager.getTempFile(`${this.packager.appInfo.productFilename}.desktop`))


mmaietta:
Thanks @m4rch3n1ng for the PR! Quick Q, is the `_` only for `pacman` version format or would other distros also make use of this?

m4rch3n1ng:
good call actually, did some quick research.

* [debian seems to allow hyphens](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version)
* the only reference i found for flatpak was this: https://github.com/flatpak/flatpak/issues/3887
* [rpm does not allow (or at least like) hyphens and encourages you to use a `~`](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_handling_non_sorting_versions_with_tilde_dot_and_caret)
* AppImage seems to work with it, though i don't know if AppImageUpdate will properly update it?
image_2023-06-25_19-43-31
* snaps also seem to support hyphens, though i also don't know how updates work then
image_2023-06-25_19-33-49
image_2023-06-25_19-34-46
* apks (which i assume are for alpine linux not android? if so you should probably rename them to `.alpine` to avoid confusion as i am seemingly not the only one that got caught of guard) [seems to be fine with hyphens](https://pkgs.alpinelinux.org/package/edge/community/riscv64/kde-cli-tools)
image
* i have literally no idea how i should even check for freebsd, i don't know how that os works sorry but the manifest seems to at least work? maybe?
image

anyways that's what i found

mmaietta:
Great investigative work!
I tried to find `apk` but don't seem to have a way to run it locally.

Can you check `apk --check 0.0.0_alpha.1_r1 && echo $?` If that's a `0`, then a ternary on `['apk', 'pacman'].includes(target)` may be best? Due to the wide variety of targets, it's hard not to have ternary without a ton of refactoring into subtargets
https://github.com/electron-userland/electron-builder/blob/8ae90611993c3b20fe2d30037ff3a68f9e5a8224/packages/app-builder-lib/src/linuxPackager.ts#L50-L57

m4rch3n1ng:
alpine seems to be a lot more annoying than i thought and i literally cannot find any proper specs on their version format apart from [this build script](https://gitlab.alpinelinux.org/alpine/abuild/-/blob/master/abuild.in)
image

i also need to figure out how flatpak does it which is currently annoying me because i can't get it to build for me (it just gives me a `stackTrace=Error: flatpak failed with status code 1` and i don't know what the problem is) and i can literally not figure out how to even set a version while manually using flatpak-builder.

my concern with a ternary would be that it would have to be nested
```
"pacman" == target ? appInfo.version.replace(/-/g, "_") : target === "rpm" ? appInfo.version.replace(/-/g, "~") : appInfo.version
```
which is ugly and hard to read, so i thought about perhaps doing it in a seperate function with a switch statement? but i wasn't sure where

m4rch3n1ng:
tested some more with alpine and surprisingly it seems to. just work like that? i don't honestly know. can't execute the file because i do not have a gui / display server here and i have no idea what would happen if you were to want to publish it, though to be fair i think it's fair to assume you don't really want to publish your pre-release versions, but it installed correctly as `testing-app` instead of how pacman did it. it should probably be fine to just leave it as is?

image

m4rch3n1ng:
super unsure on what to do with solaris and flatpak, but i think it should be fine to just fix it for rpm and pacman for now perhaps. if it's just those two i could actually just use a ternary i guess, tho it's still not ideal.

mmaietta:
Fantastic debugging work. Thank you

Okay, let's pull a helper function to `LinuxTargetHelper.ts` (we already leverage it for `getDescription`). Add this between `getDescription` and `writeDesktopEntry`.
```
  getSantitizedVersion(target: string) {
    const {
      appInfo: { version },
    } = this.packager
    switch (target) {
      case "pacman":
        return version.replace(/-/g, "_")
      case "rpm":
        return version.replace(/-/g, "~")
      default:
        return version
    }
  }
```
In `FpmTarget.ts`, called as.
```
      "--version",
      this.helper.getSantitizedVersion(target),
```

m4rch3n1ng:
done

mmaietta:
Many thanks for your contribution!

PR #7633, opened 2023-06-26
fix: change typed-emitter to tiny-typed-emitter 

s00d:
The "typed-emitter" package pulls in a number of dependencies, including the large "rxjs" library.


mmaietta:
I was under the impression typed-emitter was required in order for the emitter type-defined event names to be available in production bundle?

Issue #7579, opened 2023-05-19
How to disable preAutoEntitlements


Xing1P:
* **Electron-Builder Version**: 23.6.0

* **Node Version**: 16.15.1

* **Electron Version**:24.2.0
* **Electron Type (current, beta, nightly)**:current

* **Target**: mas

How to disable preAutoEntitlements osx-sign when build with Electron-builder

mmaietta:
Closing. It was merged into electron-builder v24.4 IIRC

PR #7642, opened 2023-06-30
feat:  Added support for overriding ‘preAutoEntitlements’ for electron/osx-sign

taozhou-glean:
We recently upgraded from v22 to v24, and start getting `Not available for testing` in testflight, after comparing the entitlements, it looks like the issue is due to auto entitlements from osx-sign, after disabling it locally the issue is resolved.

Also this should address https://github.com/electron-userland/electron-builder/issues/7579

default should be true as osx-sign is checking for false with `===`: https://github.com/electron/osx-sign/blob/cd9e8f1146610fb0f1bb0d88a5219b208594e5b1/src/sign.ts#L211

### DIFF
--- a/.changeset/four-starfishes-invite.md
+++ b/.changeset/four-starfishes-invite.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: Added support for overriding ‘preAutoEntitlements’ for electron/osx-sign
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -2679,6 +2679,11 @@
           ],
           "description": "Options to use for @electron/notarize (ref: https://github.com/electron/notarize).\nSupports both `legacy` and `notarytool` notarization tools. Use `false` to explicitly disable\n\nNote: You MUST specify `APPLE_ID` and `APPLE_APP_SPECIFIC_PASSWORD` via environment variables to activate notarization step"
         },
+        "preAutoEntitlements": {
+          "default": true,
+          "description": "Whether to enable entitlements automation from @electron/osx-sign.",
+          "type": "boolean"
+        },
         "protocols": {
           "anyOf": [
             {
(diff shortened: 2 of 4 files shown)


mmaietta:
@taozhou-glean I'm checking this again as I'm not sure about the purpose of `preAutoEntitlements` being `true`.  Should it always be `false` for `mas` builds? Wondering if `false` should be the default value.

taozhou-glean:
> @taozhou-glean I'm checking this again as I'm not sure about the purpose of `preAutoEntitlements` being `true`. Should it always be `false` for `mas` builds? Wondering if `false` should be the default value.

Honestly I did not dig too deep on what auto entitlement does for mas, but it default to true from osx-sign, and I do know one attribute it added and required for mas is the `ElectronTeamId`: https://www.electronjs.org/docs/latest/tutorial/mac-app-store-submission-guide#extra-steps-without-electron-osx-sign, so if disable `preAutoEntitlements`, then `ElectronTeamId` should be provided through `extendInfo`.

pushkin-:
I'm trying to follow along, but am confused. Is the current thought that `preAutoEntitlements` should be disabled for apps we want to upload to to the app store, but we don't exactly know why? From experimentation, keeping it enabled causes TestFlight to fail the verification step, but if I disable it, it goes through. But I don't know if I should keep it disabled since it seems to work, or figure out how to make it work with auto entitlements being enabled.

taozhou-glean:
> From experimentation, keeping it enabled causes TestFlight to fail the verification step, but if I disable it, it goes through. 

this is exactly what happened to me before and what triggered me to send out this change. My assumption and also after comparing the final entitlements between two builds are, the autoentitlement adds the team id to every file but essentially it should just add it to the app file only. So to me it seems something with the osx-sign package, but did not have time to dig deeper, and also we want to have more granular control over entitlement, so disabling auto entitlement just fits us better. But feel free to dig deeper and propose some fix to the osx-sign ;)

PR #7646, opened 2023-07-04
feat: add support for riscv64 architecture

kxxt:
Hi, I would like to add riscv64 support for electron-builder.

I would really appreciate it if you could make a new release shortly after this PR is merged.

### DIFF
--- a/test/src/helpers/downloadElectron.ts
+++ b/test/src/helpers/downloadElectron.ts
@@ -48,6 +48,10 @@ export function downloadAllRequiredElectronVersions(): Promise<any> {
         ? ["ia32", "x64"]
         : require(`${path.join(__dirname, "../../..")}/packages/builder-util/out/util`).getArchCliNames()
     for (const arch of archs) {
+      if (arch === "riscv64") {
+        // No prebuilt electron for riscv64
+        continue
+      }
       versions.push({
         version: ELECTRON_VERSION,
         arch,
--- a/.changeset/violet-days-refuse.md
+++ b/.changeset/violet-days-refuse.md
@@ -1,0 +1,7 @@
+---
+"app-builder-lib": minor
+"builder-util": minor
+"electron-builder": minor
+---
+
+feat: add support for riscv64 architecture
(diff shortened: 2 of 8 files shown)


kxxt:
PR updated. The download of electron is skipped for riscv64 architecture because electron doesn't publish binaries for riscv64 now.

BTW it seems that something else is causing the CI to fail.

test/src/helpers/downloadElectron.ts
```diff
@@ -48,6 +48,9 @@ export function downloadAllRequiredElectronVersions(): Promise<any> {
         ? ["ia32", "x64"]
         : require(`${path.join(__dirname, "../../..")}/packages/builder-util/out/util`).getArchCliNames()
     for (const arch of archs) {
+      if (arch === "riscv64")
+        // No prebuilt electron for riscv64
+        continue
```
mmaietta, line 53: Please wrap this with `{}`

```
      if (arch === "riscv64") {
         // No prebuilt electron for riscv64
         continue
      }
```

kxxt: Done

mmaietta:
Would you be willing to add this test to `linuxPackagerTest.ts`?
```
test.ifNotWindows(
  "AppImage riscv64",
  app({
    targets: Platform.LINUX.createTarget("appimage", Arch.riscv64),
  })
)
```
I'd like to make sure all logic passes with the new arch added

kxxt:
Hi, @mmaietta. Sorry for my late reply.

> I'm not really in favor of fielding github issues that would relate to this

I don't expect there will be such github issues because for now AFAIK the number of developers building electron for riscv64 is very small. And you can ping me for such issues :). 

> If you're looking to provide for a specific arch, I'm wondering if you can use the electronDist function callback directly to pass in your custom electron build?

It almost works. I am packaging wire-desktop for Arch linux riscv64. [It uses electron-builder as a library](https://github.com/wireapp/wire-desktop/blob/linux/3.31.3060/bin/build-tools/lib/build-linux.ts).

The `architecture` becomes `undefined` somehow and caused app-builder to fail. Maybe it is caused by missing riscv64 in the `Arch` enum, which can be fixed by this PR.

mmaietta:
Okay, still not in favor of adding a specific arch that doesn't have an official download supported by electron upstream.

What if we added an `Arch.custom` that requires a `zip` file of the custom electron build? You could therefore use it for any arch and the approach would be reusing the current `electronDist: string | function` in the electron builder config. I think that'd be a more extensible approach.
What are your thoughts?

PR #7648, opened 2023-07-05
fix: re-enable changeDir step for assisted, perMachine installs

l3m0nqu1z:
Hi,
There is no option to change the target directory during the installation process(assisted+perMachine) for MSI. It says 'You can change the default installation folder' but, actually, you can't. This PR should fix it.
1. 

2.


mmaietta:
Thanks for contributing!

Issue #7665, opened 2023-07-13
Beta/Alpha channels do not fetch Latest version from GitHub when attempting to update


sethjray:
* **Electron-Builder Version**: 23.6.0

* **Node Version**: 18.10.0

* **Electron Version**: 23.2.1

* **Electron Type (current, beta, nightly)**: current

* **Electron-Updater Version**: 5.3.0

* **Target**: Windows

We are attempting to use the [channels feature](https://www.electron.build/tutorials/release-using-channels.html) that was added to electron-updater in #6505. However, there is an issue where if our channel is set to beta or alpha, (v1.4.3-beta on the beta channel for example) autoUpdater does not see the latest release (v1.5.0 for example). 

After looking at GitHubProvider.ts, it seems that `isCustomChannel` is incorrectly being set to true when the release it is checking is latest. 
`const isCustomChannel = !["alpha", "beta"].includes(String(hrefChannel))`
Because of this, all of our latest/stable releases are treated as custom channels and are not downloaded when beta or alpha attempts to look for them and the following error is logged:

```
autoUpdater error:  Error: No published versions on GitHub
    at newError (C:\Program Files (x86)\Datalogic\Scan2Deploy Studio\resources\app.asar\node_modules\electron-updater\node_modules\builder-util-runtime\out\index.js:47:19)
    at GitHubProvider.getLatestVersion (C:\Program Files (x86)\Datalogic\Scan2Deploy Studio\resources\app.asar\node_modules\electron-updater\out\providers\GitHubProvider.js:94:55)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async NsisUpdater.getUpdateInfoAndProvider (C:\Program Files (x86)\Datalogic\Scan2Deploy Studio\resources\app.asar\node_modules\electron-updater\out\AppUpdater.js:319:19)
    at async NsisUpdater.doCheckForUpdates (C:\Program Files (x86)\Datalogic\Scan2Deploy Studio\resources\app.asar\node_modules\electron-updater\out\AppUpdater.js:333:24)
```

I believe this can be fixed by adding a check for `null` when assigning a value to `isCustomChannel`, since that is what `hrefChannel` will equal when looking at latest/stable releases.
`const isCustomChannel = hrefChannel !== null && !["alpha", "beta"].includes(String(hrefChannel));`

Please let me know if there's anything I may have misread or left out, and thanks for your time.

PR #7666, opened 2023-07-14
fix: for isCustomChannel in GitHubProvider.ts #7665

sethjray:
This pull request implements the fix I suggested in #7665

### DIFF
--- a/.changeset/smart-balloons-sip.md
+++ b/.changeset/smart-balloons-sip.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix: check null for `isCustomChannel` in GitHubProvider.ts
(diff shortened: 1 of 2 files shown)


mmaietta:
Hi @sethjray, thanks for contributing! Just to double-check, have you tested this locally? Such as via `patch-package`?

Wondering if this is also related #7669

sethjray:
Hi @mmaietta, yes I tested it locally by editing the package within the node_modules directory. It fixed the problem I was having, and began properly detecting when a branch did not have a prerelease tag associated with it. 

And yes, I would imagine that #7669 is caused by the same thing. Their stable branch is being mislabeled as a custom branch (named "null" probably) and it isn't allowing their alpha version to upgrade to it. 

Thanks for your help, and let me know if there's anything else I can do!

Issue #7652, opened 2023-07-05
Code-Signing Windows on Mac, `osslsigncode` produces 'Unrecognized file type' error when signing .node modules 


romansp:
* **Electron-Builder Version**: 24.4.0

* **Node Version**: 18.16.1

* **Electron Version**: 25.2.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: x64

In our electron app we're using `websocket` package that includes native prebuilt .node modules. And we're signing both Mac and Windows x64 app versions on Mac. 

After upgrading to electron-builder v24 we can no longer sign Windows app on Mac. 

```
• signing         file=dist_electron/win-unpacked/redacted.exe certificateFile=/Users/runner/work/_temp/redacted-CodeSign-Exp-2023-10-06.pfx
  • signing         file=dist_electron/win-unpacked/resources/app.asar.unpacked/node_modules/utf-8-validate/prebuilds/win32-x64/node.napi.node certificateFile=/Users/runner/work/_temp/redacted-CodeSign-Exp-2023-10-06.pfx
  • signing         file=dist_electron/win-unpacked/resources/app.asar.unpacked/node_modules/utf-8-validate/prebuilds/win32-ia32/node.napi.node certificateFile=/Users/runner/work/_temp/redacted-CodeSign-Exp-2023-10-06.pfx
  • signing         file=dist_electron/win-unpacked/resources/app.asar.unpacked/node_modules/utf-8-validate/prebuilds/linux-x64/node.napi.node certificateFile=/Users/runner/work/_temp/redacted-CodeSign-Exp-2023-10-06.pfx
  • signing         file=dist_electron/win-unpacked/resources/app.asar.unpacked/node_modules/utf-8-validate/prebuilds/darwin-x64+arm64/node.napi.node certificateFile=/Users/runner/work/_temp/redacted-CodeSign-Exp-2023-10-06.pfx
[...]

PR #7685, opened 2023-07-24
fix: add `signExts` configuration option to not sign `.node` files by default

mmaietta:
fix: allow explicit `signExts` configuration option for what additional files to sign. Do not sign .node files by default
Fixes: #7652


packages/app-builder-lib/src/winPackager.ts
```diff
@@ -355,7 +355,8 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
 
   private shouldSignFile(file: string): boolean {
     const shouldSignDll = this.platformSpecificBuildOptions.signDlls === true && file.endsWith(".dll")
-    return shouldSignDll || file.endsWith(".exe") || file.endsWith(".node")
+    const shouldSignExplicit = !!this.platformSpecificBuildOptions.signExts?.some(ext => file.endsWith(ext))
+    return shouldSignDll || shouldSignExplicit || file.endsWith(".exe")
```
johnpyp, line 359: Why not make `.exe` part of the default `signExts` array?

mmaietta: I was thinking of that, but since it has been long-established functionality, I didn't want to change it.

`.exe` is required for any signing to work AFAIK, so having it as default would technically work **until** someone explicitly sets an additional `signExts` value, in which case they'd need to also know that they'd still need to include `.exe`. This would require deeper knowledge of electron-builder and need to be explicitly called out in the docs, which can easily be overlooked.

johnpyp: Makes sense

PR #7690, opened 2023-07-26
fix: force rebuild all docker images for latest pnpm

mmaietta:
fix: adding `IMAGE_VERSION` arg for docker `node` image so that it force rebuilds when pnpm is updated on any new date

### DIFF
--- a/docker/node/Dockerfile
+++ b/docker/node/Dockerfile
@@ -1,6 +1,7 @@
-FROM electronuserland/builder:base
+ARG IMAGE_VERSION=base
+FROM electronuserland/builder:$IMAGE_VERSION
 
-ARG NODE_VERSION 18.16.0
+ARG NODE_VERSION 18.15.0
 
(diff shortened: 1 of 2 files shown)


docker/node/Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM electronuserland/builder:base
+ARG IMAGE_VERSION=base
+FROM electronuserland/builder:$IMAGE_VERSION
 
-ARG NODE_VERSION 18.16.0
+ARG NODE_VERSION 18.15.0
```
nikskiz, line 4: @mmaietta Is there a reason why the node version was rolled back?

mmaietta: Another user commented that the node version needed to be lower?

Could probably just up date it to latest 18.x. Would you mind opening a PR for that change?

nikskiz: PR has been raised https://github.com/electron-userland/electron-builder/pull/7787

PR #7715, opened 2023-08-04
fix: duplicated uploads

guohaolay:
Only schedule upload for unique files, otherwise it would try to upload the same files multiple times.

### DIFF
--- a/packages/app-builder-lib/src/index.ts
+++ b/packages/app-builder-lib/src/index.ts
@@ -89,6 +89,10 @@ export function build(options: PackagerOptions & PublishOptions, packager: Packa
       }
 
       for (const newArtifact of newArtifacts) {
+        if (buildResult.artifactPaths.includes(newArtifact)) {
+          log.warn({ newArtifact }, "skipping publish of artifact, already published");
+          continue;
+        }
         buildResult.artifactPaths.push(newArtifact)
         for (const publishConfiguration of publishConfigurations) {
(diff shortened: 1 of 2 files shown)


mmaietta:
Thanks for contributing!

I'd like to delve a little further into this, what are being duplicated during the upload phase?
Are you doing anything within the `afterAllArtifactBuild` build hook?

guohaolay:
We noticed electron builder was uploading the same file twice.

In the `afterAllArtifactBuild` hook, we are returning additional files and noticed a mistake where we included files already in the artifacts list. This caused the upload to happen twice for the same file.

The suggested fix is to catch that error in the builder to prevent bad coders (ie myself) from accidentally doing this.

packages/app-builder-lib/src/index.ts
```diff
@@ -89,6 +89,9 @@ export function build(options: PackagerOptions & PublishOptions, packager: Packa
       }
 
       for (const newArtifact of newArtifacts) {
+        if (buildResult.artifactPaths.includes(newArtifact)) {
+          continue;
```
mmaietta, line 93: Let's add a log here for the console:
```
log.warn({ newArtifact }, "skipping publish of artifact, already published")
```

guohaolay: Good call, updated and merged new changes, please review.

Issue #7587, opened 2023-05-26
`electron-builder` not working with modules developed using node-addon-api when the target is s multiple and contains mas


Koppel-Zhou:
* **Electron-Builder Version**: 23.3.3\24.4.0

* **Node Version**: 16.19.1

* **Electron Version**: ^20.1.0 (20.3.8 in yarn.lock)
* **Electron Type (current, beta, nightly)**: current

* **Electron-Updater Version**: 4.0.6

* **Target**: 
'darwin': { mac: ['mas', 'dmg', 'zip'] },
'win32': { win: ['nsis'], ia32: true, x64: false }

This error only occurs when the target product is multiple and contains mas. If the target product is only mas or does not contain mas, everything works normally. **This error is because the rebuild task of the same native module is carried out in two rebuild tasks at the same time, and they conflict.**

Hope That Helps.

***The error info with `electron-builder@23.3.3`***
```
• electron-builder  version=23.3.3 os=22.4.0
  • loaded parent configuration  file=/Users/koppel/my-app/node_modules/electron-webpack/out/electron-builder.js
  • writing effective config  file=dist/online/builder-effective-config.yaml
  • rebuilding native dependencies  dependencies=node-systeminformation@1.0.8, better-sqlite3@8.3.0, bufferutil@4.0.7, utf-8-validate@5.0.10 platform=darwin arch=x64
  • install prebuilt binary  name=better-sqlite3 version=8.3.0 platform=darwin arch=x64 napi=
  • rebuilding native dependencies  dependencies=node-systeminformation@1.0.8, better-sqlite3@8.3.0, bufferutil@4.0.7, utf-8-validate@5.0.10 platform=darwin arch=x64
  • install prebuilt binary  name=better-sqlite3 version=8.3.0 platform=darwin arch=x64 napi=
  • rebuilding native dependency  name=bufferutil version=4.0.7
  • rebuilding native dependency  name=bufferutil version=4.0.7
  • rebuilding native dependency  name=node-systeminformation version=1.0.8
  • rebuilding native dependency  name=node-systeminformation version=1.0.8
  • rebuilding native dependency  name=utf-8-validate version=5.0.10
  • rebuilding native dependency  name=utf-8-validate version=5.0.10
  ⨯ cannot execute  cause=exit status 1
$ node-gyp rebuild
      CXX(target) Release/obj.target/systeminformation/src/systeminformation.o
[...]

PR #7744, opened 2023-08-26
fix(mac): fix errors using native modules that require rebuild when the mac target is multiple and contains mas

Koppel-Zhou:
For details about the problem, please see [Issue-7587](https://github.com/electron-userland/electron-builder/issues/7587), [this](https://github.com/Koppel-Zhou/electron-rebuild-error-demo) is a demo created to reproduce the problem

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -181,10 +181,12 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       }
     }
 
-    if ((!hasMas || targets.length > 1) && !prepackaged) {
-      const appPath = path.join(this.computeAppOutDir(outDir, arch), `${this.appInfo.productFilename}.app`);
-      await this.doPack(outDir, path.dirname(appPath), this.platform.nodeName as ElectronPlatformName, arch, this.platformSpecificBuildOptions, targets)
-      .then(() => this.packageInDistributableFormat(appPath, arch, targets, taskManager))
+    if (!hasMas || targets.length > 1) {
+      const appPath = prepackaged == null ? path.join(this.computeAppOutDir(outDir, arch), `${this.appInfo.productFilename}.app`) : prepackaged
+      if (prepackaged == null) {
+        await this.doPack(outDir, path.dirname(appPath), this.platform.nodeName as ElectronPlatformName, arch, this.platformSpecificBuildOptions, targets)
+      }
+      this.packageInDistributableFormat(appPath, arch, targets, taskManager)
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/macPackager.ts
```diff
@@ -192,6 +192,15 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       }
     }
 
+    async function processTargetsPack(targets: Array<Target>, index: number) {
+      if (index < targets.length) {
+        await processTargetPack(targets[index]);
+        await processTargetsPack(targets, index + 1);
```
mmaietta, line 198: I love that you extracted the logic to a single `processTargetPack`

This logic should fail if there are more than two targets specified though, such as `"target":  ["mas", "dmg", "zip"],`, right?

How is this recursion different than the current loop of `for (const target of targets)`?

I don't see anything different here from this code
```
for (const target of targets) {
      await processTargetPack(target)
}
```

Koppel-Zhou:
```
for (const target of targets) {
      await processTargetPack(target)
}
```
This code causes the asynchronous tasks inside the loop to fire at the same time, rather than one after the other

Koppel-Zhou: Oh, sorry, I made a mistake when debugging, which made me think I found the real cause of this problem. Now I have resubmitted my edits, please check

packages/app-builder-lib/src/macPackager.ts
```diff
@@ -192,8 +181,10 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       }
     }
 
-    if (nonMasPromise != null) {
-      await nonMasPromise
+    if ((!hasMas || targets.length > 1) && !prepackaged) {
+      const appPath = path.join(this.computeAppOutDir(outDir, arch), `${this.appInfo.productFilename}.app`);
+      await this.doPack(outDir, path.dirname(appPath), this.platform.nodeName as ElectronPlatformName, arch, this.platformSpecificBuildOptions, targets)
+      .then(() => this.packageInDistributableFormat(appPath, arch, targets, taskManager))
```
mmaietta, line 187: This logic is no different than before, right? It's simply been moved.

Creating `nonMasPromise` initially does not actually execute it until we run `await nonMasPromise` here. So moving that logic from line 164 to here (line 195) by simply not preallocating the promise does nothing different in logic flow.

Not sure how this fixes the issue you're describing.

mmaietta: Try replacing this logic with this 👇 . I think this replicates previous logic. From what I can tell, the `.then(() => this.packIndDistributableFormat(...))` is the issue?
```
    if (!hasMas || targets.length > 1) {
      const appPath = prepackaged == null ? path.join(this.computeAppOutDir(outDir, arch), `${this.appInfo.productFilename}.app`) : prepackaged
      if (prepackaged == null) {
        await this.doPack(outDir, path.dirname(appPath), this.platform.nodeName as ElectronPlatformName, arch, this.platformSpecificBuildOptions, targets)
      }
      this.packageInDistributableFormat(appPath, arch, targets, taskManager)
    }
```

PR #7767, opened 2023-09-08
fix: When error code is ENOENT, try to use electron.shell.openPath to…

jackple:
I found that some Windows machines will report the errorCode `ENOENT` even `options.installerPath` is exist, but electron-updater never handle it! I think `openPath` can solve the problem. 

log:
```
[2023-09-07 11:54:11.500] [main] [info]  Executing: C:\Users\Administrator\AppData\Local\xx-updater\pending\xx-ia32-1.9.1-3.exe with args: --updated,/S,--force-run
[2023-09-07 11:54:11.502] [main] [info]  Cannot run installer: error code: ENOENT, error message: "spawn C:\Users\Administrator\AppData\Local\xx-updater\pending\xx-ia32-1.9.1-3.exe ENOENT", will be executed again using elevate if EACCES"
[2023-09-07 11:54:11.503] [main] [error] Error: Error: spawn C:\Users\Administrator\AppData\Local\xx-updater\pending\xx-ia32-1.9.1-3.exe ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
    at onErrorNT (node:internal/child_process:477:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)

```

### DIFF
--- a/.changeset/few-eagles-learn.md
+++ b/.changeset/few-eagles-learn.md
@@ -1,0 +1,6 @@
+---
+"electron-updater": patch
+---
+
+
+fix: When error code is ENOENT, try to use electron.shell.openPath to run installer on Windows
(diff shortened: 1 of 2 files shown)


jackple:
@mmaietta I have tested on several WIndows machine, also Win7/10. And I do not think this change will affect linux or mac.

PR #7813, opened 2023-10-06
docs: add snap base core22 as an option

jgresham:
https://snapcraft.io/docs/base-snaps#Choosing%20a%20base

> core22 is the currently recommended base for the majority of snaps. But much like choosing a distribution base for a project or server, the best base for an application is dependent on an application’s requirements and which plugins or extensions a base supports. If there are specific dependencies that cannot be easily met with core22 then core20 is a valid and supported alternative, as is the older core18.

### DIFF
--- a/docs/configuration/snap.md
+++ b/docs/configuration/snap.md
@@ -3,7 +3,7 @@ The top-level [snap](configuration.md#Configuration-snap) key contains set of op
 <!-- do not edit. start of generated block -->
 <ul>
 <li>
-<p><code id="SnapOptions-base">base</code> String | “undefined” - A snap of type base to be used as the execution environment for this snap. Examples: <code>core</code>, <code>core18</code>, <code>core20</code>. Defaults to <code>core18</code></p>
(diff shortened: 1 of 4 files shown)


jgresham:
will add changeset if this is approved!

docs/configuration/snap.md
```diff
@@ -3,7 +3,7 @@ The top-level [snap](configuration.md#Configuration-snap) key contains set of op
 <!-- do not edit. start of generated block -->
 <ul>
 <li>
-<p><code id="SnapOptions-base">base</code> String | “undefined” - A snap of type base to be used as the execution environment for this snap. Examples: <code>core</code>, <code>core18</code>, <code>core20</code>. Defaults to <code>core18</code></p>
+<p><code id="SnapOptions-base">base</code> String | “undefined” - A snap of type base to be used as the execution environment for this snap. Examples: <code>core</code>, <code>core18</code>, <code>core20</code>, <code>core22</code>. Defaults to <code>core18</code></p>
```
mmaietta, line 6: Similar to https://github.com/electron-userland/electron-builder/pull/7814#discussion_r1359493735

You'll need to add it to the file here: https://github.com/electron-userland/electron-builder/blob/2ca621289e025a6d0856c7c588aaac40d267cb5f/packages/app-builder-lib/src/options/SnapOptions.ts#L6
Then run `pnpm generate-all`

jgresham:
@mmaietta hopefully did this one right too

PR #7841, opened 2023-10-23
fix: remove com.apple.quarantine attribute from downloaded update files on mac

abarsov:


abarsov:
I am vin a process of verifying the fix now.

mmaietta:
Just a heads up, there's a compiler error with the current changes:
```
packages/electron-updater/src/MacUpdater.ts(102,77): error TS2339: Property 'trim' does not exist on type 'Buffer'.
```

abarsov:
@mmaietta thanks for your comment. 
I have fixed compilation and tried it today and it still doesn't work. 
As soon as I added `"LSFileQuarantineEnabled": true` to the plist of my application and try to perform autoupdate,
it start asking about quarantined app - whether I'd like to run app downloaded from the net or not.

I have not managed to find how it could be fixed. It seems that it might be on Squirrel.Mac side.
I mean I do see in logs that xattr executes successfully and clear quarantine bits from downloaded ZIP archive, but 
then after update finished new version of the app is started as if it is from quarantine anyway.

PR #7806, opened 2023-10-01
fix: compat with newest @types/node

AviVahl:
due to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66783 we got a downstream type error on httpExecutor's RequestHeaders
fixed by using OutgoingHttpHeader from node
bumped repo's @types/node to verify fix

PR #7852, opened 2023-10-26
fix: Multiple types were added, in RequestHeaders interface

ArMinasyan:


packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -12,7 +12,7 @@ import { ProgressCallbackTransform, ProgressInfo } from "./ProgressCallbackTrans
 const debug = _debug("electron-builder")
 
 export interface RequestHeaders extends OutgoingHttpHeaders {
-  [key: string]: string
+  [key: string]: string | string[] | number
```
mmaietta, line 15: What about using?
```
[key: string]: OutgoingHttpHeader | undefined
```
It's a standard @node-type that would update as node versions update. It's a type alias for
```
    type OutgoingHttpHeader = number | string | string[];
```

mmaietta: Also, after changing the value. Please run `pnpm generate-all` to update relevant documentation (if any is needed)

mmaietta: Please pull in latest `master`, changes were made to `httpExecutor` in another PR

mmaietta:
Resolving this, it was accomplished in another PR https://github.com/electron-userland/electron-builder/pull/7806

Issue #7859, opened 2023-11-01
Notarization security issue


mifi:
Apologies if creating an issue is not the right thing to do, but the discussions page doesn't seem to exist anymore. I'm trying to understand this whole Apple App Specific password thing. It's [vaguely described by Apple](https://support.apple.com/en-us/102654), but I'm very confused and cannot find the answer to this question:

Which access privileges does an App Specific Password actually give the user? Surely it gives access to notarizing any apps contained within that Apple ID. And I see from the poor Apple description that:

> the app can access information like mail, contacts, and calendars that you store in iCloud

So surely it gives access to more things than just notarizing (and notarizing is not even described there). So I'm wondering, does an App Specific password basically give full access to that Apple ID, to perform any action on behalf of the account owner? Does it give access to all teams/apps that the Apple ID has access to? (I don't see any way to restrict the app specific password in any way).

Now the reason why I think this could be a security issue for electron-builder:

If developers (like me) use [`electron-builder` with an app-specific password](https://www.electron.build/configuration/mac#MacConfiguration-target) on their CI server like github actions, doesn't this mean that **anyone with push access** to any repository that uses this has full access to the Apple ID of the developer who happened to put their app-specific password into a `APPLE_APP_SPECIFIC_PASSWORD` environment variable (github secret or similar)? Also guests can submit PRs that will take-over the owner's Apple ID (if the PR slips through a review).

If this is the case (even if only access to mail, contacts, and calendars), I think good idea to strongly recommend developers to NOT use `APPLE_APP_SPECIFIC_PASSWORD` in CI like github actions. Maybe we should instead recommend `apiKey`/`apiIssuer` which is [supported by `@electron/notarize`](https://github.com/electron/notarize), because those can be at least limited to a specific project. Or am I missing something else?

PR #7861, opened 2023-11-01
feat: allow api key and keychain notarize

mifi:
see also #7859

I've tested this here https://github.com/mifi/lossless-cut/actions/runs/6719326573/job/18260738394
 with a [patched macPackager.js](https://github.com/mifi/lossless-cut/blob/master/.yarn/patches/app-builder-lib-npm-24.8.0-51e1f5cd3f.patch)

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -481,7 +481,7 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     return true
   }
 
-  private async getNotarizeOptions(appPath) {
+  private getNotarizeOptions(appPath: string) {
     const appleId = process.env.APPLE_ID
     const appleIdPassword = process.env.APPLE_APP_SPECIFIC_PASSWORD
 
@@ -501,6 +501,9 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
     const appleApiKeyId = process.env.APPLE_API_KEY_ID;
     const appleApiIssuer = process.env.APPLE_API_ISSUER;
     if (appleApiKey || appleApiKeyId || appleApiIssuer) {
+      if (!appleApiKey || !appleApiKeyId || !appleApiIssuer) {
+        throw new InvalidConfigurationError(`Env vars APPLE_API_KEY, APPLE_API_KEY_ID and APPLE_API_ISSUER need to be set`)
+      }
       return this.generateNotarizeOptions({ appPath, appleApiKey, appleApiKeyId, appleApiIssuer });
     }
 
@@ -522,7 +525,9 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
       return
     }
     const options = this.getNotarizeOptions(appPath)
-    if (options == null) return
+    if (options == null) {
+      return
+    }
     await notarize(options)
     log.info(null, "notarization successful")
   }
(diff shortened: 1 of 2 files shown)


mmaietta:
`tsc` build step is failing with these errors:
```
Error: packages/app-builder-lib/src/macPackager.ts(484,36): error TS7006: Parameter 'appPath' implicitly has an 'any' type.
Error: packages/app-builder-lib/src/macPackager.ts(504,54): error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
Error: packages/app-builder-lib/src/macPackager.ts(504,67): error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
Error: packages/app-builder-lib/src/macPackager.ts(504,82): error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```
Can you please add explicit types to the consts?

packages/app-builder-lib/src/macPackager.ts
```diff
@@ -481,31 +481,55 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
...
+      }
+      return this.generateNotarizeOptions({ appPath, appleId, appleIdPassword })
+    }
+
+    // option 2: API key
+    const appleApiKey = process.env.APPLE_API_KEY;
+    const appleApiKeyId = process.env.APPLE_API_KEY_ID;
+    const appleApiIssuer = process.env.APPLE_API_ISSUER;
+    if (appleApiKey || appleApiKeyId || appleApiIssuer) {
+      return this.generateNotarizeOptions({ appPath, appleApiKey, appleApiKeyId, appleApiIssuer });
```
mmaietta, line 504: Do these need to be `&&` in order for it to be of type `string`?

packages/app-builder-lib/src/macPackager.ts
```diff
@@ -481,31 +481,55 @@ export default class MacPackager extends PlatformPackager<MacConfiguration> {
...
-    }
-    if (!appleId) {
-      throw new InvalidConfigurationError(`APPLE_ID env var needs to be set`)
-    }
-    if (!appleIdPassword) {
-      throw new InvalidConfigurationError(`APPLE_APP_SPECIFIC_PASSWORD env var needs to be set`)
-    }
-    const options = this.generateNotarizeOptions(appPath, appleId, appleIdPassword)
+    const options = this.getNotarizeOptions(appPath)
+    if (options == null) return
```
mmaietta, line 525: Please make this multi-line to keep coding styles consistent
```
if (options == null) {
   return
}
```

mifi:
fixed some types. however i'm really struggling to fix the types for `generateNotarizeOptions`. when `notarize: true` i'm struggling to distinguish between legacy and non legacy options. and if I don't distinguish, it complains that `notarytool` option is missing, which makes sense because @electron/notarize types [requires `tool` to be `'notarytool'`](https://github.com/electron/notarize/blob/c9af67d0dcbc1d4ba86cf6108c810e828d1adefd/src/types.ts#L62), which contradicts with their readme which says:

> tool String - The notarization tool to use, default is notarytool

not sure if i'm misunderstanding something

bcomnes:
This is a great improvement, however at the moment, it isn't functional. I opened a small fix to make it work: https://github.com/electron-userland/electron-builder/pull/7951

tldr The notarization tool never runs when the API key is set and notarize is set to `true`, and there are no other workarounds that work without the fix. Hope we can land it!

Issue #5977, opened 2021-06-22
Diacritics in the product name breaks macOS builds


pedro-pedrosa:
* **Version**: 22.11.7

* **Electron Version**: 12.0.1
* **Electron Type (current, beta, nightly)**: current

* **Target**: macOS dmg

Codesign step fails if my `productName` contains diacritics in NFC form.

My `package.json`
```json
{
  "name": "export",
  "version": "1.0.0",
  "description": "á",
  "author": "example",
  "main": "main.js",
  "scripts": {
    "start": "electron .",
    "build": "electron-builder build"
  },
  "build": {
    "productName": "á",
    "appId": "com.example",
    "mac": {
      "icon": "./assets/icon.icns",
      "target": "dmg"
    },
    "dmg": {
      "background": "./assets/background.png",
      "contents": [
        {
          "x": 320,
          "y": 240
        }
      ]
    }
  },
  "dependencies": {
    "electron": "^12.0.1",
    "electron-builder": "^22.11.7"
  },
  "devDependencies": {
  }
}
```

Produces the following:
```
> export@1.0.0 build
> electron-builder build
  • electron-builder  version=22.11.7 os=20.4.0
  • artifacts will be published if draft release exists  reason=CI detected
  • loaded configuration  file=package.json ("build" field)
  • packaging       platform=darwin arch=x64 electron=11.2.0 appOutDir=dist/mac
  • signing         file=dist/mac/á.app identityName=xxx identityHash=xxx provisioningProfile=none
  ⨯ Command failed: codesign --sign xxx --force --keychain /var/folders/56/nd8phrjx1hq6rmtbrcr5709r0000gn/T/a113b488461a02a1fd8e06e1f5f6fa8d33b6d59095fab3f4126103ba906b4588.keychain --timestamp --options runtime --entitlements /Users/travis/build/example/node_modules/app-builder-lib/templates/entitlements.mac.plist /Users/travis/build/example/export/dist/mac/á.app/Contents/Frameworks/á Helper (GPU).app/Contents/MacOS/á Helper (GPU)
/Users/travis/build/example/export/dist/mac/á.app/Contents/Frameworks/á Helper (GPU).app/Contents/MacOS/á Helper (GPU): code object is not signed at all
In subcomponent: /Users/travis/build/example/export/dist/mac/á.app/Contents/Frameworks/á Helper (GPU).app/Contents/MacOS/á Helper (GPU)
[...]

Issue #4697, opened 2020-02-24
Convert unicode filenames to NFD for MacOS build


Strangerxxx:
* **Version**: 22.3.2

* **Target**: dmg

[Unicode on Mac is insane. Mac OS X uses NFD while everything else uses NFC. This fixes that.](https://gist.github.com/JamesChevalier/8448512)

I tried to build electron app using electron-builder with the dependency of [mmmagic](https://github.com/mscdex/mmmagic)
The problem was showed up when I added package, which uses test file with filename containing unicode symbol. Since it was added I always got broken builds of my app. macOs always marking app as broken and advice me to put it into trash.
![error](https://user-images.githubusercontent.com/1136003/75123920-b2185180-56bc-11ea-88a8-cd112edc02f7.png)
It could be forced to run by calling `xattr -cr /path/to/my.app`
But even after that by checking codesign it shows me the problem
<img width="856" alt="console" src="https://user-images.githubusercontent.com/1136003/75123979-2521c800-56bd-11ea-91da-ede34c7a8180.png">
I decided to avoid it by fork problem package and remove unicode char in the filename, then open PR mscdex/mmmagic#143 for upstream to fix up the issue. And @mscdex (mmmaintainer, ha-ha) showed me the roots of the problem. 
But it seems a bit tricky to avoid this problem on mac and to not brake cross-platform capabilities of electron-builder as I suppose to in [the PR comment](https://github.com/mscdex/mmmagic/pull/143#issuecomment-590127677)

Strangerxxx:
It have possible workaround to convert files using some pre-sign or pre-pack hooks.

Issue #3169, opened 2018-07-25
code object is not signed at all


TomRomanowski:
* **Version**: 20.25.0

* **Target**: macos

When I build for windows using NSIS, I have no issue signing the app. However, when I try to build for macos, I get `code object is not signed at all` because of my LICENSE.txt being in the Contents directory. It's the same issue here:  https://github.com/electron-userland/electron-builder/issues/2522
but it doesn't seem like it was fixed

In my package.json, I have:
```
{
   ...
   "build": {
      ...
      "extraFiles": "LICENSE.txt"
   }
}
```
I created an afterPack script to move the file myself but the same issue persists

pronebird:
We experience the same with additional binary (executable) added to `extraFiles`. Shouldn't electron-builder sign it?

fishear:
I also have the same problem, does anybody have a resolution?

fishear:
> We experience the same with additional binary (executable) added to `extraFiles`. Shouldn't electron-builder sign it?

I also encountered the same problem, could you please tell me do you have a workaround? thanks~~

fishear:
> * **Version**: 20.25.0
> 
> * **Target**: macos
> 
> When I build for windows using NSIS, I have no issue signing the app. However, when I try to build for macos, I get `code object is not signed at all` because of my LICENSE.txt being in the Contents directory. It's the same issue here: #2522
> but it doesn't seem like it was fixed
> 
> In my package.json, I have:
> 
> ```
> {
>    ...
>    "build": {
>       ...
>       "extraFiles": "LICENSE.txt"
>    }
> }
> ```
> I created an afterPack script to move the file myself but the same issue persists

Hi, did you fix it? I tried my best, but still failed.

X140Yu:
Same issue here, has anyone figured out how to deal with it?

Christilut:
Same problem here

Circuit8:
Same here, any solution?

Christilut:
My solution was simple, don't use a file without extension. It seems to assume that without extension it is a binary that must be signed. I just changed my `LICENSE` file to `LICENSE.txt`

If you can't do that, I'd suggest signing it manually with the `codesign` CLI tool, it's really easy.

Circuit8:
@Christilut Thanks a lot thats really helpful. My file was an exectuable, I thought electron-builder would handle signing it, but turns out youre right, I had to sign it first, then it gets signed again as part of the final .app file by electron-builder.

[...]

PR #7901, opened 2023-11-27
fix(mac): normalize filename to NFD form

### LINKED ISSUE #6912: Helper (GPU).app/Contents/MacOS/myapp Helper (GPU).app: code object is not signed at all
v23.0.3

* **Node Version**: 
v16.15.0

* **Electron Version**: v18.2.3

Current

* **Target**: macOS

So basically, when I try to sign with a valid Apple Dev Id, I get this error.
Here is workflow run: https://github.com/ZeotYT/PokeLauncher-v2/runs/6684752383
Can you help me?
Thanks in advance for your answers

jebibot:
Fixes issues when `productName` and `executableName` contain characters that have different NFC and NFD form, e.g., diacritics and Hangul (Korean). These seem to be internally normalized to the NFD form at some point, failing string comparison and other operations. Specifically,
- fixes codesign failing with `code object is not signed at all In subcomponent:`,
  - fixes #6912
  - https://github.com/electron-userland/electron-builder/issues/3169#issuecomment-800113654
  - #5977
  - https://github.com/electron/osx-sign/issues/155
- fixes DMG layout,
  - #1234
  - #1369
  - #6282
- and possibly related:
  - #2125
  - #4697

### DIFF
--- a/packages/app-builder-lib/src/util/filename.ts
+++ b/packages/app-builder-lib/src/util/filename.ts
@@ -2,8 +2,9 @@
 import * as _sanitizeFileName from "sanitize-filename"
 import * as path from "path"
 
-export function sanitizeFileName(s: string): string {
-  return _sanitizeFileName(s)
+export function sanitizeFileName(s: string, normalizeNfd = false): string {
+  const sanitized = _sanitizeFileName(s)
+  return normalizeNfd ? sanitized.normalize("NFD") : sanitized
 }
 
 // Get the filetype from a filename. Returns a string of one or more file extensions,
--- a/.changeset/thick-flowers-bathe.md
+++ b/.changeset/thick-flowers-bathe.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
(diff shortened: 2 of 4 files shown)


packages/app-builder-lib/src/util/filename.ts
```diff
@@ -2,8 +2,9 @@
 import * as _sanitizeFileName from "sanitize-filename"
 import * as path from "path"
 
-export function sanitizeFileName(s: string): string {
-  return _sanitizeFileName(s)
+export function sanitizeFileName(s: string, normalizeNfd = false): string {
+  const sanitized = _sanitizeFileName(s)
+  return normalizeNfd ? sanitized.normalize("NFD") : sanitized
```
mmaietta, line 7: What is `normalize("NFD")`? I don't see it in the API or type definitions: https://github.com/parshap/node-sanitize-filename/blob/master/index.d.ts

mmaietta: Thank you for clarifying, and I appreciate the links as well!

Issue #7935, opened 2023-12-08
require() of ES Module not supported


lutzroeder:
* **Electron-Builder Version**: 24.9.1
* **Node Version**: 21.4.0
* **Electron Version**: 28.0.0
* **Target**: macOS

1. Set `package.json` to `type: module`.
2. Create `afterSign: notarize.js` using ESM imports.
3. Run `npx electron-builder --mac --universal --publish never`.

```
⨯ require() of ES Module ~/Projects/Test/notarize.js from ~/Projects/Test/node_modules/app-builder-lib/out/platformPackager.js not supported.
Instead change the require of notarize.js in ~/Projects/Test/node_modules/app-builder-lib/out/platformPackager.js to a dynamic import() which is available in all CommonJS modules.
```

omarabdula:
@lutzroeder  HELLO DID YOU MANAGE TO FIND A SOLUTION?

briantolsen:
Running into this issue now as well using 24.13.3

mmaietta:
So the current code detects if the package.json is of type `module`, but it automatically has a fallback of `require` from the `try-catch` logic
https://github.com/electron-userland/electron-builder/blob/29f6504e181cef2c43ac070a80f24ee79dc52481/packages/app-builder-lib/src/platformPackager.ts#L775-L787

It should be fully functional for handling both ES and CJS modules in the `next` alpha version, would you mind giving it a try?

leo4life2:
> Running into this issue now as well using 24.13.3

added some debug logs to platformPackager.js today, and turns out that the try part failed not because it's not an es module, but because there's bugs in my `notarize.js`'s code, so it threw exceptions on import. for me it was because i gave bad arguments to the `notarize` function.

Here's my working `afterSign` script:
```
import { notarize } from '@electron/notarize';

export default async function afterSign(context) {
  const { appOutDir, electronPlatformName } = context;

  if (electronPlatformName !== 'darwin') {
    return;
  }

  const appName = context.packager.appInfo.productFilename;

  await notarize({
    appPath: `${appOutDir}/${appName}.app`,
    appleApiKey: '~/private_keys/notarize.p8', // Path to your .p8 file
    appleApiKeyId: process.env.API_KEY_ID,
    appleApiIssuer: process.env.API_KEY_ISSUER_ID,
  });
}
```

mmaietta:
@leo4life2 that's a great callout and something I'll implement.

[...]

PR #7936, opened 2023-12-08
feat: ES module support

lutzroeder:
Use dynamic `import()` when `package.json` is set to [ES modules](https://nodejs.org/api/packages.html#type).

#7935

### DIFF
--- a/packages/app-builder-lib/src/appInfo.ts
+++ b/packages/app-builder-lib/src/appInfo.ts
@@ -22,6 +22,7 @@ export function smarten(s: string): string {
 export class AppInfo {
   readonly description = smarten(this.info.metadata.description || "")
   readonly version: string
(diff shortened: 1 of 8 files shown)


packages/app-builder-lib/src/appInfo.ts
```diff
@@ -39,6 +40,7 @@ export class AppInfo {
     normalizeNfd = false
   ) {
     this.version = info.metadata.version!
+    this.type = info.metadata.type!
```
mmaietta, line 43: What happens if `type` is not specified in the `package.json`? Seems we should leave this as optional?

lutzroeder: @mmaietta updated. Any insights why `pull_request_target` is failing?

PR #7949, opened 2023-12-18
fix(docs/configuration): broken links

afonsojramos:


mmaietta:
So that section of the docs is autogenerated via `pnpm generate-all`. What part of the docs isn't working currently?

afonsojramos:
@mmaietta Just some of these links as some have the id as all lowercase and then the link is camelCase, which breaks the links

PR #7967, opened 2024-01-05
feat(nsis): add NsisOption to specify selectPerMachineByDefault

jmeinke:
Add the possibility to create an assisted NSIS installer and select per Machine as default option for installation.
We need this for our installer.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh
+++ b/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh
@@ -143,7 +143,11 @@
       !ifdef INSTALL_MODE_PER_ALL_USERS
         !insertmacro setInstallModePerAllUsers
       !else
-        !insertmacro setInstallModePerUser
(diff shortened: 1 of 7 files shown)


jmeinke:
@mmaietta I too have had problems setting up the tools needed to re-create the docs. I therefore could need your help here. Thank you!

packages/app-builder-lib/templates/nsis/assistedInstaller.nsh
```diff
@@ -143,7 +143,11 @@
       !ifdef INSTALL_MODE_PER_ALL_USERS
```
mmaietta, line 143: I'm usually adverse to having nested if-statements if possible to avoid.

Seems like the logic here could be `INSTALL_MODE_PER_ALL_USERS || INSTALL_MODE_PER_ALL_USERS_DEFAULT`

Is that possible to do in nsis scripting?

jmeinke: Unfortunately, it does not seem to be possible with NSIS.

PR #7971, opened 2024-01-07
feat(archive): skip archive when up to date

OrbitZore:
I discovered that when I use the `--dir` option and `--prepackaged -c` to build different installers with various NSIS configurations, electron-builder repeatedly compresses the unpacked directory with 7zip. This process is time-consuming - in my case, it takes about 5 minutes to build one installer. To improve this, I implemented a modification time (mtime) check before the compression step, which is a common practice in build systems like GNU make. This change reduces the build time to 15 seconds per installer, making this use case more practical. It should not break the API.

### DIFF
--- a/.changeset/forty-hats-press.md
+++ b/.changeset/forty-hats-press.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+feat(archive): skip archive when destination file is already up to date
--- a/packages/app-builder-lib/src/targets/archive.ts
+++ b/packages/app-builder-lib/src/targets/archive.ts
@@ -1,5 +1,5 @@
 import { debug7z, exec, log } from "builder-util"
-import { exists, unlinkIfExists } from "builder-util/out/fs"
+import { exists, unlinkIfExists, statOrNull } from "builder-util/out/fs"
 import { move } from "fs-extra"
 import * as path from "path"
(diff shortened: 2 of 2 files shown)


LoveLuanQi1202:
Hi OrbitZore,

Thank you for your work on this project. I wanted to report an issue I encountered while using Electron Builder.

In my project, I have configured artifactName as "name.${ext}". When packaging for Mac with the universal architecture, the process first generates an arm zip and then creates a universal zip, which overwrites the original arm zip. However, the skipped archiving step prevents this from happening.

This issue causes complications in my packaging process, and I believe it might be affecting others as well.

Thank you for your attention to this matter.

OrbitZore:
> Hi OrbitZore,
> 
> Thank you for your work on this project. I wanted to report an issue I encountered while using Electron Builder.
> 
> In my project, I have configured artifactName as "name.${ext}". When packaging for Mac with the universal architecture, the process first generates an arm zip and then creates a universal zip, which overwrites the original arm zip. However, the skipped archiving step prevents this from happening.
> 
> This issue causes complications in my packaging process, and I believe it might be affecting others as well.
> 
> Thank you for your attention to this matter.

Perhaps we should consider assigning distinct names for the arm zip and universal zip files.The overwriten is ridiculous.

PR #7998, opened 2024-01-21
fix: Use full path to macOS signing utilities

### LINKED ISSUE #7997: macOS code signing fails due to finding wrong codesign on PATH
* **Electron-Builder Version**: 24.11.0

* **Node Version**: v21.5.0

* **Electron Version**: 28.1.3

* **Target**: macOS

Homebrew's `miniforge` package includes a `codesign`:

```
$ which codesign
/opt/homebrew/Caskroom/miniforge/base/bin/codesign
```

Unsurprising this fails when `electron-builder` uses it for code signing. The error message is:
```
The following arguments were not expected: --options --timestamp
```

A fix is to adjust PATH so that `/usr/bin` is ahead of `/opt/homebrew/Caskroom/miniforge/base/bin`.

dbrnz:
Rather than using something arbitrary, found on `PATH`, this makes sure we run Apple's signing utilities

Fixes #7997.

### DIFF
--- a/.changeset/thirty-bobcats-beg.md
+++ b/.changeset/thirty-bobcats-beg.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+Use full path to macOS signing utilities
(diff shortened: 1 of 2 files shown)


mmaietta:
Thanks for your contribution!

PR #8017, opened 2024-01-30
feat: added `includeExtensions` option

PavelMor25:
Hello,  
After transitioning from version 22.10.5 to the latest version, we encountered an issue where the component crashes because files with the extension `.d.ts` are not included in the build due to the `default exclude pattern`.

This option `includeExtensions` will allow us to exclude this extension from the `default exclude pattern` and also prevent similar problems for users with comparable requests.

Please let me know if I've missed anything

### DIFF
--- a/.changeset/shy-brooms-run.md
+++ b/.changeset/shy-brooms-run.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: added includeExtensions option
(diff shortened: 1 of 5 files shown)


mmaietta:
Closing in favor of https://github.com/electron-userland/electron-builder/pull/8043

PavelMor25:
Hello @mmaietta,

Could you please clarify if this option will only work with node modules files or any directory?

mmaietta:
Anything that the node module copier encounters, which is a different implementation flow that the `files` copier

PR #8035, opened 2024-02-05
fix(mac): merge `fileAssociations` with existing `CFBundleDocumentTypes` if defined in `mac.extendInfo`

davej:
`mac.extendInfo` may already contain entries for `CFBundleDocumentTypes`. Before this change, `fileAssociations` would overwrite previous `CFBundleDocumentTypes` properties. Now it merges.

I'm using this functionality so that I can add support for `public.folder`, so users can drag a folder onto the dock icon and the Electron app will recognise it.

### DIFF
--- a/packages/app-builder-lib/src/electron/electronMac.ts
+++ b/packages/app-builder-lib/src/electron/electronMac.ts
@@ -201,7 +201,7 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
 
   const fileAssociations = packager.fileAssociations
   if (fileAssociations.length > 0) {
-    appPlist.CFBundleDocumentTypes = await BluebirdPromise.map(fileAssociations, async fileAssociation => {
+    const documentTypes = await BluebirdPromise.map(fileAssociations, async fileAssociation => {
       const extensions = asArray(fileAssociation.ext).map(normalizeExt)
       const customIcon = await packager.getResource(getPlatformIconFileName(fileAssociation.icon, true), `${extensions[0]}.icns`)
       let iconFile = appPlist.CFBundleIconFile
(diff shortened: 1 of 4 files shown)


packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -223,6 +223,9 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
       }
       return result
     })
+
+    // `CFBundleDocumentTypes` may be defined in `mac.extendInfo`, so we need to merge it in that case
+    appPlist.CFBundleDocumentTypes = [...(appPlist.CFBundleDocumentTypes || []), ...documentTypes]
```
mmaietta, line 228: Probably should also handle duplicate entries.
```
Array.from(new Set([...(appPlist.CFBundleDocumentTypes || []), ...documentTypes]))
```
Not sure if there's a more pretty way to write that, can't recall the `Set` constructor off the top of my head

davej: Thanks @mmaietta! 🙏

The items in the `CFBundleDocumentTypes` array are objects (not primitives) so I don't think this will work. The objects will not be recognised as equal.

```js
Array.from(new Set([{ foo: 'bar'}, { foo: 'bar'}]))
// -> [{ "foo": "bar" }, { "foo": "bar" }]
```

I can create a custom function to do de-duplication if you like. I think it's probably overkill in this case but happy to do it if you would like me to?

mmaietta:
Can you please add a changeset file for the release notes via `pnpm generate-changeset`?

davej:
Thanks @mmaietta, changeset added.

PR #8059, opened 2024-02-20
fix: execute %SYSTEMROOT% cmd.exe directly during NSIS installer

mmaietta:


liudonghua123:
SO what's the difference between `%SYSTEMROOT%\cmd.exe` and `%SYSTEMROOT%\System32\cmd.exe`?

It seems `%SYSTEMROOT%\cmd.exe` is not exists.

image

mmaietta:
In powershell you can check env vars via `gci env:`

```
SystemRoot                     C:\WINDOWS
```

PR #8069, opened 2024-02-21
fix: Windows ES module support

lutzroeder:
Use `pathToFileUrl` to fix this error on Windows:

```
Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'  failedTask=build stackTrace=Error: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
```

#7935

### DIFF
--- a/.changeset/orange-goats-brake.md
+++ b/.changeset/orange-goats-brake.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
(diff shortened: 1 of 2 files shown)


mmaietta:
Thanks @lutzroeder !

Issue #6077, opened 2021-07-22
[Feature request] Downgrade Wixtoolset to v3 (3.14)


GeekCornerGH:
* **Electron-Builder Version**: 22.11.7

* **Node Version**: 14

* **Electron Version**: 13.1.7
* **Electron Type (current, beta, nightly)**:

* **Target**: 
Windows

I wanted to build my Electron app for arm64, with msi format, but the version you are using doesn't support arm64. Wix toolset v3.14 supports arm64. Can you change it? Thank you

mmaietta:
It looks like it pulls the toolset from https://github.com/electron-userland/electron-builder-binaries/tree/master/wix
So you'll need to move this request to that repo and/or contribute to that upstream project first.

GeekCornerGH:
Done :)

mrksr:
@mmaietta Hi! I've found this problem as well and also added a new reference in @GeekCornerGH's issue in the other repo. There has not been any activity there however. I'm quite new to the Electron ecosystem so I'm not sure how things work. Do you have an idea how the issue could be resolved? As I see it, this issue blocks _all_ Electron msi-builds for Windows on Arm, so it's not negligible. I'm happy to help things along, but I'm not sure who to get in contact with. Thanks!

PR #8086, opened 2024-02-27
feat(msi): build emulated arm64 MSI installers

Allan-Kerr:
Build x64 MSI installers when targeting the arm64 architecture. This results in an x64 MSI installer that installs an arm64 version of the application.

An x64 MSI installer is an improvement because building arm64 MSI targets errors with:
```
error CNDL0264 : The parameter 'arch' is missing or has an invalid value arm64.
Possible values are x86, x64, or ia64.
```

This is a stopgap until the electron-builder-binaries wix version is upgraded to a version that supports arm64 which is currently blocked by a number of breaking changes between 4.0.0.5512.2 and 4.0.1. This upgrade is discussed in #6077.

### DIFF
--- a/packages/app-builder-lib/src/targets/MsiTarget.ts
+++ b/packages/app-builder-lib/src/targets/MsiTarget.ts
@@ -70,9 +70,15 @@ export default class MsiTarget extends Target {
 
     const commonOptions = getEffectiveOptions(this.options, this.packager)
 
+    // wix 4.0.0.5512.2 doesn't support the arm64 architecture so default to x64 when building for arm64.
+    // This will result in an x64 MSI installer that installs an arm64 version of the application. This is a
+    // stopgap until the electron-builder-binaries wix version is upgraded to a version that supports arm64:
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/targets/MsiTarget.ts
```diff
@@ -70,17 +70,22 @@ export default class MsiTarget extends Target {
 
     const commonOptions = getEffectiveOptions(this.options, this.packager)
 
+    // wix 4.0.0.5512.2 doesn't support the arm64 architecture so default to x64 when building for arm64.
+    // This will result in an x64 MSI installer that installs an arm64 version of the application. This is a
+    // stopgap until the electron-builder-binaries wix version is upgraded to a version that supports arm64.
+    const wixArch = arch == Arch.arm64 ? Arch.x64 : arch;
```
mmaietta, line 76: Thanks for the detailed comment! Would you mind also adding this line to the comment just so that it's easier to track when we can upgrade?
`// https://github.com/electron-userland/electron-builder/issues/6077`

PR #7709, opened 2023-08-03
feat: add support for differential zip updates on macOS

beyondkmp:
Fix #7547

PR #8091, opened 2024-03-03
fix(mac): revert mac differential autoupdate

mmaietta:
Revert "feat: add support for differential zip updates on macOS"

Reverts electron-userland/electron-builder#7709


charlag:
There's no info on why it is reverted anywhere?

mmaietta:
@charlag an issue wasn't filed to track this against, but a comment was left in the originating PR
https://github.com/electron-userland/electron-builder/pull/7709#issuecomment-1974945802

PR #8093, opened 2024-03-03
fix: mac differential updater

mmaietta:
Reimplements: #7709

### DIFF
--- a/.changeset/thin-tigers-drop.md
+++ b/.changeset/thin-tigers-drop.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix: Reimplements: #7709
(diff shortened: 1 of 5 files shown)


beyondkmp:
This should have the same issue; the argument passed in should already be the temp file name.

https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/AppUpdater.ts#L685

mmaietta:
@beyondkmp could you please elaborate?
I was under the impression that `await this.httpExecutor.download(zipFileInfo.url, destinationFile, downloadOptions)` was never executed due to `canDownloadDifferentially` resolving the incorrect boolean value.
https://github.com/electron-userland/electron-builder/blob/cb335ecfef1f4fd1aef94020c1eaf5ce91bef574/packages/electron-updater/src/MacUpdater.ts#L104-L109

PR #8095, opened 2024-03-03
fix: mac differential updater

beyondkmp:
Fix copyfilesync error.

We should place copyFileSync inside the done function, ensuring that it is executed only after the file has been successfully downloaded and definitely exists within the corresponding folder


beyondkmp:
Currently, there are two issues, one is with canDownloadDifferentially, and the other is with the temp file.

1. canDownloadDifferentially always returns false, which prevents full download.
2. The destinationFile is a temp file, not the final file. If an error occurs, such as an incomplete download or removed, this temp file should not be used directly. It should only be copied after a successful download.

I have fixed these two issues. And it has been tested locally that both the initial download and the subsequent incremental downloads are normal.

beyondkmp:
@mmaietta Please help review again.

mmaietta:
LGTM, nice work!

willemdjong:
FYI: The differential download on mac is quite slow compared to windows!

beyondkmp:
In my local testing, it was faster than on Windows. This is also related to the network and device performance.

willemdjong:
Much slower, only have small updates of 8GB  or 13 GB...but compared to my full download of 250 GB it's like 5 times slower. 

My initial guess is that the diff for mac zips are larger than windows? => this is exactly what I thought...

It requires saving the update.zip locally first, and only supports incremental updates during the second update. => discovered this already! But that's not a big deal for me.

beyondkmp:
Is your app size 250GB? It's very large. Incremental updates require time to calculate. Your package is quite large, which likely took a lot of time to determine which parts were incremental.

Issue #7608, opened 2023-06-10
electron-builder: Linux deb ,rpm arch cross-complile wont work. arm64 target contains x64 binaries


jkroepke:
* **Electron-Builder Version**: 23.6.0

* **Node Version**: 16

* **Electron Version**: 22.3.3
* **Electron Type (current, beta, nightly)**: current

* **Target**:  Linux (AppImage, deb, rpm) and x86, AMD64

While cross-compile for macOS works fine, cross-compile for linux has some issues where I cant identity the root cause

Here is the log output with running file at the end:

```
> electron-builder --publish onTag --x64 --arm64
  • electron-builder  version=23.6.0 os=5.15.0-1038-azure
  • publishing will be skipped  reason=current build is a part of pull request solution=set env PUBLISH_FOR_PULL_REQUEST to true to force code signing
There are serious security concerns with PUBLISH_FOR_PULL_REQUEST=true (see the  CircleCI documentation (https://circleci.com/docs/1.0/fork-pr-builds/) for details)
If you have SSH keys, sensitive env vars or AWS credentials stored in your project settings and untrusted forks can make pull requests against your repo, then this option isn't for you.
  • loaded configuration  file=package.json ("build" field)
  • rebuilding native dependencies  dependencies=node-pty@0.10.1 platform=linux arch=x64
  • packaging       platform=linux arch=x64 electron=22.3.3 appOutDir=dist/linux-unpacked
  • downloading     url=https://github.com/electron/electron/releases/download/v22.3.3/electron-v22.3.3-linux-x64.zip size=93 MB parts=4
  • downloaded      url=https://github.com/electron/electron/releases/download/v22.3.3/electron-v22.3.3-linux-x64.zip duration=1.223s
  • building        target=AppImage arch=x64 file=dist/OpenLens-6.5.2-312.x86_64.AppImage
  • rebuilding native dependencies  dependencies=node-pty@0.10.1 platform=linux arch=arm64
  • downloading     url=https://github.com/electron-userland/electron-builder-binaries/releases/download/appimage-12.0.1/appimage-12.0.1.7z size=1.6 MB parts=1
  • downloaded      url=https://github.com/electron-userland/electron-builder-binaries/releases/download/appimage-12.0.1/appimage-12.0.1.7z duration=454ms
  • packaging       platform=linux arch=arm64 electron=22.3.3 appOutDir=dist/linux-arm64-unpacked
[...]

PR #8107, opened 2024-03-07
fix(linux): native modules overwritten with wrong arch

megahertz:
Fixes: #7608

**Description**
When building an app for multiple Linux architectures, native modules are hardlinked to the modules in `appDir/node_modules/**/*.node` if `CI` env is set. 

When the builder processes the first arch, it works fine. The second arch overwrites `appDir/node_modules/**/*.node`. Since these files are hardlinked to `linux-*unpacked/resources/app.asar.unpacked/node_modules/**/*.node` it overwrites native modules for all archs, so now the first arch distributive is broken.

`FileCopier` in `builder-util/out/fs` copies files by default, but once it detects a CI environment, it switches to hadlinking mode instead.

<details>
  <summary>Example package.json which reproduces the issue</summary>
  
```json
{
  "name": "builder-arm64-linux-issue",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "build": "env CI=1 electron-builder --linux --x64 --arm64 && find dist/linux-unpacked -name realm.node -exec file {} \\;"
  },
  "build": {
    "linux": {
      "target": {
        "arch": ["x64", "arm64"],
        "target": "dir"
      }
    }
  },
  "devDependencies": {
    "electron": "^29.1.0",
    "electron-builder": "^24.13.3"
  },
  "dependencies": {
    "realm": "^12.6.2"
  }
}
```
  
</details>

### DIFF
--- a/.changeset/weak-trains-flash.md
+++ b/.changeset/weak-trains-flash.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(linux): native modules overwritten with wrong arch
--- a/packages/app-builder-lib/src/asar/asarUtil.ts
+++ b/packages/app-builder-lib/src/asar/asarUtil.ts
@@ -1,5 +1,5 @@
 import { AsyncTaskManager, log } from "builder-util"
-import { FileCopier, Filter, MAX_FILE_REQUESTS } from "builder-util/out/fs"
+import { DO_NOT_USE_HARD_LINKS, FileCopier, Filter, MAX_FILE_REQUESTS } from "builder-util/out/fs"
 import { symlink, createReadStream, createWriteStream, Stats } from "fs"
 import { writeFile, readFile, mkdir } from "fs/promises"
 import * as path from "path"
@@ -71,7 +71,7 @@ export class AsarPackager {
 
     const transformedFiles = fileSet.transformedFiles
     const taskManager = new AsyncTaskManager(packager.cancellationToken)
-    const fileCopier = new FileCopier()
+    const fileCopier = new FileCopier(DO_NOT_USE_HARD_LINKS)
 
     let currentDirNode: Node | null = null
     let currentDirPath: string | null = null


mmaietta:
I'm wary of this change as I'm assuming the owner/author of this project had a reason for not having hard links always disabled.
For your local project, why not use env var `USE_HARD_LINKS=false`?

megahertz:
> I'm wary of this change as I'm assuming the owner/author of this project had a reason for not having hard links always disabled. For your local project, why not use env var `USE_HARD_LINKS=false`?

It's faster to use hardlinks instead of copying. But this change affects only *.node files inside app.asar.unpacked, so it can't impact performance too much. Anyway, it would be nice to ask @develar whether there are some pitfalls or hadrlinks are used only for speed reasons. 

Sure, I use this env as a workaround. But it took pretty much time for me to find the reason of this issue, so it would be nice to include this fix by default.

megahertz:
Maybe it makes sense to disable hardlinking on the Linux build only since there is no such issue on macOS build?

mmaietta:
I really need @develar to chime in before moving forward with this PR. I do understand the need to resolve this issue, but I don't have the historical context to understand the reasoning for it to be disabled by default.

thebalaa:
@develar ?

mmaietta:
@thebalaa can you still reproduce this on `next` v25.0.1? I'm trying to get a local project set up with the package.json you specify but I think I'm doing something wrong. Would you be willing to put together a sample minimum reproducible repo and I can take a deeper look if the issue still persists?

xyloflake:
How does anyone contact @develar? I don't think he's been actively checking out this repo since 2022. @mmaietta Also, one more thing, I think this repo is community maintained as of 2024 and the current maintainers have the full rights to make new changes since stuff in the software world changes and we have to adapt according to that. 

Even though there might be a very good reason behind whatever was implemented in this repo during the time when it was actively owned and maintained by @develar, it does not necessary mean imo that it should always stay like this. AFAIK, @develar did not also want autoupdates on targets other than AppImage, which @mmaietta implemented and was merged anyways.

There are a lot of things in this repo that are have a very unclear and vague "why" to them. That is why ***imo*** this repo deserves to move forward, and of course, if issues are discovered, you can always undo a change (before a release).

mmaietta:
Can you try this on latest electron-builder? Should be 25.0.5 at the time of writing this. If a minimum reproducible repo could be provided, I'd be happy to investigate further as well

megahertz:
@mmaietta I will try it next week.

megahertz:
It looks like it's fixed in 25.0.5 or earlier

PR #8120, opened 2024-03-11
feat: support additionalLightArgs

### LINKED ISSUE #7853: Support additional `light.exe` arguments
* **Electron-Builder Version**: 24.6.4

* **Node Version**: 18.17.1

* **Electron Version**: 26.2.2

* **Target**:  Windows

Hi, I use Electron to build an application, that target Windows's MSI.
Below is summary of our `package.json`

```json
{
  "build": {
    "win": {
      "target": [
        "msi"
      ]
    },
    "msi": {
      "oneClick": false,
      "runAfterFinish": true,
      "shortcutName": "Something"
    }
  }
}
```

Everything is good, until our customers ask us to provide them a Japanese MSI installer.
After some research, I found that I can add `-cultures:ja-JP` to `light.exe` command
https://wixtoolset.org/docs/v3/howtos/ui_and_localization/specifying_cultures_to_build/

I'd happy that `electron-builder` already exposed `additionalWixArgs` to customize the args
https://github.com/electron-userland/electron-builder/blob/db424e8e876e6ac1985668bf78bd52a02824dd7f/packages/app-builder-lib/src/targets/MsiTarget.ts#L142-L144
But unfortunately, the args are used by both `candle.exe` and `light.exe`, and it ends up throwing this error
```bash
candle.exe : error CNDL1098 : '-cultures:ja-jp' is not a valid command line argument.
```

https://github.com/electron-userland/electron-builder/blob/db424e8e876e6ac1985668bf78bd52a02824dd7f/packages/app-builder-lib/src/targets/MsiTarget.ts#L107-L118
https://github.com/electron-userland/electron-builder/blob/db424e8e876e6ac1985668bf78bd52a02824dd7f/packages/app-builder-lib/src/targets/MsiTarget.ts#L83

Can we add an option like `additionalLightArgs` or something to provide more flexibility for MSI build? I'm ready to make a pull request to support it.

Thanks for reading.

beyondkmp:
fix #7853

### DIFF
--- a/.changeset/neat-bottles-wash.md
+++ b/.changeset/neat-bottles-wash.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: support `additionalLightArgs` for msi target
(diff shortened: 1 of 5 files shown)


mmaietta:
Thanks again @beyondkmp !

PR #8123, opened 2024-03-11
feat: add disableSanityCheckPackage to asar to allow custom electron fork asar integrity implementations

### LINKED ISSUE #8122: disable `sanityCheckPackage` in the `doPack` step
* **Electron-Builder Version**: 24.13.3 

* **Node Version**: v18.19.0

* **Electron Version**: 28.2.3

* **Target**: win32 & darwin

Is it somehow possible to disable `sanityCheckPackage` check? I am using a custom electron fork where I implemented encrypted asar support. When packing my app I encrypt the asar archive in the `afterPack` hook, but the `sanityCheckPackage` then fails because it can't open the encrypted asar.

I'm happy to contribute and implement `sanityCheckPackage` disabling through the config if it's not possible already.

beyondkmp:
fix #8122

### DIFF
--- a/.changeset/dirty-rats-happen.md
+++ b/.changeset/dirty-rats-happen.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: add disableSanityCheckPackage to allow encrypted asars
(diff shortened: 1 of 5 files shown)


beyondkmp:
@mmaietta In window, there's an error when running pnpm generate-all, so I did not update the config yesterday.
image (7)

PR #8128, opened 2024-03-12
fix: order files in asar for incremental updates

indutny-signal:
ASAR file begins with a header that list all files and an offset to each file in the rest of the file. When a file placed early in ASAR changes its length - it means that all subsequent file declarations in the header will have their offsets updated. While harmless by itself, this negatively affects the incremental download size as more of the installer binary is different from what it used to be.

In this change we order files in asar such that:

- Dependencies/node_modules come first (they change least often)
- Main app files come last (they change more frequently)

Additionally, files in asar are now ordered alphabetically within each fileset to guarantee stable output.

All of above results in 2x improvement of incremental download size for updates that don't change dependencies.

### DIFF
--- a/.changeset/fair-pants-talk.md
+++ b/.changeset/fair-pants-talk.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: order files within asar for smaller incremental updates
--- a/packages/app-builder-lib/src/asar/asarUtil.ts
+++ b/packages/app-builder-lib/src/asar/asarUtil.ts
@@ -38,10 +38,18 @@ export class AsarPackager {
     }
     await mkdir(path.dirname(this.outFile), { recursive: true })
     const unpackedFileIndexMap = new Map<ResolvedFileSet, Set<number>>()
-    for (const fileSet of fileSets) {
+    const orderedFileSets = [
+      // Write dependencies first to minimize offset changes to asar header
+      ...fileSets.slice(1),
+
+      // Finish with the app files that change most often
+      fileSets[0],
+    ].map(orderFileSet)
+
+    for (const fileSet of orderedFileSets) {
       unpackedFileIndexMap.set(fileSet, await this.createPackageFromFiles(fileSet, packager.info))
     }
-    await this.writeAsarFile(fileSets, unpackedFileIndexMap)
+    await this.writeAsarFile(orderedFileSets, unpackedFileIndexMap)
   }
 
   private async createPackageFromFiles(fileSet: ResolvedFileSet, packager: Packager) {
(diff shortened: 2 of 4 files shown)


beyondkmp:
This is so cool.

indutny-signal:
Now that I think more about it - is there any point of having separate file sets? Could we merge them all together without a risk?

beyondkmp:
@indutny-signal I don't quite understand; could you explain in more detail? Does this involve merging all files together?

indutny-signal:
@beyondkmp no, only merging the file sets. The resulting ASAR should be the same if maybe ordered a bit more consistently.

mmaietta:
Nice work!

Related note re: the filesets and merging them, I've been yet to figure out their true purpose for being separated. I've been working in my off time to migrate us to the official electron/asar, but still not to any success yet due to the way we handle filesets

indutny-signal:
@mmaietta thanks!

Yeah, it is... funky. I think we have separate file sets for separate matchers because matchers can specify "from" and "to" paths, but I'm not entirely sure why these have to be separate. If we ever migrate to electron/asar - that I mean that I should send a PR to them as well :joy:

indutny-signal:
Blerg, sorry for snapshots failing. It makes sense that all offsets are changed, but I couldn't get tests running properly on my arm64 macbook.

mmaietta:
@indutny-signal let me take a shot at updating them from my side on my arm64 mac
cmd to run is `UPDATE_SNAPSHOT=true TEST_FILES=HoistedNodeModuleTest pnpm -w run test-linux`

indutny-signal:
@mmaietta done

indutny-signal:
I'm actually not entirely sure why it is failing in CI. It seems to pass locally...

mmaietta:
Will make this into the next release once I can get all snapshots updated

Related note re: differential updates, MacOS differential is now supported in [electron-updater@6.3.0-alpha.1](https://github.com/electron-userland/electron-builder/releases/tag/electron-updater%406.3.0-alpha.1). As with all electron-updater changes, please make sure to thoroughly test from your side

indutny-signal:
Aww, thanks so much!

We use our own differential update system so we won't get a chance to test it :(

mmaietta:
Very cool! Out of curiosity, does your differential update system work with dmgs?

indutny-signal:
@mmaietta nope, we use zip files!

mmaietta:
@indutny-signal, wanted to provide a quick update on the `electron/asar migration`. It is complete! Along with electron/fuses integration. It's in release [v26.0.0-alpha.2](https://github.com/electron-userland/electron-builder/releases/tag/v26.0.0-alpha.2).

>Yeah, it is... funky. I think we have separate file sets for separate matchers because matchers can specify "from" and "to" paths, but I'm not entirely sure why these have to be separate. If we ever migrate to electron/asar - that I mean that I should send a PR to them as well

AFAICT, I was able to retain your ordered filesets implementation (all test snapshots passed) as I order the filesets first, before pushing it through the detect library/executables and copy logic, which then retains the order of the original filesets. I create a temporary `ordering.txt` file that I then pass to `electron/asar` and it logs that 100% coverage was achieved. Great work to you on this PR!

Would love if you'd be willing to give the new release a shot and see how it works for you.

PR #8139, opened 2024-03-18
chore(deps): update app-builder-bin to 5.0.0-alpha

mmaietta:


mmaietta:
@beyondkmp , `mksquashfs` is working now, great call on the downgrade. Back to the same dependency issue in app-builder now 🙁 .
```
 FAIL  src/HoistedNodeModuleTest.ts (22.464 s)
  ✕ conflict versions (4538 ms)

  ● conflict versions

    ENOENT: no such file or directory, scandir '/tmp/et-db411cd21f3bbd882f3a5a30c58db6ff/t-jNdi69/test-project-1/packages/test-app/node_modules/foo'
```

beyondkmp:
@mmaietta I have fixed this in https://github.com/beyondkmp/electron-builder/actions/runs/8339546872/job/22821735838?pr=2.
image


But there are some other failed test cases.

mmaietta:
Nice work! Yeah, those are some hefty snapshots that are failing. I'll take a look at them as well and see what I can report back with

PR #8147, opened 2024-03-22
feat: App builder bin 5.0

beyondkmp:


beyondkmp:
@mmaietta Please help review.

packages/app-builder-lib/src/util/appFileCopier.ts
```diff
@@ -18,26 +18,73 @@ const BOWER_COMPONENTS_PATTERN = `${path.sep}bower_components${path.sep}`
...
+  const parts = file.split(path.sep)
+
+  // Find the index of the '.pnpm' folder
+  const pnpmIndex = parts.findIndex(part => part === ".pnpm")
+
+  // If '.pnpm' is found, and there are at least two more folders after it
+  if (pnpmIndex >= 0 && parts.length > pnpmIndex + 2) {
+    // Remove '.pnpm' and the next two folders from the parts array
+    parts.splice(pnpmIndex, 3)
+  }
```
mmaietta, line 32: Are we able to make this guarantee that the two nested folders within `.pnpm` folder are okay to be removed? Are they unnecessary symlinks or something?

beyondkmp:
I have a question about conflict versions 2.
                 |------ ms 2.1.1
test-app--- 
                 |------ foo 1.0.0 ----| --- ms 2.0.0
                 
The expected result is 
```
exports[`conflict versions 2`] = `
Object {
  "files": Object {
    "index.html": Object {
      "offset": "9018",
      "size": 841,
    },
    "index.js": Object {
      "offset": "9859",
      "size": 2501,
    },
    "node_modules": Object {
      "files": Object {
        "foo": Object {
          "files": Object {
            "package.json": Object {
              "offset": "4310",
              "size": 127,
            },
          },
        },
        "ms": Object {
          "files": Object {
            "index.js": Object {
              "offset": "4437",
              "size": 3034,
            },
            "license.md": Object {
              "offset": "7471",
              "size": 1077,
            },
            "package.json": Object {
              "offset": "8548",
              "size": 470,
            },
          },
        },
      },
    },
    "package.json": Object {
      "offset": "12360",
      "size": 301,
    },
  },
}
`;
```

I think the expected result is wrong, the right expected result is following:
```
exports[`conflict versions 2`] = `
Object {
  "files": Object {
    "index.html": Object {
      "offset": "9018",
      "size": 841,
    },
    "index.js": Object {
      "offset": "9859",
      "size": 2501,
    },
    "node_modules": Object {
      "files": Object {
        "foo": Object {
          "files": Object {
            "package.json": Object {
              "offset": "4310",
              "size": 127,
            },
          },
           "files": Object {
            "node_modules": Object {
                  "ms": Object {
          "files": Object {
            "index.js": Object {
              "offset": "4437",
              "size": 3034,
            },
            "license.md": Object {
              "offset": "7471",
              "size": 1077,
            },
            "package.json": Object {
              "offset": "8548",
              "size": 470,
            },
          },
        },
            },
          },
        },
        "ms": Object {
          "files": Object {
            "index.js": Object {
              "offset": "4437",
              "size": 3034,
            },
            "license.md": Object {
              "offset": "7471",
              "size": 1077,
            },
            "package.json": Object {
              "offset": "8548",
              "size": 470,
            },
          },
        },
      },
    },
    "package.json": Object {
      "offset": "12360",
      "size": 301,
    },
  },
}
`;
```

mmaietta:
It looks like we're copying the same assets twice now, but I find it odd that the `offset` and `size` values are the same. AFAICT, the packaged app size shouldn't be changed since the total `offset` is still the same for the asar. As such, I think just updating the snapshot should be fine and we can proceed with the PR

test/snapshots/HoistedNodeModuleTest.js.snap
```diff
@@ -22,7 +22,7 @@ Object {
         "foo": Object {
           "files": Object {
             "package.json": Object {
-              "offset": "0",
+              "offset": "4310",
```
mmaietta, line 25: This is an odd change. What's at offset `0` now?

packages/app-builder-lib/src/util/appFileCopier.ts
```diff
@@ -18,28 +18,61 @@ const BOWER_COMPONENTS_PATTERN = `${path.sep}bower_components${path.sep}`
...
+  return parts.join(path.sep)
+}
+
+function getHoistedModulePath(file: string, destination: string): string {
+  const filePathParts: string[] = file.split(path.sep)
+  const destinationParts: string[] = destination.split(path.sep)
+
+  const nodeModulesIndicesFilePath: number[] = filePathParts.reduce((acc: number[], part: string, index: number) => {
+    if (part === "node_modules") acc.push(index)
```
mmaietta, line 36: Minor: Please move this to new line
```
if (part === "node_modules") { 
   acc.push(index)
}
```

Same comment applies to line 41

beyondkmp:
> It looks like we're copying the same assets twice now, but I find it odd that the `offset` and `size` values are the same. AFAICT, the packaged app size shouldn't be changed since the total `offset` is still the same for the asar. As such, I think just updating the snapshot should be fine and we can proceed with the PR

You might have misunderstood. My example above was just that, directly copying data from MS modules. The approach of electron-builder is to copy two different versions of MS to the same directory, with the last copied version overriding the previous ones.

I think there might be some issues with this approach, but it seems like no one has ever reported a problem.

mmaietta:
Ohh, interesting! I was curious as to how electron-builder has handled conflicting versions. Agreed, no one has ever reported a problem over the few years I've managed this project, so we may be able to proceed with your suggested/expected result as-is as it seems like it'd be the correct approach to me.

beyondkmp:
> Ohh, interesting! I was curious as to how electron-builder has handled conflicting versions. Agreed, no one has ever reported a problem over the few years I've managed this project, so we may be able to proceed with your suggested/expected result as-is as it seems like it'd be the correct approach to me.

Under the current implementation, this is not achievable. I have made modifications to the app-builder's implementation to achieve this. Additionally, after making these changes, we no longer need to handle pnmp separately.

Taking test-app-yarn-workspace-version-conflict as an example, the old app-builder output format was like this: 
```json
[
	{
		"dir": "~\\app-builder\\pkg\\node-modules\\yarn-demo\\node_modules",
		"deps": [
			{
				"name": "ms",
				"version": "2.0.0"
			}
		]
	},
	{
		"dir": "~\\app-builder\\pkg\\node-modules\\yarn-demo\\packages\\test-app\\node_modules",
		"deps": [
			{
				"name": "foo",
				"version": "1.0.0"
			},
			{
				"name": "ms",
				"version": "2.1.1"
			}
		]
	}
]
```


After the changes, the output format is like this: 
```
[
	{
		"name": "foo",
		"version": "1.0.0",
		"dir": "~\\app-builder\\pkg\\node-modules\\yarn-demo\\packages\\foo",
		"conflictDependency": [
			{
				"name": "ms",
				"version": "2.0.0",
				"dir": "~\\app-builder\\pkg\\node-modules\\yarn-demo\\node_modules\\ms"
			}
		]
	},
	{
		"name": "ms",
		"version": "2.1.1",
		"dir": "~\\app-builder\\pkg\\node-modules\\yarn-demo\\packages\\test-app\\node_modules\\ms"
	}
] 
```


In the new format, it's clearer to see which packages are depended on and which ones are conflicting.

beyondkmp:
> I have already integrated the new implementation method in this PR([beyondkmp#2](https://github.com/beyondkmp/electron-builder/pull/2)) and passed all the tests. When you have time, please review it. If there are no issues, I can first submit a PR to app-builder and then submit this PR to electron-builder.

Here's app-builder PR(https://github.com/develar/app-builder/pull/116)

PayneFuRC:
The app-builder PR(https://github.com/develar/app-builder/pull/116) is merged. Please help publish a alpha build(v5.0.0-alpha.3) for app-builder.

beyondkmp:
The app-builder PR(https://github.com/develar/app-builder/pull/116) is merged. Please help publish a alpha build(v5.0.0-alpha.3) for app-builder.

mmaietta:
Done https://github.com/develar/app-builder/releases/tag/v5.0.0-alpha.3

beyondkmp:
closed as #8190.

PR #8152, opened 2024-03-25
fix: should not chmod for 7za when process.env.USE_SYSTEM_7ZA is true

### LINKED ISSUE #8151: Setting USE_SYSTEM_7ZA environment variable breaks the build
electron-builder version: **24.13.3**
electron version: **27.3.1**
node-js version: **20.10.0**

If we set the `USE_SYSTEM_7ZA=true` environment variable, then the `7zip-bin` package will return `7za` as path to the executable: https://github.com/develar/7zip-bin/blob/master/index.js#L7.

This is used here: https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util/src/7za.ts#L5.

The `chmod` call is failing because we do not specify the absolute path to the executable.

This is a regression from v24.6.4 where no call to `chmod` was made.

beyondkmp:
fix #8151

### DIFF
--- a/.changeset/selfish-experts-wonder.md
+++ b/.changeset/selfish-experts-wonder.md
@@ -1,0 +1,5 @@
+---
+"builder-util": patch
+---
+
+should not chmod for 7za when process.env.USE_SYSTEM_7ZA is true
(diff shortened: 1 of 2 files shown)


mmaietta:
Many thanks @beyondkmp !

Issue #5688, opened 2021-03-08
Security concerns with the GH_TOKEN


namdao2000:
In order to deploy the application and release it on GitHub, my GH_TOKEN requires `package:write` scope on, which also gives a person the permission to delete/modify/add files to the private repo. On the other hand, the only privilege I need in the GH_TOKEN for distribution is for `package:read`, which only allows the person to download the packages. Any chance I can use the `package:write` GH_TOKEN for deployment and `package:read` token for the distributed app? 

Thank you all in advance :D

mmaietta:
That's a great concern.

Looks like it is not possible to use separate tokens at this time.
https://github.com/electron-userland/electron-builder/blob/master/packages/electron-publish/src/gitHubPublisher.ts#L45
```
token = process.env.GH_TOKEN || process.env.GITHUB_TOKEN
```

https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/providerFactory.ts#L23-L28
```
const token = (githubOptions.private ? process.env.GH_TOKEN || process.env.GITHUB_TOKEN : null) || githubOptions.token
```

I'd be happy to review a PR if you're interested in discussing a solution.
Contributing doc: https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#to-setup-a-local-dev-environment

namdao2000:
Thank you for your reply mmaietta,

I am not good enough to contribute to the project, but I do hope someone out there can see this and fix this issue.

jaydoherty:
I agree with this idea _in theory_. In my opinion, something like this should work:
1. Set your GH_TOKEN or GITHUB_TOKEN environment variable to your "private" token with publishing privileges. 
2. Set "build": {"publish":{"token": "put your public lesser-privileged readonly token here"}}.

Sadly this doesn't work right now because electron-builder will completely ignore GH_TOKEN if "token" is specified directly in the configuration files. I know this from trying it out on electron-builder 22.14.5.

However...
Does GitHub allow tokens with only readonly scope? I thought GitHub only allowed users to create tokens with a minimum of **BOTH** pull **AND** push permission. I figured this was why electron-builder had never bothered to support separate tokens. At least, there is a long discussion on the topic here (with no progress as of December 2021): https://github.community/t/can-i-give-read-only-access-to-a-private-repo-from-a-developer-account/441

**TL;DR:** this feature might be impossible until GitHub allows more granularity in token permissions.

[...]

PR #8173, opened 2024-04-12
feat: Add Support for a separate Github release token to the auto-update token

AndrewEQ:
Reference: https://github.com/electron-userland/electron-builder/issues/5688 - Discussion about keeping two separate tokens; one for publishing to Github releases and the other used by the app to make requests for auto-update updates.

Now you can set a release token that has write permissions to publish your release.

The release token will be used instead of a GH_TOKEN || GITHUB_TOKEN ONLY during publishing.

The Github token defined via the Github options or environment variable will still be used as normal.

mac: ``` export GITHUB_RELEASE_TOKEN=<my token> ```

I used the Contents permission for a New fine-grained personal access token with "Read and write". "Read-only" for the usual app-update token.

So even if the app-update token is inside your app-update.yml its only read-only, yay! (Mac: you can find the app-update.yml by right-click > Show Package Contents > Contents > Resources)

### DIFF
--- a/.changeset/serious-pumpkins-suffer.md
+++ b/.changeset/serious-pumpkins-suffer.md
@@ -1,0 +1,5 @@
+---
+"electron-publish": minor
+---
+
+feat: Support for a separate Github publish token to the auto-update token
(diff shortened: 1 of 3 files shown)


mmaietta:
Love this idea, thank you for opening a PR!

Do we need to update the docs anywhere for this change? Looks like this may need to be updated to include your new logic/description
https://www.electron.build/configuration/publish

Issue #8188, opened 2024-04-26
Modifying files before package/build using distributed node version


rafaberaldo:
* **Electron-Builder Version**: 24.13.3
* **Node Version**: 20.11.1
* **Electron Version**: 29.3.0
* **Electron Type (current, beta, nightly)**: stable/current
* **Target**: win32

Related: https://github.com/electron-userland/electron-builder/issues/5619

I need to modify some files before packing using the distributed node/electron, the problem is that I'm building it on a x64 machine and the distributable is x86 (meaning I can't just use ``node_modules/electron``).

``beforePack`` hook doesn't work because distributable electron hasn't been downloaded/unziped yet, and ``afterPack`` already has the asar archive.

I did find a solution using the `@electron/asar` package:

- set ``asar: true`` and ``asarUnpack: '**/*'`` this is for electron-builder don't throw error saying the main.js doesn't exist.
- then the afterPack hook:
```js
import asar from '@electron/asar';
import { FuseV1Options, FuseVersion, flipFuses } from '@electron/fuses';
import fs from 'node:fs';
import path from 'node:path';

async function afterPack (context) {
    const fileName = context.packager.appInfo.productFilename;
    const binPath = path.join(context.appOutDir, `${fileName}.exe`);

    // Modifying some files
    await compileFile({
      filename: path.join(context.appOutDir, 'resources/app.asar.unpacked/dist/main.js'),
      node: binPath,
    });

    // Packing to asar
    await asar.createPackage(
      path.join(context.appOutDir, 'resources/app.asar.unpacked'),
      path.join(context.appOutDir, 'resources/app.asar'),
    );
    
    // Remove leftover files
    fs.rmSync(path.join(context.appOutDir, 'resources/app.asar.unpacked'), {
      recursive: true,
      force: true,
    });

    // Flip fuses
    await flipFuses(binPath, {
      version: FuseVersion.V1,
      [FuseV1Options.RunAsNode]: false,
      [FuseV1Options.OnlyLoadAppFromAsar]: true,
    });
  },
  ```

It appears to have worked correctly, and the build size is smaller, but I would like to know if there's a better solution or any downsides to it, does it break the NSIS auto updater?

Thanks!

rafaberaldo:
After playing a bit with `@electron/packager`, I could use theirs `afterCopy` hook to accomplish that.

Another option I think of would be using `@electron/get` to download the distributed version and modify the files before electron-builder. Don't like this because it'd modify the source.

[...]

Issue #5709, opened 2021-03-13
[Feature request]: AfterDependenciesInstalled hook


cawa-93:
I need a method that would allow me to control exactly which dependencies come in the distributable package and which should be excluded.

For this I propose to hang a `afterDependenciesInstalled ` hook that will fire immediately after the dependencies installed.

### Use case:
Using bundles, such as rollup or Vite, some runtime dependencies are optimized and bundled into source code. This means that there is no need to keep these optimized dependencies in node_modules.

With  `afterDependenciesInstalled ` hook, we could remove dependencies that have already been optimized and bundled.

mmaietta:
What about just manually specifying a negation in the `files` config?
```
    files: [
        "out",
        "!node_modules/some-package",
        "!**/*.node",
        "!**/*.LICENSE.txt",
        "!**/*.map"
    ],
```
The Documentation site says that all prod dependencies are automatically included because `**/node_modules/**/*` is appended, but my project actually excludes all dependencies via `!node_modules` due to the way we have to bundle the source code. Any negations seem to be prioritized over the 'allow' regexes, possibly order-based.

cawa-93:
@mmaietta I don't really understand how this should work.

Let's say I have a Vue production dependency

Vue depends on @ vue / compiler-dom, and that on @ vue / compiler-core

`Vue` -> `@vue/compiler-dom` -> `@vue/compiler-core`

I can exclude Vue from the package, as you described, but all other dependencies will remain in the package

```
files: [
        "dist",
        "!node_modules/vue",
]
```
`X` -> `@vue/compiler-dom` -> `@vue/compiler-core`

cawa-93:
> Is this still relevant?

Yeap

cawa-93:
> Is this still relevant?

Ага

mmaietta:
Please consider opening a PR to contribute
https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#to-setup-a-local-dev-environment

Issue #5619, opened 2021-02-08
Feature request: beforePack callback (or: using distributed Node version to run some scripts before packaging)


ravicious:
* **Version**: 22.9.1
* **Electron Version**: 11.3.1
* **Electron Type (current, beta, nightly)**: current
* **Target**: darwin, windows

electron-builder already offers the `afterPack` callback. However, before all the files are packaged into an asar, I need to run [bytenode](https://github.com/OsamaAbbas/bytenode/) on some of them. Now, in order for bytenode to work properly, it needs to be executed [using the Node binary which will be distributed with the app](https://github.com/OsamaAbbas/bytenode/issues/73#issuecomment-745100734).

As of now, when I use asar, `afterPack` executes when the files are already in the asar archive. The current workarounds I came up with:

1. From within `afterPack` run code that will unpack asar, compile the files with bytenode and then create the asar archive again.
   * This already doesn't play well with `asarUnpack`, as extracting the asar archive into an `app` folder copies those unpacked files there as well. So before creating the archive again I would need to remove paths from `asarUnpack` from my `app` folder. But I worry that in the future it's going to cause problems with `smartUnpack` anyway.
2. Use `extraResources` so that the files which need to be compiled with bytenode are not in the asar archive at all. Then compile them in `afterPack`.
   * If those compiled files happen to require some node modules, then most probably I'd need to manipulate the `NODE_PATH` env variable so that Node knows where to find the modules (in the asar archive). I'm yet to check if this actually works.

With something like `beforePack`, I could run bytenode on selected files and then builder could just include them in the asar as usual, without needing to worry about `asarUnpack` and `smartUnpack` myself.

This is somewhat similar to #4600, but the author of that issue seems to have asked about sth like `afterPack`.

Perhaps there's a more straightforward way to solve my problem than `beforePack`, in which case I'd highly appreciate any advice!

ravicious:
Welp, my 2nd workaround doesn't work, as `NODE_PATH` is used as a location of last resort if Node.js can't find `node_modules` with [its default algorithm](https://nodejs.org/docs/latest-v12.x/api/modules.html#modules_loading_from_node_modules_folders). In my case it'll "exit" the `dist` folder while looking for `node_modules` and stumble upon `node_modules` in my project's root directory, where it'll encounter node modules compiled for the wrong version of Node.

[...]

Issue #4422, opened 2019-11-15
[feature request] provide a hook :beforeAsar


bruceauyeung:
* **Version**: 21.2.0

I am using  https://github.com/nklayman/vue-cli-plugin-electron-builder
I wanna add/modify/delete static resources in `public` directory before electron-builder packing those resources into app.asar. 
for example, electron-builder automatically rebuild native modules,but don't pack them into the final application. i wanna to copy those `.node` native module files into `public/build/` so electron-builder can automatically pack them into `app.asar`

i tried the hook `onNodeModuleFile`, but it happens after that electron-builder has already done fetching resources from `public` directory.

PR #8194, opened 2024-05-02
feat: add `afterExtract` hook

rafaberaldo:
Add a hook that triggers after the prebuilt Electron binary has been extracted to the output directory.

What I was looking for is a `beforeAsar` (or `afterCopy`) hook, but looks like the file copying function is tightly coupled with asar transformation, this is the next best option.

Partially solves https://github.com/electron-userland/electron-builder/issues/8188 + https://github.com/electron-userland/electron-builder/issues/5619 + https://github.com/electron-userland/electron-builder/issues/4422 + https://github.com/electron-userland/electron-builder/issues/5709

### DIFF
--- a/.changeset/tough-clouds-think.md
+++ b/.changeset/tough-clouds-think.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: add `afterExtract` hook to build process with the same payload interface as `beforePack` and `afterPack`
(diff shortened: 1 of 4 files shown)


rafaberaldo:
Not sure why `changeset` detected major version bump, I've commited a minor change

mmaietta:
@rafaberaldo not too sure either why it's flagged as `major`, but since we're in an alpha release state, I think it should be fine. Thanks for your contribution!

rafaberaldo:
Have any idea why is it failing on mac?

mmaietta:
I deactivated the mac test on `master`. For some reason, all previous versions of electron-builder for FilesTest on mac are broken now. I think a mac update occurred since the test failures randomly appeared last week with no changes to electron-builder.

Issue #8199, opened 2024-05-04
AppX identityName error


ifurther:
My package's identityName has some number first, but the identityName didn't support it.
the identityName  information
https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-identity

* **Electron-Builder Version**:  24.13.3

* **Node Version**: 

* **Electron Version**:
* **Electron Type (current, beta, nightly)**:

* **Target**:  win appx

PR #8203, opened 2024-05-06
feat(appx): Update identityName for windows 10

ifurther:
Updated the identityName rule
https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-identity

Fixes: #8199


mmaietta:
Need to revert this PR, it seems it also broke Appx unit tests. Please take a look when you have a chance. Thank you!

PR #8206, opened 2024-05-07
feat(appx): Update identityName for windows 10

ifurther:
- Split the [applicationid](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-application) and [identityname](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-identity) detect function
- Add auto remove the identityname in Microsoft store has some number in first

### DIFF
--- a/packages/app-builder-lib/src/targets/AppxTarget.ts
+++ b/packages/app-builder-lib/src/targets/AppxTarget.ts
@@ -224,13 +224,13 @@ export default class AppXTarget extends Target {
           return appInfo.getVersionInWeirdWindowsForm(options.setBuildNumber === true)
 
         case "applicationId": {
-          const result = options.applicationId || options.identityName || appInfo.name
           const validCharactersRegex = /^([A-Za-z][A-Za-z0-9]*)(\.[A-Za-z][A-Za-z0-9]*)*$/
-          const identitynumber = parseInt(result, 10)
-          if (!isNaN(identitynumber)) {
+          const identitynumber =  parseInt(options.identityName as string, 10) || NaN
+          const result = (!isNaN(identitynumber) && options.identityName !== null && options.identityName !== undefined) ? options.identityName.replace(identitynumber.toString(),'') : options.applicationId || options.identityName || appInfo.name
+
+          if (!isNaN(identitynumber) ) {
             log.warn(`Remove the ${identitynumber}`)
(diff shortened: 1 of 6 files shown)


ifurther:
@mmaietta I updated the test file.

mmaietta:
Looks like there's some compilation issues now:
```
Error: test/src/windows/appxTest.ts(58,9): error TS2353: Object literal may only specify known properties, and 'applicationId' does not exist in type 'WindowsConfiguration'.
```

ifurther:
@mmaietta  I removed it

ifurther:
@mmaietta Sorry I find the error which is for loop logical.

ifurther:
@mmaietta Sorry I fixed the if loop

ifurther:
@mmaietta I check the applicationid and identityname rule and update it.

ifurther:
@mmaietta I updated the replace and find number detection

ifurther:
@mmaietta I found the error and updated it

ifurther:
@mmaietta I found the error and updated it

packages/app-builder-lib/src/targets/AppxTarget.ts
```diff
@@ -199,19 +224,57 @@ export default class AppXTarget extends Target {
...
+          const result = options.identityName || appInfo.name
+          const validCharactersRegex = /^[a-zA-Z0-9.-]+$/
+          if (result.length < 3 || result.length > 50) {
+            const message = `Appx identityName.Id with a value between 3 and 50 characters in length`
+            throw new InvalidConfigurationError(message)
+          } else if (!validCharactersRegex.test(result)) {
+            const message = `AppX identityName.Id cat be consists of alpha-numeric, period, and dash characters"`
+            throw new InvalidConfigurationError(message)
+          } else if (restrictedApplicationIdValues.includes(result.toUpperCase())) {
+            const message = `AppX identityName.Id can not be some values`
```
mmaietta, line 269: Let's be more explicit in our error message in this case otherwise users won't know what was wrong with their Identity Name

Maybe we can use?
```
const message = `AppX identityName.Id can not include ${JSON.stringify(restrictedApplicationIdValues)}`
```

ifurther:
Thank you

PR #8245, opened 2024-06-03
feat: write asar integrity resource on windows

indutny-signal:
Electron 30-x-y added support for ASAR integrity fuse on Windows. When enabled the app would fetch the ELECTRONASAR resource out of the executable file and use it to verify the integrity of the ASAR when reading the data from it.

### DIFF
--- a/.changeset/thick-flies-carry.md
+++ b/.changeset/thick-flies-carry.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+write asar integrity resource on windows
--- a/packages/app-builder-lib/package.json
+++ b/packages/app-builder-lib/package.json
@@ -70,7 +70,7 @@
     "lazy-val": "^1.0.5",
     "minimatch": "^5.1.1",
     "read-config-file": "6.4.0",
-    "resedit": "^2.0.2",
+    "resedit": "^1.7.0",
     "sanitize-filename": "^1.6.3",
     "semver": "^7.3.8",
     "tar": "^6.1.12",
--- a/packages/app-builder-lib/package.json
+++ b/packages/app-builder-lib/package.json
@@ -70,6 +70,7 @@
     "lazy-val": "^1.0.5",
     "minimatch": "^5.1.1",
     "read-config-file": "6.4.0",
+    "resedit": "^2.0.2",
     "sanitize-filename": "^1.6.3",
     "semver": "^7.3.8",
     "tar": "^6.1.12",
--- a/packages/app-builder-lib/src/electron/ElectronFramework.ts
+++ b/packages/app-builder-lib/src/electron/ElectronFramework.ts
@@ -10,6 +10,7 @@ import { LinuxPackager } from "../linuxPackager"
 import { MacPackager } from "../macPackager"
 import { getTemplatePath } from "../util/pathManager"
 import { createMacApp } from "./electronMac"
+import { addWinAsarIntegrity } from "./electronWin"
 import { computeElectronVersion, getElectronVersionFromInstalled } from "./electronVersion"
 import * as fs from "fs/promises"
 import injectFFMPEG from "./injectFFMPEG"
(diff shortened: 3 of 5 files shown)


indutny-signal:
I'm not sure if I'm supposed to commit the lock file. On one hand, it has to be updated because I installed a new dependency, on the other - it seems like the format has changed and as a result the diff is too large. Any thoughts?

mmaietta:
@indutny-signal what version of pnpm are you running? Right now, the CI is locked to 8.9.0 so the pnpm lockfile should be regenerated with that version.

Separately, I can look into updating our pnpm version in the CI, but am swamped with work and some life events right now and probably won't be able to pick that up in the near future.

indutny-signal:
@mmaietta thank you for the exact version number! It should be fixed now.

mmaietta:
Out of curiosity (I'd love to learn more on this topic), how/where did you find the information for `resedit`? Does it work when cross-compiling on other OS'?

indutny-signal:
@mmaietta electron folks gave me a link to https://github.com/electron/packager/blob/main/src/resedit.ts :)

indutny-signal:
FWIW, I believe that both resedit not pe-edit (its dependency) are pure-JS libraries.

mmaietta:
Thanks for the link! I'll TAL. Looks like there's an import error during compilation though
```
Cannot find module 'resedit/cjs' from '../packages/app-builder-lib/out/electron/electronWin.js'
```

indutny-signal:
@mmaietta thanks! I'm trying to figure it out... It seems to be something specific to how jest does the resolution...

indutny-signal:
Alright, I substituted the cjs import with an async esm import. Let me know if this works!

indutny-signal:
That obviously didn't work. I'm surprised that test-windows/test-linux fails while test-mac works fine (as confirmed locally)

indutny-signal:
@mmaietta I didn't realize that CI was running on previous commit (before force-push). Do you mind restarting it?

mmaietta:
Sure thing! I've restarted the CI, not sure what commit it'll pull from though, so we'll see what happens. At worst, you may just need to push an empty commit to kick the CI to behave 🙂

indutny-signal:
Thank you! Looks like empty commit is the way to go sadly.

indutny-signal:
@mmaietta we are good, it seems!

mmaietta:
So I checked this out locally and tried moving this to the top-level
```
import { NtExecutable, NtExecutableResource, Resource } from "resedit"
```

Seems the error is coming from within the package though.
Screenshot 2024-06-06 at 10 55 24 AM
```
    /project/node_modules/.pnpm/resedit@2.0.2/node_modules/resedit/dist/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import { NtExecutable, NtExecutableResource, Format } from 'pe-library';
```

mmaietta:
Bah, even this fails with the same error:
```
const { NtExecutable, NtExecutableResource, Resource } = require("resedit")
```

indutny-signal:
@mmaietta downgraded. Looks like 1.7.0 was the last CJS release.

I just tried the test command locally and got:
```
    /project/node_modules/.pnpm/app-builder-bin@5.0.0-alpha.3/node_modules/app-builder-bin/linux/x64/app-builder process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
    Exit code:
    2

      250 |       }
      251 |     } else {
    > 252 |       reject(new ExecError(command, code, out, errorOut))
          |              ^
      253 |     }
      254 |   })
      255 | }
```

indutny-signal:
Ah, the true error was above:
```
  ⨯ cannot execute  cause=signal: segmentation fault
                    errorOut=qemu: uncaught target signal 11 (Segmentation fault) - core dumped

                    command=wine /root/.cache/electron-builder/winCodeSign/winCodeSign-2.6.0/rcedit-ia32.exe '/tmp/et-db411cd21f3bbd882f3a5a30c58db6ff/t-AhvwVh/test-project-8/dist/win-unpacked/Test App ßW.exe' --set-version-string FileDescription 'Test Application (test quite “ #378)' --set-version-string ProductName 'Test App ßW' --set-version-string LegalCopyright 'Copyright © 2024 Foo Bar' --set-file-version 1.1.0.42 --set-product-version 1.1.0.42 --set-version-string InternalName 'Test App ßW' --set-version-string OriginalFilename '' --set-version-string CompanyName 'Foo Bar' --set-icon /tmp/et-db411cd21f3bbd882f3a5a30c58db6ff/t-AhvwVh/test-project-8/build/icon.ico
                    workingDir=
```

indutny-signal:
Thank you!

nikitakot:
@indutny-signal is it possible to completely disable this feature? I disable the electron fuse in a custom electron-after-pack.js hook and the electron-builder library obviously doesn't detect that, so `updating asar integrity executable resource` is still being executed.

indutny-signal:
@nikitakot good question, after a quick check I don't see any way to opt out. Could you elaborate on why you need it? With fuses disabled it sounds like adding integrity check data should be relatively harmless?

Issue #8252, opened 2024-06-07
25.0.0-alpha.9 bundles @electron/remote wrongly into Windows app.asar


pbi-qfs:
* 25.0.0-alpha.9

* **Node Version**: v18.13.0

* **Electron Version**: 30.1.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: windows

The javascript files of @electron/remote are all placed into `app.asar/node_modules/@electron`, without any inner directory structure, so the internal relative `requires` fail.

mmaietta:
@beyondkmp can you take a look at this? I believe it's due to the new app-builder-bin alpha changes

beyondkmp:
Okay, I'll try to reproduce it first.

PR #8254, opened 2024-06-09
fix: @electron/remote wrongly into Windows app.asar

beyondkmp:
Fix https://github.com/electron-userland/electron-builder/issues/8252

### DIFF
--- a/.changeset/nice-weeks-sip.md
+++ b/.changeset/nice-weeks-sip.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: @electron/remote wrongly into Windows app.asar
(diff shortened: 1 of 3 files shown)


beyondkmp:
@mmaietta I found a new bug if we alias node module name. Please help revivew: https://github.com/develar/app-builder/pull/119

beyondkmp:
@mmaietta It seems like the CI has a cache, comparing with the previous snapshot. Often, the first run fails, and then it succeeds on a rerun without any changes. I haven't found the root cause yet.

mmaietta:
@beyondkmp thank you! Reran the CI and it all passes, I'm not sure re: the caching issue but it is indeed leading to a flaky CI (which needs fixing for sure)

Does this PR depend on a release of the app-builder-bin changes as well?

beyondkmp:
There's no interdependency. However, I think we can integrate the new app-builder-bin in this ticket for testing. Please review this PR(https://github.com/develar/app-builder/pull/120) again, and if there are no issues, merge it so we can release an alpha version. I'll integrate it into this ticket for testing.

beyondkmp:
@mmaietta Please help rerun the failed job.

PR #8271, opened 2024-06-23
fix(deps): Update notarize tool

kochie:
Updates @electron/notarize to v2.3.2 which should fix the issue with parsing errors created by `notarytool`

https://github.com/electron/notarize/pull/191

### DIFF
--- a/.changeset/tender-countries-drive.md
+++ b/.changeset/tender-countries-drive.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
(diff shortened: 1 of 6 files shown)


mmaietta:
Thanks for your contribution! You'll need to regenerate the pnpm-lock.yaml via `pnpm install` since we're updating a npm package here.

PBK-B:
Cool, thanks @kochie for your work. But I still want to know when this branch can be merged. I have been stuck on this problem since yesterday.

PR #8282, opened 2024-06-29
fix: change abort to aborted event

beyondkmp:
There are many SimpleURLLoaderWrapper server error in sentrty, I think they might be caused by this event.

Here's a simple example:
```javascript
// test.js
'use strict';

// The same issue occurs with `http`, too.
const https = require('https');

const req = https.request('https://nodejs.org/dist/v13.7.0/node-v13.7.0.tar.gz');
req.on('response', (res) => {
  res
    .on('end', () => console.log('end'))
    .on('close', () => console.log('close'))
    .on('aborted', () => console.log('aborted'))
    .on('error', (err) => console.log(err));

  setTimeout(() => {
    console.log('start');
    req.abort();
  }, 500);
});
req.end();
```

### DIFF
--- a/packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
+++ b/packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
@@ -233,10 +233,6 @@ export abstract class DifferentialDownloader {
 
         const request = this.httpExecutor.createRequest(requestOptions, response => {
           response.on("error", reject)
-          // for old nodejs:https://nodejs.org/docs/latest-v16.x/api/http.html#event-abort
-          response.on("abort", () => {
-            reject(new Error("response has been abort by the server"))
-          })
           response.on("aborted", () => {
             reject(new Error("response has been aborted by the server"))
           })
--- a/packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
+++ b/packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
@@ -296,10 +296,6 @@ export abstract class DifferentialDownloader {
         }
 
         response.on("error", reject)
-        // for old nodejs:https://nodejs.org/docs/latest-v16.x/api/http.html#event-abort
-        response.on("abort", () => {
-          reject(new Error("response has been abort by the server"))
-        })
         response.on("aborted", () => {
           reject(new Error("response has been aborted by the server"))
         })
--- a/packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
+++ b/packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
@@ -233,6 +233,10 @@ export abstract class DifferentialDownloader {
 
         const request = this.httpExecutor.createRequest(requestOptions, response => {
           response.on("error", reject)
+          // for old nodejs:https://nodejs.org/docs/latest-v16.x/api/http.html#event-abort
+          response.on("abort", () => {
+            reject(new Error("response has been abort by the server"))
+          })
           response.on("aborted", () => {
             reject(new Error("response has been aborted by the server"))
           })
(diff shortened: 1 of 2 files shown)


packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
```diff
@@ -233,7 +233,7 @@ export abstract class DifferentialDownloader {
 
         const request = this.httpExecutor.createRequest(requestOptions, response => {
           response.on("error", reject)
-          response.on("abort", () => {
+          response.on("aborted", () => {
```
mmaietta, line 236: So it looks like `response` is an `IncomingMessage` object (not sure why it is typed as any), but with it typed correctly the only events that can be listened to according to intellisense are:
Screenshot 2024-06-29 at 8 40 14 AM

I'm wondering if this implementation is legacy for older versions of nodejs?
Note: I think older versions of nodejs still need to be supported though otherwise a bunch of electron versions that electron-updater supports would need to be dropped.
Node 14 - https://nodejs.org/docs/latest-v14.x/api/http.html#http_event_abort
Node 16 - https://nodejs.org/docs/latest-v16.x/api/http.html#event-abort
> Stability: 0 - Deprecated. Listen for the 'close' event instead.

So it looks like `abort` is indeed correct, but for newer versions of nodejs, we might need to leverage something akin to:
```
response.on('close', () => {
   const isAborted = response.aborted
   // other logic
})
```

beyondkmp:
Here's a minimal example for you, with an 'aborted' event. However, I'm currently not sure why this event isn't documented.
```javascript
// test.js
'use strict';

// The same issue occurs with `http`, too.
const https = require('https');

const req = https.request('https://nodejs.org/dist/v13.7.0/node-v13.7.0.tar.gz');
req.on('response', (res) => {
  res.on('aborted', () => console.log('aborted'))
  res.on('abort', () => console.log('abort'))

  setTimeout(() => {
    console.log('start');
    req.abort();
  }, 500);
});
req.end();
```

mmaietta:
Okay, that's super weird indeed. Thanks for the test script! Maybe we should have both listeners (just in case for backward compatibility)?

Any chance you'd be willing to post a screenshot of what error you're seeing in Sentry?

beyondkmp:
@mmaietta Here's the detailed sentry error.  Because we recently enabled differential udpate, such errors have increased, so I suspect there is a connection here.

image

mmaietta:
Any chance you'd be willing to test this on your app locally using [patch-package](https://www.npmjs.com/package/patch-package) first? I don't immediately see any harm/side-effects in the changes in this PR, but I don't have a way to verify locally via unit tests AFAICT.

beyondkmp:
Unfortunately, I cannot reproduce this issue locally at the moment. It only seems to occur frequently in production when there is a large user base. I'll have to wait for production to test it out.

mmaietta:
@beyondkmp any luck with your patched production build rollout? Looking to determine next steps with this PR 🙂

beyondkmp:
It hasn't been released yet. Once it is released and we have results, I will provide immediate feedback.

beyondkmp:
Based on the official Node.js issues and the full download code from electron-builder, it seems that the `aborted` event is correct.

https://github.com/nodejs/node/issues/31630

packages/electron-updater/src/differentialDownloader/DifferentialDownloader.ts
beyondkmp: From the Node.js type definitions, not only are there these events, but you can also customize other events. 
image

Looking at the response definitions, you can see there is an `aborted` event but no `abort` event.
image

beyondkmp: Based on the official Node.js issues and the full download code(7 years ago) from electron-builder, it seems that the `aborted` event is correct.

https://github.com/nodejs/node/issues/31630

https://github.com/electron-userland/electron-builder/blob/1c4ec128ae877ad3a3a3c2765e98e59329fd900d/packages/builder-util-runtime/src/httpExecutor.ts#L140

mmaietta: Nice investigation work! I'm in favor of proceeding with this PR change then

beyondkmp:
@mmaietta Based on the release results, network errors have not increased; instead, they have decreased, although they have not been completely eliminated.

From the results, my suggestion is that this can be merged.

Issue #8284, opened 2024-06-29
Folder's named "constructor" not being included in asar


StrangeBytesDev:
* **Electron-Builder Version**:  v24.12.3

* **Node Version**:  v20.10.0

* **Electron Version**: 31.1.0

* **Electron Type (current, beta, nightly)**: current

* **Target**: linux-x64, windows-x64

Any directory in node_modules called "constructor" will not be included in the app.asar. This was previously a known issue with @electron/asar, however it has been fixed since v3.2.7 and I have verified that I am using v3.2.10. Running asar manually will include folders named "constructor" so I believe this is an issue with electron-builder itself. I've also tested with v25.0.0-alpha.10 and this is still an issue there. The specific package that is breaking is @sinclair/typebox.
For now, I'm able to create working builds by disabling asar, but that is of course non ideal.

PR #8286, opened 2024-06-30
fix: Folder's named "constructor" not being included in asar 

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8284

Refers to https://github.com/electron/asar/pull/253


beyondkmp:
I've verified it locally, and it's working fine.
image

Issue #8280, opened 2024-06-28
Unable to code sign DMG + MAS builds from a single configuration


IsaacAderogba:
* **Electron-Builder Version**: 24.13.3

* **Node Version**: 18.20.3

* **Electron Version**: 31.1.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: Mac `zip` and `mas`

I'm building an application for the conventional Mac artifacts (`dmg`, `zip`, etc...) and Mac App Store (`mas`).  This *should* require providing the different configuration keys with the correct certificate files. E.g.

```
 ...
  mac: {
    target: [
      { arch: ["x64", "arm64"], target: "zip" },
      { target: "mas", arch: "universal" },
    ],
    cscLink: "path-to-developer-id-certificate",
    cscKeyPassword: "...",
  },
  mas: {
    cscLink: "path-to-distribution-certificate",
    cscKeyPassword: "...",
    cscInstallerLink: "path-to-installer-certificate",
    cscInstallerKeyPassword: "...",
  },
...
```

When doing this, however, the certificate file path on `mas.cscLink` seems to be ignored. I was under the impression that, when building for Mac App Store, the keys on the `mas` configuration object would have precedence over those that were defined on the `mac` object. 

From digging through the source code for [macPackager](https://github.com/electron-userland/electron-builder/blob/88bbbdbe81936df1701f26138170e0f337c4f0d4/packages/app-builder-lib/src/macPackager.ts#L29-L49), what seems to be happening is that the `codeSigningInfo` property is only being resolved once and then its cached value is being reused for all of the different artifacts (even if those artifacts specify that a different `cscLink` should be used).

This seems counterintuitive and unintentional. I should be able to patch this out, but wanted to see whether it's being tracked internally first.

PR #8291, opened 2024-07-02
feat: add MemoLazy to fix codeSigningInfo not responding to changed args

IsaacAderogba:
Designed to fix the issue with `codeSigningInfo` not working when both `mac.cscLink` and `mas.cscLink` are defined with separate values. It currently only creates the `codeSigningInfo` value once and then reuses that across all other targets. Outlined here https://github.com/electron-userland/electron-builder/issues/8280.

A simpler approach would be to stop wrapping the execution of `codeSigningInfo` within the `Lazy` class, but this PR tries to preserve most of the existing behaviour.

Please close/ignore this if you deem it is working as intended. I can otherwise use patch-package to fix this, without maintaining a forked repo.

Thanks for your time!


packages/app-builder-lib/src/util/MemoLazy.ts
```diff
@@ -0,0 +1,36 @@
...
+    private creator: (selected: S) => Promise<V>
+  ) {}
+
+  get hasValue() {
+    return this._value !== undefined
+  }
+
+  get value(): Promise<V> {
+    const selected = this.selector()
+    if (this._value !== undefined && JSON.stringify(this.selected) === JSON.stringify(selected)) {
```
mmaietta, line 21: This is an awesome class, thank you! Quick Q here, what if `this.selected` is not serializable into JSON?

It might be worth opening a PR for https://github.com/develar/lazy-val/blob/master/src/main.ts as well so that we could get this pushed upstream. I can ping develar to get it merged (he's usually pretty quick on those once I email him 🙂 )

IsaacAderogba: Yeah, I would have preferred to use either a shallow or deep equals, but I wanted to avoid adding too much new code for the PR.

I'll make this PR upstream and I'll switch to a shallow equals implementation over there! If develar is happy with it, I'll come back here to bump the package and use the newly-exported `MemoLazy`

mmaietta:
Made some minor modifications, such as adding deep equals functionality via recursion and a unit test. Also updated `winPackager` csc signing logic to use the MemoLazy class now since I think it'll run into the same issue when having multiple Window targets.
Largest changes in the PR are just from regenerating the API docs.

Figured we could merge this in to resolve the issue and once the upstream PR is accepted, we can remove the local MemoLazy in favor of upstream.

PR #8295, opened 2024-07-04
fix: verify LiteralPath of update file during windows signature verification

mmaietta:
To prevent env var expansion during the signature verification step when executed via cmd.exe -> powershell, we need to verify the LiteralPath of the scanned asset and compare the string against the original **intended** update filename

### DIFF
--- a/.changeset/nervous-carrots-begin.md
+++ b/.changeset/nervous-carrots-begin.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix: verify LiteralPath of update file during windows signature verification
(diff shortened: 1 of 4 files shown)


basil:
Are there any plans for a non-alpha release or backport? This is now showing up as https://nvd.nist.gov/vuln/detail/CVE-2024-39698

mmaietta:
Yep, I'll be converting to non-alpha release when I return from vacation on Monday. Unfortunately, I cannot backport with the current CI/CD setup using `changeset` package.

defunctzombie:
I see that 6.3.1 is on npm but under the `next` tag rather than `latest`. Is it held back for more testing? When updating to 6.3.1 to resolve this security notice do we also need to update to a specific version of electron or electron builder?

mmaietta:
It needs broader adoption before I can post to `latest`. For instance, electron-builder v25 has a blocking issue with AppImages that I'm currently working on resolving.

To leverage the fix in electron-updater 6.3.1, it does not require a specific version of electron or electron-builder.

PR #8310, opened 2024-07-09
fix: binary checking

beyondkmp:
If a module contains a folder with a name like `test.app` or `test.framework`, and there are binary files inside, it will be incorrectly processed and the module will be placed in `app.asar` instead of being smartly unpacked into `app.asar.unpacked`.

### DIFF
--- a/packages/app-builder-lib/src/asar/unpackDetector.ts
+++ b/packages/app-builder-lib/src/asar/unpackDetector.ts
@@ -82,7 +82,7 @@ export async function detectUnpackedDirs(fileSet: ResolvedFileSet, autoUnpackDir
     const moduleName = path.basename(packageDir)
     if (moduleName === "ffprobe-static" || moduleName === "ffmpeg-static" || isLibOrExe(file)) {
       shouldUnpack = true
-    } else if (!file.includes(".", nextSlashIndex)) {
+    } else {
       shouldUnpack = !!isBinaryFileSync(file)
     }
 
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/asar/unpackDetector.ts
```diff
@@ -82,7 +82,11 @@ export async function detectUnpackedDirs(fileSet: ResolvedFileSet, autoUnpackDir
     const moduleName = path.basename(packageDir)
     if (moduleName === "ffprobe-static" || moduleName === "ffmpeg-static" || isLibOrExe(file)) {
       shouldUnpack = true
-    } else if (!file.includes(".", nextSlashIndex)) {
+    }
+
+    const fileBaseName = path.basename(file)
+    const hasExtension = path.extname(fileBaseName)
+    if (!hasExtension) {
```
mmaietta, line 89: I think this needs to remain an `else-if` otherwise `isBinaryFileSync` could override `shouldUnpack` for the file-specific logic of
```
moduleName === "ffprobe-static" || moduleName === "ffmpeg-static" || isLibOrExe(file)
```

beyondkmp: Correct.

beyondkmp:
@mmaietta I'm curious why these files are being checked separately. They should all be binary files as well.

```
moduleName === "ffprobe-static" || moduleName === "ffmpeg-static" || isLibOrExe(file)
```

Do you know why this is happening? This approach doesn't seem very smart.

mmaietta:
> @mmaietta I'm curious why these files are being checked separately. They should all be binary files as well.
> 
> ```
> moduleName === "ffprobe-static" || moduleName === "ffmpeg-static" || isLibOrExe(file)
> ```
> 
> Do you know why this is happening? This approach doesn't seem very smart.

Honestly, I don't know either. That logic predates me joining the maintainer crew.

mmaietta:
Side note, thank you @beyondkmp for all your contributions! I really appreciate all the efforts you've been doing to help improve electron-builder (and app-builder as well)!

beyondkmp:
@mmaietta   I have a question: how can I become a maintainer?

mmaietta:
If you have discord, contact me at `onegoldfish` and we can chat further and be able to stay more in sync for collaboration 😃

PR #8344, opened 2024-07-17
feat: use electron-winstaller instead of self module

beyondkmp:
1. Add `customSquirrelExePath` option, which allows using custom squirrel windows
2. Use official electron-winstaller
3. Support Windows ARM
4. Fix some bugs(https://github.com/electron-userland/electron-builder-binaries/issues/33, https://github.com/electron-userland/electron-builder-binaries/issues/52)


patches/electron-winstaller@5.3.1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/lib/index.js b/lib/index.js
+index d666a6316273b96cb67213e40e17c4a30625f63b..0e328caa0e33d13d6081f8aa42662bc28aaa7116 100644
+--- a/lib/index.js
++++ b/lib/index.js
+@@ -161,7 +161,7 @@ function createWindowsInstaller(options) {
+                     _f.label = 9;
+                 case 9:
+                     Object.assign(metadata, {
+-                        exe: "".concat(appMetadata.name, ".exe"),
++                        exe: "".concat(appMetadata.productName, ".exe"),
```
mmaietta, line 10: What's this for? Can we continue using `name` in `options`? I see the `productName` is removed from options here: https://github.com/electron-userland/electron-builder/pull/8344/files#diff-0d44b799fbfefe8d2452d2b7395cb069788adcde5e199b3e83b6de88fc0c97a4L105

Stanzilla:
Is this package actually still supported by the electron team? Many packages in that org have been dead for a long time.

mmaietta:
> Is this package actually still supported by the electron team? Many packages in that org have been dead for a long time.

That's a great point. @beyondkmp what bug fixes is this solving for? (Can't tell from the description)

mmaietta:
@Stanzilla looks like updates are still being pushed to `electron-winstaller`, last update was 3mo ago. https://www.npmjs.com/package/electron-winstaller

patches/electron-winstaller@5.3.1.patch
beyondkmp: `electron-winstaller` provides an exe parameter configuration, so I have removed this patch.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -102,22 +101,21 @@ export default class SquirrelWindowsTarget extends Target {
...
       version: appInfo.version,
       description: appInfo.description,
-      // better to explicitly set to empty string, to avoid any nugget errors
+      exe: `${this.packager.platformSpecificBuildOptions.executableName || this.options.name || appInfo.productName}.exe`,
       authors: appInfo.companyName || "",
       iconUrl,
       extraMetadataSpecs: projectUrl == null ? null : `\n    <projectUrl>${projectUrl}</projectUrl>`,
       copyright: appInfo.copyright,
       packageCompressionLevel: parseInt((process.env.ELECTRON_BUILDER_COMPRESSION_LEVEL || packager.compression === "store" ? 0 : 9) as any, 10),
-      vendorPath: await getBinFromUrl("Squirrel.Windows", "1.9.0", "zJHk4CMATM7jHJ2ojRH1n3LkOnaIezDk5FAzJmlSEQSiEdRuB4GGLCegLDtsRCakfHIVfKh3ysJHLjynPkXwhQ=="),
```
mmaietta, line 115: If we're removing this, we need to validate the validity/integrity of the new package. To replicate the previoius functionality, electron-builder should validate the hash value of the exe's in the npm package by hardcoding the hashes in our repo. This will replicate what the current `getBinFromUrl` is doing with the hardcoded hash

beyondkmp:
> > Is this package actually still supported by the electron team? Many packages in that org have been dead for a long time.
> 
> That's a great point. @beyondkmp what bug fixes is this solving for? (Can't tell from the description)



> > Is this package actually still supported by the electron team? Many packages in that org have been dead for a long time.
> 
> That's a great point. @beyondkmp what bug fixes is this solving for? (Can't tell from the description)

It's a bit much; we can get it from the release page(https://github.com/Squirrel/Squirrel.Windows/releases). 

However, what I want most is to update to the latest commit to include this fix: [https://github.com/Squirrel/Squirrel.Windows/pull/1862](https://github.com/Squirrel/Squirrel.Windows/pull/1862). My plan is to first update to version 2.0.1. If there are no issues, then update to the latest.

beyondkmp:
This library is being updated, but some updates are not timely, and some executable files are almost 10 years old. I have identified two issues and submitted PRs for them.

https://github.com/electron/windows-installer/pull/524
https://github.com/electron/windows-installer/pull/525

Currently, I have forked the repository and published `electron-squirrel-winstaller` to npm. I want to update this library more quickly, and I will start using the `electron-squirrel-winstaller` library in our company's app production.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: This is a bit difficult. In the [https://github.com/electron/windows-installer](https://github.com/electron/windows-installer) library, the vendor(https://github.com/electron/windows-installer/tree/main/vendor) is already included, and it contains all the exe files. We no longer need to download them separately as they are included with the package download. So, it seems we don't need to validate them again, right?

mmaietta: We need to validate them from our side for security reasons. An npm package can change hashes for the same version with a force push IIRC. That's what electron-builder-binaries is used for.

An alternative route is to move the vendor files into this monorepo as separate github release artifacts (but not published to npm). 

The previous logic only downloads the necessary vendor packages when they are necessary. We shouldn't be downloading all resources on `npm install`

packages/electron-builder-squirrel-windows/package.json
```diff
@@ -16,9 +16,8 @@
   ],
```
mmaietta, line 16: I think `template.nuspectemplate` needs to be added to `files` array

packages/app-builder-lib/src/options/SquirrelWindowsOptions.ts
```diff
@@ -40,6 +40,11 @@ export interface SquirrelWindowsOptions extends TargetSpecificOptions {
    */
   readonly useAppIdAsId?: boolean
 
+  /**
+   *  The custom squirrel.windows.exe path. Defaults to `https://github.com/electron-userland/electron-builder-binaries/tree/master/Squirrel.Windows`.
+   */
+  readonly vendorDirectory?: string
+
```
mmaietta, line 47: Please update doc comment
```
/**
*  The custom file path to `squirrel.windows.exe` for usage with [electron-winstaller](https://github.com/electron/windows-installer) `vendorDirectory`.
*/
```

>Defaults to `https://github.com/electron-userland/electron-builder-binaries/tree/master/Squirrel.Windows`

Also, is this true with the current setup? I don't see any default value being set for `vendorDirectory`. As this is currently coded, it looks like it'll default to the vendor files that come with `electron/windows-installer`

beyondkmp: updated.

/**
   * The custom squirrel.windows.exe path. If not specified will use the Squirrel.Windows that is shipped with electron-installer(https://github.com/electron/windows-installer/tree/main/vendor).
   * After https://github.com/electron-userland/electron-builder-binaries/pull/56 merged, will add `electron-builder-binaries` to get the latest version of squirrel.
   */
 
 The squirrel windows in this  https://github.com/electron-userland/electron-builder-binaries/pull/56 is the same as the one in https://github.com/electron/windows-installer/tree/main/vendor, so when this value is empty, it defaults to using the squirrel windows from windows-installer.

test/src/helpers/packTester.ts
```diff
@@ -1,17 +1,18 @@
...
 import { computeArchToTargetNamesMap } from "app-builder-lib/out/targets/targetFactory"
 import { getLinuxToolsPath } from "app-builder-lib/out/targets/tools"
-import { convertVersion } from "electron-builder-squirrel-windows/out/squirrelPack"
+import { convertVersion } from "electron-winstaller"
 import { PublishPolicy } from "electron-publish"
 import { emptyDir, writeJson } from "fs-extra"
 import * as fs from "fs/promises"
 import { load } from "js-yaml"
 import * as path from "path"
+import AdmZip from "adm-zip"
```
mmaietta, line 14: Out of curiosity, why'd we switch from `decompress-zip`?

beyondkmp: Newer versions of squirrel windows use the cp437 compression encoding method, which is not supported by DecompressZip, and will be messed up after decompression, so it is replaced with AdmZip.

beyondkmp:
@mmaietta  Resolved the code conflict now, but this UT keeps failing, running locally is no problem. It looks like there is a problem with this UT on the master, help me out if you have time.
image

mmaietta:
I'll take a look at this today. There is no problem on `master` though...as the commit test history is clean: https://github.com/electron-userland/electron-builder/commits/master/

mmaietta:
I pushed your branch to `origin/win-installer` within the electron-builder repo and the tests all pass correctly: https://github.com/electron-userland/electron-builder/actions/runs/13076539224

My best guess is that running the job from your fork is not able to access that repo secret specifically?

beyondkmp:
@mmaietta Yes, that's correct. The reason is the same as what you mentioned. The conflicts in this MR have been resolved. When you have time, please help review it again. If there are no issues, it can be merged.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -98,24 +109,35 @@ export default class SquirrelWindowsTarget extends Target {
...
+      vendorDirectory,
+      nuspecTemplate: path.join(__dirname, "..", "template.nuspectemplate"),
       ...(this.options as any),
```
mmaietta, line 128: So this is obfuscating a ton of misconfigured properties (just remove the `as any` for compiler to throw errors)

- `extraMetadataSpecs` doesn't exist in `electron-winstaller`, nor can be added to the current nuspec template, so we'll need to inject it manually, if and only if `projectUrl != null`.
Something like:
```
    const nuspecTemplate = await this.packager.info.tempDirManager.getTempFile({ prefix: "template", suffix: ".nuspectemplate" })
    let templateContent = await readFile(path.resolve(__dirname, "..", "template.nuspectemplate"), "utf8")
    if (projectUrl != null) {
      const searchString = "<copyright><%- copyright %></copyright>"
      templateContent = templateContent.replace(searchString, `${searchString}\n    <projectUrl>${projectUrl}</projectUrl>`)
    }
    await writeFile(nuspecTemplate, templateContent)
```
- `remoteReleases` is only a stringin `electron-winstaller`, so we need to determine how to handle `true` for auto-selecting github URL (as the docs state), otherwise this breaks current functionality
- `msi` doesn't exist in `electron-winstaller`, but `noMsi` does, we'll need to negate the current value.
- `packageCompressionLevel` doesn't exist in `electron-winstaller`

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -98,24 +109,35 @@ export default class SquirrelWindowsTarget extends Target {
     const appInfo = packager.appInfo
     const projectUrl = await appInfo.computePackageUrl()
     const appName = this.appName
+    // If not specified will use the Squirrel.Windows that is shipped with electron-installer(https://github.com/electron/windows-installer/tree/main/vendor)
+    // After https://github.com/electron-userland/electron-builder-binaries/pull/56 merged, will add `electron-builder-binaries` to get the latest version of squirrel.
+    const vendorDirectory = this.options.customSquirrelExePath
```
mmaietta, line 114: We need to check this. Also `customSquirrelExePath` is different than a `vendorDirectory`, so we should rename the variable as well:
```
    let vendorDirectory = this.options.customSquirrelVendorDir
    if (isEmptyOrSpaces(vendorDirectory) || !(await exists(vendorDirectory))) {
      log.warn({ vendorDirectory }, "unable to access Squirrel.Windows vendor directory, falling back to default electron-winstaller")
      vendorDirectory = undefined
    }
```

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -98,24 +109,35 @@ export default class SquirrelWindowsTarget extends Target {
...
+      options.windowsSign = {
+        hookFunction: async (file: string) => {
+          await packager.sign(file)
+        },
+      }
+    }
+
     if (isEmptyOrSpaces(options.description)) {
-      options.description = options.productName
+      options.description = this.options.name || appInfo.productName
```
mmaietta, line 140: Why this change? This may have unintended side effects

packages/electron-builder-squirrel-windows/src/squirrelPack.ts
```diff
@@ -1,295 +0,0 @@
...
-  }
-}
-
-function syncReleases(outputDirectory: string, options: SquirrelOptions) {
-  log.info("syncing releases to build delta package")
-  const args = prepareArgs(["-u", options.remoteReleases!, "-r", outputDirectory], path.join(options.vendorPath, "SyncReleases.exe"))
-  if (options.remoteToken) {
-    args.push("-t", options.remoteToken)
-  }
-  return spawn(process.platform === "win32" ? path.join(options.vendorPath, "SyncReleases.exe") : "mono", args)
```
mmaietta, line 26: What was `SyncReleases.exe` doing? This seems important

packages/electron-builder-squirrel-windows/src/squirrelPack.ts
```diff
@@ -1,295 +0,0 @@
...
-  archive.append(nuspecContent.replace(/\n/, "\r\n"), { name: `${options.name}.nuspec` })
-
-  //noinspection SpellCheckingInspection
-  archive.append(
-    `<?xml version="1.0" encoding="utf-8"?>
-<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
-  <Relationship Type="http://schemas.microsoft.com/packaging/2010/07/manifest" Target="/${options.name}.nuspec" Id="Re0" />
-  <Relationship Type="http://schemas.openxmlformats.org/package/2006/relationships/metadata/core-properties" Target="/package/services/metadata/core-properties/1.psmdcp" Id="Re1" />
-</Relationships>`.replace(/\n/, "\r\n"),
-    { name: ".rels", prefix: "_rels" }
```
mmaietta, line 189: Where did this go? I don't see it in our template or in `electron-winstaller`
https://github.com/electron/windows-installer/blob/e722a519f8625c939087de2602b2b5b594886bc4/vendor/template.wxs

packages/electron-builder-squirrel-windows/src/squirrelPack.ts
```diff
@@ -1,295 +0,0 @@
...
-  <Default Extension="exe" ContentType="application/octet" />
-  <Default Extension="dat" ContentType="application/octet" />
-  <Default Extension="psmdcp" ContentType="application/vnd.openxmlformats-package.core-properties+xml" />
-  <Default Extension="diff" ContentType="application/octet" />
-  <Default Extension="bsdiff" ContentType="application/octet" />
-  <Default Extension="shasum" ContentType="text/plain" />
-  <Default Extension="mp3" ContentType="audio/mpeg" />
-  <Default Extension="node" ContentType="application/octet" />
-</Types>`.replace(/\n/, "\r\n"),
-    { name: "[Content_Types].xml" }
```
mmaietta, line 211: Where did this go? I don't see it in our template or in `electron-winstaller`
https://github.com/electron/windows-installer/blob/e722a519f8625c939087de2602b2b5b594886bc4/vendor/template.wxs
https://github.com/electron/windows-installer/blob/e722a519f8625c939087de2602b2b5b594886bc4/src/index.ts

mmaietta (changes requested):

packages/electron-builder-squirrel-windows/src/squirrelPack.ts
beyondkmp: The windows-installer calls the nuget command to generate the nupkg file, and the nuget command will automatically generate the [Content_Types].xml. The specific code is sourced from  
https://github.com/NuGet/NuGet.Client/blob/fa6bd3682bb698604a09d63601a708f2b357b7a8/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs#L1301

packages/electron-builder-squirrel-windows/src/squirrelPack.ts
beyondkmp: The windows-installer calls the nuget command to generate the nupkg file, and the nuget command will automatically generate the .rels. The specific code is sourced from

https://github.com/NuGet/NuGet.Client/blob/fa6bd3682bb698604a09d63601a708f2b357b7a8/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs#L1274

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: The original code was like this:
```
options.description = options.productName;
options.productName = this.options.name || appInfo.productName;
```

After this MR, there is no productName in the SquirrelOptions, so it should now be directly changed to:

```
options.description = this.options.name || appInfo.productName;
```

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: Good catch. Fixed.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: Fixed.

test/snapshots/filesTest.js.snap
```diff
@@ -47,23 +49,33 @@ exports[`extraResources on Windows 2`] = `
   "lib/net45/LICENSES.chromium.html",
   "lib/net45/resources.pak",
   "lib/net45/snapshot_blob.bin",
+  "lib/net45/squirrel.exe",
   "lib/net45/Test App ßW.exe",
   "lib/net45/Test App ßW_ExecutionStub.exe",
-  "lib/net45/Update.exe",
```
mmaietta, line 52: With this change, we need to update our documentation. This actually breaks previously functionality in changing this 😢 
https://github.com/electron-userland/electron-builder/blob/198551af4453adada8be6e6e7e50881145f10502/pages/squirrel-windows.md?plain=1#L8

The example it references is:
https://github.com/mongodb-js/electron-squirrel-startup

Which has the update code: `Update.exe`
https://github.com/mongodb-js/electron-squirrel-startup/blob/fcb9674bb9fce2c0241e75b2f57a9b70fa2b6c52/index.js#L7

I think the final steps required for this change are to update the docs and copy the startup code, so that new devs know how to set up the Squirrel auto-update.
(I can do this portion if you'd like)

What are your thoughts on the `major` semver bump though? As this technically requires code changes on the developer's side as well. It also means that anyone using `electron-squirrel-startup` will no longer be able to, so we probably need to detect that during some config validation and actually log+throw an error. That way an unsuspecting dev won't install this updated squirrel package (without updating their electron-updater version) and know that their auto-update won't break suddenly (assuming they don't thoroughly test, and it does still happen sadly)

mmaietta (changes requested):

test/snapshots/filesTest.js.snap
beyondkmp: No changes are needed. This is because when installing, squirrel window will copy `squirrel.exe` to `update.exe`. The specific code can be found at https://github.com/Squirrel/Squirrel.Windows/blob/51f5e2cb01add79280a53d51e8d0cfa20f8c9f9f/src/Update/Program.cs#L179. 

Moreover, in the official Windows Installer documentation, `update.exe` is also used as an example.
https://github.com/electron/windows-installer
image

mmaietta: Nice work, love it.

mmaietta (dismissed):
Great work!

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -96,39 +108,55 @@ export default class SquirrelWindowsTarget extends Target {
     checkConflictingOptions(this.options)
 
     const appInfo = packager.appInfo
-    const projectUrl = await appInfo.computePackageUrl()
-    const appName = this.appName
+    // If not specified will use the Squirrel.Windows that is shipped with electron-installer(https://github.com/electron/windows-installer/tree/main/vendor)
+    // After https://github.com/electron-userland/electron-builder-binaries/pull/56 merged, will add `electron-builder-binaries` to get the latest version of squirrel.
+    let vendorDirectory = this.options.customSquirrelExePath
```
mmaietta, line 113: `customSquirrelExePath` this needs to be renamed as it doesn't correspond to the tsdoc definition. (`The custom squirrel.windows.exe path.`)

Please update to something like `customSquirrelVendorDir` or `Path` or something like that.

mmaietta (approved):

PR #8368, opened 2024-07-22
fix(linux): Don't setuid chrome-sandbox when not required

### LINKED ISSUE #7545: Do we really need a setuid root chrome-sandbox for every platform?
* **Electron-Builder Version**:  23.6.0

* **Node Version**:  16

* **Electron Version**: 21.3.1

* **Target**: x86_64 linux

I opened a ticket for the Bitwarden client, which uses electron-builder, and they asked me to raise this issue here.

The Bitwarden client package package uses the electron-builder tool to build its packages. The resulting package has a post script that sets a setuid root chrome-sandbox. I'm seeing this in the latest RHEL/Fedora and Debian packages.  On the RPM packages being created, the chrome-sandbox executable is modified as part of the post-install process, which results in the package failing RPM validation because the mode on the file is different than what the package manifest claims. Any tool that relies on the RPM database validation would fail on that package. 

Also, I really would prefer to keep the number of setuid root executables on my systems to a minimum.

Looking through the Issues raised here, it looks like the setuid root executable is only needed  for sandboxing on platforms that do not support namespaces.  Perhaps the 'chmod 4755' command could have some logic to detect whether the kernel supports it before setting the mode? As far as I know, user namespaces are enabled by default in 5.10 and later Linux kernels. RHEL8 and later have it enabled by default, as does every supported Fedora.  I expect many other distros do as well.

pimterry:
Fix #7545 (already closed but as stale, not actually fixed)

Setuid for chrome-sandbox is not needed on the vast majority of systems where user namespaces are available, as this more secure mechanism is used for sandboxing instead. Root setuid binaries like this are a general security risk, and appear to cause specific problems in various cases today (e.g. making a deb package unusable on Ubuntu 24.04: https://github.com/httptoolkit/httptoolkit/issues/602, making rpm packages fail validation on Fedora: https://github.com/bitwarden/clients/issues/5153)

User namespaces were implemented in 2013 (kernel 3.8), and have been enabled by default widely for many years. Chrome themselves stopped setting this in 2016, describing it as unnecessary on all supported Linux platforms: https://issues.chromium.org/issues/40462640.

The only other case I'm aware of where setuid is required despite user namespace support is Electron <5. This electron version has been end-of-life since 2019 (more than 4 years ago) so hopefully that's not a concern, but if Electron Builder supports very old electron versions then this may be a breaking change.

This PR checks whether they're supported (i.e. `/proc/self/ns/user` is a symlink) and briefly tests that they work correctly (running `true` in a separate user namespace), and then sets the setuid bit _only_ if that fails. This is the same test approach as used to detect sandboxing support in Nix: https://github.com/NixOS/nix/blob/40f80e1b5cf2bebb6d1d8c9efac1d982a540d555/tests/functional/common/vars-and-functions.sh#L180-L182

### DIFF
--- a/.changeset/purple-terms-sing.md
+++ b/.changeset/purple-terms-sing.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: don't setuid chrome-sandbox when not required
--- a/packages/app-builder-lib/templates/linux/after-install.tpl
+++ b/packages/app-builder-lib/templates/linux/after-install.tpl
@@ -10,8 +10,13 @@ else
     ln -sf '/opt/${sanitizedProductName}/${executable}' '/usr/bin/${executable}'
 fi
 
-# SUID chrome-sandbox for Electron 5+
-chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
+# Check if user namespaces are supported by the kernel and working with a quick test:
+if ! { [[ -L /proc/self/ns/user ]] && unshare --user true; }; then
+    # Use SUID chrome-sandbox only on systems without user namespaces:
+    chmod 4755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
+else
+    chmod 0755 '/opt/${sanitizedProductName}/chrome-sandbox' || true
+fi
 
 if hash update-mime-database 2>/dev/null; then
     update-mime-database /usr/share/mime || true
--- a/test/snapshots/linux/debTest.js.snap
+++ b/test/snapshots/linux/debTest.js.snap
@@ -530,8 +530,12 @@ else
     ln -sf '/opt/foo/Boo' '/usr/bin/Boo'
 fi
 
-# SUID chrome-sandbox for Electron 5+
-chmod 4755 '/opt/foo/chrome-sandbox' || true
+if ! { [[ -L /proc/self/ns/user ]] && unshare --user true; }; then
+    # Use SUID chrome-sandbox only on systems without user namespaces:
+    chmod 4755 '/opt/foo/chrome-sandbox' || true
+else
+    chmod 0755 '/opt/foo/chrome-sandbox' || true
+fi
 
 if hash update-mime-database 2>/dev/null; then
     update-mime-database /usr/share/mime || true
--- a/test/snapshots/linux/debTest.js.snap
+++ b/test/snapshots/linux/debTest.js.snap
@@ -530,6 +530,7 @@ else
     ln -sf '/opt/foo/Boo' '/usr/bin/Boo'
 fi
 
+# Check if user namespaces are supported by the kernel and working with a quick test:
 if ! { [[ -L /proc/self/ns/user ]] && unshare --user true; }; then
     # Use SUID chrome-sandbox only on systems without user namespaces:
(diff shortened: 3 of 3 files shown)


pimterry:
I've patched the various commit/changeset warnings here, but also I've just discovered the wider underlying cause of these recent issues: https://github.com/electron/electron/issues/41066. All Electron apps are effectively broken on Ubuntu 24.04 right now due to related user namespacing restrictions, and it may be that while actually skipping this setuid is useful, since that would imply the test here will fail in those environments, this doesn't actually solve my problem... I'll keep digging there.

This isn't a blocker for this though, so this may still be useful to explore and/or merge here regardless, since setuid really shouldn't be required here anyway! I strongly suspect Ubuntu will revert this change from 24.04 (Mint and others have already done so) so it'll work there too soon.

xmedeko:
@pimterry The problem with apparmor in Ubuntu 24 is twofold: one with SETUID and one with execvp and space in path, see my comment https://github.com/electron-userland/electron-builder/issues/8440#issuecomment-2434563606 and my comment in [Comment 157 for bug Ubuntu aparmor 2046844](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2046844/comments/157)

So, setting SUID always as before is workaround for apps without space in installation path.

pimterry:
@xmedeko Hmm, the test here should cover that though - on my machine, if user namespaces are disabled (`sysctl -w kernel.apparmor_restrict_unprivileged_userns=1`) then the check here detects that (specifically, `unshare --user true` fails) and so this correctly sets SUID just as it used to.

This should only disable SUID in environments where user namespaces are genuinely available. Is that not what you're seeing?

xmedeko:
@pimterry Do you mean set to `0` to disable it: `sysctl -w kernel.apparmor_restrict_unprivileged_userns=0` ? Note: this change is only until next reboot, while file SUID is permanent. (To switch off `kernel.apparmor_restrict_unprivileged_userns` permanently, see the mentioned Electron issue.

On my Ubuntu 24 fresh install and update:
```
$ if unshare --user true; then echo yes; else no; fi
yes
```
And this check returns `yes` either I switch `sysctl -w kernel.apparmor_restrict_unprivileged_userns=0` or to `1`. So, the deb installation never sets SUID on Ubuntu 24.

xmedeko:
@pimterry Ubuntu 24 after reboot:
```
# Get current value
> sudo sysctl kernel.apparmor_restrict_unprivileged_userns
kernel.apparmor_restrict_unprivileged_userns = 1
> unshare --user true
> echo $?
0
```
So, even if I set `kernel.apparmor_restrict_unprivileged_userns` to whatever, I got `0` return value from the `unshare --user true` test. (I run the fresh Ubuntu install in VirtualBox, but I think it does not affect the result.)

By this advice https://github.com/electron-userland/electron-builder/issues/8440#issuecomment-2435927188 electron-builder has `afterInstall` param to customize `postinst` script. I'll give it a try if Ubuntu guys won't fix that bug soon.

xmedeko:
@pimterry FYI proper solution by Ubuntu guys is to add AppArmor profile, I have filled a feature request for that #8635. Until it's solved, it's possible to do it by `afterInstall` script, e.g. see https://github.com/gravitational/teleport/pull/43595

PR #8371, opened 2024-07-23
fix: delete the symlink file when the target is empty

beyondkmp:
Add the following filtering rule in the `files` section to exclude all files in the `headers` directory.
```
 "files": [
      "package.json",
      "!**/*.{h,cpp}"
    ]
```
image

 This will result in an empty directory, which will not generate a corresponding empty directory in `app.asar.unpack`. However, the symbolic links for `headers` will still be present. This will cause an error during signing, stating that the corresponding files cannot be found.

### DIFF
--- a/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
+++ b/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
@@ -161,7 +161,7 @@ export class NodeModuleCopyHelper extends FileCopyHelper {
         result[index] = undefined
       }
     }
-    const finalResult = result.filter(it => it !== undefined)
+    const finalResult: Array<string> = result.filter(it => it !== undefined)
     return finalResult
   }
 }
--- a/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
+++ b/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
@@ -137,11 +137,16 @@ export class NodeModuleCopyHelper extends FileCopyHelper {
       for (const child of sortedFilePaths) {
         if (child != null) {
           result.push(child)
-          this.metadata.get(child)?.isSymbolicLink() && symlinkFiles.set(child, result.length - 1)
+          if (this.metadata.get(child)?.isSymbolicLink()) {
+            symlinkFiles.set(child, result.length - 1)
+          }
           isEmpty = false
         }
       }
-      isEmpty && emptyDirs.add(dirPath)
+
+      if (isEmpty) {
+        emptyDirs.add(dirPath)
+      }
 
       dirs.sort()
       for (const child of dirs) {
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
```diff
@@ -130,17 +133,29 @@ export class NodeModuleCopyHelper extends FileCopyHelper {
...
 
+      let isEmpty = true
       for (const child of sortedFilePaths) {
         if (child != null) {
           result.push(child)
+          this.metadata.get(child)?.isSymbolicLink() && symlinkFiles.set(child, result.length - 1)
+          isEmpty = false
         }
       }
+      isEmpty && emptyDirs.add(dirPath)
```
mmaietta, line 144: To maintain code consistency, let's continue using bracketed if-statements.
```
if (isEmpty) {
   emptyDirs.add(dirPath)
}
```

packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
```diff
@@ -130,17 +133,29 @@ export class NodeModuleCopyHelper extends FileCopyHelper {
         CONCURRENCY
       )
 
+      let isEmpty = true
       for (const child of sortedFilePaths) {
         if (child != null) {
           result.push(child)
+          this.metadata.get(child)?.isSymbolicLink() && symlinkFiles.set(child, result.length - 1)
```
mmaietta, line 140: To maintain code consistency, let's continue using bracketed if-statements.
```
if (this.metadata.get(child)?.isSymbolicLink()) {
   symlinkFiles.set(child, result.length - 1)
}
```

packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
```diff
@@ -130,17 +133,29 @@ export class NodeModuleCopyHelper extends FileCopyHelper {
...
         queue.push(dirPath + path.sep + child)
       }
     }
-    return result
+
+    for (const [file, index] of symlinkFiles) {
+      const resolvedPath = realpathSync(file)
+      if (emptyDirs.has(resolvedPath)) {
+        // delete symlink file if target is a empty dir
+        result[index] = ""
```
mmaietta, line 156: Minor nit: Can we use `undefined` here to avoid using an empty string as a false-y value?

packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
beyondkmp: Fixed

packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
beyondkmp: Fixed.

packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
beyondkmp: There can't be any empty strings in the result, as these are paths, and the paths passed in are not empty strings.

However, from a coding perspective, using `undefined` would be better.

Fixed it.

Issue #7063, opened 2022-08-14
Auto-update fails with ENOENT: no such file or directory, rename


pietrovismara:
* **Electron-Builder Version**:  23.3.2

* **Node Version**:  16.14.2

* **Electron Version**: Electron 19.0.8
* **Electron Type (current, beta, nightly)**: current

* **Target**: Windows NSIS, MacOS DMG Zip

This is the same exact issue as https://github.com/electron-userland/electron-builder/issues/3622, which was ignored and then closed. 

This is the error I get on macOS:

```
Error: Error: ENOENT: no such file or directory, rename '/Users/user/Library/Application Support/Caches/@mydesktopapp-updater/pending/temp-MyDesktopApp-0.0.46-mac.zip' -> '/Users/user/Library/Application Support/Caches/@mydesktopapp-updater/pending/MyDesktopApp-0.0.46-mac.zip'
```

And on Windows:

```
ENOENT: no such file or directory, rename 'C:\Users\39380\AppData\Local\@mydesktopapp-updater\pending\temp-MyDesktopApp Setup 0.0.46.exe' -> 'C:\Users\39380\AppData\Local\@mydesktopapp-updater\pending\MyDesktopApp Setup 0.0.46.exe'
```

The error is also non deterministic: retrying to update several times usually works after the 3rd tentative.

I'm using the autoUpdater as follows:

```
autoUpdater.autoDownload = false;
autoUpdater.autoInstallOnAppQuit = false;

const updateApp = async () => {
    const quit = new Promise<void>((resolve, reject) => {
      const quitAndInstall = () => {
        try {
          autoUpdater.quitAndInstall();
          resolve();
        } catch (e: any) {
          reject(e);
        }
      };
      return setTimeout(quitAndInstall, isMac ? 10000 : 5000);
    });
    try {
      await quit;
    } catch (e: any) {
      console.log(e);
      Sentry.captureException(e);
    }
};

export const checkAppUpdates = async () => {
    return new Promise<string | null>((resolve, reject) => {
      autoUpdater.once('error', (error) => {
        reject(error);
      });

      autoUpdater.once('update-available', async () => {
        console.log('update-available');
        autoUpdater.once('update-downloaded', (info) => {
          console.log('update-downloaded');
          resolve(info.version);
        });
        autoUpdater.downloadUpdate();
      });
      autoUpdater.once('update-not-available', () => {
        console.log('update-not-available');
        resolve(null);
      });
[...]

PR #8378, opened 2024-07-25
fix: Add global download promise to limit concurrent downloads to 1

s77rt:
Calling `autoUpdater.checkForUpdates()` twice initiates two download processes that conflict and results in update failure. This PR adds a global download promise limiting the download processes to 1.

Fixes: https://github.com/electron-userland/electron-builder/issues/7063

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -521,24 +521,21 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
       return e
     }
 
-    try {
-      this.downloadPromise = this.doDownloadUpdate({
-        updateInfoAndProvider,
-        requestHeaders: this.computeRequestHeaders(updateInfoAndProvider.provider),
-        cancellationToken,
-        disableWebInstaller: this.disableWebInstaller,
-        disableDifferentialDownload: this.disableDifferentialDownload,
+    this.downloadPromise = this.doDownloadUpdate({
+      updateInfoAndProvider,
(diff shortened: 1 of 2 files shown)


packages/electron-updater/src/AppUpdater.ts
```diff
@@ -514,15 +522,20 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
...
       })
+        .catch((e: any) => {
+          throw errorHandler(e)
+        })
+        .finally(() => {
+          this.downloadPromise = null
+        })
+      return this.downloadPromise
     } catch (e: any) {
       return Promise.reject(errorHandler(e))
```
mmaietta, line 540: I think we need to add `this.downloadPromise = null` before this `return` as well as a failsafe.

beyondkmp: It's already been added in the finally block, so it doesn't need to be added here.

mmaietta: Oh interesting. The `finally` block still executes even when we throw inside of the `catch`?

s77rt: Yes. Though I just found out that the error we throw on catch is not caught afterwards. The thrown error in a promise can only be caught in the promise chain i.e. the try-catch block won't catch it.

```js
try {
    Promise.reject("err")
        .catch((e) => {
            console.log("caught in promise");
            throw e;
        })
        .finally(() => console.log("finally promise"))
} catch (error) {
    console.log("caught in try-catch")
}
```

Regardless the try-catch block is not needed.

Issue #7729, opened 2023-08-15
Issue with code signing with an EV certificate on Windows


stephanedupont:
I was using a code signing certificate until now, using `CSC_LINK` and `CSC_KEY_PASSWORD` as specified [here](https://www.electron.build/code-signing.html) and it was working fine.

I was not able to renew my certificate as they stopped issuing certificates without a hardware token due to new regulations (see [here](https://signmycode.com/blog/code-signing-certificates-price-hike-up-to-3x-to-4x)), so I ordered instead an EV code signing certificate that comes with a USB hardware token.

I installed the SafeNet driver, removed `CSC_LINK` and `CSC_KEY_PASSWORD` and added `win.certificateSubjectName` as mentioned in the documentation.

When executing the `electron-builder build --win` command, it now fails with the following output:

```
$ electron-builder build -c.extraMetadata.main=electron.js --win
  • electron-builder  version=24.6.3 os=10.0.22621
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist\1.1.9\builder-effective-config.yaml
  • packaging       platform=win32 arch=x64 electron=22.3.12 appOutDir=dist\1.1.9\win\x64\win-unpacked
  ⨯ Exit code: 1. Command failed: powershell.exe -NoProfile -NonInteractive -Command Get-ChildItem -Recurse Cert: -CodeSigningCert | Select-Object -Property Subject,PSParentPath,Thumbprint | ConvertTo-Json -Compress
Get-ChildItem : A parameter cannot be found that matches parameter name 'CodeSigningCert'.
At line:1 char:30
+ Get-ChildItem -Recurse Cert: -CodeSigningCert | Select-Object -Proper ...
+                              ~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Get-ChildItem], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand


Get-ChildItem : A parameter cannot be found that matches parameter name 'CodeSigningCert'.
At line:1 char:30
+ Get-ChildItem -Recurse Cert: -CodeSigningCert | Select-Object -Proper ...
+                              ~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Get-ChildItem], ParameterBindingException
[...]

PR #8384, opened 2024-07-29
fix(windows,code-sign): cannot sign binary files in Github Actions

BlackHole1:
Fixed: #7729 #8055

Signed-off-by: Kevin Cui <bh@bugs.cc>

### DIFF
--- a/packages/app-builder-lib/src/codeSign/windowsCodeSign.ts
+++ b/packages/app-builder-lib/src/codeSign/windowsCodeSign.ts
@@ -110,9 +110,9 @@ export interface CertificateFromStoreInfo {
 export async function getCertificateFromStoreInfo(options: WindowsConfiguration, vm: VmManager): Promise<CertificateFromStoreInfo> {
   const certificateSubjectName = options.certificateSubjectName
   const certificateSha1 = options.certificateSha1 ? options.certificateSha1.toUpperCase() : options.certificateSha1
-  // ExcludeProperty doesn't work, so, we cannot exclude RawData, it is ok
-  // powershell can return object if the only item
-  const rawResult = await vm.exec("powershell.exe", [
+
+  const ps = await getPSCmd(vm)
+  const rawResult = await vm.exec(ps, [
     "-NoProfile",
     "-NonInteractive",
     "-Command",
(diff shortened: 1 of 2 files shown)


NoahAndrews:
Note that unlike powershell.exe, pwsh.exe has to be manually installed.

BlackHole1:
@NoahAndrews Thank you for your addition, but I think it is not unacceptable to require developers to install pwsh.exe, just as it is necessary to use Parallels Desktop when packaging on a Mac :)

beyondkmp:
@BlackHole1  `verifyUpdateCodeSignature` is exported and can be customized. You can definitely define your own authentication function that uses pwsh.exe.

https://www.electron.build/configuration/win.html#how-do-use-a-custom-verify-function-to-enable-nsis-signature-verification-alternatives-instead-of-powershell

beyondkmp:
Making this configurable by default to use PowerShell, and allowing users to set their own configurations, might be a good solution.

BlackHole1:
@NoahAndrews @beyondkmp I spent several hours investigating this matter, and the conclusion is: this is a BUG of GitHub Actions.

Although GitHub officially switched the default Shell for Windows to PowerShell on 10/23/2019, this is still different from explicitly using `shell: PowerShell`.

After investigation, I found that the difference between the two lies in the value of `PSModulePath` in the environment variables, as follows:

```
# Use Default Shell
PSModulePath Env: "...;c:\\program files\\powershell\\7\\Modules;C:\\Windows\\system32\\WindowsPowerShell\\v1.0\\Modules"

# Use shell: PowerShell
PSModulePath Env: "...;C:\\Windows\\system32\\WindowsPowerShell\\v1.0\\Modules"
```

The fundamental cause of this problem lies in the fact that PowerShell prioritized the module `c:\\program files\\powershell\\7\\Modules\\Microsoft.PowerShell.Security` over `C:\\Windows\\system32\\WindowsPowerShell\\v1.0\\Modules\\Microsoft.PowerShell.Security`.
When I delete/rename `c:\\program files\\powershell\\7\\Modules\\Microsoft.PowerShell.Security`, the command will execute successfully.

So I believe this issue only occurs when: **using the windows-runner in Github Actions, and the default shell is used.**

In the latest fix, I solved the issue by prioritizing the use of pwsh.exe (falling back to powershell.exe if it doesn't exist).

PTAL @mmaietta

packages/app-builder-lib/src/codeSign/windowsCodeSign.ts
```diff
@@ -319,3 +319,10 @@ async function getToolPath(isWin = process.platform === "win32"): Promise<ToolIn
     return { path: path.join(vendorPath, process.platform, "osslsigncode") }
   }
 }
+
+async function getPSCmd(vm: VmManager): Promise<string> {
+  return await vm
+    .exec("powershell.exe", ["-NoProfile", "-NonInteractive", "-Command", `Get-Command pwsh.exe`])
+    .then(() => "pwsh.exe")
+    .catch(() => "powershell.exe")
+}
```
mmaietta, line 328: I think it may be worthwhile to add some debug logging here to help any debugging in the future. What are your thoughts?
```
vm
.exec("powershell.exe", ["-NoProfile", "-NonInteractive", "-Command", `Get-Command pwsh.exe`])
.then(() => {
    log.debug(null, "identified pwsh.exe for executing code signing")
    return "pwsh.exe"
})
.catch(() => {
    log.debug(null, "unable to find pwsh.exe, falling back to powershell.exe")
    return "powershell.exe"
})
```

BlackHole1: Good idea! Done.

PR #8392, opened 2024-08-02
feat: unpacked native node file

beyondkmp:
All npm C++ native modules generate a file ending in .node, which is an executable file. Currently, it is default practice to place these inside app.asar, which seems problematic for several reasons:

1. .node files are executables that need to be signed. If packed directly into app.asar, they cannot be signed.
2. According to official documentation, placing .node files inside app.asar can lead to performance issues and file not found errors.
 https://www.electronjs.org/docs/latest/tutorial/asar-archives
image

4. Looking at applications like VSCode and Slack, they currently default to placing .node files in app.asar.unpack.

Therefore, I suggest that during the smart unpack process, we automatically detect files ending in .node and place them in the unpack directory.

### DIFF
--- a/test/snapshots/BuildTest.js.snap
+++ b/test/snapshots/BuildTest.js.snap
@@ -907,16 +907,18 @@ Object {
           "files": Object {
             "LICENSE.md": Object {
               "size": "<size>",
+              "unpacked": true,
             },
             "build": Object {
               "files": Object {
                 "Release": Object {
                   "files": Object {
                     "keytar.node": Object {
-                      "executable": true,
                       "size": "<size>",
+                      "unpacked": true,
                     },
                   },
+                  "unpacked": true,
                 },
               },
             },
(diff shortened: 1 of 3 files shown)


test/snapshots/BuildTest.js.snap
```diff
@@ -4853,6 +4859,54 @@ Object {
...
+    ]
+  },
+  \\"config\\": {
+    \\"runtime\\": \\"napi\\",
+    \\"target\\": 3
+  }
+}",
+    "name": "app.asar.unpacked/node_modules/keytar/package.json",
+  },
+  "app.asar.unpacked/node_modules/keytar/lib/keytar.js",
```
mmaietta, line 4908: Why are the `LICENSE.md`, `package.json`, and `keytar.js` showing up in the unpacked folder?

beyondkmp: In smartunpack, if an executable file is detected, the entire folder containing it will automatically be placed into app.asar.unpack, not just the file itself.

https://github.com/electron-userland/electron-builder/blob/84f290990915d7ffa0d2dfe0a1d739aeb2b21560/packages/app-builder-lib/src/asar/unpackDetector.ts#L29

mmaietta (approved):

Issue #8388, opened 2024-07-30
GitHubProvider: Doesn't use custom channel that was set in the electron-builder config


jermy-c:
* **Electron-Builder Version**: 24.13.3

* **Node Version**: v20.15.0

* **Electron Version**: 25.9.8
* **Electron Type (current, beta, nightly)**: current

* **Target**: Mac OS X arm64

**Reproduction**

1. Set `electron-builder.yml` with the following settings:
```yml
publish:
  - provider: github
    protocol: https
    releaseType: release
    channel: customChannel
```

**Actual Result**

1. The build generates a `customChannel-mac.yml`
2. The executable from the build still tries to fetch `latest-mac.yml` and gives an error

**Expected Result**

1. The build generates a `customChannel-mac.yml`
2. The executable from the build still tries to fetch `customChannel-mac.yml`
3. Use the `customChannel` name as the latest channel

**Question**

1. Is there a way to have the executable fetch `customChannel-mac.yml` as the update metadata?

mmaietta:
I haven't utilized custom channels before, but I believe the next step would be to set the channel of the autoupdater before checking for an update.
https://github.com/electron-userland/electron-builder/blob/84f290990915d7ffa0d2dfe0a1d739aeb2b21560/packages/electron-updater/src/AppUpdater.ts#L132-L149

jermy-c:
Ok, thanks a lot for the help and the new feature.

benbucksch:
@beyondkmp , @mmaietta Thanks a lot!

PR #8393, opened 2024-08-03
feat: add custom channel in github provider

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8388

### DIFF
--- a/packages/electron-updater/src/providers/GitHubProvider.ts
+++ b/packages/electron-updater/src/providers/GitHubProvider.ts
@@ -48,6 +48,11 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
     super(options, "github.com", runtimeOptions)
   }
 
+  private get channel(): string {
+    const result = this.updater.channel || this.options.channel
+    return result == null ? this.getDefaultChannelName() : this.getCustomChannelName(result)
+  }
+
   async getLatestVersion(): Promise<GithubUpdateInfo> {
     const cancellationToken = new CancellationToken()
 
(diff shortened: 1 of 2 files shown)


mmaietta (approved):

packages/electron-updater/src/providers/GitHubProvider.ts
```diff
@@ -136,7 +141,7 @@ export class GitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
     }
 
     try {
-      const channel = this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0] || "latest")) : this.getDefaultChannelName()
+      const channel = this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0] || "latest")) : this.channel
```
benbucksch, line 144:
```suggestion
      const channel = this.updater.allowPrerelease ? this.getCustomChannelName(String(semver.prerelease(tag)?.[0] || this.channel)) : this.channel
```

benbucksch: Even if we deploy 0.4.0alpha1, we still want it to use the channel name that we set in the config file. The `latest` will not be deployed (or if it's deployed, it belongs to a different build mode).

In our use case, we use the channels for different brand builds of the same application, like "Yahoo" and "AOL" builds (just as an example). Even if we deploy a "Yahoo News 0.4.1beta1" build, we still want it to update from the "yahoo" channel, not the "aol" channel, nor the "latest" channel.

beyondkmp: Fixed

benbucksch: @beyondkmp Thank you! :smile:

benbucksch (approved):

packages/electron-updater/src/providers/GitHubProvider.ts
benbucksch: (I don't see the button for resolving this conversation. But it's resolved for me.)

Issue #8397, opened 2024-08-04
Cannot include `node_modules/*/examples/**/*`?


Bug-Reaper:
* **Electron-Builder Version**:  24.13.3
* **Node Version**: 16.20.2
* **Electron Version**: 29.4.5
* **Electron Type (current, beta, nightly)**: current
* **Electron Updater**: 6.2.1
* **Target**: MacOS - ARM-64 - DMG

Popular lib THREE.js includes a variety of optional (but useful) modules in `node_modules/three/examples/**/*`. My app uses these and they're not getting bundled into my ASAR.

I see the default [file filters](https://www.electron.build/configuration/contents#files) cause these to be filtered out but I cannot figure out how to stop this behavior. I've tried below config and a few variants:
```javascript
{
   files : [
      "**/*",
      "**/node_modules/*/examples/**/*",
    ]
}
```
From the docs it appears I'm getting f*cked by "**All default ignores are added in any case**" which seems to be the case even if the default-ignores are explicitly included.

Is there some way to override the default ignores so I can build properly?

Bug-Reaper:
I went snooping and I couldn't locate where these "default-ignores" are appended. Ironically, I did discover a test that accomplishes a work-around for my exact issue lol: https://github.com/electron-userland/electron-builder/blob/84f290990915d7ffa0d2dfe0a1d739aeb2b21560/test/src/BuildTest.ts#L336-L339


This does force the relevant files to be included but it feels hacky and borderline undocumented. Are we open for an easier way for users to override the "default-ignore" patterns?

beyondkmp:
By default, these files will be excluded.

https://github.com/electron-userland/electron-builder/blob/84f290990915d7ffa0d2dfe0a1d739aeb2b21560/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts#L17-L33


We should provide an option that allows users to decide for themselves, or perhaps we could remove these examples. @mmaietta , what do you think?

mmaietta:
@beyondkmp I just realized this may have a different underlying issue.

Even if there's a default excludes, I think we should allow specifications in the `files` to be able to override that behavior. Basically, check if filematcher matches the `files` array as more important than the default excluded files.

```
   files : [
      "**/*",
      "**/node_modules/*/examples/**/*",
    ]
```
[...]

PR #8398, opened 2024-08-04
fix: add disableDefaultIgnoredFiles option

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8397

### DIFF
--- a/packages/app-builder-lib/src/options/PlatformSpecificBuildOptions.ts
+++ b/packages/app-builder-lib/src/options/PlatformSpecificBuildOptions.ts
@@ -50,6 +50,13 @@ export interface PlatformSpecificBuildOptions extends TargetSpecificOptions {
    */
   readonly compression?: CompressionLevel | null
 
+  /**
+   * Whether to ignore all default files(https://www.electron.build/configuration/contents#files) and options. Defaults to `false`.
+   *
+   * @default false
+   */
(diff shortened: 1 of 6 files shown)


packages/app-builder-lib/src/options/PlatformSpecificBuildOptions.ts
```diff
@@ -50,6 +50,13 @@ export interface PlatformSpecificBuildOptions extends TargetSpecificOptions {
    */
   readonly compression?: CompressionLevel | null
 
+  /**
+   * Whether to ignore all default files(https://www.electron.build/configuration/contents#files) and options. Defaults to `false`.
+   *
+   * @default false
+   */
+  ignoreDefaultFiles?: boolean
```
mmaietta, line 58: Minor: I think we may want to change the name of this variable, such `excludeDefaultFiles` (default true) or `disableDefaultIgnoredFiles` (default false. Might be too long of a var name though)

Bug-Reaper: +1 for `disableDefaultIgnoredFiles` very clear and you write this field once as part of a config.

beyondkmp: Fixed

PR #8394, opened 2024-08-03
feat: implement autoupdates for pacman

xyloflake:
Introducing autoupdate support for pacman!

I have updated docs & tests plus ran all tests.

I have also pushed an update to the pnpm lockfile which was auto-generated during `pnpm update`

PR #8400, opened 2024-08-05
fix(electron-updater,deb): Handle spaces in application artifact name for deb

Ryan432:


packages/electron-updater/src/DebUpdater.ts
```diff
@@ -29,10 +29,8 @@ export class DebUpdater extends BaseUpdater {
 
   protected doInstall(options: InstallOptions): boolean {
     const sudo = this.wrapSudo()
-    // pkexec doesn't want the command to be wrapped in " quotes
-    const wrapper = /pkexec/i.test(sudo) ? "" : `"`
-    const cmd = ["dpkg", "-i", options.installerPath, "||", "apt-get", "install", "-f", "-y"]
-    this.spawnSyncLog(sudo, [`${wrapper}/bin/bash`, "-c", `'${cmd.join(" ")}'${wrapper}`])
+    const installCommand = `'dpkg -i ${options.installerPath} || apt-get install -f -y'`
```
xyloflake, line 32: I'm not sure why you're removing the wrapper. It does not seem ideal. Can you please tell us what your wrapper is?

xyloflake (changes requested):

packages/electron-updater/src/DebUpdater.ts
Ryan432: After my changes the `wrapper` variable doesn't have any meaning, so to simplify the code, I have just made it "wrapper" inside the `installCommand`.

The `wrapper` is basically always a single quote `'`.


And since `/pkexec/` does require a single quote, the condition became meaningless after my change as well.

xyloflake: @Ryan432 that makes sense. If you're changing this for one platform, you should change it for the others as well.

@mmaietta he's not wrong. You should always use single quotes because it's supported by all 3 (kdesudo, gksudo, pkexec) and does not create ANY problems with any of them. Which is why single quotes must be used.

xyloflake: @Ryan432 even though the point about the wrapper being single quotes is justified, what I meant is this does NOT change anything. You're basically pushing the same command just without modifying it. That command is NOT different from what would have been generated through the string operations. In the issue the image you sent suggests that you're infact running the wrong command. The command in the repo is FINE and it works with our application. On ubuntu 22 as well, I just tested it.

Ryan432: Is linux pc you are testing is using `/pkexec/`?
Can be checked with `which gksudo || which kdesudo || which pkexec || which beesu`

The install command generated won't be the same as after my change either way, and if the linux pc using `/pkexec/` there won't be quote at all due the condition in line 33.

```
Case 1 -> const wrapper = "" // If linux pc is /pkexec/
Case 2 -> const wrapper = `''` // If linux pc is other then /pkexec/
```

I believe the auto updater works just fine on linux pcs who isn't using `/pkexec/`

xyloflake: > Is the linux pc you are testing is using `/pkexec/`? Can be checked with `which gksudo || which kdesudo || which pkexec || which beesu`
> 
> The install command generated won't be the same as after my change either way, and if the linux pc using `/pkexec/` there won't be quote at all due the condition in line 33.
> 
> ```
> Case 1 -> const wrapper = "" // If linux pc is /pkexec/
> Case 2 -> const wrapper = `''` // If linux pc is other then /pkexec/
> ```
> 
> I believe the auto updater works just fine on linux pcs who isn't using `/pkexec/` but I can't confirm.

Yes, it's pkexec. I worked with pkexec for #8394. It works fine as I've stated.

xyloflake: image

This shows why you must NOT use single quotes all the time. gksudo/kdesudo NEED those double quotes. Which is why removing the wrapper is not justified.

xyloflake: If you do indeed get an error with dpkg @Ryan432 this fix should not work either, because it's basically the same command, just a lil more hard-coded.

Ryan432: I see your point, and I agree you are right, and my fix is irrelevant, but another fix should be done.
I made a few more tests, and I know what caused my issues with auto-updater.

Sorry for the confusion, and thank you for noticing.
---

You will face issues with auto-updater when the `artifactName` having spaces, for example `"artifactName": "${productName}-${version}-${arch}.${ext}"`, assume `productName` having spaces, e.g `My Product`.

It will result in artifact name like so - `My Product-2.0.6-arm64.deb`.

Now take that code snipped:
image

If you will try executing that command, you will face an error:

```
user@user-Apple-Virtualization-Generic-Platform:/usr/bin$ /usr/bin/pkexec /bin/bash -c 'dpkg -i My Product-2.0.6-arm64.deb || apt-get install -f -y'
dpkg: error: cannot access archive 'My': No such file or directory

```

The solution is to replace ' ' with '\ '
The valid command should be:

```
/usr/bin/pkexec /bin/bash -c 'dpkg -i My\ Product-2.0.6-arm64.deb || apt-get install -f -y'
```

This is the issue I was facing, I pushed a fix to the PR, I have tested it after the fix, looks good.
And thank you again :)

xyloflake: > I see your point, and I agree you are right, and my fix is irrelevant, but another fix should be done.
> I made a few more tests, and I know what caused my issues with auto-updater.
> 
> Sorry for the confusion, and thank you for noticing.
> 
> ---
> 
> You will face issues with auto-updater when the `artifactName` having spaces, for example `"artifactName": "${productName}-${version}-${arch}.${ext}"`, assume `productName` having spaces, e.g `My Product`.
> 
> It will result in artifact name like so - `My Product-2.0.6-arm64.deb`.
> 
> Now take that code snipped:
> image
> 
> If you will try executing that command, you will face an error:
> 
> ```
> user@user-Apple-Virtualization-Generic-Platform:/usr/bin$ /usr/bin/pkexec /bin/bash -c 'dpkg -i My Product-2.0.6-arm64.deb || apt-get install -f -y'
> dpkg: error: cannot access archive 'My': No such file or directory
> 
> ```
> 
> The solution is to replace ' ' with '\ '.
> A valid command should be:
> 
> ```
> /usr/bin/pkexec /bin/bash -c 'dpkg -i My\ Product-2.0.6-arm64.deb || apt-get install -f -y'
> ```
> 
> This is the issue I was facing, I pushed a fix to the PR, I have tested it after the fix, looks good.
> And thank you again :)

Cool! Absolutely super nice that you noticed what was going wrong! Now this, makes complete sense. 

May I ask you to change the title to `fix: handle spaces in artifact name for dpkg`? That would be easier for the maintainers to figure out what the PR is actually doing.

Thank you for agreeing and actually admitting your mistake, figuring out the problem and pushing a correct fix! Have a nice day!

packages/electron-updater/src/DebUpdater.ts
```diff
@@ -31,7 +31,9 @@ export class DebUpdater extends BaseUpdater {
     const sudo = this.wrapSudo()
     // pkexec doesn't want the command to be wrapped in " quotes
     const wrapper = /pkexec/i.test(sudo) ? "" : `"`
-    const cmd = ["dpkg", "-i", options.installerPath, "||", "apt-get", "install", "-f", "-y"]
+    // application artifact names may includes spaces in their name which will cause error when executing install command.
```
xyloflake, line 34: Please consider changing it to this for grammatical correct-ness.
```suggestion
    // application artifact names may include spaces in their name which leads to an error when executing install command
```

xyloflake (changes requested):

packages/electron-updater/src/DebUpdater.ts
Ryan432: Done.

xyloflake (approved):
LGTM! Ready for review from @mmaietta imo

xyloflake:
@Ryan432 you should prolly add a changeset btw

Ryan432:
> @Ryan432 you should prolly add a changeset btw

Not sure what you mean by changeset.

xyloflake:
> > @Ryan432 you should prolly add a changeset btw
> 
> Not sure what you mean by changeset.

Have a look at the comment made by the changeset bot.

xyloflake:
> > @Ryan432 you should prolly add a changeset btw
> 
> Not sure what you mean by changeset.

https://github.com/electron-userland/electron-builder/pull/8400#issuecomment-2268419761

Ryan432:
> > > @Ryan432 you should prolly add a changeset btw
> > 
> > 
> > Not sure what you mean by changeset.
> 
> Have a look at the comment made by the changeset bot.
> 
> [#8400 (comment)](https://github.com/electron-userland/electron-builder/pull/8400#issuecomment-2268419761)

changeset added.

PR #8403, opened 2024-08-07
fix: handle spaces in artifact name for all linux platforms instead of only .deb

xyloflake:
So @Ryan432 just discovered an amazing bug (and made a PR for the fix #8400) that produces problems with the auto updates when an application has spaces in the artifact name. However, it was only implemented for .deb specifically and the same issues exists with other package managers. 

So I moved the handling of spaces in installer paths to the point where the installerPath is generated, before the `doInstall` function is called so it's pre-formatted.

To address this, I updated the logic for managing installer paths to ensure compatibility across various Linux package managers:
- **Linux:** Spaces in paths are now escaped with `\ ` to ensure proper handling in shell commands.

This fix ensures that auto-updates work seamlessly regardless of whether the artifact name includes spaces.

### DIFF
--- a/packages/electron-updater/src/BaseUpdater.ts
+++ b/packages/electron-updater/src/BaseUpdater.ts
@@ -48,7 +48,20 @@ export abstract class BaseUpdater extends AppUpdater {
     }
 
     const downloadedUpdateHelper = this.downloadedUpdateHelper
-    const installerPath = downloadedUpdateHelper == null ? null : downloadedUpdateHelper.file
+    
+    // Get the installer path, ensuring spaces are escaped on Linux
+    // 1. Check if downloadedUpdateHelper is not null
+    // 2. Check if downloadedUpdateHelper.file is not null
+    // 3. If both checks pass:
+    //    a. If the platform is Linux, replace spaces with '\ ' for shell compatibility
+    //    b. If the platform is not Linux, use the original path
+    // 4. If any check fails, set installerPath to null
+    const installerPath = downloadedUpdateHelper && downloadedUpdateHelper.file 
+      ? (process.platform === 'linux' 
+        ? downloadedUpdateHelper.file.replace(/ /g, '\\ ') 
+        : downloadedUpdateHelper.file) 
+      : null;
+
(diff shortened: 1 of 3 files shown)


packages/electron-updater/src/BaseUpdater.ts
```diff
@@ -48,7 +48,20 @@ export abstract class BaseUpdater extends AppUpdater {
...
+    // Get the installer path, ensuring spaces are escaped on Linux
+    // 1. Check if downloadedUpdateHelper is not null
+    // 2. Check if downloadedUpdateHelper.file is not null
+    // 3. If both checks pass:
+    //    a. If the platform is Linux, replace spaces with '\ ' for shell compatibility
+    //    b. If the platform is not Linux, use the original path
+    // 4. If any check fails, set installerPath to null
+    const installerPath = downloadedUpdateHelper && downloadedUpdateHelper.file 
+      ? (process.platform === 'linux' 
+        ? downloadedUpdateHelper.file.replace(/ /g, '\\ ')
```
mmaietta, line 61: Will this be an issue for AppImages? Would this break previous functionality with the `emit` logic?
https://github.com/electron-userland/electron-builder/blob/12c52a81420f04ec0e205dd83798c2b0b773011d/packages/electron-updater/src/AppImageUpdater.ts#L89-L96

xyloflake: > Will this be an issue for AppImages? Would this break previous functionality with the `emit` logic?
> https://github.com/electron-userland/electron-builder/blob/12c52a81420f04ec0e205dd83798c2b0b773011d/packages/electron-updater/src/AppImageUpdater.ts#L89-L96

Thank you for pointing that! It looks like it shouldn't create any problems but in fact fix an error caused by the mv command if this was not implemented

xyloflake: @mmaietta simulated this in node REPL.
image

Does not seem to create any problems and correctly `path.basename` recognized the escape sequence correctly and figured out `\ ` implies a space.

PR #8405, opened 2024-08-08
refactor: change quitAndInstall to asynchronous to prevent blocking main thread while updating

xyloflake:


xyloflake:
@mmaietta here's the PR but I need help for testing this. Labelled as `BREAKING CHANGE` as quitAndInstall is now async and ***might*** break previous callback methods(?).

I have not tested on any single platform and this time I'll need you to help me with this :)

xyloflake:
@mmaietta I just checked out the PR on NSIS and no, quitAndInstall does NOT work. It says the update will be installed after quit but it never gets updated. So I'm converting this to a draft PR

xyloflake:
right, i don't have the time to work on this currently (hopefully within 2 months i make out some time for this)

PR #8143, opened 2024-03-19
feat: allow `riscv64` support via custom electron dist

mmaietta:
Original contributor @kxxt in https://github.com/electron-userland/electron-builder/pull/7646

Added throwing error when `riscv64` binary is not provided via `electronDist` configuration function or string path

PR #8427, opened 2024-08-20
fix: Revert "feat: allow `riscv64` support via custom electron dist"

mmaietta:
Reverts electron-userland/electron-builder#8143


mmaietta:
Need to revert as for some reason the changeset is declaring the release as a `major` semver, which is wholly incorrect.

Issue #6031, opened 2021-07-07
electron builder update emit error: {"code":"EBUSY","dest":"C:\\Users\\andyc\\AppData\\Local\\vipthink-updater\\pending\\vipthink_win_2.13.0.exe","errno":-4082,"path":"C:\\Users\\andyc\\AppData\\Local\\vipthink-updater\\pending\\temp-vipthink_win_2.13.0.exe","syscall":"rename"}


StitchC:
* **Electron-Builder Version**:  22.7.0

* **Node Version**: 12.22.0

* **Electron Version**: 4.2.12
* **Electron Type (current, beta, nightly)**: 

* **Target**: win32

and another error emit so:

{"code":"EPERM","dest":"C:\\Users\\David\\AppData\\Local\\vipthink-updater\\pending\\vipthink_win_2.13.0.exe","errno":-4048,"path":"C:\\Users\\David\\AppData\\Local\\vipthink-updater\\pending\\temp-vipthink_win_2.13.0.exe","syscall":"rename"}

should i upgrade my electron builder version to resovle this ? or how to avoid this problem?

mmaietta:
You're running into a file lock issue so this may not explicitly be electron-builder: https://stackoverflow.com/a/44014485

Please update to latest 22.11.7 first though 🙂

StitchC:
> You're running into a file lock issue so this may not explicitly be electron-builder: https://stackoverflow.com/a/44014485
> 
> Please update to latest 22.11.7 first though 🙂

sorry, i thought my word let you puzzle. i say clear now and let you know the detail

i use the autoUpdater of electron-builder to let our app update, and some of our user will emit the error, not  emit when npm install 😂

so, this error will be resolved in the latest version ? or how i can resolve this error, thanks~

mmaietta:
>emit the error after downloaded the latest version app .zip and ready to install, not emit when npm install

Whoops, I definitely had misunderstood that. My bad 😅 

> so, this error will be resolved in the latest version ? or how i can resolve this error, thanks~

I just recall various electron-updater fixes occurring between 22.7 and 22.11, aaaand it's more efficient for the two maintainers of this project to only investigate on latest versions. 22.7.0 is also long before I joined the team.

StitchC:
> > emit the error after downloaded the latest version app .zip and ready to install, not emit when npm install
> 
> Whoops, I definitely had misunderstood that. My bad 😅
> 
[...]

PR #8437, opened 2024-08-30
fix: retry renaming update file when EBUSY error occurs due to file lock

juwonjung-hdj:
This PR addresses the issue reported in electron-userland/electron-builder#6031.

Problem:
Auto-update occasionally fails when another process (e.g., antivirus scanner) locks the temp update file during the rename operation.

Solution:
Update the logic to retry the rename operation for approximately 30 seconds if the error is EBUSY.

### DIFF
--- a/packages/electron-updater/package.json
+++ b/packages/electron-updater/package.json
@@ -16,7 +16,6 @@
     "out"
   ],
   "dependencies": {
-    "builder-util": "workspace:*",
     "builder-util-runtime": "workspace:*",
     "fs-extra": "^10.1.0",
     "js-yaml": "^4.1.0",
--- a/packages/electron-updater/package.json
+++ b/packages/electron-updater/package.json
@@ -16,6 +16,7 @@
     "out"
   ],
   "dependencies": {
+    "builder-util": "workspace:*",
     "builder-util-runtime": "workspace:*",
     "fs-extra": "^10.1.0",
     "js-yaml": "^4.1.0",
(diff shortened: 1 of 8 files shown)


packages/electron-updater/package.json
```diff
@@ -16,6 +16,7 @@
     "out"
   ],
   "dependencies": {
+    "builder-util": "workspace:*",
```
mmaietta, line 19: I don't think we're able to import `builder-util` into electron-updater as it is meant for local building. I think `retry` may need to be moved into `builder-util-runtime` in order for this PR to proceed.

juwonjung-hdj: I moved the core retry function to builder-util-runtime, but exported a wrapper function of `retry` in `builder-util` that adds the original logging behavior, since the logging module remains in `builder-util`.

mmaietta:
Just noticed a changeset file is missing from the PR. Can you please generate one using `pnpm generate-changeset` when you have a chance?
I tried to directly via the PR since I have the write permissions to PRs, but I received an error that the fork is out of date.

juwonjung-hdj:
> Just noticed a changeset file is missing from the PR. Can you please generate one using `pnpm generate-changeset` when you have a chance? I tried to directly via the PR since I have the write permissions to PRs, but I received an error that the fork is out of date.

Oh, my organization has access limited to specific IPs. That's probably the reason.
I've added a changeset file as a patch bump.

Issue #8426, opened 2024-08-20
error in rebuilding native dependency  on win10


kerwinxu:
kindly the below error 
```
rebuilding native dependencies  dependencies=@serialport/bindings-cpp@12.0.1 platform=win32 arch=x64
  • rebuilding native dependency  name=@serialport/bindings-cpp version=12.0.1
  ⨯ cannot execute  cause=fork/exec C:\Users\kerwin\.yarn\releases\yarn-1.22.19.cjs: %1 is not a valid Win32 application.
                    command='C:\Users\kerwin\.yarn\releases\yarn-1.22.19.cjs' run install
                    workingDir=C:\Users\kerwin\Desktop\com_helper\node_modules\@serialport\bindings-cpp
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```
but i can Manually build.
```
node-gyp configure
gyp info it worked if it ends with ok
gyp info using node-gyp@10.2.0
gyp info using node@20.16.0 | win32 | x64
gyp info find Python using Python version 3.11.5 found at "C:\Users\kerwin\AppData\Local\Programs\Python\Python311\python.exe"

gyp info find VS using VS2022 (17.5.33424.131) found at:
gyp info find VS "d:\Program Files\Microsoft Visual Studio\2022\Community"
gyp info find VS run with --verbose for detailed information
gyp info spawn C:\Users\kerwin\AppData\Local\Programs\Python\Python311\python.exe
gyp info spawn args [
gyp info spawn args 'C:\\Users\\kerwin\\AppData\\Local\\Yarn\\Data\\global\\node_modules\\node-gyp\\gyp\\gyp_main.py',
gyp info spawn args 'binding.gyp',
gyp info spawn args '-f',
gyp info spawn args 'msvs',
gyp info spawn args '-I',
gyp info spawn args 'C:\\Users\\kerwin\\Desktop\\com_helper\\node_modules\\@serialport\\bindings-cpp\\build\\config.gypi',
gyp info spawn args '-I',
gyp info spawn args 'C:\\Users\\kerwin\\AppData\\Local\\Yarn\\Data\\global\\node_modules\\node-gyp\\addon.gypi',
gyp info spawn args '-I',
gyp info spawn args 'C:\\Users\\kerwin\\AppData\\Local\\node-gyp\\Cache\\20.16.0\\include\\node\\common.gypi',
gyp info spawn args '-Dlibrary=shared_library',
[...]

PR #8450, opened 2024-09-05
fix: correct native dependency tree mismatch in app-builder rebuild

leey0818:
This PR fixes the mismatch in the dependency tree structure provided by the rebuild-node-modules command in the app-builder binary.

### DIFF
--- a/.changeset/chatty-rice-hunt.md
+++ b/.changeset/chatty-rice-hunt.md
@@ -1,0 +1,6 @@
+---
+"app-builder-lib": patch
+"electron-builder": patch
+---
+
+fix: correct native dependency tree mismatch in app-builder rebuild
--- a/packages/app-builder-lib/src/packager.ts
+++ b/packages/app-builder-lib/src/packager.ts
@@ -132,8 +132,8 @@ export class Packager {
 
   private nodeDependencyInfo = new Map<string, Lazy<Array<any>>>()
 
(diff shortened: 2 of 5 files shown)


mmaietta:
@beyondkmp can you take a look at this when you have a chance? This PR is proposing conditionally applying the `--flatten` flag.

mmaietta:
@leey0818 do you have a github issue that this is fixing? Or can you provide a minimal repro repo as to what the issue is originally? I'd like to create a unit test for this if possible

beyondkmp:
@mmaietta This PR should be able to fix the issue(https://github.com/electron-userland/electron-builder/issues/8426#issuecomment-2336636232)

mmaietta:
> I have a question: will the legacy configuration for nativeRebuilder be deprecated in the future, or will it be retained permanently?

Probably retained for quite some time, but eventually deprecated and finally removed. Just want to make it easy to debug any issues that arise between app-builder-bin migration to electron-rebuild. electron-builder v25 still doesn't have broad adoption as `latest` tag

leey0818:
@mmaietta I have created a GitHub issue (#8459).

Using the legacy option to utilize pre-built native dependencies in a cross-platform build. It seems that @electron/rebuild does not yet support cross-platform builds.

beyondkmp:
LGTM

leey0818:
Fixed #8459

Issue #8389, opened 2024-07-30
ESM Config filetype causes import issues in external dep (pkg read-config-file)


jbool24:
* **Electron-Builder Version**: 24.13.3

* **Node Version**: 20.12.2

* **Electron Version**: 31.3.1
* **Electron Type (current, beta, nightly)**: current

* **Electron Updater Version**: 5.2.1

* **Target**: ANY

The current external dependency used to load the config file is not prepared to handle ESM and therefore breaks the build if using native ESM modules and an external config (ESM js file). 

[util/config.ts](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/util/config.ts#L43) calls to --> [read-config-file-main.ts#L19](https://github.com/develar/read-config-file/blob/master/src/main.ts#L19) to load the config file which causes an error as there are no checks if we passed an ESM module type *.js file and does not use dynamic imports to load the file.

How to reproduce. 
1. Create a vanilla project with an external config and electron-builder installed
2. add `type: "module"` to package.json
3. Run `yarn run electron-builder build -c ./PATH_TO_ESM_JS_FILE` to start build

The following error is presented:

```shell
❯ yarn run electron-builder build -c ./electron-builder.config.js                                                                                                                                                                  ✗ staging 
yarn run v1.22.21
$ /home/{{PATH_REDACTED}}/node_modules/.bin/electron-builder build -c ./electron-builder.config.js
  • electron-builder  version=24.13.3 os=6.5.0-44-generic
  ⨯ require() of ES Module /home/{{PATH_REDACTED}}/electron-builder.config.js from /home/{{PATH_REDACTED}}/node_modules/read-config-file/out/main.js not supported.
electron-builder.config.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead either rename electron-builder.config.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /home/{{PATH_REDACTED}}/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
  failedTask=build stackTrace=Error [ERR_REQUIRE_ESM]: require() of ES Module /home/{{PATH_REDACTED}}/electron-builder.config.js from /home/{{PATH_REDACTED}}p/node_modules/read-config-file/out/main.js not supported.
[...]

Issue #8451, opened 2024-09-05
electron-builder@25 crashes if electron-builder.env exists


Dozorengel:
* **Electron-Builder Version**: 25.0.0 - 25.0.5

* **Node Version**: 20.17.0

* **Electron Version**: 31.4.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: mac, windows

If there is `electron-builder.env` (even empty), any build command throws an error with the following stacktrace:
```
❯ npx electron-builder
  ⨯ require(...) is not a function  failedTask=build stackTrace=TypeError: require(...) is not a function
    at loadEnv (/Users/dozorengel/projects/glabix-screen-recorder/node_modules/read-config-file/src/main.ts:140:27)
```

If I delete this file, build is working fine. The latest working version is 24.13.3

mmaietta:
I'll take a look at this asap. There's other issues appearing with the updated dependency read-config-file, so I may have to just copy the logic directly into electron-builder for more control of the implementation.

mmaietta:
Quick note: I've extracted read-config-file into electron-builder, updated `dotenv-expand` in this PR https://github.com/electron-userland/electron-builder/pull/8455 and was unable to reproduce the issue, so hopefully that was just a simple fix.

emeryao:
+1 same here

mmaietta:
This should be fixed in electron-builder 25.1 (with app-builder-lib 25.1.1), can you please give that a shot?

emeryao:
`electron-builder@25.1.2` works fine with `.env` file for my project 👍

Issue #8435, opened 2024-08-29
Cannot read properties of undefined (reading 'extends')


ficristo:
* **Electron-Builder Version**: 25.0.5

* **Node Version**: 20.16.0

* **Electron Version**: 31.4.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: all

I wanted to move the configuration of electron-builder from `package.json` to a typescript file named `electron-builder.ts`.
Running the script `electron-builder --dir` throws the following error:
`Cannot read properties of undefined (reading 'extends')`
I tried to rename the config file but had the same error.
See also:
https://github.com/quadre-code/quadre/commit/9c02b17a31188af27639c68999f38c2e4fd393ce
https://github.com/quadre-code/quadre/actions/runs/10614792861/job/29421489405

Using the js version seems to work (just needs to have a differente name from the default).

mmaietta:
Can you try using `export default` instead of `module.exports`?

This is how I've set up the typescript config file for my test project. Wondering if the `satifies Configuration` is causing the subdependency `read-config-file` (or internally the https://www.npmjs.com/package/config-file-ts package) to throw the error.
```
import { Configuration } from "app-builder-lib"

const options: Configuration = {
    // options
};

export default options;

```

ficristo:
@mmaietta thank you!
It worked https://github.com/quadre-code/quadre/commit/b07a2c8ffefaad4bd6bee2a022a2543e463848ca
I just used the `Configuration` type from `electron-builder` instead of `app-builder-lib`.

I would be nice to add this to the documentation and how to add the types in the jsdoc comment in a js file.
Also it seems that the configuration file can be something like `electron-builder.config.ts` even without using the `--config` command.

mmaietta:
You mean like this? https://www.electron.build/api/programmatic-usage

ficristo:
Oh, didn't see that! 
I was only looking here https://www.electron.build/configuration/configuration

PR #8455, opened 2024-09-07
fix: allow usage of `module` type config files

mmaietta:
Fixes: #8389 #8435 #8451


Matse2005:
When will a new version be released with this fix?

Issue #8276, opened 2024-06-26
Support for Azure Trusted Signing in Electron Builder


iliakolev:
* **Electron-Builder Version**:  24.13.3

* **Node Version**: 18.16.0

* **Electron Version**: 31.0.2
* **Electron Type (current, beta, nightly)**:  electron-updater: 6.2.1

* **Target**: Windows

I'm inquiring about the possibility of integrating Azure Trusted Signing within the Electron Builder workflow. Currently, my build process involves:

1. Signing the executable using the [azure/trusted-signing-action@v0.3.20](https://github.com/Azure/trusted-signing-action) GitHub Action.
2. Packaging and releasing the signed executable, ensuring auto-update functionality with s3 provider.

If you could provide guidance on how to integrate Azure Trusted Signing more seamlessly with Electron Builder or if there are plans to support this in the future, it would be greatly appreciated.

Thank you for your time and assistance.

jmeinke:
We're also interested in this. There is a nice guide for implementation of trusted EV code signing using Azure here: https://melatonin.dev/blog/code-signing-on-windows-with-azure-trusted-signing/

Basically, we'd need to call the action multiple times: 
1. on the executable files after they have been built
2. on the build installer executable after everything has been packaged in

Currently, it seems that steps 1 and 2 are done in one action and therefore it's not possible to execute code signing after each of the steps using the GitHub action, right?

MikeJerred:
From the [github action code](https://github.com/Azure/trusted-signing-action/blob/main/action.yml) it looks like a powershell script that installs the TrustedSigning module and then invokes it with appropriate params is all that is required:

```pwsh
Install-Module -Name TrustedSigning -RequiredVersion 0.4.1 -Force -Repository PSGallery
Invoke-TrustedSigning @params
```

This should be possible to add to electron-builder with some extra env vars?

mmaietta:
I quickly put together a basic copy of the action code for an `afterSign` hook, which maybe you'd be able to test with.

Basically, we skip the signing stage for the regular signing process. Instead, we leverage the `afterSign` for running the powershell commands.

Would you be willing to do some investigating from there? Not sure if the params need to be passed in as a var or as a literal string. I don't have any azure account to test this myself

```
// electron-builder-config.ts
import { AfterPackContext, Configuration } from "app-builder-lib"
import { CustomWindowsSignTaskConfiguration } from "app-builder-lib/out/codeSign/windowsCodeSign"
[...]

PR #8458, opened 2024-09-09
feat(win): implement Azure Trusted Signing

mmaietta:
Implements: https://github.com/electron-userland/electron-builder/issues/8276

Adds signtool.exe-related args to sub-config `signtoolOptions` so that a new sub-config `azureOptions` can be used for configuring Azure Trusted Signing. To maintain backward compatibility, the previous config has been marked deprecated and the values passed there will be fallback values for the new `signtoolOptions.<field_name>`


packages/app-builder-lib/src/codeSign/windowsCodeSign.ts
```diff
@@ -74,6 +115,39 @@ export async function sign(options: WindowsSignOptions, packager: WinPackager):
...
+  if (options.options.signDlls === true) {
+    signExts.add("dll")
+  }
+  if (options.options.signExts) {
+    options.options.signExts.forEach(v => signExts.add(v))
+  }
+
+  const params = {
+    ...config,
+    FilesFolder: options.path,
```
andersk, line 137: `options.path` is a file, not a folder, right? We can pass it using `-Files` instead of `-FilesFolder` and `-FilesFolderFilter`.

mmaietta: Great callout. Will update that.

mmaietta:
Okay, got this working up until the point it does `Invoke-TrustedSigning` as then the parallels VM prompts for `Endpoint`, which I don't have an account with which to test with. Requires `NuGet` package provider to be installed and `TrustedSigning` module, but both also required `"-Scope", "CurrentUser"` since the cmd prompt that is automatically executed within a parallels VM is not elevated to admin.

Logs below with `DEBUG=electron-builder`
```
  • signing         file=dist/win-unpacked/electron-quick-start-typescript.exe certificateFile=Foo Bar.pfx
  • signing with Azure Trusted Signing  path=/Users/dev/Development/electron-builder-test-2/dist/win-unpacked/electron-quick-start-typescript.exe
  • executing       file=prlctl args=list -i -s name
  • executing       file=prlctl args=exec {6db0fa46-4f04-432a-a546-f8584beac98f} --current-user powershell.exe -NoProfile -NonInteractive -Command Get-Command pwsh.exe
  • ensure that 'Share folders' is set to 'All Disks', see https://goo.gl/E6XphP
  • unable to find pwsh.exe, falling back to powershell.exe
  • executing       file=prlctl args=exec {6db0fa46-4f04-432a-a546-f8584beac98f} --current-user powershell.exe Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force -Scope CurrentUser
  • executed        file=prlctl stdout=
Name                           Version          Source           Summary                                               
----                           -------          ------           -------                                               
nuget                          2.8.5.208        https://onege... NuGet provider for the OneGet meta-package manager    
                      
  • executing       file=prlctl args=exec {6db0fa46-4f04-432a-a546-f8584beac98f} --current-user powershell.exe Install-Module -Name TrustedSigning -RequiredVersion 0.4.1 -Force -Repository PSGallery -Scope CurrentUser
  • executed        file=prlctl
  • executing       file=prlctl args=exec {6db0fa46-4f04-432a-a546-f8584beac98f} --current-user powershell.exe Invoke-TrustedSigning -Files /Users/dev/Development/electron-builder-test-2/dist/win-unpacked/electron-quick-start-typescript.exe
```

The current solution is a bit disorganized, likely will need a large refactor since things like `cscInfo` are no longer required. Then the tedious task of updating the docs will need to happen as well.

Issue #8507, opened 2024-09-19
Unable to copy file for caching: ENOENT


theboolean:
* **Electron-Builder Version**: 24.13.3

* **Node Version**: 18.x

* **Electron Version**: 31
* **Electron Type (current, beta, nightly)**: current
Which version of electron-updater are you using (if applicable)? 6.3.4

* **Target**:  macOS

After updating electron-updater from 6.2.0 to 6.3.4 we are getting errors like:
`Unable to copy file for caching: ENOENT: no such file or directory, copyfile '/Users/{USER_NAME_REDACTED}/Library/Caches/{APP_NAME_REDACTED}-updater/pending/{APP_NAME_AND_VERSION_REDACTED}-arm64-mac.zip' -> ''`
every time users check for updates. this line seems to be the problem: https://github.com/electron-userland/electron-builder/blob/3b03f27245458a0b595b7a65c8634f184d32f2fa/packages/electron-updater/src/MacUpdater.ts#L123

mmaietta:
Great find in identifying the root cause. Looks like it's non-blocking of the update due to the try-catch, but definitely something that warrants further investigation.
https://github.com/electron-userland/electron-builder/blob/3b03f27245458a0b595b7a65c8634f184d32f2fa/packages/electron-updater/src/MacUpdater.ts#L122-L127
The var is set on this line.
https://github.com/electron-userland/electron-builder/blob/3b03f27245458a0b595b7a65c8634f184d32f2fa/packages/electron-updater/src/MacUpdater.ts#L104

I'll continue diving into this, but curious on @beyondkmp's thoughts here? Why would this not be getting executed?
https://github.com/electron-userland/electron-builder/blob/3b03f27245458a0b595b7a65c8634f184d32f2fa/packages/electron-updater/src/MacUpdater.ts#L104

MelonCode:
hey, any updates on this?

beyondkmp:
@mmaietta 
https://github.com/electron-userland/electron-builder/blob/3b03f27245458a0b595b7a65c8634f184d32f2fa/packages/electron-updater/src/AppUpdater.ts#L698-L702
The problem likely lies here. If it already exists, the task function won't run; instead, the done function will be executed directly.

mmaietta:
Released in 25.1.7, please feel free to give it a shot!

PR #8541, opened 2024-09-27
fix: Unable to copy file for caching: ENOENT

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8507

https://github.com/electron-userland/electron-builder/blob/3b03f27245458a0b595b7a65c8634f184d32f2fa/packages/electron-updater/src/AppUpdater.ts#L698-L702
The problem lies here. If cachedUpdateFile already exists, the task function won't run; instead, the done function will be executed directly.

### DIFF
--- a/packages/electron-updater/src/MacUpdater.ts
+++ b/packages/electron-updater/src/MacUpdater.ts
@@ -94,13 +94,13 @@ export class MacUpdater extends AppUpdater {
 
     const provider = downloadUpdateOptions.updateInfoAndProvider.provider
     const CURRENT_MAC_APP_ZIP_FILE_NAME = "update.zip"
-    const cachedUpdateFilePath = path.join(this.downloadedUpdateHelper!.cacheDir, CURRENT_MAC_APP_ZIP_FILE_NAME)
 
     return this.executeDownload({
       fileExtension: "zip",
       fileInfo: zipFileInfo,
       downloadUpdateOptions,
       task: async (destinationFile, downloadOptions) => {
+        const cachedUpdateFilePath = path.join(this.downloadedUpdateHelper!.cacheDir, CURRENT_MAC_APP_ZIP_FILE_NAME)
         const canDifferentialDownload = () => {
           if (!pathExistsSync(cachedUpdateFilePath)) {
             log.info("Unable to locate previous update.zip for differential download (is this first install?), falling back to full download")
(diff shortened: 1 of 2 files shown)


packages/electron-updater/src/MacUpdater.ts
```diff
@@ -94,16 +94,15 @@ export class MacUpdater extends AppUpdater {
...
     return this.executeDownload({
       fileExtension: "zip",
       fileInfo: zipFileInfo,
       downloadUpdateOptions,
       task: async (destinationFile, downloadOptions) => {
-        cachedUpdateFile = path.join(this.downloadedUpdateHelper!.cacheDir, CURRENT_MAC_APP_ZIP_FILE_NAME)
+        const cachedUpdateFilePath = path.join(this.downloadedUpdateHelper!.cacheDir, CURRENT_MAC_APP_ZIP_FILE_NAME)
         const canDifferentialDownload = () => {
-          if (!pathExistsSync(cachedUpdateFile)) {
+          if (!pathExistsSync(cachedUpdateFilePath)) {
```
mmaietta, line 105: This looks good, but I'm curious as to why `cacheDir` is an empty string. Won't this `!pathExistsSync(...)` always return false and fallback to full download?

beyondkmp: https://github.com/electron-userland/electron-builder/blob/3b03f27245458a0b595b7a65c8634f184d32f2fa/packages/electron-updater/src/AppUpdater.ts#L698-L702
The problem likely lies here. If cachedUpdateFile already exists, the task function won't run; instead, the done function will be executed directly.

packages/electron-updater/src/MacUpdater.ts
```diff
@@ -120,7 +119,8 @@ export class MacUpdater extends AppUpdater {
       },
       done: event => {
         try {
-          copyFileSync(event.downloadedFile, cachedUpdateFile)
+          const cachedUpdateFilePath = path.join(this.downloadedUpdateHelper!.cacheDir, CURRENT_MAC_APP_ZIP_FILE_NAME)
+          copyFileSync(event.downloadedFile, cachedUpdateFilePath)
```
mmaietta, line 123: What are your thoughts on converting this to a `log.warn` since it's not a critical error and still allows the update to proceed? Also, I think we should check if `disableDifferentialDownload` is `false` since caching is unnecessary if differentials are disabled.
```
done: event => {
  if (!downloadUpdateOptions.disableDifferentialDownload) {
    try {
      const cachedUpdateFilePath = path.join(this.downloadedUpdateHelper!.cacheDir, CURRENT_MAC_APP_ZIP_FILE_NAME)
      copyFileSync(event.downloadedFile, cachedUpdateFilePath)
    } catch (error: any) {
      this._logger.warn(`Unable to copy file for caching for future differential downloads: ${error.message}`)
    }  
  }
  return this.updateDownloaded(zipFileInfo, event)
},
```

beyondkmp: Agreed. Fixed.

mmaietta (approved):

Issue #8460, opened 2024-09-10
"!node_modules/?angular${/*}" in files doesn't exclude node_modules/@angular folder


tim-wilkins:
25.0.5

* **Electron-Builder Version**: 

v20.12.0
* **Node Version**: 

32.0.2

* **Electron Version**:
* **Electron Type (current, beta, nightly)**:
n/a

 Windows, msi
* **Target**: 

 This worked with 24.13.3 and electron 32.0.1, but I noticed that the asar file had grown a lot and extracting it find the angular folder with contents.  All versions are updated using "ng update electron" etc. when "npm outdated" shows an outdated version.  I do also have  electron-reload@1.5.0 (not updated since only alpha).

mmaietta:
The only change I was able to find in [the diff](https://github.com/electron-userland/electron-builder/compare/v24.13.3...v25.0.6) that might be related is an upgrade from `"minimatch": "^5.1.1"`, to `"^10.0.0"`

Can you try force-resolving/overwriting the version of minimatch in your package.json to 5.1.1 and see if that resolves the issue? If so, I can open a PR asap and get it into the next release

absidue:
@mmaietta I tried overriding the minimatch version with yarn but it complained about the versions being to far apart and ignored the override. In our case we tell electron-builder to ignore the node_modules folder entirely because we use webpack to bundle the files first, but going from 24.13.3 to 25.0.5, resulted in electron-builder pulling in packages in the node_modules folder that start with `@`, even things like babel which are listed in the development dependencies which electron-builder is supposed to ignore by default. For the moment we have reverted back to the previous electron-builder version that we were using.

On an unrelated note the newer version of electron-builder logs that it is adding asar integrity data, even though we don't even use that and the docs don't list any obvious way to stop electron-builder doing that unnecessary work (bit like how there doesn't seem to be a way to turn off code signing completely).

absidue:
With that version `electron-builder install-app-deps` takes forever. According to task manager `app-builder.exe` is using a bunch of CPU and has very slowly increasing RAM usage (been running for over 10 minutes and currently sitting at 230MB of RAM usage), so it is going to be a while before I can report back on whether that upgraded version fixes the original problem or not.

[...]

PR #8547, opened 2024-09-30
fix: fix the main matcher patterns for !node_modules/@test/xxxx

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8460

Reproducible demo
```
{
  "name": "TestApp",
  "productName": "Test App ßW",
  "description": "My Electron application description",
  "keywords": [],
  "main": "./main.js",
  "version": "1.0.0",
  "author": "beyondkmp",
  "scripts": {
    "start": "electron .",
    "dist": "electron-builder"
  },
  "license": "MIT",
  "build": {
    "appId": "electron-blog-example",
       "files": [
        "!node_modules/@electron/remote/*"
       ],
    "win": {
      "target": "nsis"
    }
  },
  "devDependencies": {
    "electron": "32.1.0",
    "electron-builder": "25.1.6"
  },
  "dependencies": {
     "tar": "7.4.3",
     "@electron/remote": "2.1.2"
  }
}
```

### DIFF
--- a/packages/app-builder-lib/src/util/appFileCopier.ts
+++ b/packages/app-builder-lib/src/util/appFileCopier.ts
@@ -206,7 +206,7 @@ export async function computeNodeModuleFileSets(platformPackager: PlatformPackag
     return path.dirname(parentDir)
   }
 
-  const collectNodeModules = async (dep: NodeModuleInfo,realSource:string, destination: string) => {
+  const collectNodeModules = async (dep: NodeModuleInfo, realSource: string, destination: string) => {
     const source = dep.dir
     const matcher = new FileMatcher(realSource, destination, mainMatcher.macroExpander, mainMatcher.patterns)
     const copier = new NodeModuleCopyHelper(matcher, platformPackager.info)
@@ -223,7 +223,7 @@ export async function computeNodeModuleFileSets(platformPackager: PlatformPackag
   for (const dep of deps) {
     const destination = path.join(mainMatcher.to, NODE_MODULES, dep.name)
     const realSource = getRealSource(dep.name, dep.dir)
-    await collectNodeModules(dep,realSource, destination)
+    await collectNodeModules(dep, realSource, destination)
   }
 
   return result
--- a/packages/app-builder-lib/src/util/appFileCopier.ts
+++ b/packages/app-builder-lib/src/util/appFileCopier.ts
@@ -201,23 +201,23 @@ export async function computeNodeModuleFileSets(platformPackager: PlatformPackag
     return path.dirname(path.normalize(parentPath))
   }
 
-  const collectNodeModules = async (dep: NodeModuleInfo, source: string, destination: string) => {
-    const matcher = new FileMatcher(source, destination, mainMatcher.macroExpander, mainMatcher.patterns)
+  const collectNodeModules = async (dep: NodeModuleInfo, destination: string) => {
+    const source = dep.dir
+    const matcher = new FileMatcher(getRealSource(dep.name, source), destination, mainMatcher.macroExpander, mainMatcher.patterns)
     const copier = new NodeModuleCopyHelper(matcher, platformPackager.info)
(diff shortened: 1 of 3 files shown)


packages/app-builder-lib/src/util/appFileCopier.ts
```diff
@@ -188,32 +188,40 @@ export async function computeNodeModuleFileSets(platformPackager: PlatformPackag
   const result = new Array<ResolvedFileSet>()
   let index = 0
   const NODE_MODULES = "node_modules"
-  const getRealSource = (source: string) => {
-    const parentDir = path.dirname(source)
+  const getRealSource = (name: string, source: string) => {
+    const normalizedPath = source.split(path.sep).join("/")
```
mmaietta, line 192: I think we can use `path.normalize` here or `path.resolve`. If it needs to be `"/"`, then we can `import * as pathPosix from "path/posix"` IIRC. What are your thoughts?

packages/app-builder-lib/src/util/appFileCopier.ts
```diff
@@ -188,32 +188,40 @@ export async function computeNodeModuleFileSets(platformPackager: PlatformPackag
...
+
+    if (!normalizedPath.endsWith(name)) {
+      throw new Error("Path does not end with the package name")
+    }
+    const parentDir = path.normalize(normalizedPath.slice(0, -name.length - 1))
```
mmaietta, line 197: I'm not sure I understand this logic. (e.g. the `-name.length-1`)
Can you explain here or add a code comment as to what this is doing?

packages/app-builder-lib/src/util/appFileCopier.ts
```diff
@@ -188,32 +188,40 @@ export async function computeNodeModuleFileSets(platformPackager: PlatformPackag
...
+        await collectNodeModules(c, realSource, path.join(destination, NODE_MODULES, c.name))
       }
     }
   }
 
   for (const dep of deps) {
     const destination = path.join(mainMatcher.to, NODE_MODULES, dep.name)
-    await collectNodeModules(dep, destination)
+    const realSource = getRealSource(dep.name, dep.dir)
+    await collectNodeModules(dep, realSource, destination)
```
mmaietta, line 224: Could you add a unit test for the issue this PR is fixing for? I'd like to start tracking the issues related to node module collector in unit tests if possible

beyondkmp: https://github.com/electron-userland/electron-builder/blob/d2e59f1bbdf3f7b270f23d8e1bafe805d3ec1c59/test/src/globTest.ts#L171

added.

packages/app-builder-lib/src/util/appFileCopier.ts
beyondkmp: Here's the English translation of the provided Chinese text:

"""
The source we currently provide is in this format: /root/path/node_modules/test or /root/path/node_modules/@electron/remote. We need to extract /root/path so that the filter in the fileMatcher can extract node_modules/test or node_modules/@electron/remote for filtering.

The issue now is with modules named like @electron/remote. The original code directly uses path.dir(path.dir) to obtain the path, which works fine for node_modules/test, but for @electron/remote, it retrieves /root/path/node_modules instead of /root/path/.

Therefore, to truncate the source using the module name, we first need to convert path.sep to / before we can perform the truncation process.
"""

packages/app-builder-lib/src/util/appFileCopier.ts
beyondkmp: This approach is indeed a bit confusing. I've modified it by changing name(@electron/remote) to `@electron + path.sep + remote`, and then truncating the package name from the suffix in the source. This should work.

packages/app-builder-lib/src/util/appFileCopier.ts
beyondkmp: Without this fix, adding this UT (Unit Test) case directly would fail.
image

Issue #8512, opened 2024-09-20
The build process will change the encoding of the license.rtf file.


TianbinTobin:
* **Electron-Builder Version**: v25.1.4

* **Node Version**: v20.16.0

* **Electron Version**: v31.6.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: windows

The original encoding of my license.rtf file is UTF-8, but after the build process, the encoding has changed to UTF-8 with BOM.
This behavior will cause the License display on the installation interface to appear as garbled text.
(My language is Chinese.)
![image](https://github.com/user-attachments/assets/7997b526-5108-45f2-9aa0-62ab15ba96a5)

mmaietta:
@TianbinTobin what OS are you packaging your Windows app on?

@beyondkmp looks like this may be related to the changes in https://github.com/electron-userland/electron-builder/pull/8314. Wondering if we need to have conditional logic when building on windows vs mac w.r.t file encoding

TianbinTobin:
> what OS are you packaging your Windows app on?

@mmaietta  Windows

beyondkmp:
@mmaietta  In https://github.com/electron-userland/electron-builder/issues/8294,  I have only fixed the issue of garbled text display on Windows. 

The problem describing seems to be a different issue on Windows. @TianbinTobin  Could you provide a minimal demo to reproduce the problem?

TianbinTobin:
> Could you provide a minimal demo to reproduce the problem?

[https://github.com/TianbinTobin/electron-builder-8512.git](https://github.com/TianbinTobin/electron-builder-8512.git)

The first build will change the encoding of the rtf file, and the second build of the installation package will appear garbled on windows. @beyondkmp

PR #8551, opened 2024-10-01
fix: check if the file already starts with a UTF-8 BOM

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8512

If the file is already in UTF-8 BOM format, there's no need to convert it again. If it's converted again, it will result in garbled characters.

### DIFF
--- a/packages/app-builder-lib/src/targets/nsis/nsisLicense.ts
+++ b/packages/app-builder-lib/src/targets/nsis/nsisLicense.ts
@@ -8,7 +8,6 @@ import { NsisScriptGenerator } from "./nsisScriptGenerator"
 import { nsisTemplatesDir } from "./nsisUtil"
 import * as fs from "fs"
 
-
 function convertFileToUtf8WithBOMSync(filePath: string): boolean {
   try {
     const UTF8_BOM_HEADER = Buffer.from([0xef, 0xbb, 0xbf])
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/src/targets/nsis/nsisLicense.ts
```diff
@@ -10,16 +11,25 @@ import * as fs from "fs"
 function convertFileToUtf8WithBOMSync(filePath: string): boolean {
   try {
     const data = fs.readFileSync(filePath)
-    // UTF-8 BOM is EF BB BF
     const BOM = Buffer.from([0xef, 0xbb, 0xbf])
+
+    // Check if the file already starts with a UTF-8 BOM
+    if (data.length >= 3 && data[0] === 0xef && data[1] === 0xbb && data[2] === 0xbf) {
```
mmaietta, line 17: Can we extract the common header config and then compare that here? Like this:
```
// UTF-8 BOM is EF BB BF
const header = [0xef, 0xbb, 0xbf];

// Check if the file already starts with a UTF-8 BOM
log.debug({ file: log.filePath(filePath) }, "checking file for BOM header")
if (data.length >= 3 && data[0] === header[0] && data[1] === header[1] && data[2] === header[2]) {
    log.debug({ file: log.filePath(filePath) }, "file is already in BOM format, skipping conversion.")
}

// If not, add the BOM
try {
    const BOM = Buffer.from(header)
    const dataWithBOM = Buffer.concat([BOM, data])
    fs.writeFileSync(filePath, dataWithBOM)
    log.debug({ file: log.filePath(filePath) }, "file successfully converted to UTF-8 with BOM")
    return true
} catch (err: any) {
    log.error({ file: log.filePath(filePath), message: err.message ?? err.stack }, "unable to convert file to UTF-8 with BOM")
    return false
}
```

Also, I'd recommend changing the `log.info` to `log.debug` so that we don't pollute the main console logs. I also added `file: log.filePath(...)` to the logging object so that it's easy to track down which file fails (might be overkill, but still informative enough for debug logs)

beyondkmp: Excellent suggestion. I've already made the changes. Using subarray for extraction and comparison, so that even if bytes are added to the BOM header later, we only need to modify the UTF8_BOM_HEADER constant. There won't be any performance issues because the array is very small.

mmaietta (approved):

PR #8562, opened 2024-10-06
fix: support including node_modules in other subdirectories

beyondkmp:
fix https://github.com/electron-userland/electron-builder/pull/6080  and support including node_modules in other subdirectories

**How to fix**

1. No file patterns

    -  The final File patterns are `["**/*", "!**/node_modules/**", "!dist{,/**/*}", ...] `

2. File patterns with other sub node_modules `["/*", "**/sub/node_modules/**"]`

   - File patterns are `["**/*", "!**/node_modules/**", "**/sub/node_modules/**", "!dist{,/**/*}", ...]`

3. File patterns without sub node_modules

   -  The final File patterns are `["**/*", "!**/node_modules/**", "!dist{,/**/*}", ...]`

The final patterns above all filter out the node_modules in the app root directory. In filter.ts, we handle this by returning false by default if relative === 'node_modules', so it won't be filtered out. 

https://github.com/electron-userland/electron-builder/blob/e2c79819751454dbd1a939610d66e940b5dfb73d/packages/app-builder-lib/src/util/filter.ts#L60-L62

However, if you really want to filter out the node_modules in the app directory, you can use `["!node_modules/**/*"]` to filter it.

**Two points to note**

1. Now `["**/*", "**/sub/node_modules"]` cannot match. Only `["**/*", "**/sub/node_modules/**"]` will match. 

   For a relative path of` sub/node_modules`, if  don't add a `/ ` at the end, we can only use `**/sub/node_module` to match. If  add a `/` at the end, we can only use `**/sub/node_module/**` to match.
   
   **I currently prefer that `**/sub/node_modules/**` matches, as this aligns more with conventional usage.**

2. `*/sub/node_modules/**`  cannot match because the relative path is `sub/node_modules`, and there's no extra directory at the beginning.

### DIFF
--- a/.changeset/tricky-files-speak.md
+++ b/.changeset/tricky-files-speak.md
@@ -1,6 +1,6 @@
 ---
-"app-builder-lib": patch
-"builder-util": patch
+"app-builder-lib": major 
+"builder-util": major 
 ---
 
 support including node_modules in other subdirectories
--- a/.changeset/tricky-files-speak.md
+++ b/.changeset/tricky-files-speak.md
@@ -1,0 +1,6 @@
+---
+"app-builder-lib": patch
+"builder-util": patch
+---
+
+support including node_modules in other subdirectories
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -7142,11 +7142,6 @@
       "description": "Whether to include PDB files.",
       "type": "boolean"
     },
-    "includeSubNodeModules": {
-      "default": false,
-      "description": "Whether to include *all* of the submodules node_modules directories",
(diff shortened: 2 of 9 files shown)


.changeset/tricky-files-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+"app-builder-lib": patch
+"builder-util": patch
```
mmaietta, line 3: Since the configuration schema is removing `includeSubNodeModules`, this needs to be a major semver bump

beyondkmp: updated

mmaietta (approved):

PR #8568, opened 2024-10-07
fix: ignore symlink file if it links outside module

beyondkmp:
If there is a symlink file in some node module that points to `../../../../../../etc/passwd`, this would cause the system's passwd file to be packaged along with it. 

The solution is to restrict symlinks within modules to only link to other files within their directory. Any symlinks pointing outside of this directory will be ignored, and a warning log will be generated.

### DIFF
--- a/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
+++ b/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
@@ -160,8 +160,8 @@ export class NodeModuleCopyHelper extends FileCopyHelper {
       const resolvedPath = realpathSync(file)
       const absolutePath = path.resolve(resolvedPath)
       if (!absolutePath.startsWith(path.resolve(tmpPath))) {
-        // delete symlink file if it links outside node_modules (e.g.
-        log.debug({ file: file, module_path: tmpPath, resolvedPath: resolvedPath }, `deleting symlink outside node_modules`)
+        // delete symlink file if it links outside module (e.g.
+        log.debug({ module: moduleName, file: file, resolvedPath: resolvedPath }, `deleting symlink outside module`)
         result[index] = undefined
       }
 
--- a/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
+++ b/packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
@@ -1,5 +1,5 @@
 import BluebirdPromise from "bluebird-lst"
-import { CONCURRENCY } from "builder-util"
+import { log, CONCURRENCY } from "builder-util"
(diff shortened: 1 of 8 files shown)


test/src/ignoreTest.ts
```diff
@@ -279,4 +280,39 @@ test.ifDevOrLinuxCi(
       },
     }
   )
+)
+
+test.ifDevOrLinuxCi(
+  "ignore symlink file if it links outside node_modules",
+  app(
```
mmaietta, line 287: We already have a test that should be failing if the symlink is outside of the project dir (doesn't need to be a node_module specifically, just any symlinked file)

https://github.com/electron-userland/electron-builder/blob/cd1e3b06c43bf8c2f133b7dcd5ed93cd64466b3b/test/src/globTest.ts#L107-L125

packages/app-builder-lib/src/util/NodeModuleCopyHelper.ts
```diff
@@ -108,6 +108,15 @@ export class NodeModuleCopyHelper extends FileCopyHelper {
...
             }
 
+            if (stat.isSymbolicLink()) {
+              const resolvedLinkTarget = realpathSync(filePath)
+              if (!isSubPath(resolvedLinkTarget, tmpPath)) {
+                // delete symlink file if it links outside module
+                log.warn({ module: moduleName, file: filePath, resolvedLinkTarget }, `deleting symlink outside module`)
+                return null
+              }
+            }
```
mmaietta, line 118: I think this logic needs to be in `asarUtil` instead of `NodeModuleCopyHelper` as the symlink outside the project dir can't be allowed for both file assets or node modules.

Let me finish up the `electron/asar` migration and I can incorporate your changes here into that PR branch

beyondkmp: If it's done in asarUtil, any files linked to other workspaces within the project will all fail. I'm not sure how common this usage pattern is.

mmaietta: >any files linked to other workspaces within the project will all fail

Can you elaborate on this? Are you referring to a two-package.json project structure?

>I'm not sure how common this usage pattern is.

It's less so about how common it is and more so about enforcing best practices. For instance, electron/asar already validates making sure symlinks are within the package during asar extraction: https://github.com/electron/asar/blob/464e83436967438c74d8ac184b088fc780706b2d/src/asar.ts#L249-L253

beyondkmp: Agreeded. I will close it.

PR #8571, opened 2024-10-08
feat: switch app-builder-bin to node-module-collector to get all production node modules

beyondkmp:
change app-builder-bin to   node-module-collector(typescript)

## Design

1. Get node modules tree from "npm list"(npm & yarn) or "pnpm list"
2. Transform the tree to dependency graph
3. Transform the graph to hoisted tree
4. Hoisted tree to node modules array

## Support

- [x] npm
- [x] pnpm
- [x] pnpm with hosited
- [x] yarn1
- [x] yarn berry(with node-modules)

## Performance vs app-builder-bin

This is an IO-intensive task, primarily involving reading files and traversing the entire dependency tree. Node.js is capable of handling such IO-intensive tasks without issues. 

Based on my personal testing, there's essentially no difference in performance

## Advantages

1. Perfectly supports pnpm
2. Optimize packages in the node_modules within the asar file
    unlike the previous app-builder-bin which only searched for all node_modules without optimizing. For example, if a module has one version in the dev node_modules dependencies, another in the root node_modules, and yet another version with multiple dependencies in the production node_modules, it results in multiple duplicate modules in the production node_modules. Now, hoisting is applied in such situations, reducing these duplicate packages.


packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (hasBun) {
+    return "bun"
+  }
+  return "npm"
+}
+
+export { detect }
+
+export function getNpmVersion(pm: PM) {
+  return execa(`${pm} --version`).then(res => res.stdout)
```
mmaietta, line 107: Might want to use `res.stdout.trim()` here

packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -146,17 +145,26 @@ export async function nodeGypRebuild(platform: NodeJS.Platform, arch: string, fr
...
+  switch (pm) {
+    case "yarn":
+      return process.platform === "win32" ? "yarn.cmd" : "yarn"
+    case "pnpm":
+      return process.platform === "win32" ? "pnpm.cmd" : "pnpm"
+    case "npm":
+      return process.platform === "win32" ? "npm.cmd" : "npm"
+    default:
+      return process.platform === "win32" ? "npm.cmd" : "npm"
   }
```
mmaietta, line 162: Minor: Could probably simplify this to:
```
const suffix = process.platform === "win32" ? ".cmd" : ""
switch (pm) {
  case "yarn":
    return "yarn" + suffix 
  case "pnpm":
    return "pnpm" + suffix
  case "npm":
    return "npm" + suffix
  default:
    return "npm" + suffix
}
```

test/snapshots/mac/macPackagerTest.js.snap
```diff
@@ -766,6 +766,14 @@ Object {
...
                 "Release": Object {
                   "files": Object {
+                    "node-addon-api": Object {
+                      "files": Object {
+                        "node_addon_api_except.stamp": Object {
+                          "size": "<size>",
+                          "unpacked": true,
+                        },
+                      },
+                    },
```
mmaietta, line 776: Do you know why these are being added now? I noticed this in my electron/asar PR as well, but could only replicate it on local builds, not on CI builds

beyondkmp: This PR used yarn to reinstall for this test case. Previously, the package manager on CI was using electron-builder's pnpm by default to reinstall. The difference between these two might be the cause of the issue.

packages/app-builder-lib/src/util/yarn.ts
beyondkmp: fixed

packages/app-builder-lib/src/node-module-collector/packageManager.ts
beyondkmp: Yes. Fixed

packages/app-builder-lib/src/node-module-collector/hoist.ts
```diff
@@ -0,0 +1,1108 @@
+// copy from https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-nm/sources/hoist.ts
```
mmaietta, line 1: You can install https://www.npmjs.com/package/@yarnpkg/nm directly so we can track upstream updates to hoisting logic 🙁 
It'll be located here: node_modules/@yarnpkg/nm/lib/hoist.js

test/src/globTest.ts
```diff
@@ -131,19 +131,26 @@ test.ifDevOrLinuxCi("failed peer dep", () => {
     "test-app-one",
     {
       targets: Platform.LINUX.createTarget(DIR_TARGET),
+      config: {
+        files: ["!pnpm-lock.yaml"],
```
mmaietta, line 135: `pnpm-lock.yaml` is actually a default excludes now. https://github.com/electron-userland/electron-builder/blob/e9eef0c1c7f73a5edfe3026f044c6278641077cb/packages/app-builder-lib/src/fileMatcher.ts#L18

Not sure what this is testing for, so I'm wondering if we should have a different excludes here

test/src/helpers/packTester.ts
```diff
@@ -120,7 +121,9 @@ export async function assertPack(fixtureName: string, packagerOptions: PackagerO
 
       if (checkOptions.isInstallDepsBefore) {
         // bin links required (e.g. for node-pre-gyp - if package refers to it in the install script)
-        await spawn(process.platform === "win32" ? "npm.cmd" : "npm", ["install", "--production", "--legacy-peer-deps"], {
+        const pm = await detect({ cwd: projectDir })
+        let cmd = process.platform === "win32" ?  pm+".cmd":pm
```
mmaietta, line 125: Please run `pnpm prettier` when ready for review 🙂

packages/app-builder-lib/src/node-module-collector/hoist.ts
beyondkmp: I considered directly importing it(@yarnpkg/nm), but that would result in importing almost all of the yarnpkg packages. Currently, we only need hoist.ts, which has zero dependencies.

image

If you think importing these additional yarnpkg packages isn't a problem, I can change it to a direct import method.

test/src/globTest.ts
beyondkmp: Cool, I will delelet it. I didn't add this for any special purpose; I just didn't want to modify the snapshot.

test/src/helpers/packTester.ts
beyondkmp: No problem, I've been using `pnpm lint --fix`, but it seems this didn't format this code.

packages/app-builder-lib/src/node-module-collector/hoist.ts
mmaietta: Oomph, that's a great callout. I think it'd be safer to track `yarnpkg/nm` upstream though in case the dependency tree logic changes on their side (or they patch a bug)

Curious, can we leverage the result from https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-nm/sources/buildNodeModulesTree.ts `buildNodeModulesTree` since it's in `yarnpkg/nm` instead of the custom logic in packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts? Or does `nodeModulesCollector.ts` have additional logic or requirements for us?

beyondkmp: > Oomph, that's a great callout. I think it'd be safer to track `yarnpkg/nm` upstream though in case the dependency tree logic changes on their side (or they patch a bug)
> 
image
Looking at the historical commits, changes to this file have been infrequent, so we can record each commit in the file header. When there's a new one, we can just copy it. This effort should be minimal.

> Curious, can we leverage the result from https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-nm/sources/buildNodeModulesTree.ts `buildNodeModulesTree` since it's in `yarnpkg/nm` instead of the custom logic in packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts? Or does `nodeModulesCollector.ts` have additional logic or requirements for us?

Currently, we can't use buildNodeModulesTree, as it depends on the pnp API, which is only suitable for yarn, not for npm and pnpm. At present, 'yarn list' doesn't return the path for each module, so for the yarn package manager, I'm currently using 'npm list'. The two are compatible, and so far, I haven't encountered any issues.

I'm referencing the approach of pnpm real hoisting and UT, which allows for a relatively unified implementation.
https://github.com/pnpm/pnpm/blob/main/pkg-manager/real-hoist/src/index.ts#L8

https://github.com/yarnpkg/berry/blob/d63d411bcc5adcbffd198b8987c5a14c81eaf669/packages/yarnpkg-nm/tests/hoist.test.ts#L3-L23

test/src/helpers/packTester.ts
beyondkmp: On windows, there's an error when running `pnpm prettier`.
image

beyondkmp:
ready for review. @mmaietta Please help review when you have time.

packages/app-builder-lib/src/node-module-collector/hoist.ts
mmaietta: Should we pull in the full unit test suite for specifically the hoist logic as well then? That way we prevent regressions if we need to make patches on our side.

I'd also recommend adding those links as a comment to the code so we know where it came from and can check back if they make patches upstream 😅

beyondkmp: > Should we pull in the full unit test suite for specifically the hoist logic as well then? That way we prevent regressions if we need to make patches on our side.
> 
Good suggestion. I have pulled it:
image

> I'd also recommend adding those links as a comment to the code so we know where it came from and can check back if they make patches upstream 😅
Yes. I have linked both the code links and commit information here.
image

beyondkmp:
@mmaietta  Please help review again.

packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
```diff
@@ -0,0 +1,25 @@
...
+    super(rootDir)
+  }
+
+  getPMCommand(): string {
+    const cmd = process.platform === "win32" ? "npm.cmd" : "npm"
+    return `${cmd} list --omit dev -a --json --long`
+  }
+
+  getDependenciesTree(): DependencyTree {
+    const npmListOutput = execSync(this.getPMCommand(), {
```
mmaietta, line 16: For all areas using `execSync`, we have a function in `builder-util > util.ts` for this that includes debug logging. Can we utilize those to consolidate usage and leverage already-existing util code here?
https://github.com/electron-userland/electron-builder/blob/74d98d8aaecf1f6ee2fbd7b639972d3eab1cd0c5/packages/builder-util/src/util.ts#L73
https://github.com/electron-userland/electron-builder/blob/74d98d8aaecf1f6ee2fbd7b639972d3eab1cd0c5/packages/builder-util/src/util.ts#L165

Note, this would require minor signature change to the other function:
`getPMCommand(): { cmd: string, args: string[])`

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -0,0 +1,25 @@
...
+  getPMCommand(): string {
+    const cmd = process.platform === "win32" ? "pnpm.cmd" : "pnpm"
+    return `${cmd} list --prod --json --long --depth Infinity`
+  }
+
+  getDependenciesTree() {
+    const pnpmListOutput = execSync(this.getPMCommand(), {
+      cwd: this.rootDir,
+      encoding: "utf-8",
+      maxBuffer: 1024 * 1024 * 100,
```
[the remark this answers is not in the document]
mmaietta: Nvm, it looks like the util.ts exec function uses 1000mb 😅. Might be worthwhile to leverage its implementation instead of calling `execSync` directly. Curious on your thoughts though
https://github.com/electron-userland/electron-builder/blob/74d98d8aaecf1f6ee2fbd7b639972d3eab1cd0c5/packages/builder-util/src/util.ts#L73

packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -78,13 +79,16 @@ export function getGypEnv(frameworkInfo: DesktopFrameworkInfo, platform: NodeJS.
...
 }
 
-function checkYarnBerry() {
-  const npmUserAgent = process.env["npm_config_user_agent"] || ""
-  const regex = /yarn\/(\d+)\./gm
+async function checkYarnBerry(pm: PM) {
+  if (pm !== "yarn") {
+    return false
+  }
+  const version = await getNpmVersion(pm)
```
mmaietta, line 86: Might be worth renaming this function to `getPackageManagerVersion(pm)` as I was confused as to why we were checking for `getNpmVersion` when checking the version of yarn

packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -93,26 +97,21 @@ async function installDependencies(config: Configuration, appDir: string, option
   const additionalArgs = options.additionalArgs
 
   log.info({ platform, arch, appDir }, `installing production dependencies`)
-  let execPath = process.env.npm_execpath || process.env.NPM_CLI_JS
```
mmaietta, line 96: Shouldn't we allow these to be still an override to `getPackageToolPath`? I think that would allow us to handle `npx electron-builder` correctly, right? Not sure if we have a unit test for that (or if we even should have one tbh). Curious on your thoughts

packages/app-builder-lib/src/util/appFileCopier.ts
```diff
@@ -180,7 +181,8 @@ function validateFileSet(fileSet: ResolvedFileSet): ResolvedFileSet {
 
 /** @internal */
 export async function computeNodeModuleFileSets(platformPackager: PlatformPackager<any>, mainMatcher: FileMatcher): Promise<Array<ResolvedFileSet>> {
-  const deps = (await platformPackager.info.getNodeDependencyInfo(platformPackager.platform).value) as Array<NodeModuleInfo>
+  const deps = (await getNodeModules(platformPackager.info.appDir)) as NodeModuleInfo[]
```
mmaietta, line 184: Do we need the `as NodeModuleInfo[]`? Seems like it's already typed correctly for the function as `Promise<NodeModuleInfo[]>`

Note, I always try to avoid `as SomeType`. Instead, use `const deps: NodeModuleInfo[]` to make sure the compiler flags any errors on unexpected return types

packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  const manager: PM = await detect({ cwd: rootDir })
+  switch (manager) {
+    case "npm":
+      return new NpmNodeModulesCollector(rootDir)
+    case "pnpm":
+      return new PnpmNodeModulesCollector(rootDir)
+    case "yarn":
+      return new YarnNodeModulesCollector(rootDir)
+    default:
+      log.warn({ rootDir }, `Cannot detect lock file to determine node_modules structure. Assuming plain structure.`)
```
mmaietta, line 18: Please use `log.filePath` when logging file paths to reduce console log bloat.
e.g. `rootDir: log.filePath(rootDir)`

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -0,0 +1,120 @@
...
+import * as fs from "fs"
+import { NodeModuleInfo, DependencyTree, DependencyGraph } from "./types"
+import { log } from "builder-util"
+
+export abstract class NodeModulesCollector {
+  private nodeModules: NodeModuleInfo[]
+  protected rootDir: string
+  protected dependencyPathMap: Map<string, string>
+
+  constructor(rootDir: string) {
```
mmaietta, line 12: Pro tip, you can use `constructor(private readonly rootDir: string)` to instantiate vars with the class.

```
export abstract class NodeModulesCollector {
  private nodeModules: NodeModuleInfo[] = []
  protected dependencyPathMap = new Map<string, string>()

  constructor(private readonly rootDir: string) {}
```

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -0,0 +1,120 @@
...
+        }
+      }
+    }
+    return tree
+  }
+
+  public getNodeModules(): NodeModuleInfo[] {
+    const tree = this.getDependenciesTree()
+    const realTree = this.getTreeFromWorkspaces(tree)
+    const dependencyGraph = this.TransToDependencyGraph(realTree)
```
mmaietta, line 115: Can we rename this function? Not sure what `Trans` stands for and it also isn't camelcase
Perhaps `this.convertToDependencyGraph(realTree)`?

packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -0,0 +1,112 @@
...
+    return "yarn"
+  }
+  if (hasPnpm) {
+    return "pnpm"
+  }
+  if (hasBun) {
+    return "bun"
+  }
+  return "npm"
+}
```
mmaietta, line 102: Maybe we can use this to reduce number of exec commands on the system
```
if (await hasGlobalInstallation("yarn") {
   return "yarn"
}
if (await hasGlobalInstallation("pnpm") {
   return "yarn"
}
if (includeGlobalBun && await hasGlobalInstallation("bun") {
   return "bun"
}
return "npm"
```

packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    if (isYarn) {
+      value = "yarn"
+    } else if (isPnpm) {
+      value = "pnpm"
+    } else if (isBun) {
+      value = "bun"
+    } else if (isNpm) {
+      value = "npm"
+    }
```
mmaietta, line 69: Can we use this for default fallback of npm?
```
} else {
  value = "npm"
}
```
Then the function return type is always `Promise<PM>`

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    flatten(tree)
+    return result
+  }
+
+  abstract getDependenciesTree(): DependencyTree
+  abstract getPMCommand(): string
+
+  private _getNodeModules(dependencies: Set<HoisterResult>, result: NodeModuleInfo[]) {
+    if (dependencies.size === 0) return
```
mmaietta, line 82: Please use multiline `if` statements
```
if (!dependencies.size) {
   return
}
```

README.md
```diff
@@ -100,21 +100,6 @@ See the full documentation on [electron.build](https://www.electron.build).
...
-```
-```
-shamefully-hoist=true
-```
-
-Note: Setting shamefully-hoist to true is the same as setting public-hoist-pattern to *.
-
 ### Note for Yarn 3
 
 Yarn 3 use PnP by default, but electron-builder still need node-modules(ref: [yarnpkg/berry#4804](https://github.com/yarnpkg/berry/issues/4804#issuecomment-1234407305)). Add configuration in the `.yarnrc.yaml` as follows:
```
mmaietta, line 105: Do we still need this Note with the new node collector logic flow?

packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -0,0 +1,112 @@
...
+import { promisify } from "util"
+
+const execa = promisify(exec)
+
+export type PM = "npm" | "yarn" | "pnpm" | "bun"
+
+/**
+ * Check if a path exists
+ */
+async function pathExists(p: string) {
```
mmaietta, line 15: This function already exists in `builder-util > fs.ts` so we can reuse that from there and avoid duplicate code

packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -0,0 +1,112 @@
...
+  try {
+    await fs.access(p)
+    return true
+  } catch {
+    return false
+  }
+}
+
+const cache = new Map()
```
mmaietta, line 24: Might be good to rename this from generic `cache` as to what its purpose is serving, especially since it's a file-level var. e.g. something like `const packageManagerVersions = new Map<string, string>()`

packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -146,17 +145,20 @@ export async function nodeGypRebuild(platform: NodeJS.Platform, arch: string, fr
...
+  let cmd = pm
+
   if (process.env.FORCE_YARN === "true") {
-    return process.platform === "win32" ? "yarn.cmd" : "yarn"
-  } else {
-    return process.platform === "win32" ? "npm.cmd" : "npm"
+    cmd = "yarn"
   }
+
+  return cmd + suffix
```
mmaietta, line 156: Minor: Might be worth swapping this to
```
return `${cmd}${process.platform === "win32" ? ".cmd" : ""}`

File: packages/app-builder-lib/src/node-module-collector/yarnNodeModulesCollector.ts [hunk 22]
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  getDependenciesTree(): DependencyTree {
+    try {
+      const stdout = execSync(this.getPMCommand(), {
+        cwd: this.rootDir,
+        encoding: "utf-8",
+        maxBuffer: 1024 * 1024 * 100,
+      })
+      return JSON.parse(stdout) as DependencyTree
```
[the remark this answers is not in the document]
mmaietta: Actually, let's try to use `builder-util > util.ts` here if possible

mmaietta (changes requested):
Note: There's some util functions already existing in `builder-util/src/util` and `builder-util/src/fs` that we could leverage to reduce duplicated code and improve debug logging

packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -0,0 +1,112 @@
...
+      return /^\d+.\d+.\d+$/.test(res.stdout)
+    })
+    .then(value => {
+      cache.set(key, value)
+      return value
+    })
+    .catch(() => false)
+}
+
+function getTypeofLockFile(cwd = "."): Promise<PM | null> {
```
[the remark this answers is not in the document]
beyondkmp: Fixed.

packages/app-builder-lib/src/node-module-collector/yarnNodeModulesCollector.ts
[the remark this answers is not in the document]
beyondkmp: Your suggestion was good. I've changed it to use exec from the util module.

packages/app-builder-lib/src/node-module-collector/packageManager.ts
beyondkmp: Cool. Change it.

packages/app-builder-lib/src/node-module-collector/packageManager.ts
beyondkmp: Yes. We can return the default fallback pf npm.

README.md
beyondkmp: This is still needed, as Yarn 3 requires modules from zip files, which Electron doesn't support yet.

packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
beyondkmp: I'v changed it to use exec from util module.

packages/app-builder-lib/src/util/yarn.ts
beyondkmp: Even when using 'npx electron-builder', there should still be a PM (Package Manager, which currently defaults to npm), so it should work fine. I haven't tried using 'npx electron-builder' directly - currently there are no corresponding unit tests for this scenario, and we should add such unit tests.

beyondkmp:
> Note: There's some util functions already existing in `builder-util/src/util` and `builder-util/src/fs` that we could leverage to reduce duplicated code and improve debug logging

Excellent suggestions, all changes have been implemented.

packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  const manager: PM = await detect({ cwd: rootDir })
+  switch (manager) {
+    case "npm":
+      return new NpmNodeModulesCollector(rootDir)
+    case "pnpm":
+      return new PnpmNodeModulesCollector(rootDir)
+    case "yarn":
+      return new YarnNodeModulesCollector(rootDir)
+    default:
+      log.filePath(rootDir)
```
mmaietta, line 18: Please remove. `log.filePath(...)` returns a string as well, so this line isn't doing anything.

mmaietta (commented):
This is awesome, great work!

packages/app-builder-lib/src/node-module-collector/index.ts
beyondkmp: removed

packages/app-builder-lib/src/node-module-collector/hoist.ts
```diff
@@ -0,0 +1,1110 @@
...
```
mmaietta, line 1: Just realized we need the license file for copying this. Please copy this to the top of this file: https://github.com/yarnpkg/berry/blob/master/LICENSE.md
>Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

packages/builder-util/src/util.ts
```diff
@@ -121,6 +121,10 @@ export function exec(file: string, args?: Array<string> | null, options?: ExecFi
           }
           resolve(stdout.toString())
         } else {
+          // https://github.com/npm/npm/issues/17624
+          if ((file === "npm" || file === "npm.cmd") && args?.includes("--silent")) {
+            resolve(stdout.toString())
```
mmaietta, line 126: Just to confirm, this is intentional? We're still resolving with the stdout in silent mode?

mmaietta (changes requested):

packages/builder-util/src/util.ts
beyondkmp: Yes, the `npm list` command can throw errors in some cases, but these errors don't affect the result. Therefore, we still need to return the stdout, which contains the dependency tree and its paths. 

However, the way it was written before could be a bit misleading, so I've made some adjustments.

packages/app-builder-lib/src/node-module-collector/hoist.ts
beyondkmp: added.

mmaietta (approved):

Issue #6365, opened 2021-10-22
Electron Fuses support


kaatt:
I'm flipping [fuses](https://github.com/electron/fuses) in my afterPack script. Is there a better way to get the path to the final packaged Electron executable?

```js
const path = require('path')
const { flipFuses, FuseVersion, FuseV1Options } = require('@electron/fuses')

module.exports = async function afterPack(context) {
  const ext = {
    darwin: '.app',
    win32: '.exe',
  }[context.electronPlatformName]

  const electronBinaryPath = path.join(context.appOutDir, context.packager.appInfo.productFilename + ext) // is there a better way?
  await flipFuses(
    electronBinaryPath,
    {
      version: FuseVersion.V1,
      [FuseV1Options.EnableNodeCliInspectArguments]: false,
    },
  )
}
```

liode1s:
Hi Brother, I also encountered this problem. Electron-builder is an installer, and you cannot run Fuses directly. I obtained the generated folder and packaged it again with other packaging software. I think this function is a very important function. option configure Fuses! This is great!

kaatt:
Relevant, please reopen.

mmaietta:
This is how I pull the executable in my work project:
```
import * as builder from "electron-builder";

const appExecutableDestinationPath = (context: builder.AfterPackContext) => {
    const { packager, appOutDir, electronPlatformName } = context;
    const resourcesDir = packager.getResourcesDir(appOutDir);
    return path.join(resourcesDir, "../", electronPlatformName === "darwin" ? "MacOS" : "");
}
```
Basically, the Resources folder is always nested one level down, so resolving to the higher directory pulls the folder I need to access the executable. Maybe an approach like that could work for your purposes?

Slapbox:
I don't think #6930 should be closed unless this issue explicitly includes support for `EnableEmbeddedAsarIntegrityValidation`, which as I understand it will require more substantive changes to `electron-builder` than the others, which can simply be added with a quick post-build script.

Slapbox:
This adapted code from @kaatt's shared code should cover most use cases with reasonable defaults and includes basic Linux workarounds.

```js
// Adapted from https://github.com/electron-userland/electron-builder/issues/6365#issue-1033809141
const afterPack = async context => {
  const ext = {
    darwin: '.app',
    win32: '.exe',
    linux: [''],
  }[context.electronPlatformName];

  const IS_LINUX = context.electronPlatformName === 'linux';
[...]

PR #8588, opened 2024-10-11
feat: add integration for `@electron/fuses`

mmaietta:
feat: adding hardcoded interface for [FuseV1Config](https://github.com/electron/fuses/blob/0cf2a177e70dd81cc74c4449847287cd65ea140b/src/config.ts#L19-L34) so that electron-builder can integrate [@electron/fuses](https://github.com/electron/fuses) while also allowing scheme.json to still generated correctly

Resolves: #6365

There will be a new configuration object `electronFuses: FuseOptionsV1` and it executes `@electron/fuses` directly before signing as is recommended on the repo instructions: https://github.com/electron/fuses?tab=readme-ov-file#apple-silicon

I've also opened it up as a convenience method should you want to keep custom logic in your `afterPack` method called via
```
await context.packager.addElectronFuses(context, { ... })
```

https://github.com/electron-userland/electron-builder/blob/f5cc9ed4352a9c42343d4e0fe4dde6067d33d952/packages/app-builder-lib/src/platformPackager.ts#L398
It allows you to pass a full `FuseConfig` into the method as opposed to having electron-builder parsing `electronFuses` config property.
Taking this ☝️  approach would also allow you to set `strictlyRequireAllFuses: true` should you so desire. (I couldn't pass any var of this in-code due to https://github.com/electron/fuses/blob/0cf2a177e70dd81cc74c4449847287cd65ea140b/src/config.ts#L31-L34)

The fuse `FuseV1Options.EnableEmbeddedAsarIntegrityValidation = true` is supported now as well.
Tested locally and app runs
```
npx @electron/fuses read --app dist/mac-universal/electron-quick-start-typescript.app
Analyzing app: electron-quick-start-typescript.app
Fuse Version: v1
  RunAsNode is Enabled
  EnableCookieEncryption is Disabled
  EnableNodeOptionsEnvironmentVariable is Enabled
  EnableNodeCliInspectArguments is Enabled
  EnableEmbeddedAsarIntegrityValidation is Enabled
  OnlyLoadAppFromAsar is Disabled
  LoadBrowserProcessSpecificV8Snapshot is Disabled
```

### DIFF
--- a/packages/app-builder-lib/src/platformPackager.ts
+++ b/packages/app-builder-lib/src/platformPackager.ts
@@ -330,35 +330,14 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions> 
 
     const isAsar = asarOptions != null
     await this.sanityCheckPackage(appOutDir, isAsar, framework, !!this.config.disableSanityCheckAsar)
-    if (sign) {
-      await this.doSignAfterPack(outDir, appOutDir, platformName, arch, platformSpecificBuildOptions, targets)
-    }
-  }
 
-  protected async doSignAfterPack(outDir: string, appOutDir: string, platformName: ElectronPlatformName, arch: Arch, platformSpecificBuildOptions: DC, targets: Array<Target>) {
-    const asarOptions = await this.computeAsarOptions(platformSpecificBuildOptions)
-    const isAsar = asarOptions != null
-    const packContext = {
-      appOutDir,
-      outDir,
-      arch,
-      targets,
-      packager: this,
-      electronPlatformName: platformName,
-    }
     // the fuses MUST be flipped right before signing
     if (this.config.electronFuses != null) {
       const fuseConfig = this.generateFuseConfig(this.config.electronFuses)
       await this.addElectronFuses(packContext, fuseConfig)
     }
-    const didSign = await this.signApp(packContext, isAsar)
-    const afterSign = await resolveFunction(this.appInfo.type, this.config.afterSign, "afterSign")
-    if (afterSign != null) {
-      if (didSign) {
-        await Promise.resolve(afterSign(packContext))
-      } else {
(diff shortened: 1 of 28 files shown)


packages/app-builder-lib/src/platformPackager.ts
```diff
@@ -327,11 +330,76 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions>
...
+    if (this.config.electronFuses != null) {
+      const fuseConfig = this.generateFuseConfig(this.config.electronFuses)
+      await this.addElectronFuses(packContext, fuseConfig)
+    }
     if (sign) {
       await this.doSignAfterPack(outDir, appOutDir, platformName, arch, platformSpecificBuildOptions, targets)
     }
   }
 
+  private generateFuseConfig(fuses: FuseOptionsV1): FuseV1Config {
```
beyondkmp, line 344: How about?
```
private generateFuseConfig(fuses: FuseOptionsV1): FuseV1Config {
  const config: FuseV1Config = {
    version: FuseVersion.V1,
    resetAdHocDarwinSignature: fuses.resetAdHocDarwinSignature,
  };

  const fuseOptions: (keyof FuseOptionsV1)[] = [
    'runAsNode',
    'enableCookieEncryption',
    'enableNodeOptionsEnvironmentVariable',
    'enableNodeCliInspectArguments',
    'enableEmbeddedAsarIntegrityValidation',
    'onlyLoadAppFromAsar',
    'loadBrowserProcessSpecificV8Snapshot',
    'grantFileProtocolExtraPrivileges',
  ];

  fuseOptions.forEach(option => {
    if (fuses[option] != null) {
      config[FuseV1Options[option as keyof typeof FuseV1Options]] = fuses[option];
    }
  });

  return config;
}
```

beyondkmp (approved):

Issue #8298, opened 2024-07-07
Option to Disable Combined NSIS Installer When Building for Multiple Architectures


PeterDaveHello:
* **Electron-Builder Version**: 24.13.3
* **Node Version**: v18.20.3
* **Electron Version**: 31.1.0
* **Electron Type (current, beta, nightly)**: current
* **Target**: nsis installer, arch: ia32, x64 on Windows 11 

When building for both 32-bit and 64-bit architectures, Electron Builder creates separate installers for each architecture as well as a **combined installer**. Is there an option to disable the creation of this combined installer?

Currently, I can't find a way on the doc(https://www.electron.build/configuration/nsis#32-bit-64-bit) to only generate the separate 32-bit and 64-bit installers without also producing the combined one. It would be helpful to have an option like `nsis.disableCombinedInstaller: true` or `nsis.enableCombinedInstaller: false` to prevent the creation of the combined installer when it's not needed.

Thank you for your assistance.

PeterDaveHello:
It seems this feature doesn't just work on 32-bit(x86/ia32) + 64-bit(amd64), as the doc mentioned, but also on arm64+amd64, arm64+amd64+ia32, and maybe other combinations.

These are the screenshots from using 7-zip to open the `$PLUGINSDIR\` folder in the NSIS installers in different combinations:

![image](https://github.com/electron-userland/electron-builder/assets/3691490/e93e4147-b02d-4bb6-a478-1880da6216ac)

![image](https://github.com/electron-userland/electron-builder/assets/3691490/f1917a63-e66a-45d9-b35a-f6b395fe1dae)

mmaietta:
So I must admit I don't have the original insight as to why the installers are always combined. I had a project in the past where it was required to have two distinct installers, one for each bitness, though and I implemented an electron-builder script using the programmatic API. I think it looked something akin to the code below (just make sure your artifact name uses the `{arch}` macro so that the second `build` command doesn't overwrite the artifact of the first one.)
https://www.electron.build/api/programmatic-usage
```
await builder.build({
  targets: platform.createTarget("nsis", 'x64')
  config: options
})
await builder.build({
  targets: platform.createTarget("nsis", 'ia32')
  config: options
})
```

PeterDaveHello:
Sure we can do it by different approach like using command line to build them separately, but a simple option would be also nice 😄

mmaietta:
Can you share your electron-builder config?

[...]

Issue #8597, opened 2024-10-13
Additional unversial exe package is generated


qishibo:
* **Electron-Builder Version**: 23.0.2

* **Node Version**: v16.20.0

* **Electron Version**:12.2.3
* **Electron Type (current, beta, nightly)**:

* **Target**:  `{"target": "nsis", "arch": ["x64", "arm64"]}`

if I build arm64 and x64 nsis target, and my build field like bellow
```json
"win": {
  "target": [
    {"target": "nsis", "arch": ["x64", "arm64"]}
  ]
},
"nsis": {
  "artifactName": "${productName}-Setup-${version}-${arch}.${ext}",
}
```

after building I got 3 exes instead of 2(arm64 and x64), it seems that an universial package is generated, and its size is about 2 times bigger than arm64 and x64. I am quite confused about this, can I remove the universal version??

<img width="931" alt="image" src="https://github.com/user-attachments/assets/ef8fd507-0845-415e-8ea4-2cba76f877b2">

qishibo:
if I don't want this universal version, how can I config to reduct this


![image](https://github.com/user-attachments/assets/73864d77-b224-48c0-9ec9-16b88123525b)

mmaietta:
I took a solid look at the code and am unable to determine the reason as to why a universal package needs to be generated. The logic is quite complex though and AFAICT I don't think it's possible to remove the universal package from being generated.
AFAICT, the logic exists here: https://github.com/electron-userland/electron-builder/blob/f941f18a003a3762949a36e4a06e98e1014d3ed1/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L237-L252

mmaietta:
Looks like there's been a request to disable the universal package before as well: https://github.com/electron-userland/electron-builder/issues/8298

mmaietta:
X-posting

>Would anyone be willing to test out this [patch-package](https://www.npmjs.com/package/patch-package)?

Patch found here: https://github.com/electron-userland/electron-builder/issues/8298#issuecomment-2432884953

m1rn:
Is there any progress?

PR #8607, opened 2024-10-17
feat(nsis): allow disabling of building a universal installer 

mmaietta:
By default, electron-builder builds a universal NSIS package when multiple archs are specified. This universal exe size is usually the sum of each individual installer.

This PR introduces a config property `buildUniversalInstaller` (default `true` for backward compatibility)
In order to build separate exe's, I needed to update the installer name template to include an arch. Unless a `defaultArch` is supplied, x64 will not have a suffix appended to the file name and the other arch builds will have suffix `-${arch}`.
https://github.com/electron-userland/electron-builder/blob/1b41b17c6d1b95ca50971dfd601d8f4c3ab5f7a6/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L140-L146

Note: Only `nsis` targets supports individual installers as `nsis-web` requires 32 bit arch to be present in the distributable package since it automatically detects OS architecture and downloads the corresponding package file. `nsis-web` logic overrides any setting to `buildUniversalInstaller` to `true`.
Ref: https://www.electron.build/nsis.html#web-installer

Resolves: #8298 and #8597


packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
```diff
@@ -73,8 +86,16 @@ export class NsisTarget extends Target {
     NsisTargetOptions.resolve(this.options)
   }
 
+  protected buildUniversalInstaller() {
+    const buildSeparateInstallers = this.options.buildUniversalInstaller === false
+    return !buildSeparateInstallers
+  }
```
beyondkmp, line 92:
``` javascript
protected buildUniversalInstaller() {
    return this.options.buildUniversalInstaller 
 }
```

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
```diff
@@ -117,7 +138,10 @@ export class NsisTarget extends Target {
...
   }
 
-  protected get installerFilenamePattern(): string {
+  protected installerFilenamePattern(primaryArch?: Arch | null, defaultArch?: string): string {
+    if (!this.buildUniversalInstaller()) {
+      return "${productName} " + (this.isPortable ? "" : "Setup ") + "${version}" + (primaryArch != null ? getArchSuffix(primaryArch, defaultArch) : "") + ".${ext}"
+    }
     // tslint:disable:no-invalid-template-strings
     return "${productName} " + (this.isPortable ? "" : "Setup ") + "${version}.${ext}"
   }
```
beyondkmp, line 147:
```javascript
protected installerFilenamePattern(primaryArch?: Arch | null, defaultArch?: string): string {
  const setupText = this.isPortable ? "" : "Setup ";
  const archSuffix = !this.buildUniversalInstaller() && primaryArch != null 
    ? getArchSuffix(primaryArch, defaultArch) 
    : "";
  
  return "${productName} " + setupText + "${version}" + archSuffix + ".${ext}";
}
```

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
mmaietta: `buildUniversalInstaller` is optional, so we must check against explicit false.

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
mmaietta: Love it, I'll update the logic.

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
beyondkmp: This is not a get function, so there shouldn't be any caching, right?

mmaietta: Do we need to cache the result? It's just reading from options.

But I agree, it could simply be a get function. I'll update the PR

Issue #8622, opened 2024-10-23
Mac only: app becomes unresponsive for a period after downloading the zip file, and the mouse shows a loading icon.


q837477816:
* **Electron-Builder Version**: 23.3.0

* **Node Version**: 18.17.1 (integrated within electron@27.2.0)

* **Electron Version**: 27.2.0
* **Electron Type (current, beta, nightly)**: current

* **Electron-Updater Version**: 6.3.4

* **Target**: Mac (Apple Silicon / Intel)

Issue: After enabling delta updates for macOS in electron-updater, we observed that once the app downloads the zip file in the background, it becomes unresponsive for a period, and the mouse shows a loading icon. Additionally, our Sentry has captured main process ANR (Application Not Responding) events reported by the app.

Analysis: Upon reviewing the code, I found that in order to implement delta updates, the MacUpdater calls the synchronous **copyFileSync** Node API to cache the file after the zip file is downloaded locally. This synchronous API blocks the Electron main process, especially when the zip file is large, causing the app to freeze for longer periods (our zip file is around 400MB).
https://github.com/electron-userland/electron-builder/blob/74d98d8aaecf1f6ee2fbd7b639972d3eab1cd0c5/packages/electron-updater/src/MacUpdater.ts#L120-L129

Solution: Replace copyFileSync with an asynchronous API, such as the copyFile method provided by the fs-extra module for file copying.

q837477816:
@beyondkmp could you help to check this [PR](https://github.com/electron-userland/electron-builder/pull/8623/files)

PR #8623, opened 2024-10-23
fix: copyFileSync operation will block the main thread

q837477816:
Fix https://github.com/electron-userland/electron-builder/issues/8622


beyondkmp:
good catch!

beyondkmp (approved):

Issue #8626, opened 2024-10-24
New Azure Trusted Signing not adding a timestamp


255kb:
* **Electron-Builder Version**: 
v25.1.8

* **Node Version**: 
v22

* **Electron Version**:
v33
* **Electron Type (current, beta, nightly)**:

current

* **Target**: 
Windows

I just finished [updating my app](https://github.com/mockoon/mockoon/pull/1567/files) to use Azure Trusted Signing. Everything went well and the exe was signed. However, it triggered a Trojan detection in Windows Defender which I suspected to be a false positive.
I looked at the exe properties and noticed that the timestamp was missing: 

![image](https://github.com/user-attachments/assets/2187f784-0224-4a45-b499-e3aa56d1c2df)

Looking at the [Azure signing code](https://github.com/electron-userland/electron-builder/blob/cfa67c01827a44c88fb8448562dbe928ba37494f/packages/app-builder-lib/src/codeSign/windowsSignAzureManager.ts) it seems two params are missing in the `Invoke-TrustedSigning` command: `TimestampRfc3161` and `TimestampDigest`. They are used by the [Trusted Signing Action](https://github.com/Azure/trusted-signing-action/blob/main/action.yml#L330-L338).

I guess it should adding a timestamp using the default timestamp URL, or maybe I'm missing something? 

I solved the issue by adding additional params to the `azureSignOptions` object:

```js
azureSignOptions: {
      endpoint: 'https://eus.codesigning.azure.net',
      certificateProfileName: 'cert-name',
      codeSigningAccountName: 'account-name',
      
      // new properties
      TimestampRfc3161: 'http://timestamp.acs.microsoft.com',
      TimestampDigest: 'SHA256'
}
```

mmaietta:
Great callout and thanks for reporting the issue. I can add those as default entries to the configuration. The Trusted Signing Action indeed uses them, but none of the params in that action are actually depicted as "required" other than `CodeSigningAccountName`. That was the main reason I left `azureSignOptions` to have a `[Key: string]: string` object so that the user could apply custom TrustedSigning vars.

Makes sense that Timestamp is provided with default values. I'll get a PR set up for it today 🙂

mmaietta:
@255kb did you perchance run into this issue during your signing+publishing? https://github.com/electron-userland/electron-builder/issues/8612 Trying to fix as many azure-related issues reported in the next release as possible

255kb:
[...]

PR #8627, opened 2024-10-24
fix(win): add rfc3161 timestamp entry as default for azure signing

mmaietta:
Fixes: https://github.com/electron-userland/electron-builder/issues/8626


255kb:
Thank you!

PR #8631, opened 2024-10-25
fix: Remove path from published binaries

olivereisenhut:
During refactoring #8596, the `basename` import was changed from "path" to "path/posix", which doesn't properly reduce to the base name.
This caused some filenames to contain the absolute path after publishing.


mmaietta:
Released in [v26.0.0-alpha.4](https://github.com/electron-userland/electron-builder/releases/tag/v26.0.0-alpha.4). Thank you @olivereisenhut !

Issue #8640, opened 2024-10-29
asarUnpack not honored with electron-builder 25.1.8, brings many unneeded files into app.asar.unpacked (source files, readmes, package.json, etc.)


ivanggq:
* **Electron-Builder Version**: 25.1.8

* **Node Version**: 20.12.1

* **Electron Version**: 33.0.2
* **Electron Type (current, beta, nightly)**: current latest

* **Target**:  Windows

Recent changes in electron-builder (not sure when exactly this started) bring all kind of unnecessary files into the `app.asar.unpacked` folder (where native dependencies are). We have an `asarUnpack` rule `node_modules/**/*.node`.

This works fine with electron-builder 23.6.0, but not with recent electron-builder versions.

The unnecessary files in `app.asar.unpacked` could be (but not limited to):
 - C++ source files (.h, .cpp), but I think not .cc for example (!)
 - .js files
 - package.json, readmes, jenkinsfiles, etc. whatever the native dependency has

Example:
<img width="296" alt="image" src="https://github.com/user-attachments/assets/7a667da9-5b23-49f3-8a10-2b59ed04701d">

Repro code: [source-files-unpacked.zip](https://github.com/user-attachments/files/17559959/source-files-unpacked.zip)
To repro:
- Open the repro code on Windows (I have not tested on Mac yet, suspect it might be the same result)
- `npm install`
- `npm run package`
- Inspect the `dist\win-unpacked\resources\app.asar.unpacked\node_modules\native-reg` folder - it has a bunch of unneeded files that are not part of `asarUnpack`
- Change `electron-builder` version in package.json to `23.6.0`, npm install and repackage - only the `.node` files are packaged

Depending on the native dependency, this issue could bring C++ source files, Jenkinsfiles, etc.
I did notice in `builder-debug.yml` that .cc extensions are excluded by `firstOrDefaultFilePatterns` and this is likely why .cc files are not there (while .h and .cpp are for example), but to my understanding, the `asarUnpack` rule should be respected here. At least it seems to be respected more in  `electron-builder` version `23.6.0`.

ivanggq:
Tried with `v26.0.0-alpha.4` - the issue is still there. @mmaietta

mmaietta:
I tested your project locally and it looks like that regressed in 24.0.0-alpha.1. Can you confirm from your side?

Looks like that came in this diff: https://github.com/electron-userland/electron-builder/compare/v23.6.0...v24.0.0-alpha.1
[...]

PR #8641, opened 2024-10-29
fix: only unpack native `.node` modules, not all files in the parent folders

mmaietta:
Fixes: https://github.com/electron-userland/electron-builder/issues/8640

Reverts: https://github.com/electron-userland/electron-builder/pull/8392

### DIFF
--- a/.changeset/famous-pugs-glow.md
+++ b/.changeset/famous-pugs-glow.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: only unpack node modules, not all files in the parent folders
--- a/packages/app-builder-lib/src/asar/unpackDetector.ts
+++ b/packages/app-builder-lib/src/asar/unpackDetector.ts
@@ -15,8 +15,7 @@ function addValue(map: Map<string, Array<string>>, key: string, value: string) {
 }
 
 export function isLibOrExe(file: string): boolean {
-  // https://github.com/electron-userland/electron-builder/issues/3038
-  return file.endsWith(".dll") || file.endsWith(".exe") || file.endsWith(".dylib") || file.endsWith(".so") || file.endsWith(".node")
+  return file.endsWith(".dll") || file.endsWith(".exe") || file.endsWith(".dylib") || file.endsWith(".so")
 }
 
 /** @internal */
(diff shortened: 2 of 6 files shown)


beyondkmp:
I have a question: Should we apply the same treatment to exe and dll files as well, only extracting these specific files without unpacking other files in their parent folders?

mmaietta:
@NoahAndrews that should still work based on this line of the `unpackDetector`
https://github.com/electron-userland/electron-builder/blob/ab50ad146143c554fb7a4470815a95966b559349/packages/app-builder-lib/src/asar/unpackDetector.ts#L17-L19

That being said, what dependency is this? I can test locally and/or whip up a unit test for it

We need a way to resolve #8640, but the caveat of a revert is that `**/*.node` will need to be manually specified in `asarUnpack`, which I'm adverse to doing. I'm trying to find a middleground that auto-unpacks only `.node` addons, while still any file that is `dll`/`so`/`dylib`/`exe` are also auto-unpacked with all parent files included

NoahAndrews:
> We need a way to resolve #8640, but the caveat of a revert is that `**/*.node` will need to be manually specified in `asarUnpack`, which I'm adverse to doing. I'm trying to find a middleground that auto-unpacks only `.node` addons, while still any file that is `dll`/`so`/`dylib`/`exe` are also auto-unpacked with all parent files included

That sounds good. Looking at the actual changes in this PR, it doesn't look to me like there will be any behavior changes yet, is that accurate? I know this is still a draft.

> That being said, what dependency is this? I can test locally and/or whip up a unit test for it

Thanks! The package is https://www.npmjs.com/package/@rev-robotics/can-bridge.

While I really like the idea of a unit test for it, we're hoping to [migrate it to use static linking](https://github.com/REVrobotics/node-can-bridge/issues/13) in the near-to-medium term, so maybe it's not worth it. It's also currently Windows-only, so it wouldn't be a particularly thorough test case for cross-platform native addons that link against libraries dynamically. Your call though.

mmaietta:
>it doesn't look to me like there will be any behavior changes yet, since isLibOrExe() is only called in the one place.

It's used in two places. `unpackDetector` and `appFileCopier`, albeit the latter has no logic changes, it was always including `.node` in its logic

mmaietta:
Closing this, I think it's more worthwhile to auto unpack `.node` files as opposed to required to be specified for `asarUnpack`. This does result in parent folders+files being unpacked as well, but it does not really change the functionality or size of the bundled app - it's the same files being included.

PR #8645, opened 2024-10-31
fix: smart unpack for local module with dll

beyondkmp:
Fix the following case: the foo module contains dll files that need to be packaged into asapUnpack
```javascript
test.ifDevOrWinCi("smart unpack local module with dll file", () => {
  return app(
    {
      targets: Platform.WINDOWS.createTarget(DIR_TARGET),
    },
    {
      isInstallDepsBefore:true,
      projectDirCreated: async (projectDir, tmpDir) => {
        const tempDir = await tmpDir.getTempDir()
        let localPath = path.join(tempDir, "foo")
        await outputFile(path.join(localPath, "package.json"), `{"name":"foo","version":"9.0.0","main":"index.js","license":"MIT","dependencies":{"ms":"2.0.0"}}`)
        await outputFile(path.join(localPath, "test.dll"), `test`)
        await modifyPackageJson(projectDir, data => {
          data.dependencies = {
            debug: "3.1.0",
            "edge-cs": "1.2.1",
            foo: `file:${localPath}`,
          }
        })
      },
      packed: async context => {
        await verifySmartUnpack(context.getResources(Platform.WINDOWS))
      },
    }
  )()
})
```


test/src/BuildTest.ts
```diff
@@ -324,6 +324,34 @@ test.ifDevOrLinuxCi("win smart unpack", () => {
...
+        await modifyPackageJson(projectDir, data => {
+          data.dependencies = {
+            debug: "3.1.0",
+            "edge-cs": "1.2.1",
+            foo: `file:${localPath}`,
+          }
+        })
+      },
+      packed: async context => {
+        await verifySmartUnpack(context.getResources(Platform.WINDOWS))
```
mmaietta, line 348: Can we use `LINUX` instead of `WINDOWS`? I'm unable to run this test on my mac laptop
```
 ● smart unpack local module with dll file

    ENOENT: no such file or directory, open '/private/var/folders/y3/r1xf5xzn5pddms2gslqh14sh0000gn/T/et-3c635798d532401ebed6f91c2f35adfb/t-qWgMDo/test-project-0/dist/win-unpacked/resources/app.asar'
```

Instead of `.dll`, we could use file extension `dylib` or `so` since those are also checked in the `isLibOrExe` logic

mmaietta:
I see this is a new test case being added, is it already broken on `master`? Trying to determine how we verify the results of the unpack refactor

test/src/BuildTest.ts
beyondkmp: """
I have tested this case on Linux, and it can run successfully. I haven't tested it on Mac yet, and currently none of the BuildTest cases include macOS.
image


The reason for specifically targeting Windows is because Windows uses a different path.sep. I noticed there's already `test.ifDevOrLinuxCi("win smart unpack")` specifically running on Linux, so I added this case with a specific designation for winCI.
"""

beyondkmp:
It was already broken on master for local modules and workspace modules.

When running the above test case on the master branch, the output is as follows, where the foo module is inside the asar file rather than in asarUnpack.
```
{
  "files": {
    "app": {
      "files": {
        "package.json": {
          "files": {
            "readme.md": {
              "size": "<size>",
            },
          },
        },
        "readme.md": {
          "size": "<size>",
        },
      },
    },
    "index.html": {
      "size": "<size>",
    },
    "index.js": {
      "size": "<size>",
    },
    "node_modules": {
      "files": {
        "debug": {
          "files": {
            "LICENSE": {
              "size": "<size>",
            },
            "Makefile": {
              "size": "<size>",
            },
            "node.js": {
              "size": "<size>",
            },
            "package.json": {
              "size": "<size>",
            },
            "src": {
              "files": {
                "browser.js": {
                  "size": "<size>",
                },
                "debug.js": {
                  "size": "<size>",
                },
                "index.js": {
                  "size": "<size>",
                },
                "node.js": {
                  "size": "<size>",
                },
              },
            },
          },
        },
        "edge-cs": {
          "files": {
            "LICENSE.txt": {
              "size": "<size>",
              "unpacked": true,
            },
            "lib": {
              "files": {
                "bootstrap": {
                  "files": {
                    "Dummy.cs": {
                      "size": "<size>",
                      "unpacked": true,
                    },
                    "project.json": {
                      "size": "<size>",
                      "unpacked": true,
                    },
                  },
                  "unpacked": true,
                },
                "edge-cs.dll": {
                  "size": "<size>",
                  "unpacked": true,
                },
                "edge-cs.js": {
                  "size": "<size>",
                  "unpacked": true,
                },
              },
              "unpacked": true,
            },
            "package.json": {
              "size": "<size>",
              "unpacked": true,
            },
            "src": {
              "files": {
                "Edge.js.CSharp": {
                  "files": {
                    "EdgeCompiler.cs": {
                      "size": "<size>",
                      "unpacked": true,
                    },
                    "gulpfile.js": {
                      "size": "<size>",
                      "unpacked": true,
                    },
                    "package.json": {
                      "size": "<size>",
                      "unpacked": true,
                    },
                    "project.json": {
                      "size": "<size>",
                      "unpacked": true,
                    },
                  },
                  "unpacked": true,
                },
                "edge-cs": {
                  "files": {
                    "EdgeCompiler.cs": {
                      "size": "<size>",
                      "unpacked": true,
                    },
                    "Properties": {
                      "files": {
                        "AssemblyInfo.cs": {
                          "size": "<size>",
                          "unpacked": true,
                        },
                      },
                      "unpacked": true,
                    },
                  },
                  "unpacked": true,
                },
              },
              "unpacked": true,
            },
            "tools": {
              "files": {
                "install.js": {
                  "size": "<size>",
                  "unpacked": true,
                },
              },
              "unpacked": true,
            },
          },
          "unpacked": true,
        },
        "foo": {
          "files": {
            "package.json": {
              "size": "<size>",
            },
            "test.dll": {
              "size": "<size>",
            },
          },
          "unpacked": true,
        },
        "ms": {
          "files": {
            "index.js": {
              "size": "<size>",
            },
            "license.md": {
              "size": "<size>",
            },
            "package.json": {
              "size": "<size>",
            },
          },
        },
      },
    },
    "package.json": {
      "size": "<size>",
    },
  },
}
```

mmaietta:
>It was already broken on master for local modules and workspace modules.

Is this new test case also broken on v25.1.8? (Test doesn't run on my mac) Trying to discern if this is due to electron/asar migration and is a high priority fix or if it was a long-standing issue and can be fixed after official v26 release as a patch.

beyondkmp:
This has been present since v25.1.8, and wasn't introduced as an electron/asar feature. However, there hasn't been any feedback about it until now, when I discovered it while adding unit tests.

mmaietta (approved):

Issue #8651, opened 2024-11-01
[Mac only] The app takes a long time to quit and apply the update after calling quitAndInstall


DamonYu6:
* **Electron-Builder Version**: 23.3.0

* **Node Version**: 18.17.1 (integrated within electron@27.2.0)

* **Electron Version**: 27.2.0
* **Electron Type (current, beta, nightly)**: current

* **Electron-Updater Version**: 6.3.4

* **Target**: Mac (Apple Silicon / Intel)

Issue: After upgrading our Electron app from Electron 23 to Electron 27, we sometimes find that after calling the quitAndInstall API, the app takes over 10 minutes to quit and apply the update.

Expected Behavior: The app should quit and apply the update immediately after calling the quitAndInstall API.

Analysis: By adding extra debug information, we found that after the app receives the downloaded event dispatched by electron-updater, it takes a long time before receiving the downloaded event from nativeUpdater. It’s currently unclear why the Squirrel server takes more than 10 minutes to pipe the update.zip file.

Solution: Upon reviewing the code, I discovered https://github.com/electron-userland/electron-builder/pull/5271, which altered the timing of the updater-downloaded event. Now, the updater dispatches the updater-downloaded event to the app layer as soon as the zip file is downloaded, but in reality, quitAndInstall only takes immediate effect when the updater-downloaded event from nativeUpdater is triggered. Therefore, I believe we should align the dispatch timing of the update-downloaded event with nativeUpdater by reverting the changes related to the dispatch timing in [PR 5271](https://github.com/electron-userland/electron-builder/pull/5271). This should also resolve the issue we’re currently encountering.

mmaietta:
> It’s currently unclear why the Squirrel server takes more than 10 minutes to pipe the update.zip file.

I feel like this would be the next steps in terms of investigation. I wouldn't expect a localhost server to be noticably limited by bandwidth.

I'm hesitant to move forward with this change. From what I understand, the reason why the update-downloaded event happens after file download and not via the nativeUpdater is to allow this functionality in https://github.com/electron-userland/electron-builder/pull/5271
> to be able to do an additional cryptographic check on downloaded files and conditionally abort the update on a failed verification.

If I understand your need correctly, I think you can attach to the native event to have separate logic instead; I think you can use some code like below. (needs the `as any` since `nativeUpdater` is currently private on MacUpdater)
```
if (autoUpdater instanceof MacUpdater) {
[...]

PR #8652, opened 2024-11-01
fix: keep the dispatch timing of update-downloaded events consistent with nativeUpdater

DamonYu6:
fix https://github.com/electron-userland/electron-builder/issues/8651

### DIFF
--- a/.changeset/spotty-toys-ring.md
+++ b/.changeset/spotty-toys-ring.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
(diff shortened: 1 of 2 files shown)


beyondkmp:
LGTM

beyondkmp (approved):

mmaietta:
Let's continue conversing here https://github.com/electron-userland/electron-builder/issues/8651#issuecomment-2452342734 I'm currently opposed to this change and believe the proper investigation should be as to why the mac localhost server is being so heavily bandwidth constrained.

mmaietta:
Closing per conversation in #8651

Issue #8345, opened 2024-07-17
Builder in 25.0.0 finds files outside project and throws error


blakebyrnes:
25.0.0

20.11.0

31.2.1 (current)

All, but testing on mac

Building in 24.x didn't have this problem. I tried to upgrade and have an issue where the filter is throwing an error about files it finds even though they're outside the directory where the build is happening (and also have all the dependencies set to no-hoist and that also didn't work)

```
      "**/electron-builder",
      "**/app-builder-lib",
      "**/builder-util",
      "**/builder-util-runtime",
      "**/dmg-builder",
```

Here's the error stack for one iteration (I've tried a number of ways to do this. This stack is before I did nohoist. The question is: How can I set a relative directory for the file filter not to go above?
```
<PATH>/platform/build-dist/cloud/main/.env.cloud.defaults must be under <PATH>/platform/build-dist/desktop/main/app/
at getRelativePath (<PATH>/platform/node_modules/app-builder-lib/src/util/filter.ts:32:13)
    at AsarPackager.unpackPattern (<PATH>/platform/node_modules/app-builder-lib/src/util/filter.ts:57:22)
    at AsarPackager.createPackageFromFiles (<PATH>/platform/node_modules/app-builder-lib/src/asar/asarUtil.ts:130:82)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at AsarPackager.pack (<PATH>/platform/node_modules/app-builder-lib/src/asar/asarUtil.ts:50:41)
    at <PATH>/platform/node_modules/app-builder-lib/src/platformPackager.ts:439:11
    at async Promise.all (index 0)
    at AsyncTaskManager.awaitTasks (<PATH>/platform/node_modules/builder-util/src/asyncTaskManager.ts:65:25)
    at MacPackager.doPack (<PATH>/platform/node_modules/app-builder-lib/src/platformPackager.ts:293:5)
    at MacPackager.pack (<PATH>/platform/node_modules/app-builder-lib/src/macPackager.ts:201:9)
    at Packager.doBuild (<PATH>/platform/node_modules/app-builder-lib/src/packager.ts:458:9)
    at executeFinally (<PATH>/platform/node_modules/builder-util/src/promise.ts:12:14)
    at Packager.build (<PATH>/platform/node_modules/app-builder-lib/src/packager.ts:392:31)
    at executeFinally (<PATH>/platform/node_modules/builder-util/src/promise.ts:12:14)```

[...]

PR #8654, opened 2024-11-01
fix(asar): check ResolvedFileSet src when verifying symlinks to be within project directory

mmaietta:
Fixes https://github.com/electron-userland/electron-builder/issues/8345


beyondkmp:
Do we need to add ut for this issue?

mmaietta:
I'm in favor, I just couldn't identify a repro setup for the issue as none was supplied in the GH issue.

Issue #8660, opened 2024-11-04
ASAR files in `extraResources` are not included in integrity calculations


t3chguy:
* **Electron-Builder Version**: 25.1.8

* **Node Version**: v23.1.0

* **Electron Version**: ^33.0.0
* **Electron Type (current, beta, nightly)**: current

* **Target**: macOS & Windows

electron-builder does a seemingly fine job supporting `EnableEmbeddedAsarIntegrityValidation` unless your `extraResources` contains additional `asar` files. The files are not considered in the ASAR integrity resolution at https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/platformPackager.ts#L305-L310 as the `extraResources` are not copied until https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/platformPackager.ts#L318 

Empirically the `computeData` function is never called for mac universal packages so the fix there would need to be different.

t3chguy:
Still an issue AFAICT

mmaietta:
I migrated electron-builder to `electron/asar` that now handles calculating file integrity. Can you please try `next` tag of electron-builder?

mmaietta:
`ElectronAsarIntegrity` is only for `asar` as the name suggests. Unpacked files are also calculated in the asar integrity IIRC as they are tracked internally within the asar header.
Extra resources (within the context of electron-builder) on the other hand are supplemental to the asar and exist outside an asar's domain boundary. 

What are these?
```
~/D/macos  ./Element.app/Contents/MacOS/Element --profile TTTT                                                                                            

/Users/t3chguy/Library/Application Support/Element-TTTT exists: no
/Users/t3chguy/Library/Application Support/Riot-TTTT exists: no
[51380:0122/133125.917907:FATAL:archive.cc(239)] Failed to get integrity for validatable asar archive: Resources/webapp.asar
```

Can you provide a sample repo for this that I could debug deeper with? AFAICT currently, there is no issue here per the reasoning I gave above.

t3chguy:
> ElectronAsarIntegrity is only for asar as the name suggests.

Indeed, but one can have multiple asars. The only rule with asar integrity is that the electron main process code lives in `app.asar`. ASAR integrity checks affect all loaded ASARs though, not just `app.asar`.

[...]

PR #8661, opened 2024-11-04
refactor: remove stale handler for `extend-info`

t3chguy:
The `extend-info` read was removed in a23e68066b3912110024d219507f5691fd14e450 so this if statement never runs


mmaietta:
Great catch. How did you determine that it was stale?
`decode-plist` is an undocumented function in https://github.com/develar/app-builder but AFAICT, it shouldn't be returning any entry at that array index either
https://github.com/develar/app-builder/blob/d971224691f6f48555e2a9bb327f99627893372c/pkg/plist/plist.go#L68-L70

t3chguy:
> Great catch. How did you determine that it was stale?

I was looking at ways to work around https://github.com/electron-userland/electron-builder/issues/8660 by injecting my own data into the asar without it being nuked by the asar integrity calculation and thought `extend-info` was my golden ticket. Sadly not.

Prior to the linked commit `extend-info` was a value on `buildMetadata` (`(buildMetadata as any)["extend-info"]`) but there is no longer such a key anywhere else in the codebase and indeed the undocumented function returns the same amount as you put in.

mmaietta:
Are you referring to this `extendInfo` functionality?
```
mac: {
  // test appId per platform
  appId: "foo",
  extendInfo: {
    LSUIElement: true,
    CFBundleDocumentTypes: [
      {
        CFBundleTypeName: "Folders",
        CFBundleTypeRole: "Editor",
        LSItemContentTypes: ["public.folder"],
      },
    ],
  },
}
```

t3chguy:
Yes, sadly the ASAR integrity thing just overwrites the field rather than merging into it

PR #8695, opened 2024-11-19
fix: use disableDifferentialDownload flag in the AppImageBuilder

peter-sanderson:
Hi, this is my first contribution here so please LMK if I did something wrong. 

- I have a quite trouble running the tests. It produces a lot of errors that seem unrelated.
- I was able to run `pnpm compile && TEST_FILES=linuxUpdaterTest.ts pnpm ci:test` and it passed
- It also seems there is missing a test for AppImage (`test AppImage download`) is commented and not working

### DIFF
--- a/.changeset/odd-trees-battle.md
+++ b/.changeset/odd-trees-battle.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix: respect `disableDifferentialDownload` flag for AppImage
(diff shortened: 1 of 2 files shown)


mmaietta:
Quick update: PR looks good! Just to double-check, did you test the full update process locally with a packaged app on your dev machine? Or was it only the unit tests? I need to double check if there's a differential update unit test for linux

Issue #8698, opened 2024-11-20
Linux AppImage update unable to move downloaded file


flexopticalHenno:
* **Electron-Builder Version**: 25.0.4 (also 25.1.7 / 25.1.8)

* **Node Version**: 20.16.0 / 18.20.5

* **Electron Version**: 32.1.2 / 32.2.5
* **Electron Type (current, beta, nightly)**: current

* **Electron-Updater Version**: 6.3.3 (also 6.3.9)

* **Target**: AppImage

Building the app works fine, but when the app finds an update, it downloads it to the cache folder, removes the old AppImage and then fails to move the downloaded file to the original AppImage location. The error states (translated) "mv: Executing stat not possible for '/home/username/.cache/app-name-updater/pending/APP Name.1.2.3-latest.4.AppImage': File not found". Copying the failed command into a terminal moves that file successfully. 

Stack trace includes AppImageUpdater.doInstall in AppImageUpdater:88:42 and AppImageUpdater.install in BaseUpdater.js:61:25 and AppImageUpdater.quitAndInstall in BaseUpdater.js:15:34

The app name contains spaces, so the issue may be related to the fix that was introduced in https://github.com/electron-userland/electron-builder/commit/1c14820b97fad802b300dd93ccd4d6a10a72360f

The update was still working with electron-builder 25.0.3 and electron-updater 6.3.2. Tested on Ubuntu 24.04 and Debian 12 VMs.

beyondkmp:
This should be caused by  https://github.com/electron-userland/electron-builder/blob/6a6bed46c428b45105ada071a9cb89b5d4f93d9e/packages/electron-updater/src/AppImageUpdater.ts#L96

flexopticalHenno:
I don't consider this issue to be stale because the linked pull-request is not merged or released yet

pierremtb:
Hey @flexopticalHenno thanks for bumping this. I believe we ended up closing the wrong issue last time. This is a duplicate to https://github.com/KittyCAD/modeling-app/issues/4474, which we worked around in https://github.com/KittyCAD/modeling-app/pull/4708 but never really fixed it long term.

I'll have a closer look and see what ticket is best fit to remain open.

PR #8706, opened 2024-11-23
fix: use renameSync instead of mv

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8698

### DIFF
--- a/packages/electron-updater/src/AppImageUpdater.ts
+++ b/packages/electron-updater/src/AppImageUpdater.ts
@@ -1,7 +1,7 @@
 import { AllPublishOptions, newError } from "builder-util-runtime"
 import { execFileSync } from "child_process"
 import { chmod } from "fs-extra"
-import { unlinkSync } from "fs"
+import { unlinkSync, renameSync } from "fs"
 import * as path from "path"
 import { DownloadUpdateOptions } from "./AppUpdater"
 import { BaseUpdater, InstallOptions } from "./BaseUpdater"
@@ -93,7 +93,7 @@ export class AppImageUpdater extends BaseUpdater {
       destination = path.join(path.dirname(appImageFile), path.basename(options.installerPath))
     }
 
-    execFileSync("mv", ["-f", options.installerPath, destination])
+    renameSync(options.installerPath, destination)
     if (destination !== appImageFile) {
       this.emit("appimage-filename-updated", destination)
     }
--- a/packages/electron-updater/src/DebUpdater.ts
+++ b/packages/electron-updater/src/DebUpdater.ts
@@ -29,10 +29,8 @@ export class DebUpdater extends BaseUpdater {
 
   protected doInstall(options: InstallOptions): boolean {
     const sudo = this.wrapSudo()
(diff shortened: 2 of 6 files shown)


imconnorngl (approved):
We desperately need this fix. Our Linux users are having to manually move the App images our auto updater downloads

FxMorin (approved):

packages/electron-updater/src/AppImageUpdater.ts
```diff
@@ -93,7 +93,7 @@ export class AppImageUpdater extends BaseUpdater {
       destination = path.join(path.dirname(appImageFile), path.basename(options.installerPath))
     }
 
-    execFileSync("mv", ["-f", options.installerPath, destination])
+    renameSync(options.installerPath, destination)
```
mmaietta, line 96: This doesn't seem like it would work due to `installerPath` being a space-escaped string for piping to the cmd shell
> If the platform is Linux, replace spaces with '\ ' for shell compatibility

https://github.com/electron-userland/electron-builder/blob/6a6bed46c428b45105ada071a9cb89b5d4f93d9e/packages/electron-updater/src/BaseUpdater.ts#L52-L60

beyondkmp: Yes, I found that this was the cause, so I changed it to draft.

I don't think it's handled well here - it shouldn't directly add /. Instead, it should be handled by each deb/rpm/pacman separately. 

These linux install commands use string concatenation, so if malicious code is added to the name, it could lead to serious security vulnerabilities.

https://github.com/electron-userland/electron-builder/blob/6a6bed46c428b45105ada071a9cb89b5d4f93d9e/packages/electron-updater/src/PacmanUpdater.ts#L34-L35

packages/electron-updater/src/DebUpdater.ts
```diff
@@ -29,10 +29,8 @@ export class DebUpdater extends BaseUpdater {
 
   protected doInstall(options: InstallOptions): boolean {
     const sudo = this.wrapSudo()
-    // pkexec doesn't want the command to be wrapped in " quotes
-    const wrapper = /pkexec/i.test(sudo) ? "" : `"`
     const cmd = ["dpkg", "-i", options.installerPath, "||", "apt-get", "install", "-f", "-y"]
-    this.spawnSyncLog(sudo, [`${wrapper}/bin/bash`, "-c", `'${cmd.join(" ")}'${wrapper}`])
+    this.spawnSyncLog(sudo, cmd)
```
mmaietta, line 33: Just to double check, was this tested? (Same for pacman and Rpm)

Pretty sure this will fail as my previous project required piping it to bash

louisroyer, line 33: Imo, this way of installing local `.deb` in 2 steps with `dpkg` and `apt-get install -f` can be updated. Apt is now able to install local `.deb` and depends in a single command if the package argument is an absolute path to a `.deb`.

You may now do `this.spawnSyncLog("sudo", ["apt-get", "install",  "-y", options.installerPath])`. This will avoid you to pipe commands, or to split it into 2 commands and checking return code.

PR #8711, opened 2024-11-27
feat(keygen): Add `host` property to support self-hosted Keygen instances

### LINKED ISSUE #8704: Add support for selfhosted Keygen
* **Electron-Builder Version**: v25.1.8

* **Node Version**: v22.3.0

* **Electron Version**: 30.0.9

* **Electron Updater Version**: 6.3.9 

* **Target**: MacOS

Hi,
we're using a [selfhosted Keygen](https://keygen.sh/docs/self-hosting/), however, the server hostnames are hardcoded [here](https://github.com/electron-userland/electron-builder/blob/6a6bed46c428b45105ada071a9cb89b5d4f93d9e/packages/electron-publish/src/keygenPublisher.ts#L83) and [here](https://github.com/electron-userland/electron-builder/blob/6a6bed46c428b45105ada071a9cb89b5d4f93d9e/packages/electron-updater/src/providers/KeygenProvider.ts#L19).

I'd like to be able to set this in the `package.json` or the `electron-builder.json`, that would be great!

hrueger:
fixes #8704 

I haven't tested this, though. Is there anything else I'd have to modify e.g. docs?

Thanks in advance!


packages/builder-util-runtime/src/publishOptions.ts
```diff
@@ -179,6 +179,11 @@ export interface KeygenOptions extends PublishConfiguration {
    */
   readonly provider: "keygen"
 
+  /**
+   * Keygen host
```
mmaietta, line 183: Let's add more details to this documentation with a default value.
```
Keygen host for self-hosted instances
@default "api.keygen.sh"
```

packages/electron-updater/src/providers/KeygenProvider.ts
```diff
@@ -16,7 +16,7 @@ export class KeygenProvider extends Provider<UpdateInfo> {
       ...runtimeOptions,
       isUseMultipleRangeRequest: false,
     })
-    this.baseUrl = newBaseUrl(`https://api.keygen.sh/v1/accounts/${this.configuration.account}/artifacts?product=${this.configuration.product}`)
+    this.baseUrl = newBaseUrl(`https://${this.configuration.host || "api.keygen.sh"}/v1/accounts/${this.configuration.account}/artifacts?product=${this.configuration.product}`)
```
mmaietta, line 19: Minor: Let's extract this to a separate variable since it's embedding logic in a template string.
```
const host = this.configuration.host || "api.keygen.sh"
this.baseUrl = newBaseUrl(`https://${host}/v1/accounts/${this.configuration.account}/artifacts?product=${this.configuration.product}`)
```

.changeset/kind-poems-joke.md
```diff
@@ -0,0 +1,8 @@
+---
+"app-builder-lib": minor
+"builder-util-runtime": minor
+"electron-publish": minor
+"electron-updater": minor
+---
+
+add host option for keycloak provider
```
mmaietta, line 8: What is `keycloak`? Maybe this could be changed to "Add `host` property for self-hosted Keygen instances

hrueger: Ups, `keycloak` is some other software I patched just before this pr 😅 So I mixed up the names ;-)

hrueger:
Hi @mmaietta,
thanks for the review, I fixed the few points 👍

mmaietta (approved):

Issue #7103, opened 2022-08-31
Does not work with NPM workspaces


ghost:
* **Electron-Builder Version**: 23.3.3

* **Node Version**: v18.3.0

* **Electron Version**: 20.1.0

* **Electron Type (current, beta, nightly)**: current

* **Target**:  win32 x64

electron builder does not work in npm workspace
it deletes app-builder-bin during "installing production dependencies" and then throws error:
```
• packaging       platform=win32 arch=x64 electron=20.1.0 appOutDir=dist\win-unpacked
The system cannot find the path specified.
  ⨯ spawn C:\Users\user\dev\test-workspace\node_modules\app-builder-bin\win\x64\app-builder.exe ENOENT
```
 also it deletes more packages from node_modules, so next run throws
 ```
'electron-builder' is not recognized as an internal or external command
```
Reproduction:
` npm run build -w a` in [test-workspace.zip](https://github.com/electron-userland/electron-builder/files/9458943/test-workspace.zip)

**Comment by paulantoine2** [comment]
For me, i just get stuck at `installing production dependencies`but the behavior stay the same.

However, it's working with electron-builder version 22 !

**Comment by RadoslavK** [comment]
I am having the same or similar problem and found it started happening with npm 8.5.0 (node 16.14.1) because when I downgrade to 8.3.1 (that came with node 16.14.1) while keeping a higher node version, it works. Npm changelog says something about workspaces but I am not sure if its related.

**Comment by Hinaser** [comment]
Try to add build option below in your `package.json` which has `main` entry for electron app:
```javascript
{
  "main": "build/electron/main.js",
  ...,
  "build": {
    ...,
    "npmRebuild": false
  }
}
```

joshtynjala:
Confirmed that setting `npmRebuild` to `false` works for my project, but I suspect it is not a universal workaround.

It looks like this code is where the _installing production dependencies_ happens:

https://github.com/electron-userland/electron-builder/blob/01c679107435c6afd0b2de8c44d3f140d20c5577/packages/app-builder-lib/src/util/yarn.ts#L86-L120

It seems to be running `node install --production`. I'm guessing that this command is deleting dependencies needed by electron-builder, including the mentioned **app-builder-bin**.

nanamicat:
setting npmRebuild to false works for me, but I don't think it's a suitable solution.
another problem is when depend on a non-exact electron version like `^23.1.0`, it can't find installed electron package and will throw.

[...]

PR #8715, opened 2024-12-01
fix: install all dependencies to fix building within NPM workspaces

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/7103.

If node_modules doesn't exist at all, I think we can directly install all packages, not just the production ones. Even if we  reinstall all packages under the workspace again, it shouldn't cause any issues. 

At most, this approach would just take some extra time but won't lead to any problems.

### DIFF
--- a/.changeset/small-windows-leave.md
+++ b/.changeset/small-windows-leave.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: does not work with NPM workspaces
--- a/packages/app-builder-lib/src/util/yarn.ts
+++ b/packages/app-builder-lib/src/util/yarn.ts
@@ -100,7 +100,6 @@ async function installDependencies(config: Configuration, appDir: string, option
     if (process.env.NPM_NO_BIN_LINKS === "true") {
       execArgs.push("--no-bin-links")
     }
-    execArgs.push("--production")
   }
 
   if (!isRunningYarn(execPath)) {


mmaietta:
Is there a test project I can verify locally or unit test we could set up for this? (If it's too complex, then no worries, just wanted to double check though)

mmaietta:
Can you check this? The main issue I think I see here is with the default configuration of electron-builder, where it includes all files in the project directory https://www.electron.build/contents#files. This means for projects already using electron-builder, this could lead to a dramatic increase in app size if they are relying on electron-builder to install their app dependencies as this PR will cause non-production dependencies to be packaged with it.

beyondkmp:
Here's a test demo: https://github.com/electron-userland/electron-builder/files/9458943/test-workspace.zip.

beyondkmp:
> Can you check this? The main issue I think I see here is with the default configuration of electron-builder, where it includes all files in the project directory https://www.electron.build/contents#files. This means for projects already using electron-builder, this could lead to a dramatic increase in app size if they are relying on electron-builder to install their app dependencies as this PR will cause non-production dependencies to be packaged with it.

This shouldn't increase, when we copy node modules to asar, we filter for production dependencies and only copy those. Dev dependencies won't be copied, so it won't have any impact even if everything is installed.

mmaietta:
Nice work!

Issue #8705, opened 2024-11-22
electron-builder fails when list of node_modules files is too big to pass in a glob


abram:
This recent change causes electron-builder to fail in my project:

https://github.com/electron-userland/electron-builder/blob/6a6bed46c428b45105ada071a9cb89b5d4f93d9e/packages/app-builder-lib/src/asar/asarUtil.ts#L49

It attempts to take a potentially large file list and put it into a glob pattern string. This causes an error due to exceeding a hard coded limit in minimatch:
```
pattern is too long
    at assertValidPattern (node_modules\minimatch\minimatch.js:280:11)
    at minimatch (node_modules\minimatch\minimatch.js:116:3)
    at isUnpackedDir (node_modules\@electron\asar\src\asar.ts:24:41)
    at handleFile (node_modules\@electron\asar\src\asar.ts:148:17)
    at next (node_modules\@electron\asar\src\asar.ts:185:5)
    at AsarPackager.executeElectronAsar (node_modules\app-builder-lib\src\asar\asarUtil.ts:77:5)
    at AsarPackager.pack (app-builder-lib\src\asar\asarUtil.ts:51:5)
```

You can see the limit code here:
https://github.com/isaacs/minimatch/blob/main/src/assert-valid-pattern.ts#L1

I'm not familiar with the asar module, but there must be a better way to pass a file list than using a giant glob.

beyondkmp:
https://github.com/electron-userland/electron-builder/pull/8645 can fix this issue. It only add the node_modules/xxx into unpackedPaths intead of all sub directories in node_modules/xxx.

beyondkmp:
@abram This #8645 has been merged into [v26.0.0-alpha.7](https://github.com/electron-userland/electron-builder/releases/tag/v26.0.0-alpha.7).  Can you help try it?

abram:
Thanks so much @beyondkmp! I can confirm that your fix is working for me.

PR #8725, opened 2024-12-06
fix: electron-builder fails when list of node_modules files is too big to pass in a glob

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8705

**root cause**

When listing all files, if there are too many files, it will trigger glob's length limit.

**how to fix**

If all files in a folder are unpacked, use the folder name directly instead of listing all files.


beyondkmp:
The main issue in this MR is caused by users directly configuring the entire node_modules in the asarUnpack pattern.

asarUnpack supports '!', but electron/asar doesn't support this '!'. This problem will occur whenever users have directories containing a large number of files.

My thought is whether electron-builder's asarUnpack should also not support '!' patterns, keeping consistent with electron/asar. @mmaietta  Do you have any good ideas or suggestions?

mmaietta:
Is there a way we can resolve `!` filters before passing to electron/asar's minimatch?

It seems like a pretty hefty decrease in electron-builder functionality if we remove the `!` filter feature?

packages/app-builder-lib/src/asar/asarUtil.ts
```diff
@@ -153,13 +154,24 @@ export class AsarPackager {
...
 
         if (taskManager.tasks.length > MAX_FILE_REQUESTS) {
           await taskManager.awaitTasks()
         }
       }
+
+      if (tmpUnpackedPaths.size === fileSet.files.length) {
+        const relative = path.relative(this.config.defaultDestination, fileSet.destination)
+        const destination = path.resolve(this.rootForAppFilesWithoutAsar, relative)
+        unpackedPaths.add(`${destination}${path.sep}**`)
```
mmaietta, line 168: Can you clarify this area of code with a comment? It looks like we're determining whether all files in the `fileSet` are being unpacked, if so, then we glob them together with an umbrella glob `destination/**` to reduce the number of unpacked paths that are being passed to `minimatch`?

beyondkmp: Yes, exactly as you mentioned.

electron/asar doesn't implement `!` filtering. Having to do the conversion at the electron-builder layer is very complicated, and there's currently no good method for conversion.

Therefore, we can only solve this problem from a different angle. Exceeding the length limit is always caused by filtering one or more directories. If we're just filtering specific files, it won't exceed the length limit. So solving these directory filtering issues would solve this problem.

Currently, there might still be some edge cases, such as when a directory contains tens of thousands of files, and the requirement is to put just a few files from this directory into asar while putting all other files into asar.unpack.
In practical use, this situation probably won't occur - it's rare to have a directory with tens of thousands of files where only a few files need to be in asar.

mmaietta: For `unpackedPaths.add(`${destination}${path.sep}**`)`, is it possible to use `${relative}${path.sep}**` to shorten the minimatch paths even further? I can't recall off the top of my head whether it requires absolute paths or if relative works.

beyondkmp: Good idea. However, we need to use `**${path.sep}${relative}${path.sep}**`. I've changed files to use this format as well, and current testing shows no issues.

mmaietta: Hmmm, why do we need to add the `**` prefix? That seems like it could accidentally glob more than expected.

Looking at electron/asar, it only is minimatching the relative path of the files `path.relative(src, ...)`
https://github.com/electron/asar/blob/76064f4aa6e0c17b01546b209be43e88e8007dd0/src/asar.ts#L161-L181

beyondkmp: Yes, directories are matched using relative paths. I have modified it and removed the relative path matching for files. Since specific files are matched using absolute paths, you used absolute paths to match files at that time.

https://github.com/electron/asar/blob/main/src/asar.ts#L151C34-L151C42

mmaietta:
This is really elegant, well done!

mmaietta (approved):

PR #8767, opened 2025-01-12
add more extra component packages for macos archive

dominhhai:
Dear maintainer!

The product archive of macos is supported more component packages.
I add this option to the pkg builder.

Thanks for your supporting!

### DIFF
--- a/packages/app-builder-lib/src/options/pkgOptions.ts
+++ b/packages/app-builder-lib/src/options/pkgOptions.ts
@@ -126,6 +126,19 @@ export interface PkgOptions extends TargetSpecificOptions {
    * @default upgrade
    */
   readonly overwriteAction?: "upgrade" | "update" | null
+
+  /**
+  * Add extra component packages for MacOs product archive.
+  * The extra component packages directory, relative to `build` (build resources directory).
+  * Example: `pkg-components`.
+  */
+  readonly extraPkgsDir?: string | null
+
+  /**
+  * The list of extra component packages, relative to `extraPkgsDir`.
+  * Example: `['virtual-audio-driver.pkg', 'obs-lib.pkg']`.
+  */
+  readonly extraPkgs?: Array<string> | null
 }
 
 /**
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -5144,6 +5144,27 @@
             "null",
             "string"
           ]
+        },
+        "extraPkgsDir": {
+          "description": "Add extra component packages for MacOs product archive.\nThe extra component packages directory, relative to `build` (build resources directory).\nExample: `pkg-components`.",
+            "type": [
+              "null",
+              "string"
+            ]
+        },
+        "extraPkgs": {
+          "anyOf": [
+            {
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
+            },
+            {
+              "type": "null"
+            }
+          ],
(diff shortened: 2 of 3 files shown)


packages/app-builder-lib/src/options/pkgOptions.ts
```diff
@@ -126,6 +126,19 @@ export interface PkgOptions extends TargetSpecificOptions {
...
+  * The extra component packages directory, relative to `build` (build resources directory).
+  * Example: `pkg-components`.
+  */
+  readonly extraPkgsDir?: string | null
+
+  /**
+  * The list of extra component packages, relative to `extraPkgsDir`.
+  * Example: `['virtual-audio-driver.pkg', 'obs-lib.pkg']`.
+  */
+  readonly extraPkgs?: Array<string> | null
```
mmaietta, line 141: Do we need to supply both `extraPkgsDir` and `extraPkgs`? I think we could just have `extraPkgsDir` and then just scan the dir for `.pkg` files (I can help with writing that code).
What are your thoughts? Wanted to double check the intent of having two properties.

dominhhai: @mmaietta 
LGTM.
Could you write that code?

mmaietta: Updated the logic and added a unit test for the packaging step. Can you take a look?

Is there a way to verify that a `.pkg` extra package was included in the `.pkg`?

dominhhai: @mmaietta 
LGTM!
Thanks for your supporting!

> Is there a way to verify that a .pkg extra package was included in the .pkg?

No, because of the final product archive can not be extracted by `pgkutitl` command.
Just believe Apple :)

mmaietta: Any chance you can link me to a HelloWorld component package or something similar I can test with? Even a short tutorial to generate one locally would be helpful. For some reason a google search isn't being fruitful.

Turns out, you can decompress a `.pkg` via command line `xar -xf electron-quick-start-typescript-1.0.4-arm64.pkg` and there's a file inside that you can read the pkg contents from.

I tried creating a test by reusing an electron-builder-built app `.pkg` and (unsurprisingly) it failed with the following error since the package-to-embed was naturally not an actual component package
```
Exit code: 1. Command failed: productbuild --synthesize --component /Users/mikemaietta/Development/electron-builder-test/dist/mac-arm64/electron-quick-start-typescript.app --package /Users/mikemaietta/Development/electron-builder-test/build/extra-packages/electron-quick-start-typescript-1.0.4-arm64.pkg /Users/mikemaietta/Development/electron-builder-test/dist/distribution.xml
productbuild: error: "/Users/mikemaietta/Development/electron-builder-test/build/extra-packages/electron-quick-start-typescript-1.0.4-arm64.pkg" is a product archive, not a component package.

productbuild: Adding component at /Users/mikemaietta/Development/electron-builder-test/dist/mac-arm64/electron-quick-start-typescript.app
productbuild: Inferred install-location of /Users/mikemaietta/Development/electron-builder-test/dist/mac-arm64

productbuild: error: "/Users/mikemaietta/Development/electron-builder-test/build/extra-packages/electron-quick-start-typescript-1.0.4-arm64.pkg" is a product archive, not a component package.
```

dominhhai: @mmaietta 
you can use BlackHold Installer file here to test
https://existential.audio/blackhole/download/download2ch.php?code=761608887

PR #8785, opened 2025-01-19
feat: Allow users to pass a custom headers URL

lamawithonel:
Prior to this `packages/app-builder-lib/src/util/yarn.ts` hard-coded the Electron headers `disturl` with no option to configure it for a local mirror.  This commit changes it to try the environment variable `npm_config_electron_headers_disturl` before using the hard-coded value. This lets users set the variable
`electron_headers_disturl=https://example.com/custom-headers-mirror` to use a local mirror.

### DIFF
--- a/packages/app-builder-lib/src/electron/ElectronFramework.ts
+++ b/packages/app-builder-lib/src/electron/ElectronFramework.ts
@@ -48,12 +48,6 @@ export interface ElectronDownloadOptions {
    */
   mirror?: string | null
 
-  /**
-   * The mirror base URL for Electron headers.
-   * @default https://electronjs.org/headers
-   */
-  headersMirror?: string | 'https://electronjs.org/headers'
-
   /** @private */
   customDir?: string | null
   /** @private */
--- a/packages/app-builder-lib/src/electron/ElectronFramework.ts
+++ b/packages/app-builder-lib/src/electron/ElectronFramework.ts
@@ -48,6 +48,12 @@ export interface ElectronDownloadOptions {
    */
   mirror?: string | null
 
+  /**
+   * The mirror base URL for Electron headers.
+   * @default https://electronjs.org/headers
+   */
+  headersMirror?: string | 'https://electronjs.org/headers'
+
   /** @private */
   customDir?: string | null
   /** @private */
--- a/packages/app-builder-lib/src/util/yarn.ts
+++ b/packages/app-builder-lib/src/util/yarn.ts
@@ -11,8 +11,6 @@ import { Configuration } from "../configuration"
 import { executeAppBuilderAndWriteJson } from "./appBuilder"
 import { NodeModuleDirInfo } from "./packageDependencies"
 import { rebuild as remoteRebuild } from "./rebuild/rebuild"
-import { ElectronDownloadOptions } from "../electron/ElectronFramework"
-import { getConfig } from "../util/config/config"
 
 export async function installOrRebuild(config: Configuration, appDir: string, options: RebuildOptions, forceInstall = false) {
   const effectiveOptions: RebuildOptions = {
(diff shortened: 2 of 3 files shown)


packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -71,7 +71,7 @@ export function getGypEnv(frameworkInfo: DesktopFrameworkInfo, platform: NodeJS.
   // https://github.com/nodejs/node-gyp/issues/21
   return {
     ...common,
-    npm_config_disturl: "https://electronjs.org/headers",
+    npm_config_disturl: common.npm_config_electron_headers_disturl || "https://electronjs.org/headers",
```
mmaietta, line 74: `npm_config_electron_headers_disturl` is this a standard env var? Or custom for electron-builder? Wondering if it needs to be explicitly documented on our docs site

lamawithonel: It would be custom for electron-builder.  I considered using using `npm_config_disturl`, but it might conflict if users need to download other headers with npm-gyp.  Namespacing like this guards against conflicts.

mmaietta: Hmmm, so electron-builder migrated to use `@electron/rebuild` and `@electron/asar` for handling dependency rebuilds, which they had internally migrated to `@electron/node-gyp`. The only locations that this PR code within `getGypEnv` are used are when fresh-installing a project (e.g. `yarn/npm install`), when using `nativeRebuilder: legacy` config, or when config property `nodeGypRebuild: true` is set.

In all cases, those are operations on the entire codebase, so I'm not sure I understand how it would conflict with other headers being downloaded by node-gyp. Could you elaborate?

I ask as I'm wondering if we could just do this instead. I recognize your desire to guard against conflicts but adding additional env vars is probably the easiest way for the project to collect tech debt and make documentation more complex?
I'm suggesting allow overriding the disturl but have a default fallback so the user doesn't necessarily have to supply a value.
```
npm_config_disturl: common.npm_config_disturl || "https://electronjs.org/headers"
```

lamawithonel: Ah, ok.  I was under the impression it could be used for other things.

Looking at the config interface (`electron-builder.yml`) a little more, how about I create the config key `electronDownload.headersMirror` instead?  That would put it right alongside the electron binary download URL.

packages/app-builder-lib/src/electron/ElectronFramework.ts
```diff
@@ -49,6 +49,12 @@ export interface ElectronDownloadOptions {
    */
   mirror?: string | null
 
+  /**
+   * The mirror base URL for Electron headers.
+   * @default https://electronjs.org/headers
+   */
+  headersMirror?: string | 'https://electronjs.org/headers'
```
mmaietta, line 56: This feels like a good place to put it, but this object is passed directly as a JSON string to upstream dependency `app-builder-bin`
https://github.com/electron-userland/electron-builder/blob/443ee8debd4e4d73d9f63a31fb93ce1dbed2538c/packages/app-builder-lib/src/electron/ElectronFramework.ts#L154-L155
https://github.com/electron-userland/electron-builder/blob/443ee8debd4e4d73d9f63a31fb93ce1dbed2538c/packages/app-builder-lib/src/electron/ElectronFramework.ts#L215

The `ElectronDownloadOptions` are supposed to be specific to https://github.com/electron/get, so I'll need to think more on this feature implementation on how we can make it easy to understand for developers without convoluting the intent of each configuration property

mmaietta:
@lamawithonel can you try setting your mirror value to env var `NPM_CONFIG_ELECTRON_MIRROR` or `npm_config_electron_mirror`.

mmaietta:
Honestly, I'd love to just migrate the entire logic to `@electron/get` but it looks like the API would require some breaking changes and we just bumped the major semver to v26.

I'll prep a PR with the changes so that we have a draft PR for the v27 (which will be timed with `@electron` upstream's upgrade to minimum node 22 that is occurring in the near future)

lamawithonel:
Changing the version here doesn't work quite how I thought it would.  There is still [a line in @electron/rebuild](https://github.com/electron/rebuild/blob/main/src/rebuild.ts#L164) that fails when it's changed.  The calling code from my `package.json` looks like this:

```json
  "scripts": {
    "ci:install": "ELECTRON_SKIP_BINARY_DOWNLOAD=1 npm install",
    "postinstall": "electron-builder install-app-deps",
```

Issue #8798, opened 2025-01-22
`electronFuses` doesn't work for universal builds


t3chguy:
The older approach using `afterPack` still works fine.

I'm wagering that this is due to the fuses being enabled on the 2 sub-packages, and never on the universal package.

`PlatformPackager::doPack` should have a flag akin to `disableAsarIntegrity` e.g. `disableFuses` which should skip `addElectronFuses` and `MacPackager::doPack` in the universal branch should set disableFuses to true for the x64 & arm64 runs, then call upon `addElectronFuses` itself on the universal package

The following error can be seen when trying to use `electronFuses` for a universal package:

```
 ~/W/element-desktop   t3chguy/asarintegrity *  yarn build:universal --publish never --dir                                                                             16.7s  Wed Jan 22 18:16:14 2025
yarn run v1.22.22
$ yarn run build:ts && yarn run build:res && electron-builder --universal --publish never --dir
$ tsc
$ tsx scripts/copy-res.ts
  • electron-builder  version=26.0.0-alpha.10 os=24.1.0
  • loaded configuration  file=/Users/t3chguy/WebstormProjects/element-desktop/electron-builder.ts
  • writing effective config  file=dist/builder-effective-config.yaml
  • executing @electron/rebuild  electronVersion=34.0.0 arch=x64 buildFromSource=false appDir=./
  • installing native dependencies  arch=x64
  • completed installing native dependencies
  • packaging       platform=darwin arch=x64 electron=34.0.0 appOutDir=dist/mac-universal-x64-temp
  • executing @electron/fuses  electronPath=dist/mac-universal-x64-temp/Element.app
  • executing @electron/rebuild  electronVersion=34.0.0 arch=arm64 buildFromSource=false appDir=./
  • installing native dependencies  arch=arm64
  • completed installing native dependencies
  • packaging       platform=darwin arch=arm64 electron=34.0.0 appOutDir=dist/mac-universal-arm64-temp
  • executing @electron/fuses  electronPath=dist/mac-universal-arm64-temp/Element.app
  • packaging       platform=darwin arch=universal electron=34.0.0 appOutDir=dist/mac-universal
[...]

PR #8799, opened 2025-01-22
fix(mac): only fuse macOS universal builds on the combined universal package

t3chguy:
Without this `@electron/universal` would error complaining that the x64 and the arm64 builds varied in an unexpected place. Could have maybe overridden `addElectronFuses` in MacPackager, but we didn't have a reference to the fact we were building a universal package from the param in `doPack` and didn't want to explore adding more properties in the area.

Fixes https://github.com/electron-userland/electron-builder/issues/8798

### DIFF
--- a/.changeset/eighty-swans-hang.md
+++ b/.changeset/eighty-swans-hang.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(mac): only fuse macOS universal builds on the combined universal package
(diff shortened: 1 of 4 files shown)


t3chguy:
Running tests locally proved a pain:

`macOS Catalina doesn't support 32-bit executables and as result Wine cannot run Windows 32-bit applications too`

mmaietta:
Very elegant of a fix, thank you!

The number of args being passed to `doPack` has me worried now though, as 3 args of all booleans can easily get swapped around. If you don't mind, I'd like to add a very small refactor to this PR converting the `doPack` args into some typed object(s). Would that be okay?

t3chguy:
> The number of args being passed to doPack has me worried now though, as 3 args of all booleans can easily get swapped around. If you don't mind, I'd like to add a very small refactor to this PR converting the doPack args into some typed object(s). Would that be okay?

Sure thing

mmaietta:
No actions needed from your side. I should be able to push to this PR directly using `gh pr checkout`, I'll let you know if it doesn't work

Issue #8812, opened 2025-01-27
Broken types since v26


t3chguy:
Regressed by https://github.com/electron-userland/electron-builder/pull/8805

`node_modules/app-builder-lib/out/asar/integrity.d.ts(2,10): error TS2305: Module '"../fileMatcher"' has no exported member 'FileMatcher'.` https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/fileMatcher.ts#L41 prevents the type from being included in the resulting types

t3chguy:
> Wait what, I'm not sure I follow. Is this a compile-time error with skipLibCheck: false? I'm trying to repro but can't seem to get the error.

Yes it is that, for consumers of `lib-app-builder` anyway - such as myself needing `computeData` to rebuild asar integrity after a universal pkg.

> Not sure why @internal is flagging anything.

`@internal` means the FileMatcher symbol is omitted from the output type, but `asar/integrity.ts` now imports and relies on that type in its method signature.

t3chguy:
Looks great, will give it a test tomorrow. Thanks again

mmaietta:
Deployed 26.0.1

t3chguy:
Thanks @mmaietta can confirm that fixed it, and that the issues around asar integrity and electron fuses are both fixed in v26. Keep up the great work

PR #8813, opened 2025-01-27
fix: allow `skipLibCheck: false`

mmaietta:
Tasks:
- Extract common `undefined | null` logic everywhere to reuse (now-exported) type `Nullish`
- Expose `FileMatcher` by removing `@internal` flag (why was it internal to begin with?)
- Created base type `ObjectMap` to reduce writing `{ [key: string]: <anything> }` everywhere.
- Updated eslint to register `test/tsconfig.json`
- Fixes `skipLibCheck: false` by forcing UUID.ts to return a `Buffer` instead of `Buffer<ArrayBuffer>`, which the compiler doesn't like depending on version of node you're on

This should fix: https://github.com/electron-userland/electron-builder/issues/8812

### DIFF
--- a/.changeset/metal-bulldogs-film.md
+++ b/.changeset/metal-bulldogs-film.md
@@ -1,0 +1,8 @@
+---
+"app-builder-lib": patch
+"builder-util": patch
+"builder-util-runtime": patch
+"dmg-builder": patch
+---
+
+chore: extract common `undefined | null` to reuse current (unexported) type `Nullish`. Expose `FileMatcher` instead of `@internal` flag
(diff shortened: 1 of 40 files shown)


t3chguy:
> Created base type ObjectMap to reduce writing { [key: string]: <anything> } everywhere.

That sounds like `Record<K, V>`, no?

mmaietta:
Haha yep, that is what I was looking for. Will swap that in

mmaietta:
Turns out `Record` doesn't work with the `scheme.json` generator, but I'll find a middle ground.

Issue #8816, opened 2025-01-28
Enhance error logging in BaseUpdater


Akatroj:
Hello, I had some troubles debugging auto-updating in my app. I finally pinpointed it to the install process failing due to some misconfigurations on my machine - but I had to modify electron-updater code to add the necessary logging to debug this.

In https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/BaseUpdater.ts#L124, when you call `spawnSync` and the process exits with an error, there are no logs about this. [`spawnSync` returns an object with an optional `error` property in case of an error](https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options) - in this case, it would be good to log the error, along with stderr (maybe even throw the error).

The changes I made were basically this:

```js
if (response.error || response.status !== 0) {
  this._logger.error(response.error);
  this._logger.error(response.stderr);
}
```

IMO, this could be very helpful for other people debugging autoupdating, so I'm opening this issue

mmaietta:
Great idea. Would you be willing to open a PR with your suggestion?

Akatroj:
There are a couple nuances, which I'm not sure about:
- should we only log the error and continue the update despite it? (this left me with a bootloop when the call errors - the app tries to update, fails, restarts despite that and repeats)
- should we throw the error and map it to `this.dispatchError` at every call site (means adding 3 try-catches to RpmUpdater, which would be a code smell)
- or should we throw the error and leave it to the user (breaking change)
- does it make sense to check for non-zero return codes, or only if `error` property exists (and what even is the difference)

I'm not really sure how this should be done best, so I think it would be better done by people more familiar with the project 🥺.

Akatroj:
I created a PR anyways, but I'm not sure if it makes sense

PR #8817, opened 2025-01-29
feat(updater): Dispatch error in updater if spawnSync fails

Akatroj:
https://github.com/electron-userland/electron-builder/issues/8816


mmaietta:
@Akatroj have you tested each/any flow on a Linux device/VM for this change?

Akatroj:
I tested this change on a ubuntu device with DebUpdater (using pkexec) and with a custom updater I'm using for an internal app (basically a wrapper for DebUpdater, but instead of calling `dpkg` directly, calls a local script which calls dpkg, but can run as sudo without prompting for password).

There was no impact for happy case when updates succeed and useful logging in case it failed:
Screenshot 2025-02-04 at 10 32 35

packages/electron-updater/src/RpmUpdater.ts
```diff
@@ -32,26 +32,31 @@ export class RpmUpdater extends BaseUpdater {
...
+        cmd = [packageManager, "-y", "install", installerPath]
+      } else {
+        cmd = [packageManager, "--no-refresh", "install", "--allow-unsigned-rpm", "-y", "-f", installerPath]
+      }
+      this.spawnSyncLog(sudo, [`${wrapper}/bin/bash`, "-c", `'${cmd.join(" ")}'${wrapper}`])
+      if (options.isForceRunAfter) {
+        this.app.relaunch()
+      }
+      return true
+    } catch (e) {
```
mmaietta, line 57: Looks like we don't need to wrap this with try-catch as it's already done on in the call to `doInstall`
https://github.com/electron-userland/electron-builder/blob/bee179b3cf8163041d280ed8dc5a5ce4f27786c6/packages/electron-updater/src/BaseUpdater.ts#L65-L75

I'll push an update to this PR

PR #8843, opened 2025-02-04
fix: Only update AppArmor profile if not chroot'ed

### LINKED ISSUE #8841: Postinst script for dpkg package does not work in chroot'ed environment
This is a dup of https://github.com/signalapp/Signal-Desktop/issues/7146 which I brought up in the wrong repository.

We use a `chroot` environment to maintain our diskless clients' boot images.  The postinst script has a call

```
apparmor_parser --replace --write-cache --skip-read-cache "$APPARMOR_PROFILE_TARGET"
```

which one doesn't want to have in a `chroot` environment, and it fails for us.  This means I typically remove the call from the postinst file manually.

Changing `packages/app-builder-lib/templates/linux/after-install.tpl`'s problematic line to

```
if ! { [ -x '/usr/bin/ischroot' ] && ischroot; } && hash apparmor_parser 2>/dev/null; then
```

fixes the issue for me.

fiesh:
This prevents `postinst` script failure when run in a `chroot` environment.

This closes #8841.

### DIFF
--- a/packages/app-builder-lib/templates/linux/after-install.tpl
+++ b/packages/app-builder-lib/templates/linux/after-install.tpl
@@ -42,7 +42,10 @@ if test -d "/etc/apparmor.d"; then
   if apparmor_parser --skip-kernel-load --debug "$APPARMOR_PROFILE_SOURCE" > /dev/null 2>&1; then
     cp -f "$APPARMOR_PROFILE_SOURCE" "$APPARMOR_PROFILE_TARGET"
 
-    if hash apparmor_parser 2>/dev/null; then
+    # Updating the current AppArmor profile is not possible and probably not meaningful in a chroot'ed environment.
+    # Use cases are for example environments where images for clients are maintained.
+    # There, AppArmor might correctly be installed, but live updating makes no sense.
+    if ! { [ -x '/usr/bin/ischroot' ] && /usr/bin/ischroot; } && hash apparmor_parser 2>/dev/null; then
       # Extra flags taken from dh_apparmor:
       # > By using '-W -T' we ensure that any abstraction updates are also pulled in.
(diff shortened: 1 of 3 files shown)


packages/app-builder-lib/templates/linux/after-install.tpl
```diff
@@ -42,7 +42,10 @@ if test -d "/etc/apparmor.d"; then
   if apparmor_parser --skip-kernel-load --debug "$APPARMOR_PROFILE_SOURCE" > /dev/null 2>&1; then
     cp -f "$APPARMOR_PROFILE_SOURCE" "$APPARMOR_PROFILE_TARGET"
 
-    if hash apparmor_parser 2>/dev/null; then
+    # Updating the current AppArmor profile is not possible and probably not meaningful in a chroot'ed environment.
+    # Use cases are for example environments where images for clients are maintained.
+    # There, AppArmor might correctly be installed, but live updating makes no sense.
+    if ! { [ -x '/usr/bin/ischroot' ] && ischroot; } && hash apparmor_parser 2>/dev/null; then
```
mmaietta, line 48: Can this be `&& /usr/bin/ischroot;`?

fiesh: Oh yes, good catch.  I amended my commit.

fiesh:
Ok, I'm not sure why

https://github.com/electron-userland/electron-builder/actions/runs/13157636243/job/36726618639

fails.  Is it trying to keep explicit executable paths out of the script?  If so, should we better use `which` to see if the executable is available and then call it without an absolute path?

mmaietta:
@fiesh I gotchu covered, it just needs a snapshot update as the unit test verifies that the `tpl` scripts are copied correctly verbatim. I'll push a snapshot update to this PR

Issue #8842, opened 2025-02-04
builds broken since v26.0.4


t3chguy:
Affects Windows & macOS at least. Linux seems unaffected https://github.com/element-hq/element-desktop/actions/runs/13135755860/job/36653217694?pr=624

Seems related to https://github.com/electron-userland/electron-builder/pull/8571

```
Run yarn build:universal --publish never
yarn run v1.22.22
$ yarn run build:ts && yarn run build:res && electron-builder --universal --publish never
$ tsc
$ tsx scripts/copy-res.ts
  • electron-builder  version=26.0.4 os=23.6.0
  • loaded configuration  file=/Users/runner/work/element-desktop/element-desktop/electron-builder.ts
  • executing @electron/rebuild  electronVersion=34.0.1 arch=x64 buildFromSource=false appDir=./
  • installing native dependencies  arch=x64
  • completed installing native dependencies
  • packaging       platform=darwin arch=x64 electron=34.0.1 appOutDir=dist/mac-universal-x64-temp
  • downloading     url=https://github.com/electron/electron/releases/download/v34.0.1/electron-v34.0.1-darwin-x64.zip size=105 MB parts=6
  • downloaded      url=https://github.com/electron/electron/releases/download/v34.0.1/electron-v34.0.1-darwin-x64.zip duration=2.323s
  ⨯ The "path" argument must be of type string. Received undefined  failedTask=build stackTrace=TypeError: The "path" argument must be of type string. Received undefined
    at Object.normalize (node:path:[12](https://github.com/element-hq/element-desktop/actions/runs/13135755860/job/36650705352?pr=624#step:13:13)05:5)
    at flatten (/Users/runner/work/element-desktop/element-desktop/node_modules/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts:63:49)
    at flatten (/Users/runner/work/element-desktop/element-desktop/node_modules/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts:68:9)
    at YarnNodeModulesCollector.convertToDependencyGraph (/Users/runner/work/element-desktop/element-desktop/node_modules/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts:72:5)
[...]

PR #8845, opened 2025-02-05
Fix: use include instead of moit for npm list

beyondkmp:
Fix https://github.com/electron-userland/electron-builder/issues/8842


test/src/HoistedNodeModuleTest.ts
```diff
@@ -97,8 +97,29 @@ test.ifAll("pnpm es5-ext without hoisted config", () =>
   )
 )
 
+test.only("yarn electron-clear-data", () =>
```
mmaietta, line 100: Please remove the `.only` :)

beyondkmp: Fixed

PR #8853, opened 2025-02-06
feat: Improve Azure Trusted Signing Docs

nathanlesage:
Hej! As announced in my recent issue, here are my suggestions to improve the Azure Trusted Signing documentation. I tried to find a balance between "not explaining everything" and ensuring that people understand the very confusing naming conventions Microsoft uses in its documentation (such as "Service Principal" versus "App Registration"). I believe that the way I tried to condense it hopefully makes it easier understandable for users to get started.

While writing the docs, I assumed as the "average" reader of this documentation someone with no experience with Azure who only wants to implement code signing. Because of this, I decided to add a "minimal viable setup" guide to help people get started, but keep the documentation in there to tell people "Hey, if you need another setup structure, here's what you need to know". In addition, I decided to copy over the descriptions from Microsoft's guides to give more security to readers who may be confused by the wording, and emphasize that especially not all environment variables are required in every scenario.

Let me know what you think and if I should change something!

### DIFF
--- a/pages/code-signing-win.md
+++ b/pages/code-signing-win.md
@@ -16,21 +16,36 @@ If you use Windows 7, please ensure that [PowerShell](https://blogs.technet.micr
 
 If you are on Linux or Mac and you want sign a Windows app using EV Code Signing Certificate, please use [the guide for Unix systems](tutorials/code-signing-windows-apps-on-unix.md).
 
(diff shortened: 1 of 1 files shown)


mmaietta:
This is fantastic, thank you @nathanlesage !

Issue #8857, opened 2025-02-09
26.0.6 node-module-collector RangeError: Maximum call stack size exceeded


absidue:
I just attempted to update from electron-builder 25.1.8 to 26.0.6 and received the `RangeError: Maximum call stack size exceeded` error from the node-module-collector (full error below). The irony of that is that we don't even need electron-builder to create a list of `node_modules` to include, as we explicitly exclude them from being packaged with `!node_modules/**/*` in the `files` array, as we bundle the JavaScript code before calling electron-builder.

So this issue is a bug report but also an ask for `electron-builder` to check the `files` array before collecting the `node_modules`, so that it can avoid doing work that isn't necessary. If checking the files array isn't possible a boolean config option to skip the node_modules collection would also work.

```
RangeError: Maximum call stack size exceeded
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:54:21)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
    at flatten (...\node_modules\app-builder-lib\src\node-module-collector\nodeModulesCollector.ts:69:9)
```

beyondkmp:
Can you provide a minimum reproducible repo or instructions on how to repro your issue?

absidue:
I'll work on trying to create a minimum reproduction.

[...]

PR #8864, opened 2025-02-12
fix: Detected circular dependencies and add debug logs for nodeModulesCollector

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8857

**Root cause**
It's caused by https://github.com/electron-userland/electron-builder/pull/8851

**How to fix**

In situations like issue #8851, a `skipCircularDeps` parameter should be added. This way, when traversing the tree, only the values inside are retrieved without further traversal, thus avoiding an infinite loop.


test/src/HoistedNodeModuleTest.ts
```diff
@@ -198,6 +198,57 @@ test.ifAll("yarn some module add by manual instead of install", () =>
...
+          modifyPackageJson(projectDir, data => {
+            data.dependencies = {
+              "npm-run-all": "^4.1.5",
+            }
+          }),
+          outputFile(path.join(projectDir, "pnpm-lock.yaml"), ""),
+        ])
+      },
+      packed: async context => {
+        expect(await readAsarJson(path.join(context.getResources(Platform.LINUX), "app.asar"), "node_modules/npm-run-all/package.json")).toMatchSnapshot()
```
mmaietta, line 246: I think for all these `"test-app-hoisted"` tests, we need to change it from `expect(await readAsarJson(` to `packed: context => verifyAsarFileTree(context.getResources(...))` to more accurately track changes we're making to the node module collector. That'll mirror how previous tests were set up

packages/app-builder-lib/src/node-module-collector/types.ts
```diff
@@ -11,6 +11,7 @@ export interface DependencyTree {
   readonly from?: string
   readonly workspaces?: string[]
   readonly path: string
+  skipCircularDeps?: boolean
```
mmaietta, line 14: I'm wary of using a boolean that is optional and only ever set true in the code logic flow.

Maybe we can use `private circularDependencyDetected = false`?

packages/app-builder-lib/src/node-module-collector/yarnNodeModulesCollector.ts
```diff
@@ -17,8 +17,10 @@ export class YarnNodeModulesCollector extends NodeModulesCollector {
   removeNonProductionDependencie(tree: DependencyTree) {
     const dependencies = tree.dependencies || {}
     const _dependencies = tree._dependencies || {}
-    if (dependencies && Object.keys(dependencies).length === 0) {
+    if (Object.keys(_dependencies).length > 0 && Object.keys(dependencies).length === 0) {
       tree.dependencies = this.allDependencies.get(`${tree.name}@${tree.version}`)?.dependencies || {}
+      tree.skipCircularDeps = true
```
mmaietta, line 22: Can we add some logging here? Probably a `debug` level would suffice as we don't know how often this will trigger for users.

```
const { name, version } = tree
tree.dependencies = this.allDependencies.get(`${name}@${version}`)?.dependencies || {}
log.debug({ name, version }, "circular dependency detected")
tree.circularDependencyDetected = true
```

mmaietta: I think this also may be a worthwhile time to refactor `yarnNodeModuleCollector` as it's the same logic and duplicated code.
```
export class YarnNodeModulesCollector extends NpmNodeModulesCollector { }
```

.changeset/swift-hotels-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+skip circular deps
```
mmaietta, line 5: Please make this more descriptive of a changeset since this is published to our changelog

packages/app-builder-lib/src/node-module-collector/types.ts
```diff
@@ -11,6 +11,7 @@ export interface DependencyTree {
   readonly from?: string
   readonly workspaces?: string[]
   readonly path: string
+  __circularDependencyDetected?: boolean
   dependencies?: {
     [packageName: string]: DependencyTree
   }
```
mmaietta, line 17: Side note: This doesn't need to be part of this PR fix. But we should really reconsider refactoring this implementation as right now, it's an interface of **all but 1** optional, which is hardly better than just using `any`

What are your thoughts that we can do there?

test/src/HoistedNodeModuleTest.ts
beyondkmp: Agreed with you. Fixed.

packages/app-builder-lib/src/node-module-collector/types.ts
beyondkmp: In this interface, `name`, `version`, and `path` should all be set as non-optional. If any of them is empty, it will cause issues. I have already fixed that.
As for the refactoring you mentioned, I don't have a good idea at the moment.

.changeset/swift-hotels-exercise.md
beyondkmp: Fixed.

packages/app-builder-lib/src/node-module-collector/yarnNodeModulesCollector.ts
beyondkmp: Very good suggestion. Fixed.

packages/app-builder-lib/src/node-module-collector/types.ts
```diff
@@ -6,11 +6,11 @@ export interface NodeModuleInfo {
...
 export interface DependencyTree {
-  readonly version?: string
-  readonly name?: string
-  readonly from?: string
-  readonly workspaces?: string[]
+  readonly name: string
+  readonly version: string
   readonly path: string
+  readonly workspaces?: string[]
+  __circularDependencyDetected?: boolean
```
mmaietta, line 13: So just to confirm, which of these are received from `const dependencyTree: DependencyTree | DependencyTree[] = JSON.parse(dependencies)`?

I think we should create an interface that is:
1. Directly corresponds to what the JSON.parse is returning.
2. Extend that interface with the `__circularDependencyDetected` and force it to be false in logic before setting it true in the logic. Again, I'm always wary of having a `property: true | false | undefined` as `false | undefined` should certainly have a manner to be coalesced into `false`

Something like this below. I recognize it's verbose, but it is much more explicit in intent and usage.
If preferred, I can push this through as a separate follow-up/refactor PR so that we isolate the bug fix here to simply the bug. Wanted to check what your thoughts on this though?
```
export interface DependencyTree extends NodeParsedDependencyTree {
  circularDependencyDetected: boolean
}

export interface NodeParsedDependencyTree extends ParsedDependencyTree {
  // for npm list --json
  _dependencies?: {
    [packageName: string]: string
  }
}
export interface ParsedDependencyTree extends Dependency<ParsedDependencyTree> {
  readonly version?: string
  readonly name?: string
  readonly from?: string
  readonly workspaces?: string[]
  readonly path: string
}

export interface Dependency<T> {
  dependencies?: {
    [packageName: string]: T
  }
  optionalDependencies?: {
    [packageName: string]: T
  }
  peerDependencies?: {
    [packageName: string]: T
  }
}
```

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -61,11 +61,26 @@ export abstract class NodeModulesCollector {
...
+        log.debug(
+          {
+            dependency: key,
+            version,
+            path: value.path,
+            parentModule: node.name,
+            parentVersion: node.version,
+          },
+          "Adding dependency to module tree"
+        )
```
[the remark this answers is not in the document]
beyondkmp: Deleted. This is duplicated with https://github.com/electron-userland/electron-builder/blob/6055b5b4b47ff3a943d2b0d49336f7bb5b4eded0/packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts#L24

packages/app-builder-lib/src/node-module-collector/types.ts
beyondkmp: This refactoring is fantastic! I agree with your suggestion; it's better to put it into a separate commit.

mmaietta: Alrighty, I'll get started on it! It'll help me better understand the logic flow too

mmaietta (approved):

PR #8873, opened 2025-02-18
feat: Exclude node modules from asar

beyondkmp:
add excludeNodeModulesFromAsar config

refers to https://github.com/electron-userland/electron-builder/issues/8857

### DIFF
--- a/packages/app-builder-lib/src/platformPackager.ts
+++ b/packages/app-builder-lib/src/platformPackager.ts
@@ -504,7 +504,7 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions> 
 
     const _computeFileSets = (matchers: Array<FileMatcher>) => {
       return computeFileSets(matchers, this.info.isPrepackedAppAsar ? null : transformer, this, isElectronCompile).then(async result => {
-        if ((!this.info.isPrepackedAppAsar && !this.info.areNodeModulesHandledExternally) || !!this.info.config.excludeNodeModulesFromAsar) {
+        if (!this.info.isPrepackedAppAsar && !this.info.areNodeModulesHandledExternally && !this.info.config.excludeNodeModulesFromAsar) {
           const moduleFileMatcher = getNodeModuleFileMatcher(appDir, defaultDestination, macroExpander, platformSpecificBuildOptions, this.info)
           result = result.concat(await computeNodeModuleFileSets(this, moduleFileMatcher))
         }
--- a/packages/app-builder-lib/src/platformPackager.ts
+++ b/packages/app-builder-lib/src/platformPackager.ts
@@ -504,7 +504,7 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions> 
 
     const _computeFileSets = (matchers: Array<FileMatcher>) => {
       return computeFileSets(matchers, this.info.isPrepackedAppAsar ? null : transformer, this, isElectronCompile).then(async result => {
-        if (!this.info.isPrepackedAppAsar && !this.info.areNodeModulesHandledExternally) {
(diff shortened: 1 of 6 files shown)


packages/app-builder-lib/src/platformPackager.ts
```diff
@@ -504,7 +504,7 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions>
 
     const _computeFileSets = (matchers: Array<FileMatcher>) => {
       return computeFileSets(matchers, this.info.isPrepackedAppAsar ? null : transformer, this, isElectronCompile).then(async result => {
-        if (!this.info.isPrepackedAppAsar && !this.info.areNodeModulesHandledExternally) {
+        if (!this.info.isPrepackedAppAsar && !this.info.areNodeModulesHandledExternally && !this.info.config.excludeNodeModulesFromAsar) {
```
mmaietta, line 507: This is actually already handled by `beforeBuild` logic (albeit only if the jsdocs are read by the end-user/dev). Should just need to return `false` from `beforeBuild` hook
https://github.com/electron-userland/electron-builder/blob/9bca417a8c27facb289dac2cfcbb4ef1c111090f/packages/app-builder-lib/src/packager.ts#L525-L533

beyondkmp: While writing this MR, I also found that returning `false` in `beforeBuild` can achieve this functionality. However, the literal meaning of `beforeBuild` is really confusing. It’s not clear that it can be used to exclude `node_modules` from being packed into the ASAR file.

mmaietta: Yeahh, that was a patch fix to avoid adding yet another configuration property. I tried to make it clear in the docs, but docs aren't always read unfortunately.

Related, for next major semver, I'm thinking it may be worth moving all native module handling (`npmRebuild`, `nativeRebuilder`, `buildDependenciesFromSource`, etc.) to nest into a single configuration object. What are your thoughts? I feel like our electron-builder-config.js is getting really nasty/complex. Could also be an optimal time to remove `nativeRebuilder` too

beyondkmp: > Related, for next major semver, I'm thinking it may be worth moving all native module handling (npmRebuild, nativeRebuilder, buildDependenciesFromSource, etc.) to nest into a single configuration object. What are your thoughts? I feel like our electron-builder-config.js is getting really nasty/complex. Could also be an optimal time to remove nativeRebuilder too

Indeed, these native module configurations should be removed. Some of them are redundant and overly complex, and many people don't use them.
Moreover, if we can eventually remove `app-builder-bin` and rewrite everything in TypeScript, we could achieve better cross-platform support, such as for the Loongson architecture.

beyondkmp: >If we have the node module collector fully working and since it's not a performance hog, I don't think it's necessary to implement this request. app-builder-bin always collected node modules before filtering, so I don't see any logic that was "broken" during the Golang->JS migration. Curious on your thoughts

I agree with your opinion. We can provide a detailed explanation in the docs about the functionality of `beforeBuild` for excluding `node_modules` from being packed into the ASAR file.

mmaietta: >Moreover, if we can eventually remove app-builder-bin and rewrite everything in TypeScript

I absolutely love this goal. It's going to take a hot minute to get there though in terms of x-platform support for things like icon generation, appimages, rcedit, and wine 😬

beyondkmp:
@mmaietta The reason for reopening this issue is that there's another problem: if the user is using a package.json or package.yml configuration, they won't be able to use the beforeBuild hook function, right? The user would need to switch to using JavaScript for the configuration in order to use the hook function.

Issue #8881, opened 2025-02-20
Error: Cannot find module '.../package.json'


staniel359:
Version: 26.0.7

Hey.
When i run `electron-builder --dir` command, I get this error:

![Image](https://github.com/user-attachments/assets/b2ad29b3-8876-46be-b199-387cf4520c28)

And `.pnpm` folder is almost empty:

![Image](https://github.com/user-attachments/assets/7d477cea-97d9-41c8-8337-29661b655c49)

Could you help me with this issue?

xyloflake:
Hello @mmaietta, I've kind of looked into this issue. Please notice how it says "node_modules/.pnpm/dayjs@1.11.13" and then it appends to the same path "node_modules/dayjs" and finally "package.json". The later part is correct however the first path is incorrect and I don't really get why it's repeated.

Here's our config 
https://github.com/staniel359/muffon/blob/fix-schema/electron-builder.json

xyloflake:
This is solved by removing `node-linker=hoisted`.

beyondkmp:
This is because we no longer need to specify the `hoist` configuration under `pnpm`. However, for those who have previously set it, we need to remove the corresponding configuration. @mmaietta  Should we notify this somewhere?

mmaietta:
Hmmm I agree with @xyloflake. We either need to:
- Support both hoist key/value setups (ideal solution)
- detect npmrc and log warning (or InvalidConfigurationError?) if key is set + recommend remediation
- Orrrrrr....? Suggestions welcome!

xyloflake:
@mmaietta I think the best solution would be to detect if node-linker is set to hoisted and then if it *is* set to hoisted, then use old hoisted packaging methods. If it's not, then use new method.

mmaietta:
Yeah, I think that would be the ideal solution as well.

@beyondkmp, is it possible for us to extend the current implementation to handle this? I'd prefer not falling back to app-builder-bin for legacy implementation as that'll just be more code pathways for us to maintain (plus I'm also collaborating with `@electron` org, and they've expressed interest in leveraging our implementation into a dedicated `@electron` package)

xyloflake:
> Yeah, I think that would be the ideal solution as well.
> 
[...]

PR #8885, opened 2025-02-22
fix: allow usage of `node-linker=hoisted` with pnpm

mmaietta:
Fixes: https://github.com/electron-userland/electron-builder/issues/8881

### DIFF
--- a/packages/app-builder-lib/src/node-module-collector/index.ts
+++ b/packages/app-builder-lib/src/node-module-collector/index.ts
@@ -10,12 +10,12 @@ async function isPnpmProjectHoisted(rootDir: string) {
   const config = await exec(command, ["config", "list"], { cwd: rootDir, shell: true })
   const lines = config
     .split("\n")
-    .map(line => line.split("="))
+    .map(line => line.trim().split("="))
     .reduce<Record<string, string>>((accum, curr) => {
       const [key, value] = curr
       return {
         ...accum,
-        [key]: value,
+        [key.trim()]: value.trim(),
       }
     }, {})
   return lines["node-linker"] === "hoisted"
--- a/packages/app-builder-lib/src/node-module-collector/index.ts
+++ b/packages/app-builder-lib/src/node-module-collector/index.ts
@@ -6,7 +6,7 @@ import { NodeModuleInfo } from "./types"
 import { exec } from "builder-util"
 
 async function isPnpmProjectHoisted(rootDir: string) {
-  const command = process.platform === "win32" ? "pnpm.cmd" : "pnpm"
+  const command = await PnpmNodeModulesCollector.pmCommand.value
   const config = await exec(command, ["config", "list"], { cwd: rootDir, shell: true })
   const lines = Object.fromEntries(config.split("\n").map(line => line.split("=").map(s => s.trim())))
   return lines["node-linker"] === "hoisted"
--- a/packages/app-builder-lib/src/node-module-collector/index.ts
+++ b/packages/app-builder-lib/src/node-module-collector/index.ts
@@ -18,7 +18,7 @@ async function isPnpmProjectHoisted(rootDir: string) {
         [key]: value,
       }
     }, {})
-  return lines["node-linker"] === "hoisted" || lines["shamefully-hoist"] === "true" || lines["public-hoist-pattern"] === "*"
+  return lines["node-linker"] === "hoisted"
 }
 
 async function getCollectorByPackageManager(rootDir: string) {
--- a/packages/app-builder-lib/src/node-module-collector/index.ts
+++ b/packages/app-builder-lib/src/node-module-collector/index.ts
@@ -3,19 +3,45 @@ import { PnpmNodeModulesCollector } from "./pnpmNodeModulesCollector"
 import { YarnNodeModulesCollector } from "./yarnNodeModulesCollector"
 import { detect, PM, getPackageManagerVersion } from "./packageManager"
 import { NodeModuleInfo } from "./types"
+import { readFile } from "fs-extra"
+import * as path from "path"
+import { exists } from "builder-util"
 
 async function getCollectorByPackageManager(rootDir: string) {
   const manager: PM = await detect({ cwd: rootDir })
+  const npmrc = await parseNpmrc(rootDir)
   switch (manager) {
+    case "pnpm":
+      if (npmrc?.["node-linker"] === "hoisted") {
+        return new NpmNodeModulesCollector(rootDir)
+      }
+      return new PnpmNodeModulesCollector(rootDir)
     case "npm":
       return new NpmNodeModulesCollector(rootDir)
-    case "pnpm":
-      return new PnpmNodeModulesCollector(rootDir)
     case "yarn":
       return new YarnNodeModulesCollector(rootDir)
     default:
       return new NpmNodeModulesCollector(rootDir)
   }
+}
+
+async function parseNpmrc(rootDir: string) {
+  const npmrc = path.join(rootDir, ".npmrc")
+  if (await exists(npmrc)) {
+    const n = await readFile(npmrc, "utf-8")
+    const lines = n
+      .split("\n")
+      .map(line => line.split("="))
+      .reduce<Record<string, string>>((accum, curr) => {
+        const [key, value] = curr
+        return {
+          ...accum,
+          [key]: value,
+        }
+      }, {})
+    return lines
+  }
+  return undefined
 }
 
 export async function getNodeModules(rootDir: string): Promise<NodeModuleInfo[]> {
--- a/packages/app-builder-lib/src/node-module-collector/index.ts
+++ b/packages/app-builder-lib/src/node-module-collector/index.ts
@@ -15,7 +15,7 @@ async function isPnpmProjectHoisted(rootDir: string) {
       const [key, value] = curr
       return {
         ...accum,
-        [key.trim()]: value.trim(),
+        [key.trim()]: value?.trim(),
       }
     }, {})
   return lines["node-linker"] === "hoisted"
--- a/packages/app-builder-lib/src/node-module-collector/index.ts
+++ b/packages/app-builder-lib/src/node-module-collector/index.ts
@@ -8,16 +8,7 @@ import { exec } from "builder-util"
 async function isPnpmProjectHoisted(rootDir: string) {
   const command = process.platform === "win32" ? "pnpm.cmd" : "pnpm"
   const config = await exec(command, ["config", "list"], { cwd: rootDir, shell: true })
-  const lines = config
-    .split("\n")
-    .map(line => line.trim().split("="))
-    .reduce<Record<string, string>>((accum, curr) => {
-      const [key, value] = curr
-      return {
-        ...accum,
-        [key.trim()]: value?.trim(),
-      }
-    }, {})
+  const lines = Object.fromEntries(config.split("\n").map(line => line.split("=").map(s => s.trim())))
   return lines["node-linker"] === "hoisted"
 }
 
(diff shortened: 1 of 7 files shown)


packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -3,21 +3,47 @@ import { PnpmNodeModulesCollector } from "./pnpmNodeModulesCollector"
...
+    case "npm":
+      return new NpmNodeModulesCollector(rootDir)
     case "yarn":
       return new YarnNodeModulesCollector(rootDir)
     default:
       return new NpmNodeModulesCollector(rootDir)
   }
 }
 
+async function parseNpmrc(rootDir: string) {
```
beyondkmp, line 28: We should preferably use `pnpm config get node-linker`, as it works even in the case of two `package.json` files or within a workspace.

image

test/src/HoistedNodeModuleTest.ts
```diff
@@ -317,3 +317,28 @@ test.ifAll("npm tar", () =>
     }
   )
 )
+
+//github.com/electron-userland/electron-builder/issues/8881
+test.ifAll("pnpm node-linker=hoisted", () =>
```
beyondkmp, line 322: We need to add more test cases for pnpm hoisted config.

image

beyondkmp: Under these two configurations, we can still use `pnpm list` instead of `npm list`.

packages/app-builder-lib/src/node-module-collector/index.ts
beyondkmp: Refers to https://github.com/electron-userland/electron-builder/pull/8886

mmaietta: That's fantastic. Definitely will use that.

test/src/HoistedNodeModuleTest.ts
mmaietta: Will do!

packages/app-builder-lib/src/node-module-collector/index.ts
mmaietta: Since we need to check `shamefully-hoist=true` `public-hoist-pattern=*` and `node-linker=hoisted`, `pnpm config list` with a `reduce` to `Record<string, string>` should do the trick.

Not sure if we can easily check variations of `public-hoist-pattern` though, I have no clue how to parse it atm
https://pnpm.io/npmrc#public-hoist-pattern

beyondkmp: Under these two configurations(`shamefully-hoist=true public-hoist-pattern=*`), we can still use pnpm list instead of npm list.

beyondkmp (approved):

packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -3,14 +3,34 @@ import { PnpmNodeModulesCollector } from "./pnpmNodeModulesCollector"
...
+  const command = process.platform === "win32" ? "pnpm.cmd" : "pnpm"
+  const config = await exec(command, ["config", "list"], { cwd: rootDir, shell: true })
+  const lines = config
+    .split("\n")
+    .map(line => line.split("="))
+    .reduce<Record<string, string>>((accum, curr) => {
+      const [key, value] = curr
+      return {
+        ...accum,
+        [key]: value,
```
beyondkmp, line 18: We'd better to perform a `trim` operation on both the beginning and end of the string, as some operating systems may introduce extra spaces.

packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -3,14 +3,34 @@ import { PnpmNodeModulesCollector } from "./pnpmNodeModulesCollector"
 import { YarnNodeModulesCollector } from "./yarnNodeModulesCollector"
 import { detect, PM, getPackageManagerVersion } from "./packageManager"
 import { NodeModuleInfo } from "./types"
+import { exec } from "builder-util"
+
+async function isPnpmProjectHoisted(rootDir: string) {
+  const command = process.platform === "win32" ? "pnpm.cmd" : "pnpm"
+  const config = await exec(command, ["config", "list"], { cwd: rootDir, shell: true })
+  const lines = config
+    .split("\n")
```
beyondkmp, line 12: If you split like this on Windows, `\r` may remain in the result, so it's necessary to `trim` the final output.

packages/app-builder-lib/src/node-module-collector/index.ts
mmaietta: Good callout, updated!

xyloflake:
I'll have a look

packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -3,14 +3,34 @@ import { PnpmNodeModulesCollector } from "./pnpmNodeModulesCollector"
 import { YarnNodeModulesCollector } from "./yarnNodeModulesCollector"
 import { detect, PM, getPackageManagerVersion } from "./packageManager"
 import { NodeModuleInfo } from "./types"
+import { exec } from "builder-util"
+
+async function isPnpmProjectHoisted(rootDir: string) {
+  const command = process.platform === "win32" ? "pnpm.cmd" : "pnpm"
```
xyloflake, line 9: Are you sure this should be pnpm.cmd when it's windows? I use windows but I mostly use powershell and I know that pnpm is set to Path environment variable during installation. pnpm.cmd ***may*** introduce unnecessary compatibility issues if the system shell is powershell (by default in win 1122H2+).

xyloflake: Also `pnpm` is valid for both powershell and command prompt but `pnpm.cmd` is only valid for command prompt.

mmaietta: I was under the impression `exec` + `shell: true` is executed in a command prompt env.

packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -3,14 +3,34 @@ import { PnpmNodeModulesCollector } from "./pnpmNodeModulesCollector"
...
+  const config = await exec(command, ["config", "list"], { cwd: rootDir, shell: true })
+  const lines = config
```
xyloflake, line 11: I believe you can just do this

```suggestion
  const lines = Object.fromEntries(
  config.split("\n").map(line =>
    line.split("=").map(s => s.trim())
  )
);

```

This keeps the code more maintainable I think

xyloflake (changes requested):

packages/app-builder-lib/src/node-module-collector/index.ts
xyloflake: > I was under the impression `exec` + `shell: true` is executed in a command prompt env.
> 
> The `PnpmNodeModuleCollector` also uses `.cmd` suffix.

I am pretty sure exec + shell: true will open powershell on win 11 22h2+

xyloflake: I believe it's an extra unnecessary step so it should be eradicated.

packages/app-builder-lib/src/node-module-collector/index.ts
mmaietta: That's beautiful. Will update the PR

packages/app-builder-lib/src/node-module-collector/index.ts
mmaietta: That's a fair callout, but the current electron-builder needs to support OS versions older than win 11 without requiring bumping a major semver.

xyloflake: > That's a fair callout, but the current electron-builder needs to support OS versions older than win 11 without requiring bumping a major semver.

pnpm installs itself and adds the binary path to the path env variable amid installation, thus `pnpm` should be fine on all windows versions (even the older ones) but `pnpm.cmd` will be only valid for versions older than win11 22h2, which is why `pnpm` must be used instead of `pnpm.cmd` is what I meant to say 😅

beyondkmp: I've been using Windows 11 24H2, and I haven't encountered the issue you mentioned.
image

xyloflake: Yes, that's expected. As I said before:
> pnpm.cmd ***may*** introduce unnecessary compatibility issues if the system shell is powershell (by default in win 1122H2+)

xyloflake: It's not certain but I believe pnpm.cmd is no longer necessary.

mmaietta: I'd prefer to play it safe until issue reports come in that `pnpm.cmd` is failing. `.cmd` is already what is used in `pnpmNodeModulesCollector`
https://github.com/electron-userland/electron-builder/blob/aafb531fddbc0863afd967c84841ff70b54ff27d/packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts#L10-L12

mmaietta: > pnpm.cmd will be only valid for versions older than win11 22h2, which is why pnpm must be used instead of pnpm.cmd

I guess I could wrap the command to determine if `pnpm --version` is available? If not, fallback to `pnpm.cmd`?

mmaietta: Implemented with `lazy-val` so that we only query the system once for which cmd to use (we take a similar approach with `pwsh.exe` versus `powershell.exe` detection)
https://github.com/electron-userland/electron-builder/pull/8885/files#diff-3781d1bcfef363df7dea345579c15ffd71e6f2e012535e5ca40c49633742990cR12-R22

xyloflake: That's great!

xyloflake (approved):
LGTM!

packages/app-builder-lib/src/node-module-collector/index.ts
xyloflake: > I guess I could wrap the command to determine if `pnpm --version` is available. If not, fallback to `pnpm.cmd`? I'd rather play as safely as possible so as not to break various OS compatibility.

@mmaietta I think you should definitely implement this!

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -1,17 +1,29 @@
...
+      try {
+        await exec("pnpm", ["--version"])
+      } catch (_error: any) {
+        log.debug(null, "pnpm not detected, falling back to pnpm.cmd")
+        return "pnpm.cmd"
+      }
+    }
+    return "pnpm"
+  })
+
```
xyloflake, line 23:
```js
static readonly pmCommand = new Lazy<string>(async () =>
  process.platform === "win32" &&
  (await exec("pnpm", ["--version"]).catch(() => false)) === false
    ? "pnpm.cmd"
    : "pnpm"
)
```
What do you think of this?

mmaietta: I like the syntactic sugar, but it's much harder for me to read tbh 😅

xyloflake: @mmaietta I believe you can also remove the `=== false`
```js
static readonly pmCommand = new Lazy<string>(async () =>
  process.platform === "win32" &&
  (await exec("pnpm", ["--version"]).catch(() => false))
    ? "pnpm.cmd"                         //        |
    : "pnpm"                             //        | --> Implement logging here if needed
)
```

mmaietta:
Thanks for the review @xyloflake and @beyondkmp !

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
xyloflake: > I like the syntactic sugar, but it's much harder for me to read tbh 😅

Cool! No problem!

PR #8889, opened 2025-02-22
chore: delete all old node module collector and legacy nativeRebuilder

beyondkmp:
Now that the new Node module collector is released, we can delete these old commands.

After deletion, it will completely replace the Node collector inside `app-builder-bin`. The next step is to replace some other commands until all are fully replaced.


packages/app-builder-lib/src/configuration.ts
```diff
@@ -129,10 +129,10 @@ export interface CommonConfiguration {
    */
   readonly npmRebuild?: boolean
   /**
-   * Use `legacy` app-builder binary for installing native dependencies, or `@electron/rebuild` in `sequential` or `parallel` compilation modes.
+   * Use `@electron/rebuild` in `sequential` or `parallel` compilation modes.
    * @default sequential
    */
-  readonly nativeRebuilder?: "legacy" | "sequential" | "parallel" | null
+  readonly nativeRebuilder?: "sequential" | "parallel" | null
```
mmaietta, line 135: This requires this PR to be a `major` semver changeset. Great PR to prep for when we want to do a Node 22 engine upgrade to match `electron` upstream

PR #8890, opened 2025-02-22
chore(plist): replace the plist functionality in app-builder-bin with plist

beyondkmp:


packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -5,7 +5,8 @@ import { filterCFBundleIdentifier } from "../appInfo"
 import { AsarIntegrity } from "../asar/integrity"
 import { MacPackager } from "../macPackager"
 import { normalizeExt } from "../platformPackager"
-import { executeAppBuilderAndWriteJson, executeAppBuilderAsJson } from "../util/appBuilder"
+import SimplePlist from "simple-plist"
```
mmaietta, line 8: Can we use `plist` npm package instead? That's the one currently used in `electron/universal`

beyondkmp: Good suggestion. Fixed.

packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -229,31 +208,34 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
...
   }
+
   if (helperLoginPlist != null) {
-    plistDataToWrite[helperLoginPlistFilename] = helperLoginPlist
+    await savePlistFile(helperLoginPlistFilename, helperLoginPlist)
   }
 
+  await savePlistFile(appPlistFilename, appPlist)
+  await savePlistFile(helperPlistFilename, helperPlist)
+
```
mmaietta, line 237: Maybe can just keep the previous logic and then do this?
```
await Promise.all(Object.entries(plistDataToWrite).map(([filename, content] => savePlistFile(filename, content)))
```

packages/app-builder-lib/package.json
```diff
@@ -77,7 +77,9 @@
     "semver": "^7.3.8",
     "tar": "^6.1.12",
     "temp-file": "^3.4.0",
-    "tiny-async-pool": "1.3.0"
+    "tiny-async-pool": "1.3.0",
+    "plist":"3.1.0",
+    "@types/plist": "3.0.5"
```
mmaietta, line 82: Can this be a `devDependency`?

packages/app-builder-lib/src/util/plist.ts
```diff
@@ -0,0 +1,37 @@
...
+    }, {})
+}
+
+export async function savePlistFile(path: string, data: any): Promise<void> {
+  const sortedData = sortObjectKeys(data)
+  const plist = build(sortedData)
+  await fs.promises.writeFile(path, plist)
+}
+
+export async function parsePlistFile(file: string): Promise<any> {
```
mmaietta, line 27: Any way we can remove the `any` types from this file and properly type the vars?

packages/app-builder-lib/src/util/plist.ts
```diff
@@ -0,0 +1,37 @@
...
+  try {
+    const data = await fs.promises.readFile(file, "utf8")
+    return parse(data)
+  } catch (e: any) {
+    if (e.code === "ENOENT") {
+      return null
```
mmaietta, line 33: Why this logic? I don't think we should have a `return null`. Also, usages of this function in the code don't seem to handle the `null` return since it's obfuscated by the `Promise<any>`

beyondkmp: This situation does occur; I've encountered it in the unit tests (UT). Currently, app-builder-bin just returns null(https://github.com/develar/app-builder/blob/580c34f2e19347061bc2243fa6ab6e57e9bbd2a6/pkg/plist/plist.go#L75-L77). If we want to ensure that the file always exists, we would need to check the file path before calling it, which would be more troublesome.
image

packages/app-builder-lib/src/electron/electronMac.ts
beyondkmp: Currently, I don't think `Promise.all` is the best choice, especially when it comes to error messages. If there are performance issues, then `Promise.all` would indeed be better.

This approach does have a slight performance overhead, but as soon as one operation fails, it stops immediately without proceeding to write others. The error message also clearly indicates which part encountered the issue.

packages/app-builder-lib/src/util/plist.ts
beyondkmp: I will try it.

packages/app-builder-lib/src/util/plist.ts
beyondkmp: https://github.com/electron-userland/electron-builder/blob/22da64425182456eb4d1243138dde27c80d6adac/packages/app-builder-lib/src/electron/electronMac.ts#L96-L97

packages/app-builder-lib/package.json
beyondkmp: Fixed.

packages/app-builder-lib/src/util/plist.ts
beyondkmp: Fixed. Use safeParsePlistFile.

packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -69,38 +70,26 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
...
-  const helperGPUPlist = plistContent[6]
-  const helperLoginPlist = plistContent[7]
+  // Replace the multiple parsePlistFile calls with:
+  const helperPlist = (await safeParsePlistFile(helperPlistFilename)) as PlistObject
+  const helperEHPlist = (await safeParsePlistFile(helperEHPlistFilename)) as string | null
+  const helperNPPlist = (await safeParsePlistFile(helperNPPlistFilename)) as string | null
+  const helperRendererPlist = (await safeParsePlistFile(helperRendererPlistFilename)) as string | null
+  const helperPluginPlist = (await safeParsePlistFile(helperPluginPlistFilename)) as string | null
+  const helperGPUPlist = (await safeParsePlistFile(helperGPUPlistFilename)) as string | null
+  const helperLoginPlist = (await safeParsePlistFile(helperLoginPlistFilename)) as PlistObject | null
```
mmaietta, line 92: I'm really not a fan of this casting. If you make the function generic, you won't need to worry about casting everything to `string` or `PlistObject`. I'm also not sure I understand the logic here. Why are some `string` and others `PlistObject`. Shouldn't they all be returning the same type?

```
export async function parsePlistFile<T>(file: string): Promise<T> {

```

beyondkmp: >I'm also not sure I understand the logic here. Why are some string and others PlistObject

This is defined based on how the parsed results are intended to be used. For `PlistObject`, properties are directly assigned as shown below. 
https://github.com/electron-userland/electron-builder/blob/f1b519e5671d6e49fcbbd5641d163916c45fb33d/packages/app-builder-lib/src/electron/electronMac.ts#L126-L129


Some of them are strings, as illustrated.
https://github.com/electron-userland/electron-builder/blob/f1b519e5671d6e49fcbbd5641d163916c45fb33d/packages/app-builder-lib/src/electron/electronMac.ts#L25-L30

beyondkmp:
```
export async function parsePlistFile<T>(file: string): Promise<T> {
```
Great suggestion, it has been updated accordingly.

packages/app-builder-lib/src/electron/electronMac.ts
```diff
@@ -222,39 +213,44 @@ export async function createMacApp(packager: MacPackager, appOutDir: string, asa
     )
 
     // `CFBundleDocumentTypes` may be defined in `mac.extendInfo`, so we need to merge it in that case
-    appPlist.CFBundleDocumentTypes = [...(appPlist.CFBundleDocumentTypes || []), ...documentTypes]
+    appPlist.CFBundleDocumentTypes = [...((appPlist.CFBundleDocumentTypes as PlistValue[]) || []), ...documentTypes]
   }
 
   if (asarIntegrity != null) {
-    appPlist.ElectronAsarIntegrity = asarIntegrity
+    appPlist.ElectronAsarIntegrity = JSON.parse(JSON.stringify(asarIntegrity))
```
mmaietta, line 220: Whats the purpose of the `stringify` -> `parse` here? This seems like it's unnecessary or might be hiding a different issue.

packages/app-builder-lib/src/util/plist.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function savePlistFile(path: string, data: PlistValue): Promise<void> {
+  const sortedData = sortObjectKeys(data)
+  const plist = build(sortedData)
+  await fs.promises.writeFile(path, plist)
+}
+
+export async function parsePlistFile<T>(file: string): Promise<T> {
+  const data = await fs.promises.readFile(file, "utf8")
+  return parse(data) as T
+}
```
mmaietta, line 37: nit: Can we import `import * as fs from "fs/promises"` instead of accessing it directly through `fs.promises`? (It'll help me later when I need to regex find+migrate from fs-extra to `fs/promises` (where we're able to at least)

mmaietta (dismissed):

packages/app-builder-lib/src/electron/electronMac.ts
beyondkmp: How about this modification?

```javascript
  const toPlistObject = (asarIntegrity: AsarIntegrity): PlistObject => {
    const result: PlistObject = {}
    for (const [filePath, headerHash] of Object.entries(asarIntegrity)) {
      result[filePath] = {
        algorithm: headerHash.algorithm,
        hash: headerHash.hash,
      }
    }
    return result
  }

  if (asarIntegrity != null) {
    appPlist.ElectronAsarIntegrity = toPlistObject(asarIntegrity)
  }
```

packages/app-builder-lib/src/util/plist.ts
beyondkmp: Fixed.

packages/app-builder-lib/src/electron/electronMac.ts
beyondkmp: >All we're testing in getAvailableHelperSuffixes is whether they're null or not though. AFAICT, parse is always supposed to return an object, not a string https://www.npmjs.com/package/plist

Yes, I checked the code, and it indeed only validates whether it is `null`, so it shouldn't be a `string`. I've updated all instances.

packages/app-builder-lib/src/electron/electronMac.ts
mmaietta: Looks great!

mmaietta (approved):
Nice work!

PR #8892, opened 2025-02-23
chore(snap): migrate snap generation from app-builder-bin to JS

mmaietta:


packages/electron-publish/src/snapStorePublisher.ts
```diff
@@ -28,14 +27,29 @@
...
+
+    // return executeAppBuilder(args)
   }
 
   toString(): string {
     return "Snap Store"
   }
+
+  async publishToStore(file: string, channels: string[]): Promise<void> {
+    const args = ["upload", file]
```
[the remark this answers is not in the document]
mmaietta: It already uses `execFile`
https://github.com/electron-userland/electron-builder/blob/e4d51475676819fc1c92f43b954171b1154d3290/packages/builder-util/src/util.ts#L98-L101

mmaietta:
Verified on ubuntu 22.04 arm64 VM
Screenshot 2025-02-25 at 8 48 53 AM

Verified on Fedora 40 arm64 VM
Screenshot 2025-02-25 at 8 53 01 AM

packages/app-builder-lib/src/targets/snap/snap.ts
```diff
@@ -234,22 +241,31 @@ export default class SnapTarget extends Target {
...
 
     const hooksDir = await packager.getResource(options.hooks, "snap-hooks")
     if (hooksDir != null) {
-      args.push("--hooks", hooksDir)
-    }
-
-    if (this.isUseTemplateApp) {
-      args.push("--template-url", `electron4:${snapArch}`)
+      args.hooksDir = hooksDir
+      // args.push("--hooks", hooksDir)
```
beyondkmp, line 264: This can be removed.

packages/app-builder-lib/src/targets/snap/snap.ts
```diff
@@ -234,22 +241,31 @@ export default class SnapTarget extends Target {
...
     }
 
-    await executeAppBuilder(args)
+    // await executeAppBuilder(args)
```
beyondkmp, line 267: This can be removed.

packages/app-builder-lib/src/targets/snap/snapBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+export interface SnapBuilderOptions extends CommandWrapperOptions {
+  appDir: string
+  output: string
+  arch: string
+
+  icon?: string
+  hooksDir?: string
+  excludedAppFiles?: string[]
+
+  compression?: "xz" | "lzo"
```
beyondkmp, line 18: Are there other compression algorithms? Gzip is very common on Linux.

packages/app-builder-lib/src/targets/snap/linuxTools.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export async function getMksquashfs(): Promise<string> {
+  if (!process.env.USE_SYSTEM_MKSQUASHFS) {
+    const result = process.env.MKSQUASHFS_PATH
+    if (!result) {
+      return await getLinuxTool("mksquashfs")
+    }
+    return result
+  }
+  return "mksquashfs"
```
beyondkmp, line 36: Wouldn't this be clearer? However, it's just a coding habit, and there's no issue if it's not changed.


```javascript
if (process.env.USE_SYSTEM_MKSQUASHFS) {
  return "mksquashfs"
}
if (process.env.MKSQUASHFS_PATH) {
  return process.env.MKSQUASHFS_PATH
}
return await getLinuxTool("mksquashfs")
```

packages/app-builder-lib/src/targets/snap/snapBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  // https://github.com/electron-userland/electron-builder/issues/3608
+  // even if electron-builder will correctly unset setgid/setuid, still, quite a lot of possibilities for user to create such incorrect permissions,
+  // so, just unset it using chmod right before packaging
+  const dirs = [stageDir, options.appDir, templateDir]
+  await Promise.all(
+    dirs
+      .filter(dir => !!dir)
+      .map(dir =>
+        exec(`chmod -R g-s ${dir}`).catch((err: any) => {
```
beyondkmp, line 128: How about using `fs.chmodSync`?

packages/app-builder-lib/src/binDownload.ts
```diff
@@ -34,7 +34,7 @@ export function getBinFromUrl(name: string, version: string, checksum: string):
       process.env.npm_package_config_electron_builder_binaries_custom_dir ||
       process.env.ELECTRON_BUILDER_BINARIES_CUSTOM_DIR ||
       dirName
-    const urlSuffix = dirName + ".7z"
+    const urlSuffix = fileName + ".7z"
```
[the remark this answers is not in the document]
mmaietta: Normally it is the same for electron-builder artifacts, but for some reason, the snap templates have a different artifact filename
- "snap-template-electron-4.0-1-armhf.tar" and "snap-template-electron-4.0-2-amd64.tar"
https://github.com/electron-userland/electron-builder/pull/8892/files#diff-9d1c29419fc0816dd25c22691e74bfd53bc266405cabaee89867716ab8dd22d3R50-R64

packages/app-builder-lib/src/targets/snap/linuxTools.ts
mmaietta: Yep, looks cleaner! I'll push a change

packages/app-builder-lib/src/targets/snap/snapBuilder.ts
mmaietta: Sure thing, I'l double check that. I was during purely a 1:1 migration from the app-builder-bin code, which was calling `chmod` directly, but I'll see if it's possible to leverage built-in fs functions here

packages/app-builder-lib/src/targets/snap/snapBuilder.ts
mmaietta: IIRC, the `mksquashfs` that comes from electron-builder-binaries doesn't support all compression algs, but this was a 1:1 copy from app-builder-bin. I'll investigate further

brew `mksquashfs`
```
Compressors available and compressor specific options:
        gzip (default)
        lzo
        lz4
        xz
        zstd
        lzma
```

electron-builder-binaries app-image tools `mksquashfs`
```
Compressors available and compressor specific options:
        gzip (default)
        xz
```

mmaietta: Confirming that `xz` | `lzo` is what the `SnapOptions` also provide: https://github.com/electron-userland/electron-builder/blob/ea86783728e5f59ac9625cd34dbc3c92a9d6f8fc/packages/app-builder-lib/src/options/SnapOptions.ts#L134-L137

mmaietta:
Confirming `isTemplateApp = false` also works on Fedora and Ubuntu arm64 VMs
Screenshot 2025-02-27 at 11 16 18 AM
Screenshot 2025-02-27 at 11 16 24 AM

beyondkmp (approved):

beyondkmp:
LGTM

beyondkmp:
The snap compiled with this version fails to install, with the following error:
```
➜  dist git:(master) ✗ sudo snap install --dangerous electron-quick-start-typescript_1.0.0_amd64.snap
error: cannot read snap file: open /tmp/read-file3906864710/unpack/meta/snap.yaml: no such file or
       directory
```

Using the same configuration with 26.0.10 works without any issues.
```
➜  dist git:(master) ✗ sudo snap install --dangerous electron-quick-start-typescript_1.0.0_amd64.snap
Ensure prerequisites for "electron-quick-start-typescript" are available                                                           -
electron-quick-start-typescript 1.0.0 installed
```

PR #8907, opened 2025-02-26
chore(test): migrating `jest` to `vitest`

mmaietta:
- Migrates project to vitest and updates snapshots
- Prep work (to reduce diff size) before introducing additional changes for test isolation + concurrency support


test/snapshots/mac/masTest.js.snap
```diff
@@ -10,13 +10,25 @@ exports[`dev 2`] = `
 {
   "CFBundleDisplayName": "Test App ßW",
   "CFBundleExecutable": "Test App ßW",
-  "CFBundleIconFile": "Test App ßW.icns",
+  "CFBundleIconFile": "icon.icns",
```
beyondkmp, line 13: Will there be any other issues since the name has been changed?

.gitignore
```diff
@@ -28,3 +28,4 @@ tsconfig.tsbuildinfo
 
 .pnpm-store/
 **/node_modules
+coverage
```
beyondkmp, line 31: I noticed an additional coverage directory. After switching to Vitest, will the test results display the UT coverage?

test/snapshots/mac/masTest.js.snap
mmaietta: No production code was changed in this PR. I was updating a snapshot localling as `masTest` was actually never running in the CI.
https://github.com/electron-userland/electron-builder/pull/8907/files#diff-e5263da2cd1e4118e9e993957ad0609307ef3c828d1a151d28fb9041545a1294L10-L14

.gitignore
mmaietta: Theoretically yes, but we don't have a way to run all the tests on a single node, so technically the coverage report will need some "merging" logic after coverage generation from each test runner.

beyondkmp (approved):

PR #8913, opened 2025-02-27
chore(refactor): electron-updater cyclical imports

mmaietta:
Refactoring code to reduce cyclical imports in order to migrate to rollup and/or vitest (which have much more strict module resolutions)


packages/electron-updater/src/exports.ts
```diff
@@ -0,0 +1,81 @@
...
+export interface ResolvedUpdateFileInfo {
+  readonly url: URL
+  readonly info: UpdateFileInfo
+
+  packageInfo?: PackageFileInfo
+}
+
+export type UpdaterEvents = "login" | "checking-for-update" | "update-available" | "update-not-available" | "update-cancelled" | "download-progress" | "update-downloaded" | "error"
+
+export type LoginHandler = (authInfo: any, callback: LoginCallback) => void
```
beyondkmp, line 81: Doesn't it make more sense to name this file types.ts since the code inside looks like type definitions?

beyondkmp (approved):

packages/electron-updater/src/types.ts
```diff
@@ -0,0 +1,81 @@
+import { CancellationToken, PackageFileInfo, ProgressInfo, UpdateFileInfo, UpdateInfo } from "builder-util-runtime"
```
lyswhut, line 1: Please export `ProgressInfo` and `UpdateInfo` in the `electron-updater` package

My app depends on `UpdateInfo` and `ProgressInfo` types, and I noticed that this change removed the exports of these two types :(

mmaietta: Oh no! Sorry about that, I must have missed the re-export. Will get that into the next release asap

PR #8920, opened 2025-03-02
chore(test): Adding test suite to cover multi-platform + multi-target + multi-arch configurations

mmaietta:
(Prep work for concurrency refactor)


mmaietta:
Note: Pausing work on this PR. We'll need this test case once parallel test execution and async targets are implemented, otherwise this added test suite will 2x the CI time duration.

PR #8957, opened 2025-03-13
fix: pnpm collection of optional dependencies

indutny-signal:
node-module-collector should not expect optional dependencies to be present on disk. If `require('dep/package.json')` throws for optional dependency - it is completley safe to continue as if it was never installed.


indutny-signal:
Sorry, it was supposed to be a patch release. I messed up the changeset at first

mmaietta:
Can you elaborate on this? I'm not sure I follow why it's safe to continue
>  If require('dep/package.json') throws for optional dependency - it is completley safe to continue as if it was never installed.

indutny-signal:
> Can you elaborate on this? I'm not sure I follow why it's safe to continue

The way pnpm module collector is written right now is that it will iterate through all dependencies both optional and not, and will `require('package.json')` of every single one of them. In our case we have `fs-xattr` in optional dependencies and it cannot be installed on Windows, so the build on Windows was failing with an error from pnpm collector saying that `fs-xattr` cannot be "require"d.

mmaietta:
Oh crap, that totally makes sense! Thanks for flagging this!

I wonder if we'll have the same issue with yarn/npm then with optional dependencies? I'll need to take a deeper look at the code now that my laptop has been repaired

mmaietta:
@indutny-signal can you share a sample package.json for a minimum repro? Basically just looking for anything I can use to write a unit test around.

Issue #8956, opened 2025-03-13
Cannot resolve node_modules require when building from yarn workspace


Lemonexe:
Application built by running `electron-builder` in a yarn workspace from repository root is broken.
Runtime dependencies are missing; it seems that `node_modules` are not crawled during build at all.
Suspected cause in https://github.com/electron-userland/electron-builder/pull/8571.
[Here is a reproduction example](https://github.com/Lemonexe/electron-builder-yarn-workspaces-repro) with branches for latest vs. last stable version (26.0.3).

### Notes

I used `is-odd` as a simplest possible dependency, but I think the usecase is mainly for packages with native code. 
I first discovered the problem with `blake-hash` package; our project uses Webpack for all of main, renderer and preload processes, so all JS-only packages are bundled, and only the native code packages remain as runtime requires, and get packed into `app.asar.unpacked`.
Nevertheless, the problem is reproducible with _any_ runtime require, even though `is-odd` goes to `app.asar` and not `app.asar.unpacked`.

### Possible duplicates

Possible duplicates of this issue, but they don't provide that much information so I can't tell 100%:
- https://github.com/electron-userland/electron-builder/issues/8858
- https://github.com/electron-userland/electron-builder/issues/8852

Also related to yarn workspaces, but internal import instead of external, so I don't believe it's the same thing:
- https://github.com/electron-userland/electron-builder/issues/8938

PR #8958, opened 2025-03-14
fix: optimize workspace package resolution in dependency tree

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8956

### DIFF
--- a/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
+++ b/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
@@ -147,12 +147,16 @@ export abstract class NodeModulesCollector<T extends Dependency<T, OptionalsType
 
   private getTreeFromWorkspaces(tree: T): T {
     if (tree.workspaces && tree.dependencies) {
+      const dependencyName = path.basename(this.rootDir)
+      const packagesName = path.basename(path.dirname(this.rootDir))
+      const scopedPackagesName = `@${packagesName}/${dependencyName}`
       for (const [key, value] of Object.entries(tree.dependencies)) {
-        if (this.rootDir.endsWith(path.normalize(key))) {
+        if (key === dependencyName || key === scopedPackagesName) {
           return value
         }
       }
     }
+
     return tree
   }
 


mmaietta:
Do we need to add a new fixture or at minimum, not commit the node_modules to the repo? Is it possible to use an existing fixture? Seems like we could simply use
```
await modifyPackageJson(projectDir, data => {
  data.dependencies = {
     "is-odd": "^3.0.1"
  }
}),
```
With:
```
isInstallDepsBefore: true
```

beyondkmp:
```
 test("yarn workspace for scope name", ({ expect }) =>
  assertPack(
    expect,
    "test-app-yarn-several-workspace",
    {
      targets: linuxDirTarget,
      projectDir: "packages/test-app",
    },
    {
      isInstallDepsBefore: true,
      projectDirCreated: projectDir => {
          return Promise.all([
            modifyPackageJson(projectDir, data => {
              console.log("fuck is here",data)
              data.name = "@scope/xxx-app"
              data.dependencies = {
                "is-odd": "3.0.1"
              }
            }),
          ])
        },

      packed: context => verifyAsarFileTree(expect, context.getResources(Platform.LINUX)),
    }
  ))
```
The modification of package.json is incorrectly targeting the project's root package.json instead of the one in packages/test-app directory. This appears to be a bug.

beyondkmp:
@mmaietta I'll have to concatenate it manually. The unit test issues have been resolved. Please help review it when you have time.

mmaietta:
>The modification of package.json is incorrectly targeting the project's root package.json instead of the one in packages/test-app directory. This appears to be a bug.

Not a bug, it's designed to work with both one-package.json and two-package.json setups but with the `projectDir` expected to be under `app` and not `test-app` (so part of the test-app) fixture. To flip to two-package.json "mode", it's simply setting the optional `isApp` boolean
https://github.com/electron-userland/electron-builder/blob/7ba4fea95825650f02749949632b351c75d3019a/test/src/helpers/packTester.ts#L546-L547

Much cleaner diff now though, thank you for making the modifications!

mmaietta (approved):

Lemonexe:
Thank you, @beyondkmp :1st_place_medal: 
With `app-builder-lib` built from `master`, I confirmed it to be working both in the minimal repro, as well as in our slightly more complex production app :heavy_check_mark: :+1:

PR #8959, opened 2025-03-14
fix: generate stubExecutableExe and sign it

beyondkmp:


packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -33,16 +32,31 @@ export default class SquirrelWindowsTarget extends Target {
...
       }
     }
-
     return tmpVendorDirectory
   }
 
+  private async generateStubExecutableExe(appOutDir: string, vendorDir: string) {
+    const files = await fs.promises.readdir(appOutDir, { withFileTypes: true })
+    for (const file of files) {
+      if (file.isFile() && file.name.toLocaleLowerCase().endsWith(".exe") && file.name.toLocaleLowerCase() !== "squirrel.exe") {
```
mmaietta, line 48: Maybe we can use `path` package for comparisons?
```
path.extname(file.name) === ".exe" && path.basename(file.name, "exe") !== "Squirrel"
```

packages/electron-builder-squirrel-windows/template.nuspectemplate
```diff
@@ -19,6 +19,7 @@
     <file src="*.pak" target="lib\net45" />
     <file src="*.exe.config" target="lib\net45" />
     <file src="*.exe.sig" target="lib\net45" />
+    <file src="*_ExecutionStub.exe" target="lib\net45" />
```
mmaietta, line 22: Why is this now being added?

beyondkmp: Since electron-builder now generates and signs *_ExecutionStub.exe, adding this configuration will move the generated *_ExecutionStub.exe to the lib\net45 directory.

t3chguy:
Seems to fail to build at least for ARM64 package: https://github.com/element-hq/element-desktop/actions/runs/13895690921/job/38875908510?pr=2211

t3chguy:
Looks like `package.json` `files` needs updating to include `vendor` dir

t3chguy:
image

Looks like it works sans the package.json not including vendor in the package - good job @beyondkmp 

I also checked that the number of signings remained the same

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -33,16 +32,31 @@ export default class SquirrelWindowsTarget extends Target {
...
-
     return tmpVendorDirectory
   }
 
+  private async generateStubExecutableExe(appOutDir: string, vendorDir: string) {
+    const files = await fs.promises.readdir(appOutDir, { withFileTypes: true })
+    for (const file of files) {
+      if (path.extname(file.name) === ".exe" && path.basename(file.name, "exe") !== "Squirrel") {
+        const filePath = path.join(appOutDir, file.name)
+        log.debug({ file: filePath }, "generating stub executable for exe")
```
mmaietta, line 50: Can use `log.filePath(filePath)` here

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -33,16 +32,31 @@ export default class SquirrelWindowsTarget extends Target {
...
+        const fileNameWithoutExt = file.name.slice(0, -4)
```
mmaietta, line 51: Can extract `path.basename(file.name, "exe")` from if-statement to a `const` and then reuse here instead of using `slice` since `basename` w/ ext supplied **should** get us the `filenameWithoutExt`

mmaietta (commented):
Adding the vendor directory adds a few megabytes (quick maffs) to the repo, which is the antipattern from what electron-builder-binaries is supposed to be used for. IIRC though, the squirrel windows target/package must be installed separately and isn't part of the electron-builder dependency tree, right? In this case, I think it's safe to add the vendor files (albeit I'll still need to verify file origin). In general though, we should avoid adding vendor files directly to electron-builder unless absolutely necessary (which is the case with this fix)

t3chguy:
> the squirrel windows target/package must be installed separately and isn't part of the electron-builder dependency tree, right? 

Yup, `electron-builder-squirrel-windows` https://www.npmjs.com/package/electron-builder-squirrel-windows is not a transitive dependency of `electron-builder`

beyondkmp:
> Looks like `package.json` `files` needs updating to include `vendor` dir

my bad. Added.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: Fixed.

beyondkmp:
I suggest directly placing the vendor into electron-builder-squirrel-windows. Later, we can upgrade the 7zip/nuget and other files within it, as these are all 10-year-old versions with security and performance issues. After updating, the packaging performance will also improve significantly.

t3chguy:
@beyondkmp / @mmaietta any update here?

beyondkmp:
@t3chguy 
We hope that the upstream will merge this MR（https://github.com/Squirrel/Squirrel.Windows/pull/1903/files） so that an official upstream version of the Squirrel Windows vendor can be released. However, it seems that there has been no activity from the upstream.

beyondkmp:
@mmaietta Do you have any thoughts on this MR?

beyondkmp:
@mmaietta I think it’s fine. I’ve already cloned it, and it can run successfully on GitHub Actions, and the build completes without issues. You can refer to this directly.

beyondkmp:
It probably won't work well, as the newer version of Squirrel Windows has changed some things. My suggestion is still to use the Windows Installer. When there’s a new version, it can be synced directly. If we use our own implementation, some issues may not be synced in time.

mmaietta:
>You can refer to this directly.
https://github.com/beyondkmp/Squirrel.Windows

I'm against using a fork for the project.

I'd rather `git clone` the source repo in a build container, apply your changes as a patch file, then build the artifact fresh for Squirrel.Windows within the build container/runner, which then would be published directly from electron-builder-binaries. (I actually have this approach ready in my current work on the electron-builder-binaries CI pipeline)
Is it possible to provide our own vendor dir for electron-winstaller? And/or leverage our own signing flow or something like that?

If changes are needed in a dedicated `electron/*` repo, I can definitely sync up with the electronHQ folks to discuss the topic live during one of our biweekly syncs

beyondkmp:
What I mean is that you can refer to how this fork uses actions to automatically generate builds, rather than saying that you should use this fork directly.

mmaietta:
The purpose of electron-builder-binaries is to provide vendor files on-demand based on the use-case/packaging targets required for a build. I'm against committing any vendor files to this project directly

beyondkmp:
> The purpose of electron-builder-binaries is to provide vendor files on-demand based on the use-case/packaging targets required for a build. I'm against committing any vendor files to this project directly. Willing to make exceptions though, just not for files that are measured in megabytes
ok, got it. I'll set it to draft status first, and once the Squirrel.Windows update in electron-builder-binaries is complete, we can come back to update this PR.

mmaietta:
@beyondkmp new squirrel.windows in electron-builder-binaries with your patch applied can be found in this "compile" PR. (The CI/CD takes care of generating and committing the artifacts post-merge of the PR through Changesets GHA)
https://github.com/electron-userland/electron-builder-binaries/pull/67

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -20,31 +22,55 @@ export default class SquirrelWindowsTarget extends Target {
...
     const files = await fs.promises.readdir(tmpVendorDirectory)
     for (const file of files) {
-      if (["Squirrel.exe", "StubExecutable.exe"].includes(file)) {
+      if (file === "Squirrel.exe") {
         const filePath = path.join(tmpVendorDirectory, file)
         log.debug({ file: filePath }, "signing vendor executable")
         await this.packager.sign(filePath)
+        break
       }
     }
```
mmaietta, line 55: Minor: Can avoid the `for-if-break` approach with a `find`
```
const file = files.find(f => f === "Squirrel.exe")
```
That being said, it looks like we're just trying to sign that one file, right?
```
const filePath = path.join(tmpVendorDirectory, "Squirrel.exe")
if (filePath) { // do we need this check or should Squirrel.exe always be present?
   log.debug({ file: filePath }, "signing vendor executable")
   await this.packager.sign(filePath)
}
```

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -20,31 +22,55 @@ export default class SquirrelWindowsTarget extends Target {
...
+        "squirrel.windows",
+        "https://github.com/electron-userland/electron-builder-binaries/releases/download/squirrel.windows@1.0.0/squirrel.windows-2.0.1-patched.7z",
+        "DWijIRRElidu/Rq0yegAKqo2g6aVJUPvcRyvkzUoBPbRasIk61P6xY2fBMdXw6wT17md7NzrTI9/zA1wT9vEqg=="
+      )
     }
 
     const tmpVendorDirectory = await this.packager.info.tempDirManager.createTempDir({ prefix: "squirrel-windows-vendor" })
-    // Copy entire vendor directory to temp directory
     await fs.promises.cp(vendorDirectory, tmpVendorDirectory, { recursive: true })
+    if (customSquirrelBin && customSquirrelBin.length > 0) {
```
mmaietta: I'm not sure what this is checking for. Are you intending?
`!isEmptyOrSpaces(customSquirrelBin) && await exists(customSquirrelBin)`

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -20,31 +22,55 @@ export default class SquirrelWindowsTarget extends Target {
...
+      if (path.extname(file.name) === ".exe" && fileNameWithoutExt !== "Squirrel") {
+        const filePath = path.join(appOutDir, file.name)
+        log.filePath(filePath)
+        const stubExePath = path.join(appOutDir, `${fileNameWithoutExt}_ExecutionStub.exe`)
+        await fs.promises.copyFile(path.join(vendorDir, "StubExecutable.exe"), stubExePath)
+        await execWine(path.join(vendorDir, "WriteZipToSetup.exe"), null, ["--copy-stub-resources", filePath, stubExePath])
+        await this.packager.sign(stubExePath)
+      }
+    }
+  }
```
mmaietta, line 73: Wait, why do we need to copy a `${fileNameWithoutExt}_ExecutionStub.exe` for every `.exe` file in the appOutDir? Is that to be expected or are there only specific files that we need the stub for?

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: Good suggestion. Fix it.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: There was a bit of ambiguity in the wording. What I meant is: if the user has specified a custom Squirrel, we should not overwrite it with our patched version.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: Fix it.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -20,31 +22,57 @@ export default class SquirrelWindowsTarget extends Target {
...
+      const filePath = path.join(appOutDir, appExe.name)
+      log.filePath(filePath)
+      const stubExePath = path.join(appOutDir, `${this.appName}_ExecutionStub.exe`)
+      await fs.promises.copyFile(path.join(vendorDir, "StubExecutable.exe"), stubExePath)
+      await execWine(path.join(vendorDir, "WriteZipToSetup.exe"), null, ["--copy-stub-resources", filePath, stubExePath])
+      await this.packager.sign(stubExePath)
+      log.debug({ file: filePath }, "signing app executable")
+      await this.packager.sign(filePath)
+    } else {
+      log.warn("App executable not found in app directory, skipping signing")
```
mmaietta: Wait, should this edge case ever exist or would it be a broken installer/app? Wondering if we should just
```
if (!appExe) {
   throw new Error("App executable not found in app directory")
}
```

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -20,31 +22,57 @@ export default class SquirrelWindowsTarget extends Target {
...
-      }
+    if (isEmptyOrSpaces(customSquirrelVendorDirectory) || !fs.existsSync(customSquirrelVendorDirectory)) {
+      log.warn({ customSquirrelVendorDirectory: customSquirrelVendorDirectory }, "unable to access custom Squirrel.Windows vendor directory, falling back to default vendor ")
+      const windowInstallerPackage = require.resolve("electron-winstaller/package.json")
+      const vendorDirectory = path.join(path.dirname(windowInstallerPackage), "vendor")
+
+      const squirrelBin = await getBin(
+        "squirrel.windows",
+        "https://github.com/electron-userland/electron-builder-binaries/releases/download/squirrel.windows@1.0.0/squirrel.windows-2.0.1-patched.7z",
+        "DWijIRRElidu/Rq0yegAKqo2g6aVJUPvcRyvkzUoBPbRasIk61P6xY2fBMdXw6wT17md7NzrTI9/zA1wT9vEqg=="
```
mmaietta, line 36: From some recent code changes, I think this should now be able to be simplified to
```
getBinFromUrl("squirrel.windows@1.0.0", "squirrel.windows-2.0.1-patched.7z", hash)
```

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: Yes, you are right. Fixed.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: Fixed

beyondkmp: There is an ut error
```
builder\Cache\winCodeSign\295934459 C:\Users\runneradmin\AppData\Local\electron-builder\Cache\winCodeSign\winCodeSign-2.6.0: Access is denied.
  • downloading     url=https://github.com/electron-userland/electron-builder-binaries/releases/download/nsis-3.0.4.1/nsis-3.0.4.1.7z size=1.3 MB parts=1
  • downloaded      url=https://github.com/electron-userland/electron-builder-binaries/releases/download/nsis-3.0.4.1/nsis-3.0.4.1.7z duration=202ms
  ⨯ Get "squirrel.windows-2.0.1-patched.7z": unsupported protocol scheme ""
```

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -20,31 +22,57 @@ export default class SquirrelWindowsTarget extends Target {
...
-        log.debug({ file: filePath }, "signing vendor executable")
-        await this.packager.sign(filePath)
-      }
+    if (isEmptyOrSpaces(customSquirrelVendorDirectory) || !fs.existsSync(customSquirrelVendorDirectory)) {
+      log.warn({ customSquirrelVendorDirectory: customSquirrelVendorDirectory }, "unable to access custom Squirrel.Windows vendor directory, falling back to default vendor ")
+      const windowInstallerPackage = require.resolve("electron-winstaller/package.json")
+      const vendorDirectory = path.join(path.dirname(windowInstallerPackage), "vendor")
+
+      const squirrelBin = await getBinFromUrl(
+        "squirrel.windows@1.0.0",
```
beyondkmp, line 34: A special `@` character in the path causes the failure.

t3chguy:
Hey @beyondkmp - any light at the end of this tunnel?

beyondkmp:
@mmaietta Please help review again when you have a chance.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -20,31 +22,57 @@ export default class SquirrelWindowsTarget extends Target {
...
 
+  private async generateStubExecutableExe(appOutDir: string, vendorDir: string) {
+    const files = await fs.promises.readdir(appOutDir, { withFileTypes: true })
+    const appExe = files.find(f => f.name === `${this.exeName}.exe`)
+    if (!appExe) {
+      throw new Error(`App executable not found in app directory: ${appOutDir}`)
+    }
+
+    const filePath = path.join(appOutDir, appExe.name)
+    log.filePath(filePath)
```
mmaietta: Can this be removed?

mmaietta (dismissed):

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
beyondkmp: deleted

Issue #8888, opened 2025-02-22
suggestion: parallel builds


xyloflake:
So recently, while working on ci, I had to configure electron builder to build for both x64 and aarch64/arm64.

I was pretty frustrated by the build times (we have to build for a lot of linux platforms + 2 architectures) which was around 16 minutes. I tried something to reduce it. I broke down the 2 architecture into 2 commands (for linux only):

1. release normally (x86_64)
```bash
electron-builder --publish always
```
2. release for aarch64/arm64
```bash
electron-builder --publish always --arm64
```

I then used bash to execute both simultaneously
```bash
electron-builder --publish always & \
electron-builder --publish always --arm64 & \
wait
````

This executed both at the same time and the total time was now down to 9 minutes from 16.

What this experiment has proven is that electron builder could explore the possibilities of building binaries as parallel jobs, which may speed up build times exponentially depending on cpu.

That about configuration on number of parallel jobs, by default I suggest setting it to `NUM_CPU_CORES` and if required, modify through electron-builder.json through some property like 
```json
{
  "parallelJobs": 4
}
```

I've also thought it out and figured out a possibility of race conditions while publishing builds to github (or similar service). For this I suggest waiting for all builds to finish and then uploading all the assets at once in one single job.

beyondkmp:
Great suggestion. We also previously patched the build process to enable parallel compilation. However, on Windows, signing would often fail due to file locks, causing the entire build to fail. We later switched to running multiple Docker containers for parallel compilation.

If we can resolve the issue of file locks during signing, it should be possible to implement concurrent compilation.

xyloflake:
> We also previously patched the build process to enable parallel compilation.

Do you mind explaining this or reference a relevant commit so I can take a look?

> If we can resolve the issue of file locks during signing, it should be possible to implement concurrent compilation.
Thank you, I believe if we do have problems with windows, we should start out with linux and slowly work the way up with supporting all platforms as we understand the challenges introduced. I am mostly worried about race conditions (such as reading the unpacked folder at the same time could result in `EBUSY` and could fail the build if not handled correctly).

I will think about that to see if I can come up with something.

xyloflake:
@beyondkmp do you think the builds could be temporarily placed in specific folders and the lockfile could be cloned for each one of the builds and then we execute signing, which can use their specific lockfiles?

xyloflake:
@mmaietta would love it if you got some suggestions

mmaietta:
[...]

PR #8962, opened 2025-03-16
chore(refactor): explore parallelization of builds

mmaietta:
Allow targets to be executed in parallel for those that do not include locking of resources (macos dmg: hdiutil, win: squirrel/nsis/msi signing, etc)

Adds a massive ConcurrentBuildsTest suite for making sure all assets are produced correctly and to measure speed improvements against

Implements https://github.com/electron-userland/electron-builder/issues/8888

CC @xyloflake if you're willing to be an early adopter of the experimental implementation


xyloflake:
looks great! will make efforts to implement this for us!
cc @staniel359 @gamersi

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
```diff
@@ -401,7 +410,8 @@ export class NsisTarget extends Target {
 
     // https://github.com/electron-userland/electron-builder/issues/2103
     // it is more safe and reliable to write uninstaller to our out dir
-    const uninstallerPath = path.join(this.outDir, `__uninstaller-${this.name}-${this.packager.appInfo.sanitizedName}.exe`)
+    // to support parallel builds, the uninstaller path must be unique to each target and arch combination
+    const uninstallerPath = path.join(this.outDir, `__uninstaller-${Math.floor(Math.random() * Date.now())}-${this.name}-${this.packager.appInfo.sanitizedName}.exe`)
```
beyondkmp, line 414: Will there be any issues during user upgrades or overlay installations? Since the paths change each time, will uninstalling previous versions fail due to missing paths?

packages/app-builder-lib/src/packager.ts
```diff
@@ -479,6 +481,18 @@ export class Packager {
       const nameToTarget: Map<string, Target> = new Map()
       platformToTarget.set(platform, nameToTarget)
 
+      let poolCount = Math.floor(packager.config.concurrency?.jobs || 1)
+      if (poolCount < 1) {
+        log.warn({ concurrency: poolCount }, "concurrency is invalid, overriding with job count: 1")
+        poolCount = 1
+      } else if (poolCount > MAX_FILE_REQUESTS) {
```
beyondkmp, line 488: Can delete `else`

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
beyondkmp: I suggest using fixed paths, such as including the type (nsis/squirrel) in the path, rather than generating random paths. Some uninstallation processes require these fixed paths for proper removal.

beyondkmp: The uninstallerPath is passed to install.nsi through UNINSTALLER_OUT_FILE.

https://github.com/electron-userland/electron-builder/blob/e41c24ca47165a363a2ebb847416058f20047b9c/packages/app-builder-lib/templates/nsis/installer.nsi#L56-L58

Apps registered through registry or in Settings/Apps/Uninstall Apps may have issues, we need to test and verify.

```
  WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\${APP_NAME}" "UninstallString" '"${UNINSTALLER_OUT_FILE}"'
```

packages/app-builder-lib/src/packager.ts
mmaietta: Wait, why?

xyloflake: > Wait, why?
Same question, I believe that is a necessary conditional

packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
mmaietta: If we use a fixed path, then we completely cannot have parallel builds of nsis target.
We could theoretically switch it back to a tmp-dir with a fixed name, but then that goes explicitly against the comment above it: `// it is more safe and reliable to write uninstaller to our out dir`

The problem is that I can't identify the **why** based off the issue linked in the comment.

packages/app-builder-lib/src/packager.ts
beyondkmp: Apologized for the misunderstanding.
I mean

```javascript
 if (poolCount < 1) {
        log.warn({ concurrency: poolCount }, "concurrency is invalid, overriding with job count: 1")
        poolCount = 1
  } 
 if (poolCount > MAX_FILE_REQUESTS) {}
 ```

File: packages/app-builder-lib/src/targets/nsis/NsisTarget.ts (hunk 1 above)
Comment (mmaietta) [reply packages/app-builder-lib/src/targets/nsis/NsisTarget.ts:414]: @beyondkmp @xyloflake what if we just had a unique directory name within the `out` directory? Then we keep the uninstaller name as a constant? That way we still respect the comment of "reliability outputting to `out` dir" while maintaining the static name. 

Previous implementations were these, so I think we'd still be satisfying the original intent while not introducing regressions:
````
packager.getTempFile("uninstaller.exe") // completely temp dir/file elsewhere on the filesystem
```
```
path.join(this.outDir, `.__uninstaller-${this.name}-${process.pid.toString(16)}-${Date.now().toString(16)}.exe`) // within outDir but non-static file name
```

xyloflake: > @beyondkmp @xyloflake what if we just had a unique directory name within the `out` directory? Then we keep the uninstaller name as a constant? That way we still respect the comment of "reliability outputting to `out` dir" while maintaining the static name.
> 
> Previous implementations were these, so I think we'd still be satisfying the original intent while not introducing regressions:
> 
> ```
> packager.getTempFile("uninstaller.exe") // completely temp dir/file elsewhere on the filesystem
> ```
> ```
> path.join(this.outDir, `.__uninstaller-${this.name}-${process.pid.toString(16)}-${Date.now().toString(16)}.exe`) // within outDir but non-static file name
> ```
> ```

I personally think that is a GREAT idea. I think it should be something like: if concurrency is enabled, then it should use the unique directory in the out folder and if it doesn't have concurrency enabled then it should fallback to this old method. But I think even if there is no fallback, this is a great idea and I think it's pretty necessary for parallelization!

mmaietta: Note: the current implementation only applies to the **temporary** uninstaller file. Changing the output directories of other artifacts would unfortunately require a breaking/major semver bump. Changing just the output when concurrency isn't enabled could indeed work, but there's no identifying target/arch at this stage of the function. I had actually investigated that but don't recall identifying any way to refactor it.

mmaietta: @beyondkmp, thinking on this further, I think we can just use the `installerPath` since it should always be unique to each installer being built, regardless of archs since that's calculated previously.
```
const uninstallerPath = path.join(this.outDir, `${path.basename(installerPath, "exe")}__uninstaller.exe`)
```

beyondkmp:
LGTM

beyondkmp (approved):

mmaietta:
@xyloflake have you been able to try this feature out? Curious how well it's working for you :)

xyloflake:
> @xyloflake have you been able to try this feature out? Curious how well it's working for you :)

@mmaietta hey. Sorry, haven't been able to test it yet, I should be able to do it soon since my examinations end tomorrow. For the record, what do I have to do to test this? Is it in the official builds already?

beyondkmp:
@mmaietta I thought of a potential issue: if you run `electron-builder --win --x64 --arm64` on a single machine to build both x64 and arm64 simultaneously while enabling parallel processing, there could be problems. During `npm rebuild`, is it possible that the native module for arm64 could end up being packaged into the x64 build?

mmaietta:
IIRC, the native compilation step happens before the parallel processing of the packaging async processing, but I can double check @beyondkmp 

@xyloflake it's in the most recent release under `26.0.12` IIRC. Just add `concurrency.jobs = <any number>` to your config

xyloflake:
@mmaietta I've started working on this. Just repatching and making sure everything works before proceeding. Will update on this.

PR #9007, opened 2025-04-03
fix(mac): allow ad-hoc identities for codesigning

### LINKED ISSUE #5850: Universal build fails on non-ARM64 hardware, and M1 target DMG is 'damaged'
I've got an app which I build hosted in an Electron container, to give me some native/filesystem interop. It works well.

However, for some reason, if I build the M1/ARM64 version (or a Universal build) on GitHub actions, I get a 'damaged' DMG file (for the ARM64 build) and an odd Javascript/Node error for the universal build:

```
TypeError [ERR_INVALID_CALLBACK]: Callback must be a function
at makeCallback (fs.js:136:11)
at rmdir (fs.js:671:14)
at Promise (internal/util.js:274:30)
at new Promise (<anonymous>)
at rmdir (internal/util.js:273:12)
at MacPackager.doPack (/Users/runner/work/Damselfly/Damselfly/Damselfly.Desktop/node_modules/app-builder-lib/src/macPackager.ts:123:15)

```
You can see the build failure for the Universal version [here](https://github.com/Webreaper/Damselfly/runs/2486907712?check_suite_focus=true#step:4:34). The ARM64 build completes, but the resulting App fails to start, with MacOS saying it's 'damaged'.

Building locally on my M1 MacBook pro works fine. My hunch is that the GitHub Actions runners aren't MacOS ARM64 hardware, so can't build the M1 versions. But that's weird, because my understanding is that electron-builder should be able to build ARM64/Universal targets on an Intel Mac (as per [this issue](https://github.com/electron-userland/electron-builder/issues/5689). I'm sure it used to work about 3-4 weeks ago, and I haven't seen any references to Electron build needing to run on M1 hardware to build M1 targets.

I'm using Electron 11.0.0, electron-builder 22.10.4, node 10.24.1. The OS that I'm running this on is 11.4 beta, on an M1 Macbook pro.

kthchew:
Fixes #5850

If no signing identity is present while an app is built and it is downloaded via a web browser, macOS on ARM shows "app is damaged" errors that are hard to bypass for the average user (see #5850 and some other duplicates). This seems to be because the default signature applied by the linker to the executables isn't valid due to something the build system is doing [1]. It is thus an error to not manually codesign the resulting app, even if no stable signing identity is available (in which case it must be an ad-hoc signature).

The damaged warning is **not** the same as the unverified warning. Not paying for a developer account does not mean your app must show "this app is damaged" warnings to users, but it will show an unverified warning. The damaged warning cannot be bypassed through the GUI (such as in System Settings) and requires running commands to remove the quarantine attribute (as of the current macOS version). An unverified warning, on the other hand, has [supported GUI methods of bypassing the warning and running the app](https://support.apple.com/en-us/102445#openanyway). A proper (free) ad-hoc signature means that the damaged warning will not show up, but the unverified warning will.

Since there are no problems for those with "proper" signing certificates, it follows that the existing signing system does make a valid signature, and if it can be repurposed to accept empty identities (for ad-hoc signatures), this will resolve the issue. This PR does two main things:

- Add an option to use ad-hoc signatures by passing `-` as the `identity`. This is the same convention that the `codesign` command uses.
- Use ad-hoc signatures as a fallback by default when building ARM or universal builds if no valid identity is present. This can be overridden by setting the `identity` to `null`.

I tested these changes with FreeTube, which previously showed "'FreeTube' is damaged" warnings.
Before these changes, the warning looks like the below and can't be bypassed in System Settings.
image

With these changes, the new warning looks like
Screenshot 2025-04-02 at 4 57 11 PM
And System Settings shows an option to bypass the warning like it does on Intel.
Screenshot 2025-04-02 at 4 57 22 PM

---

[1] For additional details if you're curious, see the [macOS 11.0.1 release notes](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-universal-apps-release-notes#New-Features). Specifically (emphasis mine):
> To reduce the impact on existing development workflows, starting with Xcode 12 beta 4, the toolchain will now automatically sign your executables whenever you build from Xcode, or use command-line utilities such as clang(1) or ld(1). Since this signing mechanism generates signatures directly at link time, and **doesn’t cover any resource other than the executable**, it is expected to be faster than a traditional codesign(1) invocation. If you **use a custom workflow involving tools that modify a binary after linking** (e.g. strip or install_name_tool) you might need to manually call codesign(1) as an additional build phase to properly ad-hoc sign your binary.

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -238,7 +238,7 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
     const isMas = masOptions != null
     const options = masOptions == null ? this.platformSpecificBuildOptions : masOptions
     const qualifier = options.identity
-    const fallBackToAdhoc = arch === Arch.arm64 || arch === Arch.universal
+    const fallBackToAdhoc = (arch === Arch.arm64 || arch === Arch.universal) && !this.forceCodeSigning
 
     if (qualifier === null) {
       if (this.forceCodeSigning) {
@@ -273,17 +273,15 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
         }
       }
 
+      const noIdentity = !options.sign && identity == null
       if (qualifier === "-") {
         identity = new Identity("-", undefined)
-      }
-
-      if (!options.sign && identity == null) {
-        await reportError(isMas, certificateTypes, qualifier, keychainFile, this.forceCodeSigning, fallBackToAdhoc)
-        if (fallBackToAdhoc) {
-          identity = new Identity("-", undefined)
-        } else {
-          return false
-        }
+      } else if (noIdentity && fallBackToAdhoc) {
+        log.warn(null, "falling back to ad-hoc signature for macOS application code signing")
+        identity = new Identity("-", undefined)
+      } else if (noIdentity) {
+        await reportError(isMas, certificateTypes, qualifier, keychainFile, this.forceCodeSigning)
+        return false
       }
     }
 
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -1,6 +1,7 @@
 import { notarize } from "@electron/notarize"
 import { NotarizeOptionsNotaryTool, NotaryToolKeychainCredentials } from "@electron/notarize/lib/types"
 import { PerFileSignOptions, SignOptions } from "@electron/osx-sign/dist/cjs/types"
+import { Identity } from "@electron/osx-sign/dist/cjs/util-identities"
 import { Arch, AsyncTaskManager, copyFile, deepAssign, exec, getArchSuffix, InvalidConfigurationError, log, orIfFileNotExist, statOrNull, unlinkIfExists, use } from "builder-util"
 import { MemoLazy, Nullish } from "builder-util-runtime"
 import * as fs from "fs/promises"
@@ -8,7 +9,7 @@ import { mkdir, readdir } from "fs/promises"
 import { Lazy } from "lazy-val"
 import * as path from "path"
 import { AppInfo } from "./appInfo"
-import { CertType, CodeSigningInfo, createKeychain, CreateKeychainOptions, findIdentity, Identity, isSignAllowed, removeKeychain, reportError, sign } from "./codeSign/macCodeSign"
+import { CertType, CodeSigningInfo, createKeychain, CreateKeychainOptions, findIdentity, isSignAllowed, removeKeychain, reportError, sign } from "./codeSign/macCodeSign"
 import { DIR_TARGET, Platform, Target } from "./core"
 import { AfterPackContext, ElectronPlatformName } from "./index"
 import { MacConfiguration, MasConfiguration } from "./options/macOptions"
@@ -229,7 +230,7 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
     }
   }
 
-  private async sign(appPath: string, outDir: string | null, masOptions: MasConfiguration | null, arch: Arch | null): Promise<boolean> {
+  private async sign(appPath: string, outDir: string | null, masOptions: MasConfiguration | null, arch: Arch): Promise<boolean> {
     if (!isSignAllowed()) {
       return false
     }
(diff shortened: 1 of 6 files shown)


packages/app-builder-lib/src/macPackager.ts
```diff
@@ -268,9 +273,17 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
...
       if (!options.sign && identity == null) {
-        await reportError(isMas, certificateTypes, qualifier, keychainFile, this.forceCodeSigning)
-        return false
+        await reportError(isMas, certificateTypes, qualifier, keychainFile, this.forceCodeSigning, fallBackToAdhoc)
+        if (fallBackToAdhoc) {
+          identity = new Identity("-", undefined)
+        } else {
+          return false
+        }
       }
```
mmaietta, line 287: This part seems a little messy as I'm generally averse to nesting logic when not necessary and I'm seeing `reportError` still having a `throw new Error()`, so the `fallBackToAdhoc` would never be executed.  Maybe this would work? 
```
const noIdentity = !options.sign && identity == null
if (qualifier === "-" || (noIdentity && fallBackToAdhoc)) {
    identity = new Identity("-", undefined)
} else if (noIdentity) {
    await reportError(isMas, certificateTypes, qualifier, keychainFile, this.forceCodeSigning)
    return false
}
```

mmaietta: I still think `reportError` should only be used for throwing if no certificates are valid, whereas this `noIdentity && fallBackToAdhoc` could just include a `log.info` logic flow saying that it was selected as a fallback 
```
const noIdentity = !options.sign && identity == null
if (qualifier === "-") {
    identity = new Identity("-", undefined)
} else if (noIdentity && fallBackToAdhoc) {
    log.info("falling back to ad-hoc signature for macOS application code signing")
    identity = new Identity("-", undefined)
} else if (noIdentity) {
    await reportError(isMas, certificateTypes, qualifier, keychainFile, this.forceCodeSigning)
    return false
}
```
Thoughts?

kthchew: I've applied those changes. I also adjusted `fallBackToAdhoc` to avoid falling back if force code signing is enabled, since allowing ad-hoc signing with that setting enabled would probably be unexpected for most people. I also set the log message to the warn level.

mmaietta: Nice catch!

kthchew:
I found confirmation that `--sign -` has existed as an option since at least 2012, and perhaps longer (all commits on this repo were on the same date). https://github.com/Apple-FOSS-Mirror/security_systemkeychain/blame/68cbc1193007950c8c49aa63423c2e0bad4410a0/src/cs_utils.cpp#L195-L197

beyondkmp:
Will cause an error
```

dyld[3352]: Library not loaded: @rpath/Electron Framework.framework/Electron Framework
  Referenced from: <4C4C44F1-5555-3144-A109-371CAA64827A> /Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/MacOS/Cherry Studio
  Reason: tried: '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (code signature in <4C4C4421-5555-3144-A1F6-E07CD7537C3E> '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs), '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (code signature in <4C4C4421-5555-3144-A1F6-E07CD7537C3E> '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs)

```

kthchew:
> Will cause an error
> 
> ```
> 
> dyld[3352]: Library not loaded: @rpath/Electron Framework.framework/Electron Framework
>   Referenced from: <4C4C44F1-5555-3144-A109-371CAA64827A> /Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/MacOS/Cherry Studio
>   Reason: tried: '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (code signature in <4C4C4421-5555-3144-A1F6-E07CD7537C3E> '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs), '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Electron Framework' (code signature in <4C4C4421-5555-3144-A1F6-E07CD7537C3E> '/Users/payne/Code/cherry-studio/dist/mac-arm64/Cherry Studio.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs)
> ```

Use the fix described at https://github.com/electron-userland/electron-builder/issues/5850#issuecomment-2887254139 if you have manually set entitlements and didn’t include the one described there.

Issue #9000, opened 2025-03-31
Required dependencies not included in app.asar


renkei:
No issues with electron-builder 25.1.8.
When switching to electron-builder 26.0.12, resulting electron app crashes on start because of the missing dependency 'dunder-proto/get'.

Electron App forks a process, running an Express.js web server.

```
Error: Cannot find module 'dunder-proto/get'
Require stack:
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\get-proto\index.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\get-intrinsic\index.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\side-channel-map\index.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\side-channel\index.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\qs\lib\stringify.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\qs\lib\index.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\express\lib\middleware\query.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\express\lib\application.js
- c:\MyElectronApp\bin\Electron\win-unpacked\resources\app.asar\node_modules\express\lib\express.js
```

I've checked the last package in the require stack 'get-proto', its package.json correctly contains 'dunder-proto' in the dependencies section:

```
// from package.json of 'get-proto'
"dependencies": {
    "dunder-proto": "^1.0.1",
    "es-object-atoms": "^1.0.0"
}
```
I've disabled `asar` just to check the node_modules folder created by the electron-builder and I can see that 'dunder-proto' is missing. Too aggressive pruning?

Switching back to electron-builder 25.1.8 fixes the issue. I kept asar disabled and run the electron-builder 25.1.8 again, just to compare the resulting node_modules folders:

electron builder 26.0.12 creates a node_modules folder with 183 packages and 31.5MB from my project.
electron builder 25.1.8 creates a node_modules folder with only 163 packages but 33.7MB from my project. And 'dunder-proto' is part of it.

[...]

Issue #9006, opened 2025-04-02
"Cannot find module" errors after upgrade to electron-builder 26


KillerCodeMonkey:
I am using the node-module "form-data" in a project and after upgrade to electron builder from v25 to v26 i get the following error after starting the final app package.

```ts
A JavaScript error occurred in the main process
Uncaught Exception:
Error: Cannot find module 'delayed-stream'
Require stack:
- /tmp/.mount_APP Co8bvpMS/resources/app.asar/node_modules/combined-stream/lib/combined_stream.js
- /tmp/.mount_APP Co8bvpMS/resources/app.asar/node_modules/form-data/lib/form_data.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1232:15)
    at s._resolveFilename (node:electron/js2c/browser_init:2:129170)
    at Module._load (node:internal/modules/cjs/loader:1062:27)
    at c._load (node:electron/js2c/node_init:2:18008)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:227:24)
    at Module.require (node:internal/modules/cjs/loader:1318:12)
    at require (node:internal/modules/helpers:136:16)
```

Seems like not all nested dependencies are correctly bundled.

https://github.com/KillerCodeMonkey/electron-builder-26-modules-not-found/tree/main

Nantris:
Any particular platform? Looks like Linux?

KillerCodeMonkey:
I just tested it on Ubuntu 24.04 and electron v35.

Since all my mac/win mates were already in their after hours. I could not verify it on other platform. 


Am 2. April 2025 21:08:04 MESZ schrieb Nantris ***@***.***>:

beyondkmp:
@KillerCodeMonkey 
Test directly using the `form-data` library; I am unable to reproduce the issue on my end. Could you help provide a minimal reproducible demo code?

KillerCodeMonkey:
Strange if i build it locally it is working, but when building via ci (github actions) those problems are popping up.

KillerCodeMonkey:
@beyondkmp we are building the app in the ci like this:

NODE_ENV=production
NPM_CONFIG_PRODUCTION=false

and then `npm ci` instead of `npm i`

if you set both envs locally i can reproduce the error, even formData is a production dep

KillerCodeMonkey:
[...]

PR #9010, opened 2025-04-04
fix: remove implicit dependencies handling

beyondkmp:
fix  https://github.com/electron-userland/electron-builder/issues/9000 and https://github.com/electron-userland/electron-builder/issues/9006

**Root Cause**

It directly copies `node_modules` into the app directory and removes `devDependencies` from `package.json`. This way, `npm list` can retrieve all dependencies. However, during the lookup process, if there are duplicate dependency packages, some `dependencies` in `npm` may be empty. A previous fix（https://github.com/electron-userland/electron-builder/pull/8864 ） introduced the concept of `implicit dependencies` to address this issue, but it only handled the first level of `implicit dependencies`. If `implicit dependencies` themselves contain further `implicit dependencies`, the problem described in #9000 will occur.

**How to fix**

The current approach directly searches the `parsedTree` and updates the global `this.productionGraph`. If there are cases where `dependencies` are empty, it retrieves the same dependency from `allDependencies` and performs a lookup for sub-dependencies. If the dependency already exists in `this.productionGraph`, no further lookup is performed. Additionally, to prevent infinite loops, an empty array is set before searching for sub-dependencies.

Now that two tree lookup has been eliminated, there will be a certain improvement in performance.


beyondkmp:
https://github.com/electron-userland/electron-builder/actions/runs/14277766371/job/40028363385

The HoistedNodeModuleTest did not run, and there are several others under this category that were also skipped. This might be caused by issues with parallel testing. When you have time, could you help take a look at this?

```
          - ArtifactPublisherTest,BuildTest,ExtraBuildTest,RepoSlugTest,binDownloadTest,configurationValidationTest,filenameUtilTest,filesTest,globTest,ignoreTest,macroExpanderTest,mainEntryTest,urlUtilTest,extraMetadataTest,linuxArchiveTest,linuxPackagerTest,HoistedNodeModuleTest,MemoLazyTest,HoistTest
```

test/snapshots/HoistedNodeModuleTest.js.snap
beyondkmp, line 1: There are two cases:
1. two packages json(newly added)
2. npm tar

    I manually compared the `npm tar`, and there are no issues. The previous collection algorithm had a problem, as it included some unnecessary packages.

packages/app-builder-lib/src/node-module-collector/types.ts
```diff
@@ -7,15 +7,12 @@ export interface NodeModuleInfo {
 
 export type ParsedDependencyTree = {
   readonly name: string
+  readonly from: string // for pnpm
```
mmaietta, line 10: Would this be more appropriate in `PnpmDependency` since that extends `ParsedDependencyTree` already?

beyondkmp: Fixed.

beyondkmp: Fixed.

test/snapshots/HoistedNodeModuleTest.js.snap
beyondkmp: Let me correct that: there was an issue with the previous comparison of the `npm tar`. During local testing, the library `ramf` was updated, and a dependency was removed, causing the discrepancy.

It has now been updated. Currently, there is a newly added unit test (UT) in the snap, and everything else remains unchanged.


image

packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
```diff
@@ -15,53 +14,34 @@ export class NpmNodeModulesCollector extends NodeModulesCollector<NpmDependency,
...
+  }
+
+  protected extractProductionDependencyGraph(tree: NpmDependency, dependencyId: string): void {
+    if (this.productionGraph[dependencyId]) return
+
+    const { _dependencies: prodDependencies = {}, dependencies = {} } = tree
+    const isDuplicateDep = Object.keys(prodDependencies).length > 0 && Object.keys(dependencies).length === 0
+    const resolvedDeps = isDuplicateDep ? (this.allDependencies.get(dependencyId)?.dependencies ?? {}) : dependencies
+    // don't delete this, it's used to prevent infinite loops
+    this.productionGraph[dependencyId] = { dependencies: [] }
```
mmaietta, line 36: Can you elaborate on this a bit more? This comment makes me feel like there's an alternative way to write this.

packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
```diff
@@ -15,53 +14,34 @@ export class NpmNodeModulesCollector extends NodeModulesCollector<NpmDependency,
...
-      dependencies,
-      implicitDependenciesInjected,
+      this.allDependencies.set(`${key}@${value.version}`, value)
+      this.collectAllDependencies(value)
     }
-    return depTree
+  }
+
+  protected extractProductionDependencyGraph(tree: NpmDependency, dependencyId: string): void {
+    if (this.productionGraph[dependencyId]) return
```
mmaietta, line 30: Minor, but codestyle for this is
```
if (...) {
   return
}
```

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -28,54 +29,51 @@ export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependenc
...
 
-  protected extractRelevantData(npmTree: PnpmDependency): PnpmDependency {
-    const tree = super.extractRelevantData(npmTree)
-    return {
-      ...tree,
-      optionalDependencies: this.extractInternal(npmTree.optionalDependencies),
-    }
-  }
+  extractProductionDependencyGraph(tree: PnpmDependency, dependencyId: string): void {
+    if (this.productionGraph[dependencyId]) return
```
mmaietta, line 33: Minor, multi-line `if` codeestyle for `return`

mmaietta:
Fantastic work on this PR!

packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
beyondkmp: Fixed.

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
beyondkmp: Fixed

packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
beyondkmp: Yes. 

 // Initialize with empty dependencies array first to mark this dependency as "in progress"
    // This prevents infinite loops when circular dependencies are encountered

beyondkmp: After this initialization, if there are libraries with the same name+version later, they will not be searched recursively again, which can prevent infinite loops.

mmaietta (approved):

Issue #9011, opened 2025-04-04
"dependency path is undefined" errors after upgrade to electron-builder 26


panther7:
We have similar issue like #9006 with `dependency path is undefined`, becase project has `node_modules` in root folder, but "build" folder is without `node_modules`.
Posible solution is define path for node_modules, like this https://www.electron.build/app-builder-lib.interface.metadatadirectories

Example:
```
.
├── node_modules
└── dist
    └── app.html
```

Works perfectly in electron-builder 25.

Testcases: 

- [testcase_issue_9011_v2.zip](https://github.com/user-attachments/files/19610279/testcase_issue_9011_v2.zip)

mmaietta:
@panther7 would you be willing to share a minimum repro repo for us to create a unit test from?

@beyondkmp can you please take a look at this?

panther7:
@mmaietta simple test case added to first desc.

beyondkmp:
Released [v26.0.13](https://github.com/electron-userland/electron-builder/releases/tag/v26.0.13)

PR #9013, opened 2025-04-07
fix: dependency path is undefined

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/9011

Under the two `package.json` framework, if there are no `node_modules` in the `app` directory and `npmRebuild` is set to `false`, dependencies will not be reinstalled. As a result, `npm list` will return empty. In this case, we should attempt to retrieve the `node_modules` using the `projectDir`, which can resolve this issue.

Using `getProjectRootPath` directly from `electron/rebuild` is not accurate because it recursively searches upward level by level and returns the first match it finds. If I set the `app` directory outside the `project dir`, such as in another directory, the result returned by `getProjectRootPath` in `electron/rebuild` will be incorrect. Therefore, the `projectDir` should be passed in directly instead of searching upward level by level.
https://github.com/electron/rebuild/blob/ff1ec40f82ca64e014079b246053f039b3cf4f23/src/search-module.ts#L76-L86
image

### DIFF
--- a/packages/app-builder-lib/src/util/yarn.ts
+++ b/packages/app-builder-lib/src/util/yarn.ts
@@ -11,7 +11,7 @@ import { PM, detect, getPackageManagerVersion } from "../node-module-collector"
 import { NodeModuleDirInfo } from "./packageDependencies"
 import { rebuild as remoteRebuild } from "./rebuild/rebuild"
 
-export async function installOrRebuild(config: Configuration, appDir: string, projectDir: string, options: RebuildOptions, forceInstall = false) {
+export async function installOrRebuild(config: Configuration, { appDir, projectDir }: DirectoryPaths, options: RebuildOptions, forceInstall = false) {
   const effectiveOptions: RebuildOptions = {
     buildFromSource: config.buildDependenciesFromSource === true,
     additionalArgs: asArray(config.npmArgs),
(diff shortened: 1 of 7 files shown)


packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -12,7 +11,7 @@ import { PM, detect, getPackageManagerVersion } from "../node-module-collector"
 import { NodeModuleDirInfo } from "./packageDependencies"
 import { rebuild as remoteRebuild } from "./rebuild/rebuild"
 
-export async function installOrRebuild(config: Configuration, appDir: string, options: RebuildOptions, forceInstall = false) {
+export async function installOrRebuild(config: Configuration, appDir: string, projectDir: string, options: RebuildOptions, forceInstall = false) {
```
mmaietta, line 14: For **all** instances of where we have `appDir: string, projectDir: string`, can we structure it as an object? Having 2 string params makes it easy to get swapped by another dev in the future.
```
installOrRebuild(config: Configuration, dirs: { appDir: string, projectDir: string }, options: RebuildOptions, forceInstall = false)
```
`{ appDir: string, projectDir: string }` could be extracted to a type somewhere too

beyondkmp: Good suggestion.

beyondkmp: Fixed.

PR #9016, opened 2025-04-07
feat(electron-updater): add ignoreStagingPercentage option for checking updates

Lemonexe:
## Description

Add `ignoreStagingPercentage` option to the interface of `AppUpdater.checkForUpdates()`, so that you can override the staging percentage and get offerred an update _if the other necessary conditions_ are met.

### Usecase
staging rollout is observed when automatically querying for updates in background, but we also have a manual "Check for updates" button in Settings. This will allow the manual check to always offer you the latest version.

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -35,7 +35,7 @@ import { gunzipSync } from "zlib"
 import { blockmapFiles } from "./util"
 import { DifferentialDownloaderOptions } from "./differentialDownloader/DifferentialDownloader"
 import { GenericDifferentialDownloader } from "./differentialDownloader/GenericDifferentialDownloader"
-import { DOWNLOAD_PROGRESS, Logger, ResolvedUpdateFileInfo, UPDATE_DOWNLOADED, UpdateCheckResult, UpdateDownloadedEvent, UpdaterSignal } from "./types"
+import { CheckForUpdatesOptions, DOWNLOAD_PROGRESS, Logger, ResolvedUpdateFileInfo, UPDATE_DOWNLOADED, UpdateCheckResult, UpdateDownloadedEvent, UpdaterSignal } from "./types"
 import { VerifyUpdateSupport } from "./main"
 
 export type AppUpdaterEvents = {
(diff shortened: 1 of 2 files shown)


Lemonexe:
### Thoughts :thinking: 

The staging percentage feature is a great tool, but I think it'd greatly benefit from having two version definitions: the stable & the new one.
If all users are on app v7, and `latest.yml` specifies v8, then it totally makes sense that the staging percentage is applied and update offered to those in the "test group".
However, if I understand it correctly, even users with a very outdated version, say v2, they will be offered v8 only if in test group, and _nothing at all_ if in control group. It'd be ideal to offer v7 to those in control groups. But I do realize it'd require a major refactoring, so that when you publish v8, both v7+v8 are published and offered alternatively.
What do you think? Is it worth creating an issue?

Lemonexe:
### Thoughts :thinking: 

The staging percentage feature is a great tool, but I think it'd greatly benefit from having two version definitions: the stable & the new one.
If all users are on app v7, and `latest.yml` specifies v8, then it totally makes sense that the staging percentage is applied and update offered to those in the "test group".

However, if I understand it correctly, even users with a very outdated version, say v2, they will be offered v8 only if in test group, and _nothing at all_ if in control group. It'd be ideal to offer v7 to those in control group. But I do realize it'd require a major refactoring, so that you can publish two versions v7 & v8 when releasing v8.

What do you think? Is it worth creating an issue, or did I misunderstand how it works?
Would love to hear your view @mmaietta :slightly_smiling_face:

mmaietta:
Thanks for submitting a PR!

I think this functionality already exists with `set isUpdateSupported(hook)` functionality, where you can override the `is UpdateSupported` default logic with your own hook/validation of the `UpdateInfo`
https://github.com/electron-userland/electron-builder/blob/106640dd42a3db08bfbe3a3a32fe333e93ba5c10/packages/electron-updater/src/AppUpdater.ts#L209-L215

Lemonexe:
@mmaietta I believe that's a different case :thinking: 
If I understand correctly, then `isUpdateSupported` lets me block update on any arbitrary criteria, in case I need some more complex logic than that provided by the default [checkIfUpdateSupported](https://github.com/electron-userland/electron-builder/blob/065c6a456e34e7f8c13cba483d433502b9325168/packages/electron-updater/src/AppUpdater.ts#L436-L450) (blocking update based on `updateInfo?.minimumSystemVersion`).

But I meant the case when `isUpdateSupported()` is `true`, but still, `isUpdateAvailable()` results to `false` because of [isStagingMatch](https://github.com/electron-userland/electron-builder/blob/065c6a456e34e7f8c13cba483d433502b9325168/packages/electron-updater/src/AppUpdater.ts#L372-L393) logic. I'd like an option to force update available when it is supported, but below the rollout threshold.

:bulb:  Though your suggestion did give me an idea – _instead_ of the new parameter, I could create similar get/set/protected logic for `isStagingMatch`, like there is for `isUpdateSupported` (so that the API is consistent).
What do you think?

mmaietta:
Closing in favor of https://github.com/electron-userland/electron-builder/pull/9021

PR #9018, opened 2025-04-09
docs(electron-publish): Mention the use of ~/.aws/config in electron-publisher s3 options

gtluszcz:
It might seem like `electron-builder` or `electron-publish` do not support global aws credential configuration via `~/.aws/config` file. They do but you have to specify `AWS_SDK_LOAD_CONFIG=1` env variable as the underlying Go SDK used for publishing [does not load this file by default](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html) 

In our use case we were trying to use IAM Identity Center SSO Credentials stored in `~/.aws/config` combined with `AWS_PROFILE=` env variable. Took us quite some time to find how to achieve that. IMO this should be mentioned in the docs.

If you're seeing:
```
NoCredentialProviders: no valid providers in chain. Deprecated.
For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```
this may apply to you


gtluszcz:
I think the labels shoudl mention app-builder-lib, eletron-publish but i don't know how to change that

mmaietta:
Thanks for adding to the docs! Could you please regenerate the schema using `pnpm generate-all` and then also create a changeset file for this change? (`pnpm generate-changeset`)

gtluszcz:
Done

Issue #9012, opened 2025-04-06
Request to bump @electron/rebuild to 3.7.1


MikeMcC399:
## What would I like

Please consider bumping [@electron/rebuild@3.7.0](https://github.com/electron/rebuild/releases/tag/v3.7.0), released Oct 8, 2024, to [@electron/rebuild@3.7.1](https://github.com/electron/rebuild/releases/tag/v3.7.1), released Nov 12, 2024 (current `latest`), in [packages/app-builder-lib](https://github.com/electron-userland/electron-builder/tree/master/packages/app-builder-lib)

https://github.com/electron-userland/electron-builder/blob/106640dd42a3db08bfbe3a3a32fe333e93ba5c10/packages/app-builder-lib/package.json#L54

## Why it's needed

When I tested an update of `electron-builder` in [cypress-io/cypress](https://github.com/cypress-io/cypress) from [electron-builder@^25.1.8](https://github.com/electron-userland/electron-builder/releases/tag/electron-builder%4025.1.8) to [electron-builder@^26.0.12](https://github.com/electron-userland/electron-builder/releases/tag/v26.0.12) (current `latest`) I was blocked by an `nx` error:

```text
lerna notice cli v8.1.9
lerna ERR!     The "nx/js/dependencies-and-lockfile" plugin threw an error while creating dependencies: Source project does not exist: npm:@electron/node-gyp@git+https://github.com/electron/node-gyp.git#06b29aafb7708acef8b3669835c8a7857ebc92d2
lerna ERR!     Error: Source project does not exist: npm:@electron/node-gyp@git+https://github.com/electron/node-gyp.git#06b29aafb7708acef8b3669835c8a7857ebc92d2
```

This doesn't make a lot of sense to me, and it could be a lerna / nx / Yarn v1 bug.

[cypress-io/cypress](https://github.com/cypress-io/cypress) already migrated to use [@electron/rebuild@3.7.1](https://github.com/electron/rebuild/releases/tag/v3.7.1) and if I downgrade to [@electron/rebuild@3.7.0](https://github.com/electron/rebuild/releases/tag/v3.7.0) in [cypress-io/cypress](https://github.com/cypress-io/cypress) then I can update to [electron-builder@^26.0.12](https://github.com/electron-userland/electron-builder/releases/tag/v26.0.12) without the lerna error that occurs during dependency installation.

[...]

PR #9019, opened 2025-04-10
chore(deps): bump @electron/rebuild to 3.7.1

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/9012


MikeMcC399:
@beyondkmp 

Many thanks for this update! It's enormously helpful!

PR #9021, opened 2025-04-10
feat(electron-updater): allow overriding AppUpdater.isStagingMatch

Lemonexe:
Alternative implementation for https://github.com/electron-userland/electron-builder/pull/9016

Add get/set/protected `isUserWithinRollout` interface to `AppUpdater` class, similarly to how it is implemented for `isUpdateSupported` so that you can override the default calculation of staging percentage.
The user will be offerred an update _if the other necessary conditions_ are met, but they would otherwise be below the rollout threshold.

### Usecase
staging rollout is observed when automatically querying for updates in background, but we also have a manual "Check for updates" button in Settings. This will allow the manual check to always offer you the latest version.

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -204,7 +204,7 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
     this.configOnDisk = new Lazy<any>(() => this.loadUpdateConfig())
   }
 
-  protected _isUpdateSupported: VerifyUpdateSupport = (updateInfo: UpdateInfo): boolean | Promise<boolean> => this.checkIfUpdateSupported(updateInfo)
+  protected _isUpdateSupported: VerifyUpdateSupport = updateInfo => this.checkIfUpdateSupported(updateInfo)
 
   /**
    * Allows developer to override default logic for determining if an update is supported.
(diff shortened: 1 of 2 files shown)


packages/electron-updater/src/AppUpdater.ts
```diff
@@ -417,8 +434,8 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
       return false
     }
 
-    const isStagingMatch = await this.isStagingMatch(updateInfo)
-    if (!isStagingMatch) {
+    const isUserWithinRollout = await this.isUserWithinRollout(updateInfo)
```
mmaietta, line 437: I think this needs to be `await Promise.resolve(this.isUserWithinRollout(updateInfo))` since the hook can support both a `boolean | Promise<boolean>`

mmaietta:
This is clean! Thank you for the contribution

Quick request, can you generate a changeset file for this using `pnpm generate-changeset` so that it's picked up by the CD pipeline? Might want to run `pnpm generate-all` as well to verify prettier.

packages/electron-updater/src/AppUpdater.ts
Lemonexe: Right, fixed. Awaiting a non-promise should not be an issue, but it's clearer this way :slightly_smiling_face:

Lemonexe:
@mmaietta Done. I ran `lint`, `lint-deps`, `generate-changeset` and `generate-all` :heavy_check_mark: 

The last one generated a diff in `app-build-lib/scheme.json` (differs only in text). Though I did not commit that, to avoid mixing multiple packages in one commit..

PR #9024, opened 2025-04-11
fix(electron-updater): allow forceDevUpdateConfig also on Linux

Lemonexe:
A DX improvement for development on Linux:
- Currently, you may override `appUpdater.forceDevUpdateConfig = true` to test fetching and parsing update config.
- But it works only on macOS and Windows, where `isUpdaterActive` is called [directly on AppUpdater](https://github.com/electron-userland/electron-builder/blob/065c6a456e34e7f8c13cba483d433502b9325168/packages/electron-updater/src/AppUpdater.ts#L331).
- On Linux, `AppImageUpdater` overrides this method and early returns `false` on dev.
- So you have to _also_ set `APPIMAGE=true` env to get the same behavior.
- This PR removes the need for this workaround.

### DIFF
--- a/.changeset/shaggy-teachers-relax.md
+++ b/.changeset/shaggy-teachers-relax.md
@@ -2,4 +2,4 @@
 "electron-updater": patch
 ---
 
-chore: allow forceDevUpdateConfig also on Linux
+fix: allow forceDevUpdateConfig also on Linux
--- a/.changeset/shaggy-teachers-relax.md
+++ b/.changeset/shaggy-teachers-relax.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+chore: allow forceDevUpdateConfig also on Linux
--- a/packages/electron-updater/src/AppImageUpdater.ts
+++ b/packages/electron-updater/src/AppImageUpdater.ts
@@ -16,7 +16,7 @@ export class AppImageUpdater extends BaseUpdater {
   }
 
   public isUpdaterActive(): boolean {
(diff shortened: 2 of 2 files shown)


mmaietta:
Based on https://github.com/electron-userland/electron-builder/pull/9024/files#diff-f670e73a74bd05ee25dafba71f25a5d9cbe1c674b424c9e7db001563348901beL79-L82

I'm under the impression that `APPIMAGE` needs to be present in order for an installation to proceed during `doInstall`, so how does this PR handle the auto-update of a dev AppImage?

Lemonexe:
@mmaietta Sure, it is not possible to follow-up with an update. This merely enables you to query for available updates on dev environment. After all, that is available on Windows & macOS dev environments by the `forceDevUpdateConfig=true` option, and there you cannot "update the dev environment" either, right?. This just aligns the possibilities on Linux dev env with Windows & macOS.
But feel free to close the PR if it does not make sense :slightly_smiling_face:

Lemonexe:
@mmaietta Ok, but feel free to close if this is off track, this is really not a big deal :slightly_smiling_face: 
The only DX inconvenience is that when I run dev env of an electron app, and I want to test querying updates , I have to do this:
```
APPIMAGE=true npm start-my-electron-app
```
Naturally, I cannot proceed with the update on dev env. But I can test querying them when I set `forceDevUpdateConfig`

mmaietta (approved):

mmaietta:
LGTM. Thanks for your contribution!

Issue #9020, opened 2025-04-10
v26.0.12 breaks build in multiple ways


Almighty-Alpaca:
After updating to the latest electron-builder version my builds keep failing. I isolated the following problems to the switch to the node-module-collector in v26.0.4:

1. **Even when using Yarn, node-module-collector assumes that npm is installed and available on the path. If this isn't the case, the `NpmNodeModulesCollector` tries to parse an empty string as JSON and fails with a non-descriptive error message**
   Solution: This can easily be fixed by the user, but should imo be documented and/or produce a more helpful error message.
2. **On Windows the node-module-collector assumes that pnpm/npm/yarn have a `.cmd` file extension. This however is not the case when using a version manager such as [Volta](https://volta.sh/) or [Proto](https://moonrepo.dev/docs/proto), which create shims with an `.exe` file extension** 
   Solution: This was already inadvertently fixed for [pnpm in #8885](https://github.com/electron-userland/electron-builder/pull/8885/files#diff-3781d1bcfef363df7dea345579c15ffd71e6f2e012535e5ca40c49633742990cR12-R23) and could be fixed for other package managers in the same way. However, in practice it wouldn't even be a problem to leave the file extension away completely as `.cmd` is in the default Windows `PATHEXT` environment variable (and removing it manually would probably break a lot of other things).
3. **The package manager [is executed inside a shell](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts#L46). At least on Windows (I haven't tested Linux) this causes Powershell to load the user's profile. If the shell profile prints something when being loaded, the command output won't be valid JSON anymore and therefore parsing the dependency tree fails**
   Solution: Using a shell shouldn't be necessary imo

beyondkmp:
> After updating to the latest electron-builder version my builds keep failing. I isolated the following problems to the switch to the node-module-collector in v26.0.4:
> 
> 1. **Even when using Yarn, node-module-collector assumes that npm is installed and available on the path. If this isn't the case, the `NpmNodeModulesCollector` tries to parse an empty string as JSON and fails with a non-descriptive error message**
>    Solution: This can easily be fixed by the user, but should imo be documented and/or produce a more helpful error message.
[...]

PR #9026, opened 2025-04-11
fix: don't assume commands end with .cmd on Windows

Almighty-Alpaca:
Fixed first two issues of #9020. If an executable cannot be found it now looks like this:
```
  • electron-builder  version=26.0.12 os=10.0.26100
  • loaded configuration  file=<project_path>\electron-builder.yml
  • writing effective config  file=dist\builder-effective-config.yaml
  • skipped dependencies rebuild  reason=npmRebuild is set to false
  • packaging       platform=win32 arch=x64 electron=35.1.5 appOutDir=dist\win-unpacked

<project_path>\node_modules\builder-util\src\util.ts:94
  throw new Error(`Cannot find executable for ${options.name}, tried: ${executables.join(", ")}`)
        ^
Error: Cannot find executable for npm, tried: npm, npm.cmd
    at findExecutable (<project_path>\node_modules\builder-util\src\util.ts:94:9)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
```

I was a bit surprised that the logic to find the package manager executables was duplicated (`util/yarn.ts` and `node-modules-collector`). But since I'm not familiar with how electron-builder works internally I didn't try to refactor anything regarding this

### DIFF
--- a/.changeset/healthy-cats-train.md
+++ b/.changeset/healthy-cats-train.md
@@ -1,5 +1,0 @@
----
-"app-builder-lib": patch
----
-
-test: adding helper function to packageMetadata to accept a relative file:// uri dependency
--- a/.changeset/little-bulldogs-thank.md
+++ b/.changeset/little-bulldogs-thank.md
@@ -1,5 +1,0 @@
----
-"electron-updater": patch
----
-
-fix: don't error out when trying to determine package manager
--- a/.changeset/many-toys-yell.md
+++ b/.changeset/many-toys-yell.md
@@ -1,7 +1,0 @@
----
-"app-builder-lib": patch
-"builder-util": patch
-"electron-builder": patch
----
-
-chore(deps): replace `is-ci` with `ci-info`
--- a/.changeset/nice-onions-yawn.md
+++ b/.changeset/nice-onions-yawn.md
@@ -1,5 +1,0 @@
----
-"app-builder-lib": patch
----
-
-fix: allow custom electron zip name to be provided when unpacking a provided electronDist
--- a/.changeset/old-games-wait.md
+++ b/.changeset/old-games-wait.md
@@ -1,5 +1,0 @@
----
-"dmg-builder": patch
----
-
-refactor(dmg): add human-readable log messages for hdiutil error codes and refactoring retry logic to only retry on transient error codes
--- a/.changeset/silly-dolls-pretend.md
+++ b/.changeset/silly-dolls-pretend.md
@@ -1,5 +1,0 @@
----
-"app-builder-lib": patch
----
-
-fix(linux): productName should be used as the default value when executableName is not set. (#8766)
--- a/.changeset/soft-pets-warn.md
+++ b/.changeset/soft-pets-warn.md
@@ -1,5 +1,0 @@
----
-"dmg-builder": patch
----
-
-fix(dmg): --force unmount dmg using hdiutil after 3sec delay when receiving error code 16 (resource is busy)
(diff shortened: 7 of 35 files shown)


Almighty-Alpaca:
It seems like some tests are currently failing, I'll take a look

mmaietta:
@Almighty-Alpaca does `TEST_FILES=BuildTest,oneClickInstallerTest pnpm ci:test` work for you? Some tests are scoped specifically to certain OS's, but those two files should allow you to test your implementation on Windows

Almighty-Alpaca:
@mmaietta thanks that really helped. 

I learned a few intricacies of `exec()` vs `execFile()` debugging this. Turns out that on Windows (and only on Windows) you _need_ to use either `exec()` or `execFile(..., ..., {shell: true})` to execute `.bat`/`.cmd` script, which the npm/Yarn/PnPm wrappers from e.g. corepack are. The extension in the PATHEXT environment variable will also only be respected when executign through a shell.

I've fixed the implementation to use `{shell: true}` like most places already do and rebased everything onto the current master

Almighty-Alpaca:
Upon further inspection this doesn't work correctly in all cases _yet_, I'll try to investigate further

As a side note un- and then reinstalling the whole node/npm/yarn/pnpm in different ways for testing slightly different implementations kinda sucks

mmaietta:
> As a side note un- and then reinstalling the whole node/npm/yarn/pnpm in different ways for testing slightly different implementations kinda sucks, I don't know of a better way to test this though...

Yeah, I don't know of another way without committing the node_modules for each test directly to the repo. It was only recently that I added lockfile snapshots to prevent transient dependencies getting version bumped and causing test cases to randomly break.

For linux docker tests, you could point mount `-v` your yarn/npm/pnpm cache directly into the container though to potentially speed things up. Haven't tested it myself yet, but could be worthwhile adding to the `test-linux` cmd

mmaietta:
@Almighty-Alpaca  do you have a sample repro repo for this that I can take a look at? I can try and help by picking up this PR to work with your setup

Almighty-Alpaca:
> @Almighty-Alpaca do you have a sample repro repo for this that I can take a look at? I can try and help by picking up this PR to work with your setup

Thanks, that would be awesome. Unfortunately I've been a bit busy with other deadlines and stuck with v26.0.3 for now. 

Due to this being an environment issue more that a project setup issue I don't have a repro repository, but on Windows you can recreate the problem using one of the examples from the docs, e.g. by running these commands (I tested this in a fresh VM just now, but any Windows machine that doesn't already have any other NodeJS environment installed that could interfere should work):
```pwsh
# Install proto https://moonrepo.dev/docs/proto/install#windows
irm https://moonrepo.dev/install/proto.ps1 | iex
& "$HOME\.proto\bin\proto.exe" setup

# If on a fresh machine install Git
winget install git.git

### Reopen terminal here to reload environment variables

# Install Node and Yarn
proto install node latest
proto install yarn latest

# Clone example repo and update electron-builder to problematic version
git clone https://github.com/electron-vite/electron-vite-react
cd electron-vite-react
Add-Content -Value 'nodeLinker: node-modules' -Path .yarnrc.yml # Add Yarn workaround
yarn add --dev electron-builder@26.0.13

# Runs into issue 1 from #9020 as there is no npm on installed
yarn build

# Install npm and reinstall packages with npm
proto install npm 10.9.2
Remove-Item -Recurse .\node_modules\
npm i

# Runs into issue 2 from #9020 as npm is installed as npm.exe not npm.cmd
npm run build

# See that the npm shim is named npm.exe
where.exe npm
```

PR #9029, opened 2025-04-13
feat: Support caching the blockmap file for the new version

beyondkmp:
Currently, by default, both the new and old blockmap files are downloaded, resulting in double the CDN requests and traffic. Since the blockmap file is very small, we can keep a copy of the current version locally, so only the new version’s blockmap file needs to be downloaded, saving half of the CDN requests and traffic.

### DIFF
--- a/.changeset/giant-hotels-shake.md
+++ b/.changeset/giant-hotels-shake.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+feat: Support caching the blockmap file for the new version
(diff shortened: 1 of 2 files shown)


mmaietta:
> we can keep a copy of the current version locally, so only the new version’s blockmap file needs to be downloaded

I think this is missing some edge cases, one of which is security-related I think?
- If the local "old"/current blockmap is manipulated before blockmap check, could that lead to a security vulnerability?
- Wondering if there's a previous reason as to why we want a "fresh" blockmap downloaded when checking for updates

https://github.com/electron-userland/electron-builder/blob/6a83607dc5da34ed2c8931477cfe07e4a2e68d5f/packages/electron-updater/src/differentialDownloader/FileWithEmbeddedBlockMapDifferentialDownloader.ts#L28-L29

PR #9032, opened 2025-04-15
feat: customNsisResources override in nsis options

indutny-signal:
Prior to this change `customNsisBinary` could be used to override the url of `nsis-*.7z` from the `electron-builder-binaries` repo to a custom location. However, there was no option to change the url of the `nsis-resources-*.7z` file.

With this chcange the url of the NSIS resources could also be overriden.

### DIFF
--- a/.changeset/four-cheetahs-wash.md
+++ b/.changeset/four-cheetahs-wash.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+Add customNsisResources override to nsis options
--- a/packages/app-builder-lib/src/targets/nsis/nsisOptions.ts
+++ b/packages/app-builder-lib/src/targets/nsis/nsisOptions.ts
@@ -3,7 +3,7 @@ import { TargetSpecificOptions } from "../../core"
 
 export interface CustomNsisBinary {
   /**
-   * @default https://github.com/electron-userland/electron-builder-binaries/releases/download
+   * @default https://github.com/electron-userland/electron-builder-binaries/releases/download/nsis-3.0.4.1/nsis-3.0.4.1.7z
    */
   readonly url: string | null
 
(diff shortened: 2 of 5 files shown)


packages/app-builder-lib/src/targets/nsis/nsisOptions.ts
```diff
@@ -3,7 +3,7 @@ import { TargetSpecificOptions } from "../../core"
 
 export interface CustomNsisBinary {
```
mmaietta, line 4: Wondering if it'd be worth extending resources interface here. What are your thoughts this? (just need to add back the jsdoc, but not sure how that works w.r.t. inherited properties... which could cancel out this idea)
```
export interface CustomNsisBinary extends CustomNsisResources {
   readonly debugLogging?: boolean | null
}
```

packages/app-builder-lib/src/targets/nsis/nsisUtil.ts
```diff
@@ -38,6 +38,19 @@ export const NSIS_PATH = () => {
   })
 }
 
+export const NSIS_RESOURCES_PATH = () => {
+  return NsisTargetOptions.then((options: NsisOptions) => {
+    if (options.customNsisResources) {
+      const { checksum, url, version } = options.customNsisResources
+      if (checksum && url) {
+        const binaryVersion = version || checksum.substr(0, 8)
```
mmaietta, line 46: Wait, what is happening here?

indutny-signal: No idea, I just copy-pasted it from above 😂 I don't think we actually need it?

packages/app-builder-lib/src/targets/nsis/nsisOptions.ts
indutny-signal: I actually decided to go into other direction in the light of the comment below, and made all properties required and non-nullable.

packages/app-builder-lib/src/targets/nsis/nsisUtil.ts
indutny-signal: Fixed.

.changeset/four-cheetahs-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+"app-builder-lib": minor
```
mmaietta, line 2: I realize this is a `feat`, but it seems that the `changesets-cli` CI/CD is detecting this as a major semver upgrade. For the interim, until I figure out wtf is happening there (again), can you set this to `patch`?

indutny-signal: You got it. Force pushed

mmaietta (approved):

PR #9034, opened 2025-04-17
feat: add buildUniversalInstaller option to NSIS portable configuration

beyondkmp:
When using `electron-builder --win --x64 --arm64` to build NSIS, we can set buildUniversalInstaller to false to avoid generating a universal setup EXE. However, a universal portable EXE is still generated. Add a configuration to control whether to generate the universal portable EXE.

### DIFF
--- a/packages/app-builder-lib/src/targets/nsis/nsisOptions.ts
+++ b/packages/app-builder-lib/src/targets/nsis/nsisOptions.ts
@@ -231,6 +231,12 @@ export interface PortableOptions extends TargetSpecificOptions, CommonNsisOption
    * The image to show while the portable executable is extracting. This image must be a bitmap (`.bmp`) image.
    */
   readonly splashImage?: string | null
+
+  /**
+   * Disable building an universal installer of the archs specified in the target configuration
+   * @default true
+   */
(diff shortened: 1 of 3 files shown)


packages/app-builder-lib/src/targets/nsis/nsisOptions.ts
```diff
@@ -231,6 +231,12 @@ export interface PortableOptions extends TargetSpecificOptions, CommonNsisOption
    * The image to show while the portable executable is extracting. This image must be a bitmap (`.bmp`) image.
    */
   readonly splashImage?: string | null
+
+  /**
+   * Disable building an universal installer of the archs specified in the target configuration
+   * @default true
+   */
+  readonly buildUniversalInstaller?: boolean
```
mmaietta, line 239: Can we just move the `buildUniversalInstaller` to `CommonNsisOptions` so that it's inherited to both Nsis and Portable interfaces?

beyondkmp: Currently, this `buildUniversalInstaller` does not support `nsis-web`, so it is not ideal to place it globally.
image

PR #9038, opened 2025-04-19
refactor: improve resource directory handling for macOS 

beyondkmp:
There are also `locales` files under the Electron frameworks on macOS. Deleting these can significantly reduce the app size.

The `locales` folders under `Contents/Resources` are all empty, so deleting them does not reduce the size.

It has already been tested in production, and deleting the `locales` files inside the Electron frameworks does not cause any issues.


mmaietta:
How does this reduce app size if "The locales folders under Contents/Resources are all empty, so deleting them does not reduce the size."?

beyondkmp:
Sorry for the misunderstanding. Electron Builder currently only deletes the `locales` files under `Contents/Resources`, but these do not contain any locale files under `lproj`, so it does not reduce the size. To significantly reduce the size, the `locales` files under `Contents/Frameworks/Electron Framework.framework/Resources` need to be deleted, as the actual `locales` files are located there in the framework.

image
image

packages/app-builder-lib/src/platformPackager.ts
```diff
@@ -608,13 +608,16 @@ export abstract class PlatformPackager<DC extends PlatformSpecificBuildOptions>
...
+    if (isElectronBased(this.info.framework)) {
       return path.join(appOutDir, "resources")
-    } else {
-      return appOutDir
     }
+    return appOutDir
   }
 
+  public getMacOsElectronFrameworkResourcesDir(appOutDir: string): string {
+    return path.join(appOutDir, `${this.appInfo.productFilename}.app`, "Contents", "Frameworks", "Electron Framework.framework", "Resources")
```
mmaietta, line 619: I think `Electron Framework.framework` may change name if a custom electron distributable is compiled with a different `BRANDING` file.

As such, I think we may need to use: https://github.com/electron-userland/electron-builder/blob/3d65267a6c53ca824f70e5b0f5d8f4ba8be38237/packages/app-builder-lib/src/electron/ElectronFramework.ts#L29-L34

packages/app-builder-lib/src/electron/ElectronFramework.ts
```diff
@@ -97,27 +97,29 @@ async function removeUnusedLanguagesIfNeeded(options: BeforeCopyExtraFilesOption
...
+        return
+      }
 
-    const language = file.substring(0, file.length - langFileExt.length)
-    if (!wantedLanguages.includes(language)) {
-      return fs.rm(path.join(dir, file), { recursive: true, force: true })
-    }
-    return
-  })
+      const language = file.substring(0, file.length - langFileExt.length)
```
mmaietta, line 108: I realize this isn't new code, but I think we could do here:
```
path.extname(file) !== langFileExt // if we remove the `.` from the extension filter
```
AND
```
const language = path.basename(file, langFileExt)
```

test/src/mac/macPackagerTest.ts
```diff
@@ -45,6 +45,9 @@ test.ifMac("two-package", ({ expect }) =>
       checkMacApp: async appDir => {
         const resources = await fs.readdir(path.join(appDir, "Contents", "Resources"))
         expect(resources.filter(it => !it.startsWith(".")).sort()).toMatchSnapshot()
+
+        const ElectronFrameworkResources = await fs.readdir(path.join(appDir, "Contents", "Frameworks", "Electron Framework.framework", "Resources"))
```
mmaietta, line 49: Minor: lowercase here `const electronFrameworkResourcesDir`

mmaietta (changes requested):

packages/app-builder-lib/src/platformPackager.ts
beyondkmp: Good catch.

packages/app-builder-lib/src/electron/ElectronFramework.ts
beyondkmp: Fixed.

test/src/mac/macPackagerTest.ts
beyondkmp: Fixed.

packages/app-builder-lib/src/electron/ElectronFramework.ts
```diff
@@ -97,27 +97,29 @@ async function removeUnusedLanguagesIfNeeded(options: BeforeCopyExtraFilesOption
...
+        return
+      }
 
-    const language = file.substring(0, file.length - langFileExt.length)
-    if (!wantedLanguages.includes(language)) {
-      return fs.rm(path.join(dir, file), { recursive: true, force: true })
-    }
-    return
-  })
+      const language = path.basename(file, langFileExt)
```
mmaietta, line 108: So I just double checked in a js playground, and in order to use `basename` with ext, we need to in clude a `.`
```
`.${langFileExt}`
```

beyondkmp: Yes. I have added `.`

mmaietta (approved):

PR #9041, opened 2025-04-21
feat: support freebsd

Martinfx:


absidue:
As electron-builder defaults to downloading the official Electron builds and they don't have FreeBSD builds, it should probably be documented that these FreeBSD options can only be used in combination with a custom Electron build e.g. the electronDist or electronDownload options.

mmaietta:
> it should probably be documented that these FreeBSD options can only be used in combination with a custom Electron build e.g. the electronDist or electronDownload options.

Love this feedback. Yes, completely agree with this.

Also, we'll need a unit test for this since it's proposing an entirely new platform target. Is there an artifact (anywhere) that we can use `electronDownload` for pulling the artifact for a unit test?

Additionally, is there any special config required for freebsd? Not sure what `Target` is supposed to be getting built with here, right now it only looks like configuration params are accepted.

mmaietta:
Closing due to lack of response. Cleaning up open PRs

Martinfx:
> Closing due to lack of response. Cleaning up open PRs

I am sorry. I want work to support for freebsd. What i need for accept  pull request ? Add some tests ?

PR #9059, opened 2025-04-25
feat: support upgrading from ARM to ARM, rather than upgrading to x64 in window and linux

beyondkmp:


packages/electron-updater/src/providers/Provider.ts
```diff
@@ -80,12 +80,19 @@ export abstract class Provider<T extends UpdateInfo> {
...
 
-  const result = files.find(it => it.url.pathname.toLowerCase().endsWith(`.${extension.toLowerCase()}`))
+  const result = files
+    .filter(file => {
+      if (!filterByArch) {
+        return true
+      }
+      return (process.arch == "arm64") === (file.url.pathname.includes("arm64") || file.info.url.includes("arm64"))
+    })
+    .find(it => it.url.pathname.toLowerCase().endsWith(`.${extension.toLowerCase()}`))
```
mmaietta, line 95: I'm not sure we need `filterByArch` param here. We could do something like:
```
const filteredFiles = files.filter(it => it.url.pathname.toLowerCase().endsWith(`.${extension.toLowerCase()}`))
const result = filteredFiles.find(it => [file.url.pathname, file.info.url].some(n => n.includes(process.arch))) ?? result.shift()
```

That way this logic applies to all arch filters and if none found, it default falls back to the previous logic

beyondkmp: Initially, my idea was the same as yours, and I also wanted to unify the implementation. However, after completing it, I realized that macOS already has a very complex filtering logic in place. 

https://github.com/electron-userland/electron-builder/blob/8e88a9bb0d6bee7bfb61c822492d90d605dbaa56/packages/electron-updater/src/MacUpdater.ts#L55-L88

That’s why I added the `filterByArch` parameter—by default, it will be used for all platforms, but macOS will continue to use the previously implemented complex filtering logic.

beyondkmp: However, your suggestion is also good. If it can be found, return it; if not, return all. This approach seems better.

beyondkmp: Fixed.

mmaietta:
This looks good! Just to confirm, which OS' has this been tested already on? Anything I need to do on my side?

mmaietta:
For Windows ARM, did you test the flow where the x64 installation is being updated to an arm64 build? I think there might be some issues there in terms of registry key cleanup?

beyondkmp:
I just tested it, and I didn’t notice any issues. The EXE installer is 32-bit and runs normally. Could you specify which registry issues you’re referring to?

mmaietta:
Sorry, did a double check and it's not the registry key

Do we need to be concerned with previous installed location since x86 and x64 are different directories? Will changing the arch of the update cause the previous installation to not be removed?
https://github.com/electron-userland/electron-builder/blob/444b791f9d2812f2a0f60481f7b25297585d9c5a/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L250

beyondkmp:
> Sorry, did a double check and it's not the registry key
> 
> Do we need to be concerned with previous installed location since x86 and x64 are different directories? Will changing the arch of the update cause the previous installation to not be removed?
> 
> https://github.com/electron-userland/electron-builder/blob/444b791f9d2812f2a0f60481f7b25297585d9c5a/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L250

The code you're referring to is for the portable EXE. Since the portable EXE doesn’t require installation, there should be no issues. Currently, NSIS installs everything under the AppData directory, and the installation directories for x86 and x64 are the same.

The only difference is when using NSIS's "Install for all users" option: x86 installs to `C:\Program Files (x86)`, while x64 installs to `C:\Program Files`. However, the uninstallation process for the EXE directly calls the old version’s EXE to remove the app, so there shouldn’t be any problems with that either.

mmaietta (approved):

mmaietta (approved):

PR #9061, opened 2025-04-25
chore: migrate fpm packaging from app builder

mmaietta:
Fairly straight-forward translation of https://github.com/develar/app-builder/blob/master/pkg/package-format/fpm/fpm.go

This will make it much easier to update fpm and innately test within electron-builder's CI/CD without interspersed updates of app-builder project.


packages/app-builder-lib/src/targets/FpmTarget.ts
```diff
@@ -292,6 +290,36 @@ export default class FpmTarget extends Target {
...
+
+    const fpmPath = await getFpmPath()
+
+    await exec(fpmPath, fpmArgs, { env }).catch(e => {
+      if (e.message.includes("Need executable 'rpmbuild' to convert dir to rpm")) {
+        const hint = "to build rpm, executable rpmbuild is required, please install rpm package on your system. "
+        if (process.platform === "darwin") {
+          log.error(null, hint + "(brew install rpm)")
+        }
+        log.error(null, hint + "(sudo apt-get install rpm)")
```
beyondkmp, line 317: If it's on Windows, will it prompt you to make a change?

mmaietta: `getFpmPath` returns `fpm` on Windows directly from PATH. Based o the Go code, it doesn't seem to hit this error message specifically on Windows

beyondkmp (approved):

Issue #8970, opened 2025-03-17
"Cannot find module 'sharp'" on Windows only after upgrading to 26.0.11


Nantris:
After upgrading to `26.0.11` it seems Sharp isn't being properly bundled or run. None of our config changed except changing `win.publisherName` to `win.signToolOptions.publisherName`.

I'm not sure how to go further in investigating this. Any advice appreciated!

mmaietta:
Hiya! Can you share a minimum reproducible repo that I can test locally with and/or create a unit test around? Would like to assist but need more information first.

Nantris:
Hey @mmaietta thanks for your reply.

I did find the problem - in `files` we had this rule, which worked fine in 25.x but not in 26.x (Sharp version didn't change):

```
!**/node_modules/sharp/install/*
```

Simply removing this is enough of a fix for us and probably only adds a tiny bit to the package size, but I do wonder why that would now be breaking in 26.x.

Thanks again!

Nantris:
Actually, I must have overlooked something because that's definitely not resolving the issue now...

Nantris:
Hm... I really thought I had this working but every attempt now results in this same issue. I'm not sure if I changed something else or if what I thought fixed it never really did. I'll update soon.

Nantris:
@mmaietta I did notice that there's nothing under `app.asar.unpacked` from Sharp (or any other Node dependencies.

mmaietta:
Yeahhhh, definitely need a minimum repro repo for this unfortunately. The node collector logic was completely migrated to TS from Golang to support pnpm amongst other things, and a migration to `electron/asar `in order to support `electron/fuses` integration

Nantris:
Thanks as always for the great work and for your help @mmaietta!

-----

I put together a repro here: https://github.com/Nantris/electron-builder-repro-8970

It requires Yarn v4 (which I think should be used automatically as long as you have corepack enabled?)

**Repro steps:**
1. `yarn install`
2. `yarn package`
3. Run executable in `win-unpacked`

-----

This isn't an exact copy of our config and may not be 100% correct in every regard, but it should at least output the `app.asar.unpacked` folder I would think. Alas, it does not.

mmaietta:
Well this is odd indeed. Your `.asar` doesn't have any node modules dir present. What version of electron-builder are you upgrading from? I'm wondering if this is a `yarn 4`-specific bug

Nantris:
[...]

PR #9067, opened 2025-04-29
refactor: update package manager detection and improve type handling

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/8970#issuecomment-2830421527

- Replaced the `detect` and `getPackageManagerVersion` functions with a new `detectPackageManager` function for better clarity and type safety.
- Updated the `PM` type to an enum for improved maintainability.
- Adjusted the logic in `getCollectorByPackageManager` and `installDependencies` to utilize the new detection method.
- Cleaned up unused code and comments in `packageManager.ts` and `yarn.ts` files.

### DIFF
--- a/packages/app-builder-lib/src/node-module-collector/packageManager.ts
+++ b/packages/app-builder-lib/src/node-module-collector/packageManager.ts
@@ -69,6 +69,14 @@ function detectPackageManagerByEnv(): PM {
   return PM.NPM
 }
 
+export function getPackageManagerCommand(pm: PM) {
+  let cmd = pm
+  if (pm === PM.YARN_BERRY || process.env.FORCE_YARN === "true") {
+    cmd = PM.YARN
+  }
+  return `${cmd}${process.platform === "win32" ? ".cmd" : ""}`
+}
+
 export function detectPackageManager(cwd: string) {
   const isYarnLockFileExists = fs.existsSync(path.join(cwd, "yarn.lock"))
   const isPnpmLockFileExists = fs.existsSync(path.join(cwd, "pnpm-lock.yaml"))
--- a/packages/app-builder-lib/src/node-module-collector/packageManager.ts
+++ b/packages/app-builder-lib/src/node-module-collector/packageManager.ts
@@ -8,7 +8,7 @@ export enum PM {
   YARN_BERRY = "yarn-berry",
 }
 
-export function detectPackageManager(cwd: string): PM {
+function detectPackageManagerByEnv(): PM {
   if (process.env.npm_config_user_agent) {
     const userAgent = process.env.npm_config_user_agent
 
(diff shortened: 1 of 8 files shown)


packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -146,16 +133,11 @@ export async function nodeGypRebuild(platform: NodeJS.Platform, arch: string, fr
 function getPackageToolPath(pm: PM) {
   let cmd = pm
   if (process.env.FORCE_YARN === "true") {
-    cmd = "yarn"
+    cmd = PM.YARN
   }
   return `${cmd}${process.platform === "win32" ? ".cmd" : ""}`
```
mmaietta, line 138: Just to double check. Does this still work if `PM.YARN_BERRY` is used here? Is the `cmd` on Windows `yarn-berry.exe`?

beyondkmp: No, it's still `yarn.cmd`.

image

beyondkmp:
@mmaietta 

Should we directly determine the package manager based on the env, and completely remove the logic that determines it based on the lock file? Could you check if this might cause any other issues?

mmaietta:
Hmmm, I think we should emulate the approach that electron-builder took prior to the `PM` refactor. IIRC, it only accepted `npm`/`yarn` though, so we are technically expanding the featureset of detecting of the package manager. I'm not sure what errors could occur if we only used env, but envs can differ across local dev vs. build machine, not sure if we want to only take into account env. If multiple lockfiles are detected, then we should use env, otherwise fall back to lockfile logic?

packages/app-builder-lib/src/util/yarn.ts
beyondkmp: @mmaietta  Sorry for the misunderstanding. Yarn Berry still uses `yarn.cmd`, not `yarn-berry.cmd`. This has already been verified.

beyondkmp:
@mmaietta The current logic has been changed: if there is only one lock file, the lock file will be used to determine the package manager (PM). If there are multiple lock files or no lock file, the environment variables (`env`) will be used instead.

packages/app-builder-lib/src/util/yarn.ts
mmaietta: This logic here still doesn't seem right to me.

```
function getPackageToolPath(pm: PM) {
  let cmd = pm
  if (process.env.FORCE_YARN === "true") {
    cmd = "yarn"
    cmd = PM.YARN
  }
  return `${cmd}${process.platform === "win32" ? ".cmd" : ""}`
}
```
When `pm = PM.YARN_BERRY && platform === win32`, it returns -> `return "yarn-berry.cmd"`

beyondkmp: My bad, I misunderstood. I’ve added this function to `packageManager`, so others can directly call it when needed, without having to create their own.

mmaietta (approved):

cinderisles:
@beyondkmp @mmaietta didn't this PR effectively remove support for Bun without notice? Bun can be used as just a drop-in replacement for npm much like using pnpm with `--shamefully-hoist` and I only see removals of references to bun in this PR. What I'd like to be able to do is just force Bun usage and have electron-builder treat it like regular npm

avarayr:
I'm curious to understand how the removal of bun slipped thru the code review, as this is actively breaking builds in repos that use bun workspaces instead of npm

avarayr:
image

installs initiated with `bun` in repos that use bun workspaces are erroneously defaulting to `yarn` because this pull request removed bun detection

packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -1,95 +1,103 @@
...
-  }
+function detectPackageManagerByEnv(): PM {
+  if (process.env.npm_config_user_agent) {
+    const userAgent = process.env.npm_config_user_agent
 
-  return Promise.all([
-    exists(resolve(cwd, "yarn.lock")),
-    exists(resolve(cwd, "package-lock.json")),
-    exists(resolve(cwd, "pnpm-lock.yaml")),
-    exists(resolve(cwd, "bun.lockb")),
```
cinderisles, line 42: Bun 1.2 changed the default lockfile format to text-based, so the extension is `.lock` and not just `.lockb`. I think the binary version `.lockb` is still supported, just not the default anymore, so there should be a check here for both

https://bun.com/blog/bun-lock-text-lockfile

avarayr: great point @cinderisles 
i've added a companion PR to the active https://github.com/electron-userland/electron-builder/pull/9262 
hopefully we'll get bun back thru that

if not i'll try to make a separate PR to do that

Issue #9009, opened 2025-04-04
NSIS killing all processes with same name


panther7:
ref: #7773 

Autoupdate(nsis) process killing all apliccations with same name.

Expected is kill only process from installed folder.

#### Example:
app1: `/folder1/app.exe`
app2: `/folder2/app.exe`
Both running, NSIS installing/updating app1, but app2 is killed too.

---

* **Electron-Builder Version**:  24.6.4, 26.0.12

* **Node Version**: 20.5.1, 22.14.0

* **Electron Version**: 26.2.0, 35.1.2
* **Electron Type (current, beta, nightly)**: current

* **Target**: win

#### Solution:

Print all processes from actual directory:

```ps
powershell -Command "Get-Process | ?{$_.Path -and $_.Path.StartsWith($(Get-Location).Path)}"
```

And kill them:
```ps
powershell -Command "Get-Process | ?{$_.Path -and $_.Path.StartsWith($(Get-Location).Path)} | Stop-Process -Force"
```

mmaietta:
I like the solution you propose. Would you be willing to open a PR for this change?

panther7:
@mmaietta #9069

panther7:
Merged via #9069

PR #9069, opened 2025-04-29
feat(nsis): terminate only processes running in installation folder

panther7:
Modified the process termination logic in NSIS scripts to only kill application processes that are running from the installation directory. This prevents accidental termination of processes with the same name running from other locations during installation or update.

issue: #9009


mmaietta:
@beyondkmp would you mind taking a look at this when you have a chance?

packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
```diff
@@ -37,14 +33,9 @@
...
-!macro FIND_PROCESS _FILE _ERR
-  !ifdef INSTALL_MODE_PER_ALL_USERS
-    ${nsProcess::FindProcess} "${_FILE}" ${_ERR}
-  !else
-    # find process owned by current user
-    nsExec::Exec `"$SYSDIR\cmd.exe" /c tasklist /FI "USERNAME eq %USERNAME%" /FI "IMAGENAME eq ${_FILE}" /FO csv | "$SYSDIR\find.exe" "${_FILE}"`
-    Pop ${_ERR}
-  !endif
+!macro FIND_PROCESS _PATH _ERR
+  nsExec::Exec `powershell -Command "if ((Get-Process | Where-Object {$$_.Path -and $$_.Path.StartsWith('${_PATH}')}).Count -gt 0) {exit 0} else {exit 1}"`
```
beyondkmp, line 37: Using PowerShell is not highly recommended, as many enterprise and institutional users have PowerShell disabled for security reasons, which would cause these scripts to fail.

beyondkmp: I suggest making this feature configurable, allowing users to choose whether to use PowerShell。

mmaietta: The "user" as in the developer using electron-builder, or the end user?

If the latter, we can't guarantee if powershell is accessible, or if they know whether it is accessible. We also need to take into account the update flow where it's non-interactive.

Not sure if possible, but is there a way to detect if powershell is available? If so, use the powershell command in this PR, if not, fallback to previous logic?

panther7: I did commit with fallback, but need more tests.

mmaietta:
@panther7 following up here, what tests have you run so far?

Wondering if we can create an install+update test somehow for this in GHA runner? Similar to how this works: https://github.com/electron-userland/electron-builder/pull/9113/files?diff=unified&w=1#diff-1d9441830c74b56db3ecc46ef9805c4519141c15e145fabb24159d86e49d526dR10-R31

mmaietta:
Working on a full "blackbox" test suite for verifying installation and updater flows with native/non-mocked Updaters
https://github.com/electron-userland/electron-builder/pull/9122

Will get that functional and then pull it into this PR for you so that we can add a black-box test for this

panther7:
After hard testing, done.

BUT, only works on powershell 7+, because powershell 5 have "bug" and sometimes are `Path`s empty, which I need for detect right process.

Anyway, `taskkill` and `tasklist` are fallback.

.changeset/flat-trees-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat(nsis): terminate only processes running in installation folder
```
mmaietta, line 5: Please change this to `patch` with `fix(nsis)`. This should partially fix a reported issue and also, the CD for some reason flags `minor` as a major semver bump, which we aren't ready to do yet.

panther7: done

panther7:
I found solution for PowerShell 5+. It must be use `Get-CimInstance -ClassName Win32_Process`, because `Get-Process` inside instance 32bit PowerShell 5 don't view `Paths` for 64bit application. `Get-CimInstance` returns `Path`s for all processes, both 32bit and 64bit.

Check fixup.

mmaietta:
Nice work @panther7 !! I need to do some hard testing on my side just to sanity check. Is there a way to force whether powershell or cmd approach is used for app termination?

Just disable powershell via policy?

packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
```diff
@@ -30,22 +30,77 @@
...
-      !insertmacro customCheckAppRunning
+  Var /GLOBAL CmdPath
+  Var /GLOBAL FindPath
+  Var /GLOBAL PowerShellPath
+  !ifmacrodef customCheckAppRunning
+    !insertmacro customCheckAppRunning
+  !else
+    StrCpy $CmdPath "$SYSDIR\cmd.exe"
+    StrCpy $FindPath "$SYSDIR\find.exe"
+    StrCpy $PowerShellPath "$SYSDIR\WindowsPowerShell\v1.0\powershell.exe"
```
panther7, line 41: @mmaietta You can edit this to `StrCpy $PowerShellPath "blablablaexe"` if you want "disable" powershell.

panther7:
Something news? Can I help with?

PR #9071, opened 2025-04-30
fix(differentialDownloader): improve error handling and timeout management in multipleRangeDownloader

### LINKED ISSUE #9063: Differential download must fail after 10 seconds
When useMultipleRangeRequest is true, differential download must fail after 10 seconds.

DifferentialDownloader.ts:
```
private async doDownloadFile(tasks: Array<Operation>, fdList: Array<OpenedFile>): Promise<any> {
    ...

    await new Promise((resolve, reject) => {
      let w: any
      if (this.options.isUseMultipleRangeRequest) {
        w = executeTasksUsingMultipleRangeRequests(this, tasks, firstStream, oldFileFd, reject)
        w(0)
        return
      }

      ...
   })
  }
```

multipleRangeDownloader.ts:
```
export function executeTasksUsingMultipleRangeRequests(
  differentialDownloader: DifferentialDownloader,
  tasks: Array<Operation>,
  out: Writable,
  oldFileFd: number,
  reject: (error: Error) => void
): (taskOffset: number) => void {
  const w = (taskOffset: number): void => {
     ...

    const nextOffset = taskOffset + 1000
    doExecuteTasks(
      differentialDownloader,
      {
        tasks,
        start: taskOffset,
        end: Math.min(tasks.length, nextOffset),
        oldFileFd,
      },
      out,
      () => w(nextOffset),
      reject
    )
  }
  return w
}

function doExecuteTasks(differentialDownloader: DifferentialDownloader, options: PartListDataTask, out: Writable, resolve: () => void, reject: (error: Error) => void): void {
  ...

  const requestOptions = differentialDownloader.createRequestOptions()
  requestOptions.headers!.Range = ranges.substring(0, ranges.length - 2)
  const request = differentialDownloader.httpExecutor.createRequest(requestOptions, response => {
    if (!checkIsRangesSupported(response, reject)) {
      return
    }

    const contentType = safeGetHeader(response, "content-type")
    const m = /^multipart\/.+?(?:; boundary=(?:(?:"(.+)")|(?:([^\s]+))))$/i.exec(contentType)
    if (m == null) {
      reject(new Error(`Content-Type "multipart/byteranges" is expected, but got "${contentType}"`))
      return
    }

    const dicer = new DataSplitter(out, options, partIndexToTaskIndex, m[1] || m[2], partIndexToLength, resolve)
    dicer.on("error", reject)
    response.pipe(dicer)

    response.on("end", () => {
      setTimeout(() => {
        request.abort()
        reject(new Error("Response ends without calling any handlers"))
      }, 10000) <-- The timer isn't canceled after data processing is completed.
    })
  })
  differentialDownloader.httpExecutor.addErrorAndTimeoutHandlers(request, reject)
  request.end()
}
```

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/9063


packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts
```diff
@@ -89,30 +89,50 @@ function doExecuteTasks(differentialDownloader: DifferentialDownloader, options:
...
 
-    const dicer = new DataSplitter(out, options, partIndexToTaskIndex, m[1] || m[2], partIndexToLength, resolve)
-    dicer.on("error", reject)
+    const dicer = new DataSplitter(out, options, partIndexToTaskIndex, m[1] || m[2], partIndexToLength, wrappedResolve)
+    dicer.on("error", wrappedReject)
     response.pipe(dicer)
 
     response.on("end", () => {
-      setTimeout(() => {
+      timeoutId = setTimeout(() => {
```
mmaietta, line 128: Do we need to store `timeoutId` for anything? Curious as to why we need to have `wrappedReject`. I was under the impression `setTimeout` without storing its ID would auto-release when GC'd

beyondkmp: The `reject` does seem unnecessary now, as it has already failed. At the time of writing, it was done for consistency, so both `resolve` and `reject` were included.

beyondkmp: Deleted.

mmaietta: I still see `timeoutId` but only being cleared on resolve now. Why can't we reuse previous logic and only change the timeout?

mmaietta: @beyondkmp can you please follow up here when you have a chance? Then we can proceed with the PR

PR #9097, opened 2025-05-12
feat: add excludeReBuildModules option to configuration and scheme

beyondkmp:
Introduced a new configuration option `excludeReBuildModules` to allow users to specify modules that should be excluded from the rebuild process. This option is now available in the `CommonConfiguration` and `PlatformSpecificBuildOptions` interfaces, as well as in the scheme.json file.

### DIFF
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -1825,6 +1825,20 @@
             "string"
           ]
         },
+        "excludeReBuildModules": {
+          "anyOf": [
+            {
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "description": "The modules to exclude from the rebuild."
+        },
         "executableArgs": {
           "anyOf": [
             {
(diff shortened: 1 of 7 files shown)


mmaietta:
Is this solving for a currently-existing issue?

Also wondering if we should put this in the next major semver bump so as to reorganize all native-rebuilding logic (npmRebuild, buildDependenciesFromSource, etc.) into a single property called `nativeRebuild` or potentially more accurately put `electronRebuild`

beyondkmp:
>Is this solving for a currently-existing issue?

Yes, but that person didn't create an issue — they asked me directly. He has a prebuilt package and doesn't want to recompile it, as it takes a lot of time, but he still wants to compile the other native packages. He mentioned that Electron Forge has such a configuration, but Electron Builder does not.

>Also wondering if we should put this in the next major semver bump so as to reorganize all native-rebuilding logic (npmRebuild, buildDependenciesFromSource, etc.) into a single property called nativeRebuild or potentially more accurately put electronRebuild

Agree to put this in the next major semver.

mmaietta:
@beyondkmp do you mind if I pick this up? I'd like to do a little refactoring and typescript magic with the electron/rebuild integration.

Quick side note though, next major semver is also the migration to ESM-only

beyondkmp:
@mmaietta  Not at all. Feel free to at me anytime if you need help.

PR #9117, opened 2025-05-19
refactor: replace `is-ci` with `ci-info`

talentlessguy:
Since the `is-ci` package is just `ci-info`.isCI, this PR just makes electron-builder use `is-ci` directly. The change is backwards compatible


packages/app-builder-lib/package.json
```diff
@@ -59,6 +59,7 @@
     "builder-util": "workspace:*",
     "builder-util-runtime": "workspace:*",
     "chromium-pickle-js": "^0.2.0",
+    "ci-info": "^3.7.0",
```
mmaietta, line 62: Can we update to a newer version of `ci-info`? [v4.2.0](https://github.com/watson/ci-info/releases/tag/v4.2.0)

In particular, [v3.9.0](https://github.com/watson/ci-info/releases/tag/v3.9.0) has Azure Signing for instance.

talentlessguy: done

PR #9119, opened 2025-05-20
feat(nsis): add support for more Windows Installer options

### LINKED ISSUE #9105: Add support for Windows Installer Properties for the Uninstall Registry Key
* **Electron-Builder Version**: 26.0.15

* **Node Version**: 22.15.1

* **Electron Version**: 36.2.0

* **Electron Updater**: 6.6.4

* **Target**: Windows

I would like to specify the Help Link, Support Link, Update Information, Comments and Readme fields shown in the Control Panel:

image

These values correspond with the [ARPURLINFOABOUT](https://learn.microsoft.com/en-us/windows/win32/msi/arpurlinfoabout), [ARPHELPLINK](https://learn.microsoft.com/en-us/windows/win32/msi/arphelplink), [ARPURLUPDATEINFO](https://learn.microsoft.com/en-us/windows/win32/msi/arpurlupdateinfo), [ARPCOMMENTS](https://learn.microsoft.com/en-us/windows/win32/msi/arpcomments) and [ARPREADME](https://learn.microsoft.com/en-us/windows/win32/msi/arpreadme) properties. 

There might be some others that would be worth to make configurable: 

- https://learn.microsoft.com/en-us/windows/win32/msi/uninstall-registry-key
- https://learn.microsoft.com/en-us/windows/win32/msi/configuring-add-remove-programs-with-windows-installer

A sensible default would be to set `ARPURLINFOABOUT` to the `homepage` property of `package.json`, the `ARPHELPLINK` to the `bugs` property of `package.json` and the `ARPCOMMENTS` to the `description` property of `package.json`.

mtdvlpr:
Closes #9105


packages/app-builder-lib/src/targets/nsis/NsisTarget.ts
```diff
@@ -207,6 +207,10 @@ export class NsisTarget extends Target {
       APP_GUID: guid,
       // Windows bug - entry in Software\Microsoft\Windows\CurrentVersion\Uninstall cannot have \ symbols (dir)
       UNINSTALL_APP_KEY: uninstallAppKey,
+      UNINSTALL_URL_HELP: options.uninstallUrlHelp || this.packager.info.metadata.homepage || undefined,
+      UNINSTALL_URL_INFO_ABOUT: options.uninstallUrlInfoAbout || this.packager.info.metadata.homepage || undefined,
+      UNINSTALL_URL_UPDATE_INFO: options.uninstallUrlUpdateInfo || this.packager.info.metadata.homepage || undefined,
+      UNINSTALL_URL_README: options.uninstallUrlReadme || this.packager.info.metadata.homepage || undefined,
```
mmaietta, line 213: I think `undefined` may have unexpected side effects. Note: `defines.ENABLE_LOGGING_ELECTRON_BUILDER = null` where `null` is used as a flag (I guess implicitly represents a boolean in the nsis scripts?)

Easy way around this - only set the defines var when provided a value
```
const { homepage } = this.packager.info.metadata
// `use` from builder-util IIRC
use(options.uninstallUrlHelp || homepage, (it) => defines.UNINSTALL_URL_HELP = it)
use(options.uninstallUrlInfoAbout || homepage, (it) => defines.UNINSTALL_URL_INFO_ABOUT = it)
use(options.uninstallUrlUpdateInfo || homepage, (it) => defines.UNINSTALL_URL_UPDATE_INFO = it)
use(options.uninstallUrlReadme || homepage, (it) => defines.UNINSTALL_URL_README = it)
```

mmaietta: Also, wondering if we need to allow disabling the field from being set if the developer doesn't want their homepage to show there. Curious on your thoughts? Basically, `uninstallUrlHelp?: string | false`

mtdvlpr: I don't think that's necessary. The only use case for defining the homepage is to show it somewhere, whether that's the npm registry or somewhere else, you want that homepage to be linked to the project.

mtdvlpr:
@mmaietta, is there anything I can still do or should I just patiently wait?

mmaietta:
Sorry, been busy with other bugs lately, but I needed to double-check whether unit tests already covered this. Looks like they already do since this is nsis + test fixture that already has `homepage` set in the `package.json`. Kicking off CI again since some of the tests are flaky (still investigating that separately as well)

PR #9125, opened 2025-05-26
Update README.md

saulotarsobc:
Add


mmaietta:
I don't think this is necessary to add. First, there's not enough reputation on the repo linked, but also, there's already examples of vite + electron in the docs

mmaietta:
It's mainly due to security reasons tbh, as adding community project links to electron-builder could be considered as "vetted" example/template projects. But the issue is that github repo's can change over time, and what once was not malicious (when PR is first reviewed by me), could then have malicious code added to it's `master` branch later down the line without me knowing.

I've been meaning to re-evaluate the requirements for adding a project to that list (as they also show up on the documentation site)

saulotarsobc:
Thanks for the detailed explanation, @mmaietta ... that makes total sense.

I completely understand the concerns regarding security and the need for caution when it comes to linking external repositories. It's reassuring to know that you're considering re-evaluating the requirements for adding community projects to the docs in the future.

In any case, I appreciate your time reviewing the PR. If the guidelines change down the line, I’d be happy to revisit this or contribute in another way.

Cheers! 😊

PR #9140, opened 2025-06-04
refactor: simplify nuspec file structure by consolidating file entries into a single pattern

### LINKED ISSUE #9092: .exe files not included in package after upgrading from 25.1.8 to 26.0.12
After upgrading from version 25.1.8 to 26.0.12, I've encountered an issue with the packaging process. In version 25.1.8, .exe extension files were properly included in the package. However, after upgrading to version 26.0.12, .exe files are no longer being included in the package.

Steps to reproduce:
1. Package a project containing .exe files using version 25.1.8 → .exe files are properly included
2. Upgrade to version 26.0.12
3. Package the same project → .exe files are missing from the package

beyondkmp:
fix #9092

### DIFF
--- a/packages/electron-builder-squirrel-windows/template.nuspectemplate
+++ b/packages/electron-builder-squirrel-windows/template.nuspectemplate
@@ -12,20 +12,6 @@
     <copyright><%- copyright %></copyright>
   </metadata>
   <files>
-    <file src="locales\**" target="lib\net45\locales" />
-    <file src="resources\**" target="lib\net45\resources" />
-    <file src="*.bin" target="lib\net45" />
-    <file src="*.dll" target="lib\net45" />
-    <file src="*.pak" target="lib\net45" />
-    <file src="*.exe.config" target="lib\net45" />
-    <file src="*.exe.sig" target="lib\net45" />
-    <file src="icudtl.dat" target="lib\net45\icudtl.dat" />
-    <file src="Squirrel.exe" target="lib\net45\squirrel.exe" />
-    <file src="LICENSE.electron.txt" target="lib\net45\LICENSE.electron.txt" />
-    <file src="LICENSES.chromium.html" target="lib\net45\LICENSES.chromium.html" />
-    <file src="<%- exe %>" target="lib\net45\<%- exe %>" />
-    <% additionalFiles.forEach(function(f) { %>
-    <file src="<%- f.src %>" target="<%- f.target %>" />
-    <% }); %>
+    <file src="**" target="lib\net45" exclude="**\.git\**;" />
   </files>
 </package>


mmaietta:
I don't think this is a necessary change. I'd prefer explicit file `includes` as opposed to a glob + exclude filter. Is there a performance benefit to this PR?

beyondkmp:
The old version used to copy all files into `\lib\net45`, so this approach doesn't seem to be an issue.

mmaietta:
> The old version used to copy all files into \lib\net45, so this approach doesn't seem to be an issue.

I don't see a **reason** for this change from the current functionality being provided - this simplification doesn't seem needed. Allowlists are always my more preferred approach when it comes to production assets, not blocklists using `exclude="**\.git` and other exclusions as we can't guarantee if other files might accidentally be included by the end-developer's app configuration/setup.

If we're needing to handle `extraFiles` in being included, then can we explicitly pipe those into the inclusion array?

beyondkmp:
@mmaietta How about we make this template file customizable? If users need to include more files, they could specify their own template. If none is specified, the current default template would be used.

mmaietta:
Hmmm, that's a great proposal to handle very advanced use cases, not sure if the average dev will understand how to go that deep into nuspec.

Is there a way we could just loop through a property on squirrel for customizing the accepted/to-be-copied files?
Similar how it already does for:
```
    <file src="<%- exe %>" target="lib\net45\<%- exe %>" />
    <% additionalFiles.forEach(function(f) { %>
    <file src="<%- f.src %>" target="<%- f.target %>" />
    <% }); %>
```

mmaietta:
@beyondkmp friendly ping on https://github.com/electron-userland/electron-builder/pull/9140#issuecomment-2993044268

Let me know what the best next steps are, but currently, I'm not a fan of having a globstar "all" regex

beyondkmp:
No better solution has been identified yet. The previous approach used by the old Squirrel.Windows version was to package all contents from the app directory into `lib\net45` without any filtering. Therefore, the current idea is to simply include the entire app directory without any filtering.

packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
```diff
@@ -161,18 +161,76 @@ export default class SquirrelWindowsTarget extends Target {
...
-      return nuspecTemplate
     }
-    return templatePath
+
+    // Replace the additionalFiles loop with the actual files
+    if (additionalFiles.length > 0) {
+      const additionalFilesContent = additionalFiles.map(f => `    <file src="${f.src}" target="${f.target}" />`).join("\n")
+      templateContent = templateContent.replace('<% file src="additionalFiles.src" target="additionalFiles.target" / %>', additionalFilesContent)
+    } else {
+      templateContent = templateContent.replace('<% file src="additionalFiles.src" target="additionalFiles.target" / %>', "")
```
mmaietta, line 183: I like this approach!

Can you create a const for this? (similar to `searchString` on line 170)
```
const additionalFilesReplaceKey = '<% file src="additionalFiles.src" target="additionalFiles.target" / %>'
```

test/snapshots/windows/squirrelWindowsTest.js.snap
```diff
@@ -272,6 +272,7 @@ exports[`squirrel.windows > squirrel window x64 msi 2`] = `
   "lib/net45/d3dcompiler_47.dll",
   "lib/net45/ffmpeg.dll",
   "lib/net45/icudtl.dat",
+  "lib/net45/index.html",
```
mmaietta, line 275: Is this expected behavior? If we're testing just `extraFiles`, then I'd rather it not be the `index.html` since it may get misconstrued as part of the normal bundle when comparing manually comparing snapshot diffs.

Maybe create a temp file in the unit test?

PR #9151, opened 2025-06-09
refactor: enabling `getBin` to accept a different repo as a download source (useful for testing)

mmaietta:


drivron:
I'm seeing an issue with this fix with the call to `doGetBin` in `binDownload.ts`. This function is now called with `cacheName`, which produces an invalid value when ELECTRON_BUILDER_CACHE is set.
For example, if ELECTRON_BUILDER_CACHE="~cache", the following command is called and fails: `app-builder.exe download-artifact --name ~cachewinCodeSign`; and the following directory is created: `~cache/~cachewinCodeSign/590582919`.
Replacing `promise = doGetBin(cacheName, url, checksum)` with `promise = doGetBin(cacheKey, url, checksum)` seems to fix this issue.

mmaietta:
@drivron would you be willing to open a PR for that code change? I'm swamped with some critical investigations currently

drivron:
Here is PR #9293.

PR #9126, opened 2025-05-28
fix: allow custom electron zip name to be provided when unpacking a provided electronDist

mmaietta:

PR #9157, opened 2025-06-10
fix: `electronDist` - detect if directory and copy electron from there, short circuit current logic path

mmaietta:
Turns out that `ExtraBuildResourcesTest` wasn't running in the CI. Now it will

Fixes: https://github.com/electron-userland/electron-builder/pull/9126#issuecomment-2948662417


mmaietta:
CC @panther7, this will be in the upcoming release. Generated a unit test for the sample configuration you provided

PR #9166, opened 2025-06-16
feat(nsis): added supports for uninstall components page

panther7:
- Added support for uninstall components page.
- Uninstall section is RO (readonly)
- Uninstall section can be rename via `UNINSTALL_SECTION_NAME` definition

Example (installer.nsh):
```nsis
!define UNINSTALL_SECTION_NAME "Odinstalovat prohlížeč"

!macro customUnInstallSection
  Section /o "un.Vymazat data aplikace"
    RMDir /r "$APPDATA\${PRODUCT_NAME}"
  SectionEnd
!macroend
; ...
```

### DIFF
--- a/packages/app-builder-lib/templates/nsis/uninstaller.nsh
+++ b/packages/app-builder-lib/templates/nsis/uninstaller.nsh
@@ -131,7 +131,12 @@ Function un.restoreFiles
     Exch $R0
 FunctionEnd
 
-Section "un.install"
+!ifndef UNINSTALL_SECTION_NAME
+  !define UNINSTALL_SECTION_NAME "Uninstall"
+!endif
+
+Section "un.${UNINSTALL_SECTION_NAME}"
+  SectionIn RO
   # for assisted installer we check it here to show progress
   !ifndef ONE_CLICK
     ${IfNot} ${Silent}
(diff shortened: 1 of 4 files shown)


packages/app-builder-lib/templates/nsis/uninstaller.nsh
```diff
@@ -131,7 +131,12 @@ Function un.restoreFiles
     Exch $R0
 FunctionEnd
 
-Section "un.install"
+!ifndef UNINSTALL_SECTION_NAME
+  !define UNINSTALL_SECTION_NAME "Uninstall"
```
mmaietta, line 135: Is this correct? Previously the section was `un.install`.

panther7: Yes.

First argument in Section is "name", if have prefix "un.", when is section for uninstaller.

I need rename this uninstall section for user (like screenshot in description) "Odinstalovat prohlížeč".

If someone only add some more uninstall section, when default value will be "Uninstall" ("un.Uninstall"), it is more better than previous "install" ("un.install"). :)

I hope that makes sense.

mmaietta: Makes sense. I guess I'm more so wondering then, is this a backward compatible change or a breaking change for existing scripts?

panther7: No breaking change for existing scripts, by default, this section is invisible.

mmaietta:
Nice work! LGTM

Issue #8811, opened 2025-01-27
Enable differential download with only new blockmap


erijo:
Instead of downloading blockmaps for both the old and the new version when updating, how about just downloading the new version and then calculate the hashes for the old version from the local binary?

In our use-case we don't include the version number in the artifact name so the old and the new blockmap filenames are the same and thus the differential download doesn't work. For now we've disabled the differential download, but it would be nice if it could be made to work.

michalzaq12:
I have the same case (artifact name without version) and differential download doesn't work.
Why is the blockmap file not included in the `resources` directory, just like the `app-update.yml` file in the local binary?

PR #9172, opened 2025-06-18
feat(updater): Cache the new blockmap file and allow customization of the old blockmap file base URL.

beyondkmp:
Currently, downloading updates from GitHub's Latest Release does not support delta updates. This is mainly due to the inability to download the old map file. Now, by adding a configuration that allows users to customize the old map file URL, users can define the old map file URL themselves. This makes it possible to achieve delta updates without modifying the server side.

Cache the new blockmap file, and during the next update, prioritize reading it from the local cache. This reduces requests for the blockmap and allows delta updates even if the old version’s blockmap file is no longer available on the server.
 
Fix https://github.com/electron-userland/electron-builder/issues/8811

### DIFF
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -806,7 +806,7 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
       if (this._testOnlyOptions != null && !this._testOnlyOptions.isUseDifferentialDownload) {
         return true
       }
-      const blockmapFileUrls = blockmapFiles(fileInfo.url, this.app.version, downloadUpdateOptions.updateInfoAndProvider.info.version, this._oldBlockMapFileBaseUrl)
+      const blockmapFileUrls = blockmapFiles(fileInfo.url, this.app.version, downloadUpdateOptions.updateInfoAndProvider.info.version, this.previousBlockmapBaseUrlOverride)
       this._logger.info(`Download block maps (old: "${blockmapFileUrls[0]}", new: ${blockmapFileUrls[1]})`)
 
       const downloadBlockMap = async (url: URL): Promise<BlockMap> => {


packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts
```diff
@@ -46,7 +46,7 @@ function doExecuteTasks(differentialDownloader: DifferentialDownloader, options:
   for (let i = options.start; i < options.end; i++) {
     const task = options.tasks[i]
     if (task.kind === OperationKind.DOWNLOAD) {
-      ranges += `${task.start}-${task.end - 1}, `
+      ranges += `${task.start}-${task.end - 1},`
```
mmaietta: Is this an intentional prod change?

packages/electron-updater/src/AppUpdater.ts
```diff
@@ -117,6 +117,12 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
    */
   forceDevUpdateConfig = false
 
+  /**
+   * The base URL of the old block map file.
+   * @default null
```
mmaietta, line 128: Can you add to the docs what `default: null` actually signifies? (for those end-user/devs not familiar with differential downloading or blockmaps)

- What happens when it's null, what does it mean
- What happens when it's set

mmaietta:
This looks good!

packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts
beyondkmp: deleted

packages/electron-updater/src/AppUpdater.ts
beyondkmp: added

packages/electron-updater/src/AppUpdater.ts
```diff
@@ -133,6 +145,14 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
     return this._channel
   }
 
+  get oldBlockMapFileBaseUrl(): string | null {
+    return this._oldBlockMapFileBaseUrl
+  }
+
+  set oldBlockMapFileBaseUrl(value: string | null) {
+    this._oldBlockMapFileBaseUrl = value
+  }
```
mmaietta: Are these custom getter/setter needed? There's no data manipulation in either function, so I think we can just have it like this. Having it also on the public property will also have it rendered correctly in the jsdoc.
```
public previousBlockmapBaseUrlOverride: string | null = null
```

mmaietta:
Left a comment, but otherwise this LGTM! Nice work!

packages/electron-updater/src/AppUpdater.ts
beyondkmp: Fixed

PR #9178, opened 2025-06-23
feat(updater): clear cached installers after update check

beyondkmp:
if there's no available update version , then we clear the pending dir to save the disk size.


mmaietta:
@beyondkmp I'm still hesitant on having automatic cleanup up updates. What if we were to open an API that allowed clearing updates programmatically? That would be solving for this issue while avoiding us having to deal with unintended side effects?
https://github.com/electron-userland/electron-builder/issues/9200

mmaietta:
Taking alternative approach for user-driven cleanup in https://github.com/electron-userland/electron-builder/pull/9363

PR #9186, opened 2025-06-28
feat(electron-updater): add GitLab provider

daihere1993:
### 🔧 **Usage Example**
```typescript
autoUpdater.setFeedURL({
  provider: 'gitlab',
  projectId: 12345,
  token: process.env.GITLAB_TOKEN // for private repos
})
```

### 🧪 **Testing**
- [x] Integration tests added
- [x] Follows existing provider patterns  
- [x] Works with GitLab.com and self-hosted instances

### 📚 **Documentation**
Let me know, if need to update.

### DIFF
--- a/.changeset/funny-bottles-occur.md
+++ b/.changeset/funny-bottles-occur.md
@@ -1,0 +1,6 @@
+---
+"builder-util-runtime": major
+"electron-updater": major
+---
+
+feat(electron-updater): add GitLab provider support
--- a/.changeset/funny-bottles-occur.md
+++ b/.changeset/funny-bottles-occur.md
@@ -1,6 +1,6 @@
 ---
-"builder-util-runtime": major
-"electron-updater": major
+"builder-util-runtime": patch
+"electron-updater": patch
 ---
 
 feat(electron-updater): add GitLab provider support
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -357,8 +357,8 @@ function parseUrl(url: string, options: RequestOptions): URL {
     return new URL(url)
   } else {
     // Relative URL - construct base URL from original options
+    const hostname = options.hostname
     const protocol = options.protocol || "https:"
-    const hostname = options.hostname || "localhost"
     const port = options.port ? `:${options.port}` : ""
     const baseUrl = `${protocol}//${hostname}${port}`
     return new URL(url, baseUrl)
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -352,10 +352,10 @@ export interface DownloadCallOptions {
 }
 
 function parseUrl(url: string, options: RequestOptions): URL {
-  if (url.startsWith("http://") || url.startsWith("https://")) {
-    // Absolute URL
+  try {
+    // Would throw exception if url is not absolute
     return new URL(url)
-  } else {
+  } catch {
     // Relative URL - construct base URL from original options
     const hostname = options.hostname
     const protocol = options.protocol || "https:"
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -319,7 +319,7 @@ Please double check that your authentication token is correct. Due to security r
     const newOptions = configureRequestOptionsFromUrl(redirectUrl, { ...options })
     const headers = newOptions.headers
     if (headers?.authorization) {
-      const parsedNewUrl = new URL(redirectUrl)
+      const parsedNewUrl = parseUrl(redirectUrl, options)
       if (parsedNewUrl.hostname.endsWith(".amazonaws.com") || parsedNewUrl.searchParams.has("X-Amz-Credential")) {
         delete headers.authorization
       }
@@ -351,9 +351,24 @@ export interface DownloadCallOptions {
   destination: string | null
 }
 
+function parseUrl(url: string, options: RequestOptions): URL {
+  if (url.startsWith("http://") || url.startsWith("https://")) {
+    // Absolute URL
+    return new URL(url)
+  } else {
+    // Relative URL - construct base URL from original options
+    const protocol = options.protocol || "https:"
+    const hostname = options.hostname || "localhost"
+    const port = options.port ? `:${options.port}` : ""
+    const baseUrl = `${protocol}//${hostname}${port}`
+    return new URL(url, baseUrl)
+  }
+}
+
 export function configureRequestOptionsFromUrl(url: string, options: RequestOptions) {
   const result = configureRequestOptions(options)
-  configureRequestUrl(new URL(url), result)
+  const parsedUrl = parseUrl(url, options)
+  configureRequestUrl(parsedUrl, result)
   return result
 }
 
--- a/packages/builder-util-runtime/src/publishOptions.ts
+++ b/packages/builder-util-runtime/src/publishOptions.ts
@@ -167,12 +167,6 @@ export interface GitlabOptions extends PublishConfiguration {
   readonly host?: string | null
 
   /**
-   * The protocol. GitLab supports both `https` and `http`.
-   * @default https
-   */
-  readonly protocol?: "https" | "http" | null
-
-  /**
    * The access token to support auto-update from private GitLab repositories. Never specify it in the configuration files. Only for [setFeedURL](./auto-update.md#appupdatersetfeedurloptions).
    */
   readonly token?: string | null
--- a/packages/builder-util-runtime/src/publishOptions.ts
+++ b/packages/builder-util-runtime/src/publishOptions.ts
@@ -1,12 +1,13 @@
 import { OutgoingHttpHeaders } from "http"
 import { Nullish } from "."
 
-export type PublishProvider = "github" | "s3" | "spaces" | "generic" | "custom" | "snapStore" | "keygen" | "bitbucket"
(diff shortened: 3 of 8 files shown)


daihere1993:
@beyondkmp I see you’ve recently refactored electron-updater, and I’d appreciate it if you could help review this new feature that adds a new GitLab provider. I’m currently using my public GitLab demo repository as the source for integration tests. If this PR is acceptable, I’m happy to create a dedicated repository for testing purposes and leave it untouched going forward.

packages/builder-util-runtime/src/publishOptions.ts
```diff
@@ -142,6 +143,47 @@ export function githubUrl(options: GithubOptions, defaultHost = "github.com") {
...
+   * The host (including the port if needed).
+   * @default gitlab.com
+   */
+  readonly host?: string | null
+
+  /**
+   * The protocol. GitLab supports both `https` and `http`.
+   * @default https
+   */
+  readonly protocol?: "https" | "http" | null
```
mmaietta: For the sake of security, do we need to allow an `http` option? Is there a usecase for it?

packages/electron-updater/src/providers/GitLabProvider.ts
```diff
@@ -0,0 +1,193 @@
...
+        throw e
+      }
+    }
+
+    try {
+      rawData = await fetchChannelData(this.channel)
+    } catch (e: any) {
+      // If custom channel fails, try default channel as fallback
+      if (this.channel !== this.getDefaultChannelName()) {
+        rawData = await fetchChannelData(this.getDefaultChannelName())
```
mmaietta, line 128: Can you elaborate on this logic? What's the use case of falling back to default channel?

It seems like we'd want to hard-check/respect the `channel` provided for pulling latest version _for that channel_, right?

.changeset/funny-bottles-occur.md
```diff
@@ -0,0 +1,6 @@
+---
+"builder-util-runtime": major
+"electron-updater": major
```
mmaietta: I don't think these need to be `major`, there's no already-existing API being altered.

These will need to be `patch` though in order for CI/CD as there's a bug in the release setup that is causing all `minor` bumps to be treated as `major` for some reason

packages/electron-updater/src/providers/GitLabProvider.ts
```diff
@@ -0,0 +1,193 @@
...
+      format: string
+      url: string
+    }>
+    links: Array<{
+      id: number
+      name: string
+      url: string
+      direct_asset_url: string
+      link_type: string
+    }>
```
mmaietta: Can you extract this subproperty into its own interface, such as `GitlabReleaseAsset`?

packages/electron-updater/src/providers/GitLabProvider.ts
```diff
@@ -0,0 +1,193 @@
...
+      // GitLab assets may have spaces replaced with dashes
+      const possibleNames = [
+        fileInfo.url, // Original filename
+        fileInfo.url.replace(/ /g, "-"), // Spaces replaced with dashes
+      ]
+
+      let assetUrl: string | undefined
+      for (const name of possibleNames) {
+        assetUrl = updateInfo.assets.get(name)
+        if (assetUrl) break
```
mmaietta: Minor: Please use multi-line `if`
```
if (assetUrl) {
   return
}
```

That being said, it looks like you can simplify this to a `find`?
```
const assetUrl = Array.from(updateInfo.assets).find(asset => possibleNames.includes(asset))
```

packages/builder-util-runtime/src/publishOptions.ts
daihere1993: It’s just aligned with `GithubOptions`. I don’t see any API using the HTTP protocol in the current official GitLab documentation.

packages/electron-updater/src/providers/GitLabProvider.ts
daihere1993:
```typescript
// GitHubProvider.ts
try {
  let channel = this.channel
  if (this.updater.allowPrerelease && semver.prerelease(tag)?.[0]) {
    channel = this.getCustomChannelName(String(semver.prerelease(tag)?.[0]))
  }
  rawData = await fetchData(channel)
} catch (e: any) {
  if (this.updater.allowPrerelease) {
    // Allow fallback to `latest.yml`
    rawData = await fetchData(this.getDefaultChannelName())
  } else {
    throw e
  }
}
```

It refers to the behavior of `GitHubProvider`, which retries fetching the default channel (`latest`) if the initial data fetch fails. The difference is there is no `prerelease` in gitlab. Personally I don't think it's necessary, we can just throw error directly.

packages/electron-updater/src/providers/GitLabProvider.ts
daihere1993: Good point, but there `Array.from(updateInfo.assets)` would convert Map to an array of `[key, value]` tuples.

I've modified this part of code to be following one:

```typescript
const matchingAssetName = possibleNames.find(name => updateInfo.assets.has(name))
const assetUrl = matchingAssetName ? updateInfo.assets.get(matchingAssetName) : undefined
```

packages/electron-updater/src/providers/GitLabProvider.ts
daihere1993: Done

.changeset/funny-bottles-occur.md
daihere1993: Done

daihere1993:
@mmaietta I found a bug in the `prepareRedirectUrlOptions()` function under the `builder-util-runtime` module. It fails to handle relative URLs properly. Specifically, the URL `https://gitlab.com/api/v4/projects/50484362/releases/permalink/latest` redirects to a relative path like `/api/v4/projects/50484362/releases/v1.4.19`. However, the current implementation of `prepareRedirectUrlOptions()` passes this relative path directly to `new URL(url)`, which throws an error since it’s not a valid absolute URL. I've corrected it in this PR then my gitlab ut cases can pass.

packages/electron-updater/src/providers/GitLabProvider.ts
mmaietta: Ah, I understand now, that makes sense. We don't want someone to be locked onto a different channel if the channel is no longer being distributed, but latest is.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -351,9 +351,24 @@ export interface DownloadCallOptions {
...
 }
 
+function parseUrl(url: string, options: RequestOptions): URL {
+  if (url.startsWith("http://") || url.startsWith("https://")) {
+    // Absolute URL
+    return new URL(url)
+  } else {
+    // Relative URL - construct base URL from original options
+    const protocol = options.protocol || "https:"
+    const hostname = options.hostname || "localhost"
```
mmaietta: Why do we have localhost as the fallback? That seems like a security vulnerability waiting to happen.

packages/builder-util-runtime/src/publishOptions.ts
mmaietta: If the Gitlab docs don't have a usage for that protocol, then let's remove it. I'd rather only enable/start with a security-first mindset, then if someone **really** needs `http`, then we can add it later.

packages/builder-util-runtime/src/httpExecutor.ts
daihere1993: Fixed, removed `localhost` and use `options.hostname` directly

packages/builder-util-runtime/src/publishOptions.ts
daihere1993: Done

daihere1993:
Fixed all of comments and I've created a dedicated gitlab repo for tests. @mmaietta 

https://gitlab.com/daihere1993/gitlab-electron-updater-test

mmaietta (approved):
This is fantastic, LGTM.

On a related note, what would it take to have a matching GitlabPublisher as well? (Happy to provide guidance on that)

daihere1993:
> This is fantastic, LGTM.
> 
> On a related note, what would it take to have a matching GitlabPublisher as well? (Happy to provide guidance on that)

@mmaietta Yes, that's what I want to do next. I'll provide a draft PR for GitlabPublisher later.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -351,9 +351,24 @@ export interface DownloadCallOptions {
   destination: string | null
 }
 
+function parseUrl(url: string, options: RequestOptions): URL {
+  if (url.startsWith("http://") || url.startsWith("https://")) {
+    // Absolute URL
+    return new URL(url)
```
mmaietta, line 357: Just noticed that this logic might actually be fragile since it only accepts http subprefix.

Since a relative URL throws when constructing a URL, what if we just `try-catch` here?
```
try {
    // will throw exception when not absolute URL
    return new URL(url)
} catch {
    // Relative URL - construct base URL from original options
    const hostname = options.hostname
    const protocol = options.protocol || "https:"
    const port = options.port ? `:${options.port}` : ""
    const baseUrl = `${protocol}//${hostname}${port}`
    return new URL(url, baseUrl)
}
```

Not sure if that'll throw a typescript compiler error though. Alternative code
```
let urlObj: URL
try {
    // will throw exception when not absolute URL
    urlObj = new URL(url)
} catch {
    // Relative URL - construct base URL from original options
    const hostname = options.hostname
    const protocol = options.protocol || "https:"
    const port = options.port ? `:${options.port}` : ""
    const baseUrl = `${protocol}//${hostname}${port}`
    urlObj = new URL(url, baseUrl)
}
return urlObj!
```

daihere1993: Good point, this is the more robust way!

PR #9193, opened 2025-07-05
fix: stack overflow in electron-builder due to cyclic dependencies in _getNodeModules

DinoscapeProgramming:
This PR fixes a "Maximum call stack size exceeded" error caused by infinite recursion in the `_getNodeModules` method within `NodeModulesCollector`. 

### Root Cause
The function did not account for cyclic dependencies in the hoisted dependency graph, causing unbounded recursion and stack overflows during the build process.

### Fix
- Added cycle detection using a `Set<string>` to track visited dependencies.
- Prevents re-processing already-seen dependencies identified by `name@version`.

### Impact
- Resolves fatal crash during `electron-builder` packaging in large or monorepo projects.
- Makes dependency resolution more stable and safe when hoisting is involved.

Originally reported in #8857.

### DIFF
--- a/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
+++ b/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
@@ -92,29 +92,45 @@ export abstract class NodeModulesCollector<T extends Dependency<T, OptionalsType
     return node
   }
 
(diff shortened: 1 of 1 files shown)


DinoscapeProgramming:
Sorry, didn't see #8864.

PR #9194, opened 2025-07-06
feat(app-builder-lib): use jiti

### LINKED ISSUE #8944: feature request: use jiti for loading electron-builder.config.ts
jiti is better than config-file-ts, in these points:
- It doesn't break `__dirname` or `import.meta.dirname`
- It can handle esmodule better

I'm using jsdoc for typing, but native .ts file is better in many ways (`new Promise` for example)

sevenc-nanashi:
This PR makes app-builder-lib to use jiti, instead of config-file-ts.
I've tested on my app, but I don't know how to add test for this...

This PR closes #8944 .


mmaietta:
Just to confirm, how did you test this locally? I agree in that a unit test may not be easily accomplished for this.

mmaietta:
Verified locally as well. LGTM

PR #9196, opened 2025-07-08
fix(nsis): implement custom function to handle /D parameter with spaces

### LINKED ISSUE #7946: /D NSIS Flag does not support paths containing white spaces (e.g. C:/Program Files/)
* **Electron-Builder Version**:  24.9.1

* **Node Version**:  18 (LTS)

* **Electron Version**: 25.6.0

N/A

* **Target**:  Windows (NSIS)

The /D flag for the NSIS installer in command line mode is not properly parsing/handling file paths containing spaces.
The issue appears to be occurring in the parsing stage of the /D flag in multiUser.nsh (https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/nsis/multiUser.nsh#L45). 
A very simple test to reproduce the issue (on a Windows device) is to attempt running the NSIS installer in "quiet" mode with an override path containing a white space, for instance in my case:
./gdotv.exe /S /D=C:\Program Files\gdotv

The expected output of this command is for gdotv.exe to be installed under C:\Program Files\gdotv, however it ends up being installed under C:\Program
I've also tried different permutations with and without quotes but to no avail, e.g.:
./gdotv.exe /S /D="C:\Program Files\gdotv" fails the installation altogether
./gdotv.exe /S /D='C:\Program Files\gdotv' also fails the installation altogether

For reference, https://nsis.sourceforge.io/Docs/Chapter3.html, section 3.2 suggests the /D flag should allow for white spaces (and therefore be the last flag in the command).
I'm unfortunately not familiar with nsh scripts but I expect the solution here would be to modify the parsing of the /D flag such that it either supports quotes or sticking to NSIS standard just captures the rest of the command (everything after /D=, effectively) as the value of the parameter.

This can be reproduced with a minimal, default NSIS build via electron-builder, no extra configuration required.

beyondkmp:
Added a new macro `GetDParameter` to improve handling of the /D command line parameter in NSIS scripts. This function extracts the installation path correctly even when it contains spaces, ensuring compatibility with the NSIS requirements for the /D parameter.

fix #7946 

test results:

image
image

### DIFF
--- a/packages/app-builder-lib/templates/nsis/multiUser.nsh
+++ b/packages/app-builder-lib/templates/nsis/multiUser.nsh
@@ -41,7 +41,7 @@ Var installMode
     ${endif}
 
     # allow /D switch to override installation path https://github.com/electron-userland/electron-builder/issues/1551
-    ${StdUtils.GetParameter} $R0 "D" ""
+    !insertmacro GetDParameter $R0
     ${If} $R0 != ""
       StrCpy $INSTDIR $R0
     ${endif}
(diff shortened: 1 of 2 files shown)


mmaietta (approved):
LGTM AFAICT.

Shot in the dark, is there a way we can unit-test various nsis script functions without having to launch an app? Like a CLI testing utility?

packages/app-builder-lib/templates/nsis/multiUser.nsh
```diff
@@ -84,10 +84,51 @@ Var installMode
...
+  # Search for /D= or /d= using a simple loop
+  StrLen $R7 $R8
+  IntOp $R7 $R7 - 2  # Don't check last 2 characters
+  StrCpy $R6 0
+
+  ${Do}
+    StrCpy $R5 $R8 3 $R6  # Get 3 characters starting at position $R6
+    ${If} $R5 == "/D="
+    ${OrIf} $R5 == "/d="
+      # Found /D= or /d=, extract everything after it
```
mmaietta, line 120: Can we ensure that `/D` is the last argument? Just wondering if it's safe to extract everything after it.

KristinaPlusPlus:
Note this broke my scripts which were running the silent install and handling the spaces with quotes.  Removing the quotes makes it work again but just a note incase anyone else runs into

Issue #9187, opened 2025-06-28
26.0.12 - 26.0.17 breaks yarn workspace


itx-niklas-baugatz:
Hey there, i have a problem with my yarn workspace.

🧩 Project Setup
Stack: Angular + Electron (shared package.json)
electron-builder: upgraded from v26.0.11 to v26.0.12 (26.0.17 has the same issue)
yarn 1.22.22

🐞 Problem
After upgrading to v26.0.12:
node_modules is not copied into the .asar bundle.

The build completes without errors, but the installed app fails at runtime with:

Cannot find module '<first-required-dependency>'

Build debug output shows:

collected node modules  nodeModules=[]
When using Yarn, nodeModules is empty, but building with npm or pnpm works as expected (logged full array of deps). Switching to pnpm resolves the issue. Specifying yarn as packagemanager, causes the build to fail.

🔗 Related Issues
https://github.com/electron-userland/electron-builder/issues/9169

✅ Expected Behavior
Building under Yarn should:
Copy node_modules into .asar
Yield a working app with no runtime missing-module errors

Run yarn electron-builder
Build succeeds

Debuglogs show collected node modules nodeModules=[]
Resulting .asar lacks dependencies
App crashes with “Cannot find module”

Switch to pnpm or npm, rebuild → Everything works

I can't switch back to 26.0.11 because 26.0.12 fixes a signing issue.

beyondkmp:
could you please provide a mini reproduce demo?

dosprn:
Look at #9161 also. I suspect the same problem is here.

itx-niklas-baugatz:
thanks for the quick response, i created a demo
https://github.com/itx-niklas-baugatz/electron-builder-issue-9187

onNodeModuleFile() {return true} is a workaround on windows and linux, but does not work on mac

everything else seems works fine, when written in the files[] property, its only the node_modules

itx-niklas-baugatz:
Do you think there is a way to fix this or a workaround for the time, so i could still use yarn?
I switched my project to npm but it brings other problems, like i need to manually install optional dependencies (dmg-license) on macos to build, which wasn't an issue with yarn.

beyondkmp:
@itx-niklas-baugatz Could you try the 26.0.18?

beyondkmp:
This is caused by https://github.com/electron-userland/electron-builder/pull/9026.

[...]

Issue #9169, opened 2025-06-17
v26.0.17 breaks for Yarn


mtdvlpr:
I'm using the following versions:

- `electron-builder`@26.0.17
- `Yarn`@4.9.1

When I try to build my project locally I get the following error:

```
Exit code: 1. Command failed: pnpm.CMD config list
This project is configured to use yarn because ...\dist\electron\UnPackaged\package.json has a "packageManager" field

This project is configured to use yarn because ...\dist\electron\UnPackaged\package.json has a "packageManager" field
```

When I try to build it in my GitHub Actions pipeline, I get the following error:

```
Exit code: 1. Command failed: yarn list -a --include prod --include optional --omit dev --json --long --silent

Usage Error: Couldn't find a script named "list".
```

`v26.0.16` works fine. I suspect it has to do with #9026.

mmaietta:
I can take a look at this issue this week for sure.

Can you share a link to your project or a minimum repro repo?

The packageManager field should be getting respected, but for some reason your GH Actions pipeline is detecting it as yarn instead of pnpm.

mtdvlpr:
https://github.com/sircharlo/meeting-media-manager/actions/runs/15712423976/job/44273753132

mtdvlpr:
I'll see if I can create a minimal repro tomorrow

beyondkmp:
@mmaietta This caused by https://github.com/electron-userland/electron-builder/pull/9026. Yarn should use npm list instead of yarn list.

<img width="2003" height="127" alt="Image" src="https://github.com/user-attachments/assets/20f7f017-43a4-4474-8365-13ff5911cc6c" />

mmaietta:
Fix in https://github.com/electron-userland/electron-builder/pull/9197

PR #9197, opened 2025-07-09
fix: change installOptions to use npm in yarn PM

beyondkmp:
fix https://github.com/electron-userland/electron-builder/issues/9187 https://github.com/electron-userland/electron-builder/issues/9169

### DIFF
--- a/.changeset/funny-horses-protect.md
+++ b/.changeset/funny-horses-protect.md
@@ -1,5 +1,0 @@
----
-"dmg-builder": patch
----
-
-fix(dmg): resolve dmg flakiness by updating dmgbuild and related python vendor files
--- a/.changeset/orange-experts-do.md
+++ b/.changeset/orange-experts-do.md
@@ -1,5 +1,0 @@
----
-"electron-updater": patch
----
-
-feat(updater): Cache the new blockmap file and allow customization of the old blockmap file base URL
--- a/.changeset/thin-geckos-dream.md
+++ b/.changeset/thin-geckos-dream.md
@@ -1,5 +1,0 @@
----
-"app-builder-lib": patch
----
-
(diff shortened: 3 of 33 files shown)


packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -77,7 +77,10 @@ export function detectPackageManager(cwd: string): PM {
   if (npm) detected.push(PM.NPM)
 
   if (detected.length === 1) {
-    return detected[0] === PM.YARN ? detectPackageManagerByEnv() : detected[0]
```
beyondkmp, line 80: The reason this bug wasn't caught by UT is because of this code change. We do have `yarn.lock` files in our UT setup, but our tests are running with `pnpm`, and this code uses `detectPackageManagerByEnv` which correctly identifies the environment as `pnpm`, causing all the tests to pass — even though it's not the expected behavior when using Yarn.

mmaietta: I wonder if there's a way we could use `corepack` to enable/disable the correct package manager within each unit test's cwd?

mmaietta (approved):

packages/app-builder-lib/src/node-module-collector/packageManager.ts
beyondkmp: That's indeed a good approach. Let's create another PR to take a look at it.

cavivie:
It does not still works for yarn 4: https://github.com/beyondkmp/electron-builder/blob/fc8ce299d411012821f15de638c1f785111df80c/packages/app-builder-lib/src/node-module-collector/packageManager.ts#L50

PR #9204, opened 2025-07-16
feat: Rename the blockmap file in the pending directory.

beyondkmp:
image

Rename the blockmap file in the `pending` directory to `${updateFileName}.blockmap`, This makes the naming more identifiable and clear.

### DIFF
--- a/.changeset/funny-horses-protect.md
+++ b/.changeset/funny-horses-protect.md
@@ -1,5 +1,0 @@
----
-"dmg-builder": patch
----
-
-fix(dmg): resolve dmg flakiness by updating dmgbuild and related python vendor files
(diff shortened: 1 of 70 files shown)


mmaietta:
Out of curiosity, why do we need to rename the file? It's working fine as-is, no?

beyondkmp:
All are functioning normally. Renaming to `ringcentral-25.3.3010.exe.blockmap` increases clarity and identifiability, but not changing it is also acceptable.

mmaietta:
I think we'd also want to sanitize the filename here - I'm worried of edge cases of unsupported characters in filenames/paths.  I appreciate this submission, but I'm going to close this PR as I'd rather not change this functionality that is already working perfectly fine.

PR #9209, opened 2025-07-19
feat(electron-publish): add gitlab publisher

daihere1993:
As title, this is a new feat about gitlab publisher referring to the existing behaviors of githubPublisher. With this feat then we can fully support gitlab platform combined with [GitlabUpdater](https://github.com/electron-userland/electron-builder/pull/9186).

Notes:
1. As the legacy test structure is pretty messy, I’ve introduced a more semantic folder structure for the GitlabPublisher that includes both unit and integration tests. By the way, we can progressively refactor legacy tests following similar structure in the future.
2. The new integration tests require the GITLAB_TOKEN environment variable, tied to the gitlab repo of [gitlab-electron-updater-test](https://gitlab.com/daihere1993/gitlab-electron-updater-test). To enable this in CI/CD, I suggest adding a new private variable for the token. Let me know if you’re okay with this approach.
(For reference, the tests have been verified successfully in my local environment.)

**Details about new interface  `getBlockMapFiles()` in `Provider`:**
The legacy behavior required the block map file to be in the same folder as the base file. However, when publishing GitLab assets via `project_upload`, GitLab generates a unique path for each asset, making the legacy approach ineffective. To address this, I introduced a new `getBlockMapFiles()` interface in `Provider` and moved the legacy logic into a default implementation. This allows `GitlabProvider` to override the method with custom logic. It’s a more flexible solution and remains backward compatible.


daihere1993:
@mmaietta  fyi

daihere1993:
@mmaietta  by the way, let me know your comments about docs update, better to provide some guide then I can know how to update.

test/src/publisher/gitlab/GitlabPublisherTest.ts
```diff
@@ -0,0 +1,365 @@
...
+          env.restore()
+        }
+      })
+
+      test("should throw error when project ID is missing", ({ expect }) => {
+        const env = GitlabTestFixtures.setupTestEnvironment()
+
+        try {
+          delete process.env.CI_PROJECT_ID
+          delete process.env.CI_PROJECT_PATH
```
mmaietta: This actually mucks with the dev environment **across** tests IIRC. Please store the original var (even if not existing already), then restore it back after the test.

packages/electron-publish/src/gitlabPublisher.ts
```diff
@@ -0,0 +1,289 @@
...
+    context: PublishContext,
+    private readonly info: GitlabOptions,
+    private readonly version: string,
+    private readonly options: PublishOptions = {}
+  ) {
+    super(context, true)
+
+    let token = info.token
+    if (isEmptyOrSpaces(token)) {
+      token = process.env.GITLAB_TOKEN || process.env.GL_TOKEN
```
mmaietta: I think we can just use `GITLAB_TOKEN` here and remove `GL_TOKEN`?

For Github, I think `GH_TOKEN` was added after `GITHUB_TOKEN` IIRC when support was added for private repos

packages/electron-publish/src/gitlabPublisher.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    const response = await httpExecutor.doApiRequest(
+      configureRequestOptions({
+        protocol: parsedUrl.protocol,
+        hostname: parsedUrl.hostname,
+        port: parsedUrl.port as any,
+        path: parsedUrl.path,
+        method: "POST",
+        headers: {
+          "PRIVATE-TOKEN": this.token || "",
```
mmaietta: I don't think we should have a fallback value for `this.token || ""`, most notably because it can make things reaaaally difficult to debug dev-side. It looks like it is always defined at this state of the code, but if compiler errors are thrown, then let's throw an error if the token doesn't exist by the time we enter this function.

packages/electron-publish/src/gitlabPublisher.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    return httpExecutor.doApiRequest(
+      configureRequestOptions({
+        protocol: parsedUrl.protocol,
+        hostname: parsedUrl.hostname,
+        port: parsedUrl.port as any,
+        path: parsedUrl.path,
+        method: "PUT",
+        headers: {
+          "PRIVATE-TOKEN": this.token || "",
```
mmaietta: Ditto https://github.com/electron-userland/electron-builder/pull/9209/files#r2226904687

packages/electron-publish/src/gitlabPublisher.ts
```diff
@@ -0,0 +1,289 @@
...
+    return parseJson(
+      httpExecutor.request(
+        configureRequestOptions(
+          {
+            protocol: baseUrl.protocol,
+            hostname: baseUrl.hostname,
+            port: baseUrl.port as any,
+            path: apiPath,
+            headers: {
+              "PRIVATE-TOKEN": token || "",
```
mmaietta: Ditto https://github.com/electron-userland/electron-builder/pull/9209/files#r2226904687

packages/electron-updater/src/providers/GitLabProvider.ts
```diff
@@ -150,22 +147,157 @@ export class GitLabProvider extends Provider<GitlabUpdateInfo> {
...
+
+    for (const releaseId of possibleReleaseIds) {
+      const releaseUrl = newUrlFromBase(`projects/${this.options.projectId}/releases/${encodeURIComponent(releaseId)}`, this.baseApiUrl)
+
+      try {
+        const releaseResponse = await this.httpRequest(
+          releaseUrl,
+          {
+            accept: "application/json",
+            "PRIVATE-TOKEN": this.options.token || "",
```
mmaietta: Ditto https://github.com/electron-userland/electron-builder/pull/9209/files#r2226904687

test/src/publisher/gitlab/GitlabTestHelper.ts
```diff
@@ -0,0 +1,175 @@
...
+  /**
+   * Get all releases from GitLab project
+   */
+  async getAllReleases(): Promise<GitlabReleaseInfo[]> {
+    try {
+      const releases = await this.gitlabRequest<GitlabReleaseInfo[]>(`/projects/${encodeURIComponent(this.projectId)}/releases`)
+      return releases || []
+    } catch (e: any) {
+      log.warn({ error: e.message }, "Failed to get all releases")
+      return []
```
mmaietta: Just to understand, why are we returning `[]` if the request fails? Wouldn't that create a separate "branch" in test flows that isn't determinant?

packages/electron-publish/src/gitlabPublisher.ts
daihere1993: Any benefit to retaining just `GITLAB_TOKEN`? 

[#2503 ] I just refer to this change, seems we only support `GH_TOKEN` at first,

test/src/publisher/gitlab/GitlabPublisherTest.ts
daihere1993: I've remove these two env variables to make it more simple that requires user must provider projectId in configuration.

packages/electron-publish/src/gitlabPublisher.ts
daihere1993: That's true, I've removed that fallback value and furthermore I've extracted a new function `setAuthHeaderForToken()` in the `httpExecutor.ts` to create auth related header correctly.

test/src/publisher/gitlab/GitlabTestHelper.ts
daihere1993: You're right, should just throw it out.

Done

mmaietta:
Did you change something in your Gitlab repo? I may need to fork it...because tests broke seem to have been continually broken on master

daihere1993:
> Did you change something in your Gitlab repo? I may need to fork it...because tests broke seem to have been continually broken on master

Do you see any gitlab related errors logs?

> cannot resolve https://github.com/electron-userland/electron-builder-binaries/releases/download/nsis-resources-3.4.1/nsis-resources-3.4.1.7z: status code 503

I see many above github 503 errors.

daihere1993:
@mmaietta  Sorry, I published a new version v1.1.0 in gitlab repo in order to test `getBlockMapFiles()` in this PR that impacts the tests on the master branch.

I've created a separate PR (https://github.com/electron-userland/electron-builder/pull/9221) to fix the failing tests, so they no longer block the CI on the master branch.

This was a silly mistake—really sorry about that. I should have created a new dedicated gitlab repo specifically for `GitlabPublisher` related tests to avoid issues like this in the future.

Do you have any thoughts or suggestions on this approach?

mmaietta:
> This was a silly mistake—really sorry about that. I should have created a new dedicated gitlab repo specifically for GitlabPublisher related tests to avoid issues like this in the future.

This is what I do. There's not really a simple way other than that since it needs to be a full e2e test with github/gitlab/etc.

packages/electron-publish/src/gitlabPublisher.ts
mmaietta: Either way works, but I'd prefer only have 1 env var able to be used for token. If we ever need to support 2 types of token (public/private), then we'd have the flexibility in doing so in the same approach as Github Publisher

daihere1993: Yeah, it's ok for me.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -601,3 +601,31 @@ export function safeStringifyJson(data: any, skippedNames?: Set<string>) {
...
+  if (token.startsWith("glpat-")) {
+    headers["PRIVATE-TOKEN"] = token
+    return
+  }
+
+  // GitLab OAuth Application Secret - use Bearer authorization
+  if (token.startsWith("gloas-")) {
+    headers.authorization = `Bearer ${token}`
+    return
+  }
```
mmaietta: So I've been thinking on this for awhile, and I'm not too comfortable having custom `token` code in the base executor, super specific to Gitlab, but at the same time, could cause adverse effects if there's any overlap with other services (idk if that'd ever happen, but potentially with Custom Providers for instance, and I just want to be safe since this is updater-runtime code)

I'm wondering if there's a way that we can have a token callback in the `httpExecutor`, where the `GitLabProvider` can configure the executor to have this token-specific logic. What are your thoughts?

daihere1993: I completely understand your concern.

I’ve reverted the changes to `httpExecutor` and, instead, created a custom GitLab auth-related header in the `GitlabPublisher/Provider`. I’ve verified these changes locally, please take a look.

daihere1993:
There are two tests(`two-package` and `pkg scripts`) failed, they seems to be flaky tests. @mmaietta please have a look.

daihere1993:
I've created a dedicated PR to fix these two flaky tests https://github.com/electron-userland/electron-builder/pull/9251

daihere1993:
@mmaietta The code for this feature is almost ready for delivery. Let me know if any documentation updates are needed I'll create separate PR for that. There may still be some bugs or room for improvements, and I’ll continue to follow up on those. 

By the way, need I share the gitlab token with you? then we can test those new integration tests in the github ci.

mmaietta:
@daihere1993 can you message me via discord @ `onegoldfish` and let's sync up on this test fixture setup

amrit-moomie:
@daihere1993 when do you expect this to release.

daihere1993:
@mmaietta any other comments? let's speed up this review

mmaietta:
I've updated the token for the Gitlab tests, however, the tests don't seem to have been configured to run in the CI.

I think you need to add `Gitlab` to the TEST_FILES and add the secret to the `env`
https://github.com/electron-userland/electron-builder/blob/38c2085e9a344812d32611d197d1df66d6644b45/.github/workflows/test.yaml#L154-L160

daihere1993:
> ✓  node  test/src/publisher/gitlab/GitlabPublisher.integration.test.ts (3 tests | 2 skipped) 337ms

I've updated env variable for job `test-updater`, but seems that two tests still got skipped. I have a quick check, seems current forked PR can't access secret variable thus all of tests that depends on secret variable would be skipped.

@mmaietta any comment from your side?

mmaietta:
Hmmm, looking at the log, is this expected logging behavior or an actual issue? (Quick side note: we shouldn't be console logging to stdout in prod or test code)
```
✓  node  test/src/publisher/gitlab/GitlabPublisherTest.ts (15 tests) 14ms
non-error logging is silenced during VITEST workflow when DEBUG=electron-builder flag is not set
Sourcemap for "/project/node_modules/.pnpm/electron-winstaller@5.4.0/node_modules/electron-winstaller/lib/index.js" points to missing source files
stdout | test/src/publisher/gitlab/GitlabPublisher.integration.test.ts > GitLab Publisher - Integration Tests > Authentication > should reject invalid token
Failed to get or create GitLab release {
  tag: 'v9.73.72.1758169751044',
  version: '9.73.72.1758169751044',
  error: '401 Unauthorized\n' +
    '"method: GET url: [https://gitlab.com/api/v4/projects/72170733/releases\\n\\n](https://gitlab.com/api/v4/projects/72170733/releases//n//n)          Data:\\n          {\\"message\\":\\"401 Unauthorized\\"}\\n          "\n' +
    'Headers: {\n' +
    '  "date": "Thu, 18 Sep 2025 04:29:11 GMT",\n' +
    '  "content-type": "application/json",\n' +
    '  "content-length": "30",\n' +
    '  "connection": "keep-alive",\n' +
    '  "cf-ray": "980e19d10d70617e-ORD",\n' +
    '  "cf-cache-status": "MISS",\n' +
    '  "cache-control": "no-cache",\n' +
    '  "strict-transport-security": "max-age=31536000",\n' +
    '  "vary": "Origin",\n' +
    `  "content-security-policy": "default-src 'none'",\n` +
    '  "gitlab-lb": "haproxy-main-17-lb-gprd",\n' +
    '  "gitlab-sv": "api-gke-us-east1-c",\n' +
    '  "nel": "{\\"max_age\\": 0}",\n' +
    '  "referrer-policy": "strict-origin-when-cross-origin",\n' +
    '  "x-content-type-options": "nosniff",\n' +
    '  "x-frame-options": "SAMEORIGIN",\n' +
    '  "x-gitlab-meta": "{\\"correlation_id\\":\\"197601ed7cce2890116ffd326db1a687\\",\\"version\\":\\"1\\"}",\n' +
    '  "x-request-id": "197601ed7cce2890116ffd326db1a687",\n' +
    '  "x-runtime": "0.025886",\n' +
    '  "set-cookie": [\n' +
    '    "_cfuvid=xXfATPcFOShCNiE3.NnFMKZG7l5hk7DCHVd853uCLAI-1758169751309-0.0.1.1-604800000; path=/; domain=.gitlab.com; HttpOnly; Secure; SameSite=None"\n' +
    '  ],\n' +
    '  "server": "cloudflare"\n' +
    '}',
  errorType: 'authentication',
  statusCode: 401
}
 ❯ Logger._doLog packages/builder-util/src/log.ts:79:17
 ❯ Logger.doLog packages/builder-util/src/log.ts:66:12
 ❯ Logger.error packages/builder-util/src/log.ts:49:10
 ❯ GitlabPublisher.getOrCreateRelease packages/electron-publish/src/gitlabPublisher.ts:79:11
 ❯ GitlabPublisher.doUpload packages/electron-publish/src/gitlabPublisher.ts:141:21
 ❯ test/src/publisher/gitlab/GitlabPublisher.integration.test.ts:73:9

 ✓  node  test/src/publisher/gitlab/GitlabPublisher.integration.test.ts (3 tests | 2 skipped) 337ms
```

I'm thinking of just merging this and then just can ping you to investigate further if the tests are still being skipped?

daihere1993:
> I'm thinking of just merging this and then just can ping you to investigate further if the tests are still being skipped?

Yes, let's merge it now, then I'll have further check for it.

mmaietta:
I'm not sure why `test-updater` test is now failing, I'll take a look at this asap

PR #9211, opened 2025-07-21
fix: strip auth headers for GitHub release assets redirects

### LINKED ISSUE #9207: Auto-update fails with 403 errors when downloading latest.yml or latest-mac.yml from GitHub Releases
We’re seeing **403 Forbidden** errors during the auto-update process in our Electron app using `electron-updater` with GitHub Releases as the update provider.

### Error Output

```
Error in auto-updater: HttpError: 403 
"method: GET url: https://release-assets.githubusercontent.com/github-production-release-asset/845531298/df5526c7-0304-40be-8ba5-f036c2a689dc?sp=r&sv=2018-11-09&sr=b&spr=https&se=2025-07-17T11%3A42%3A53Z&rscd=attachment%3B+filename%3Dlatest-mac.yml&rsct=application%2Foctet-stream&skoid=96c2d410-5711-43a1-aedd-ab1947aa7ab0&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skt=2025-07-17T10%3A42%3A39Z&ske=2025-07-17T11%3A42%3A53Z&sks=b&skv=2018-11-09&sig=acs7mv1WMOrsGVYG2%2BmpWh3IWEI%2But%2BP2Ze%2FCFH6pTQ%3D&jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmVsZWFzZS1hc3NldHMuZ2l0aHVidXNlcmNvbnRlbnQuY29tIiwia2V5Ijoia2V5MSIsImV4cCI6MTc1Mjc1MDI0OSwibmJmIjoxNzUyNzQ5OTQ5LCJwYXRoIjoicmVsZWFzZWFzc2V0cHJvZHVjdGlvbi5ibG9iLmNvcmUud2luZG93cy5uZXQifQ.548vclzwamgye3EUtoR3hKuts23TV_EwQi7-IBjMC6s&response-content-disposition=attachment%3B%20filename%3Dlatest-mac.yml&response-content-type=application%2Foctet-stream\n\n          Data:\n          ﻿<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.\nRequestId:c76df6fe-601e-004e-3a09-f7cfdd000000\nTime:2025-07-17T10:59:10.3443684Z</Message></Error>\n          "
Headers: {
  "server": "Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0",
  "x-ms-request-id": "c76df6fe-601e-004e-3a09-f7cfdd000000",
  "accept-ranges": "bytes",
  "via": "1.1 varnish, 1.1 varnish",
  "date": "Thu, 17 Jul 2025 10:59:10 GMT",
  "x-served-by": "cache-iad-kjyo7100108-IAD, cache-bom4751-BOM",
  "x-cache": "MISS, MISS",
  "x-cache-hits": "0, 0",
  "x-timer": "S1752749950.204785,VS0,VE335",
  "content-disposition": "attachment; filename=latest-mac.yml",
  "content-type": "application/octet-stream",
  "content-length": "321"
}
    at createHttpError (/Applications/Spark Cloud Studio.app/Contents/Resources/app.asar/node_modules/electron-updater/node_modules/builder-util-runtime/out/httpExecutor.js:14:12)
    at IncomingMessage.<anonymous> (/Applications/Spark Cloud Studio.app/Contents/Resources/app.asar/node_modules/electron-updater/node_modules/builder-util-runtime/out/httpExecutor.js:147:28)
    at IncomingMessage.emit (node:events:513:28)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

### Observations

This appears to be an issue with GitHub’s CDN (Azure Blob Storage), possibly related to signed URL tokens expiring or failing intermittently.

Others are also reporting similar failures in the GitHub Community discussion:  
https://github.com/orgs/community/discussions/166304

### Steps to Reproduce

1. Publish a new release using `electron-builder` targeting GitHub.
2. Attempt auto-update using `electron-updater`.
3. Observe intermittent 403 errors on fetching `latest.yml`.

### Environment

- **electron-builder**: ^23.6.0  
- **electron-updater**: ^6.2.1  
- **Target**: GitHub Releases  
- **Platform**: macOS and Windows (confirmed), may also affect Linux  

### Request

- Is there any known workaround or fix that can be implemented in `electron-builder` or `electron-updater`?
- Would you recommend implementing a retry mechanism or using a custom update server if this continues?

FringeNet:
Fixes #9207

## Problem
GitHub redirects download requests from `api.github.com` to `release-assets.githubusercontent.com` (Azure Blob Storage backend). Azure rejects GitHub auth tokens causing 403 Forbidden errors during auto-updates.

## Root Cause
Azure blob storage intermittently returns 403 errors when an unidentified bearer token is present in the Authorization header.

## Security Concern
Electron exposes the GitHub PAT to Azure blob storage, and potentially other storage providers in custom update server implementations

## Solution
Strip Auth Header from all cross origin redirects, as is the standard with most request libraries.

## Testing
- All linting and compilation checks pass
- Preserves auth headers for legitimate GitHub API endpoints
- Targeted fix maintains backward compatibility

**This is my first contribution, any and all feedback welcome**

### DIFF
--- a/.changeset/lovely-brooms-move.md
+++ b/.changeset/lovely-brooms-move.md
@@ -1,0 +1,5 @@
+---
+"electron-builder-squirrel-windows": patch
+---
+
+fix: generate stubExecutableExe and sign it for squirrel.windows using new electron-builder-binaries asset
--- a/.changeset/lovely-geckos-chew.md
+++ b/.changeset/lovely-geckos-chew.md
@@ -1,0 +1,5 @@
+---
+"electron-builder-squirrel-windows": patch
+---
+
+fix(squirrel-windows): update loadingGif handling to prioritize user-defined option
--- a/.changeset/sharp-elephants-shout.md
+++ b/.changeset/sharp-elephants-shout.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix(electron-updater): do not use quotes when installing update with sudo
--- a/packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
+++ b/packages/electron-builder-squirrel-windows/src/SquirrelWindowsTarget.ts
@@ -1,4 +1,6 @@
 import { InvalidConfigurationError, log, isEmptyOrSpaces } from "builder-util"
+import { execWine } from "app-builder-lib/out/wine"
+import { getBinFromUrl } from "app-builder-lib/out/binDownload"
 import { sanitizeFileName } from "builder-util/out/filename"
 import { Arch, getArchSuffix, SquirrelWindowsOptions, Target, WinPackager } from "app-builder-lib"
 import * as path from "path"
@@ -20,29 +22,54 @@ export default class SquirrelWindowsTarget extends Target {
   }
 
   private async prepareSignedVendorDirectory(): Promise<string> {
-    // If not specified will use the Squirrel.Windows that is shipped with electron-installer(https://github.com/electron/windows-installer/tree/main/vendor)
-    // After https://github.com/electron-userland/electron-builder-binaries/pull/56 merged, will add `electron-builder-binaries` to get the latest version of squirrel.
-    let vendorDirectory = this.options.customSquirrelVendorDir || path.join(require.resolve("electron-winstaller/package.json"), "..", "vendor")
-    if (isEmptyOrSpaces(vendorDirectory) || !fs.existsSync(vendorDirectory)) {
-      log.warn({ vendorDirectory }, "unable to access Squirrel.Windows vendor directory, falling back to default electron-winstaller")
-      vendorDirectory = path.join(require.resolve("electron-winstaller/package.json"), "..", "vendor")
-    }
-
+    const customSquirrelVendorDirectory = this.options.customSquirrelVendorDir
     const tmpVendorDirectory = await this.packager.info.tempDirManager.createTempDir({ prefix: "squirrel-windows-vendor" })
-    // Copy entire vendor directory to temp directory
-    await fs.promises.cp(vendorDirectory, tmpVendorDirectory, { recursive: true })
-    log.debug({ from: vendorDirectory, to: tmpVendorDirectory }, "copied vendor directory")
+
+    if (isEmptyOrSpaces(customSquirrelVendorDirectory) || !fs.existsSync(customSquirrelVendorDirectory)) {
+      log.warn({ customSquirrelVendorDirectory: customSquirrelVendorDirectory }, "unable to access custom Squirrel.Windows vendor directory, falling back to default vendor ")
+      const windowInstallerPackage = require.resolve("electron-winstaller/package.json")
+      const vendorDirectory = path.join(path.dirname(windowInstallerPackage), "vendor")
+
+      const squirrelBin = await getBinFromUrl(
+        "squirrel.windows@1.0.0",
+        "squirrel.windows-2.0.1-patched.7z",
+        "DWijIRRElidu/Rq0yegAKqo2g6aVJUPvcRyvkzUoBPbRasIk61P6xY2fBMdXw6wT17md7NzrTI9/zA1wT9vEqg=="
+      )
+
+      await fs.promises.cp(vendorDirectory, tmpVendorDirectory, { recursive: true })
+      // copy the patched squirrel to tmp vendor directory
+      await fs.promises.cp(path.join(squirrelBin, "electron-winstaller", "vendor"), tmpVendorDirectory, { recursive: true })
+    } else {
+      // copy the custom squirrel vendor directory to tmp vendor directory
+      await fs.promises.cp(customSquirrelVendorDirectory, tmpVendorDirectory, { recursive: true })
+    }
 
     const files = await fs.promises.readdir(tmpVendorDirectory)
-    for (const file of files) {
-      if (["Squirrel.exe", "StubExecutable.exe"].includes(file)) {
-        const filePath = path.join(tmpVendorDirectory, file)
-        log.debug({ file: filePath }, "signing vendor executable")
-        await this.packager.sign(filePath)
-      }
+    const squirrelExe = files.find(f => f === "Squirrel.exe")
+    if (squirrelExe) {
+      const filePath = path.join(tmpVendorDirectory, squirrelExe)
+      log.debug({ file: filePath }, "signing vendor executable")
+      await this.packager.sign(filePath)
+    } else {
+      log.warn("Squirrel.exe not found in vendor directory, skipping signing")
+    }
+    return tmpVendorDirectory
+  }
+
+  private async generateStubExecutableExe(appOutDir: string, vendorDir: string) {
+    const files = await fs.promises.readdir(appOutDir, { withFileTypes: true })
+    const appExe = files.find(f => f.name === `${this.exeName}.exe`)
+    if (!appExe) {
+      throw new Error(`App executable not found in app directory: ${appOutDir}`)
     }
 
-    return tmpVendorDirectory
+    const filePath = path.join(appOutDir, appExe.name)
+    const stubExePath = path.join(appOutDir, `${this.exeName}_ExecutionStub.exe`)
+    await fs.promises.copyFile(path.join(vendorDir, "StubExecutable.exe"), stubExePath)
+    await execWine(path.join(vendorDir, "WriteZipToSetup.exe"), null, ["--copy-stub-resources", filePath, stubExePath])
+    await this.packager.sign(stubExePath)
+    log.debug({ file: filePath }, "signing app executable")
+    await this.packager.sign(filePath)
   }
 
   async build(appOutDir: string, arch: Arch) {
@@ -62,6 +89,7 @@ export default class SquirrelWindowsTarget extends Target {
         arch,
       })
       const distOptions = await this.computeEffectiveDistOptions(appOutDir, installerOutDir, setupFile)
+      await this.generateStubExecutableExe(appOutDir, distOptions.vendorDirectory!)
       await createWindowsInstaller(distOptions)
 
       await packager.signAndEditResources(artifactPath, arch, installerOutDir)
@@ -118,6 +146,10 @@ export default class SquirrelWindowsTarget extends Target {
 
   private get appName() {
     return this.options.name || this.packager.appInfo.name
+  }
+
+  private get exeName() {
+    return this.packager.appInfo.productFilename || this.options.name || this.packager.appInfo.productName
   }
 
   private select7zipArch(vendorDirectory: string) {
@@ -203,7 +235,9 @@ export default class SquirrelWindowsTarget extends Target {
       options.remoteReleases = this.options.remoteReleases
     }
 
-    if (!("loadingGif" in options)) {
+    if (this.options.loadingGif) {
+      options.loadingGif = path.resolve(packager.projectDir, this.options.loadingGif)
+    } else {
       const resourceList = await packager.resourceList
       if (resourceList.includes("install-spinner.gif")) {
         options.loadingGif = path.join(packager.buildResourcesDir, "install-spinner.gif")
--- a/packages/electron-builder-squirrel-windows/template.nuspectemplate
+++ b/packages/electron-builder-squirrel-windows/template.nuspectemplate
@@ -19,6 +19,7 @@
     <file src="*.pak" target="lib\net45" />
     <file src="*.exe.config" target="lib\net45" />
     <file src="*.exe.sig" target="lib\net45" />
+    <file src="*_ExecutionStub.exe" target="lib\net45" />
     <file src="icudtl.dat" target="lib\net45\icudtl.dat" />
     <file src="Squirrel.exe" target="lib\net45\squirrel.exe" />
     <file src="LICENSE.electron.txt" target="lib\net45\LICENSE.electron.txt" />
--- a/packages/electron-updater/src/LinuxUpdater.ts
+++ b/packages/electron-updater/src/LinuxUpdater.ts
@@ -31,8 +31,11 @@ export abstract class LinuxUpdater extends BaseUpdater {
     const installComment = `"${name} would like to update"`
     const sudo = this.sudoWithArgs(installComment)
     this._logger.info(`Running as non-root user, using sudo to install: ${sudo}`)
-    // pkexec doesn't want the command to be wrapped in " quotes
-    const wrapper = /pkexec/i.test(sudo[0]) ? "" : `"`
+    let wrapper = `"`
+    // some sudo commands dont want the command to be wrapped in " quotes
+    if (/pkexec/i.test(sudo[0]) || sudo[0] === "sudo") {
+      wrapper = ""
+    }
     return this.spawnSyncLog(sudo[0], [...(sudo.length > 1 ? sudo.slice(1) : []), `${wrapper}/bin/bash`, "-c", `'${commandWithArgs.join(" ")}'${wrapper}`])
   }
 


packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,30 @@ Please double check that your authentication token is correct. Due to security r
...
+    const protocol = options.protocol || "https:"
+    const hostname = options.hostname || "localhost"
+    const port = options.port ? `:${options.port}` : ""
+    const path = options.path || "/"
+    return new URL(`${protocol}//${hostname}${port}${path}`)
+  }
+
+  private static isCrossOriginRedirect(originalUrl: URL, redirectUrl: URL): boolean {
+    return originalUrl.protocol !== redirectUrl.protocol || originalUrl.hostname !== redirectUrl.hostname || originalUrl.port !== redirectUrl.port
+  }
```
nogitting: Way better. It is a little bit more complex than this though. Take a look at how [Requests](https://github.com/psf/requests/blob/fec96cd5976ad763e45bac9a033d62cca1877a00/src/requests/sessions.py#L127C9-L154) or [HttpClient](https://github.com/apache/httpcomponents-client/blob/9d38e5fb4438e8fb6442ac8e5ff304142938cf83/httpclient5/src/main/java/org/apache/hc/client5/http/impl/DefaultRedirectStrategy.java#L110-L122) do it.

FringeNet: I followed your advice.

I also did some digging on the custom implementations which allow pass through of custom headers (like self hosting the update server). 

I fear my implementation may be too aggressive. 
I am not familiar enough with Electron's internals, so I think it's best to leave the rest to the maintainers.

nogitting (commented):
Looks clean. Kudos.

packages/builder-util-runtime/src/httpExecutor.ts
nogitting: I would not call it aggressive at all, especially that you went the _Requests_ route.
Now, it is on/up to the maintainers.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,62 @@ Please double check that your authentication token is correct. Due to security r
...
+      if (HttpExecutor.isCrossOriginRedirect(originalUrl, parsedRedirectUrl)) {
         delete headers.authorization
       }
     }
     return newOptions
   }
 
+  private static reconstructOriginalUrl(options: RequestOptions): URL {
+    const protocol = options.protocol || "https:"
+    const hostname = options.hostname || "localhost"
```
mmaietta: I'm worried about `localhost` redirects may open a security vulnerability. Is this fallback required?

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,62 @@ Please double check that your authentication token is correct. Due to security r
...
+    const port = options.port ? `:${options.port}` : ""
+    const path = options.path || "/"
+    return new URL(`${protocol}//${hostname}${port}${path}`)
```
mmaietta, line 345: I think the typical usage of URL with a root URL changed is `new URL(path, `${protocol}//${hostname}${port}`)`?  That way the `path` doesn't require a `/` prefix (since it doesn't seem like we're double checking for that either)

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,62 @@ Please double check that your authentication token is correct. Due to security r
...
+
+    // Strip auth on port change (accounting for default ports)
+    const originalPort = HttpExecutor.getEffectivePort(originalUrl)
+    const redirectPort = HttpExecutor.getEffectivePort(redirectUrl)
+    return originalPort !== redirectPort
+  }
+
+  private static getEffectivePort(url: URL): number {
+    if (url.port) {
+      return parseInt(url.port, 10)
```
mmaietta: Sanity checking myself here. Can `port` ever be returned as not a number?
Wondering if it'd be better to do
```
const defaultPort = url.protocol === "https:" ? 443 : 80
try {
   return parseInt(url.port || defaultPort, 10)
} catch { }
return defaultPort
```

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: Addressed

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: The current approach new URL(`${protocol}//${hostname}${port}${path}`) actually works correctly. Since options.path typically starts with "/" and the fallback is "/", this construction is fine. Your suggestion of new URL(path, baseUrl) would work too, but the current implementation isn't broken.

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: Addressed

nogitting: This one is not needed (defensive programming). As long as `port` is not empty, it is safely parseable to an integer.

packages/builder-util-runtime/src/httpExecutor.ts
nogitting: Both boil down to the very same thing (in terms of effort/value).

packages/builder-util-runtime/src/httpExecutor.ts
nogitting: It would be quite an edge case, but yes, it is best to err on the side of caution here.

Was it tested though, James?

FringeNet:
The new cross-origin redirect logic breaks differential downloads for private GitHub repositories.

Problem: When downloading .blockmap files, GitHub redirects api.github.com → release-assets.githubusercontent.com, which the new logic correctly identifies as cross-origin and strips the auth header. However, private repo assets require authentication even on the Azure backend, causing 403 errors and "incorrect header check" failures when trying to parse HTML error pages as gzip.

```
Downloading update (already in progress)
Cannot download differentially, fallback to full download: Error: Cannot parse blockmap "https://api.github.com/repos/FringeNet/xx-xxxxxx/releases/assets/xxxxxxxx.blockmap", error: Error: incorrect header check
    at downloadBlockMap (C:\Program Files\xxxxxx\resources\app.asar\node_modules\electron-updater\out\AppUpdater.js:623:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Promise.all (index 1)
    at async NsisUpdater.differentialDownloadInstaller (C:\Program Files\xxxxxx\resources\app.asar\node_modules\electron-updater\out\AppUpdater.js:638:38)
    at async Object.task (C:\Program Files\xxxxxx\resources\app.asar\node_modules\electron-updater\out\NsisUpdater.js:49:22)
    at async NsisUpdater.executeDownload (C:\Program Files\xxxxxx\resources\app.asar\node_modules\electron-updater\out\AppUpdater.js:590:13)
```

Impact: All private GitHub repos using auto-updates with differential downloads will fail.

Suggested Fix: Add exception for GitHub asset redirects:

// In isCrossOriginRedirect function, before hostname comparison
```
if (originalUrl.hostname === "api.github.com" && 
    (redirectUrl.hostname === "release-assets.githubusercontent.com" || 
     redirectUrl.hostname === "objects.githubusercontent.com")) {
  return false // Preserve auth for GitHub asset redirects
}
```

Reproduction: Any private GitHub repo using electron-updater with default settings will encounter this error during blockmap downloads.

Updates work in my testing, but fallback to full download

It seems we are back to the yucky domain specific filtering, appreciate any feedback on this

nogitting:
@FringeNet, aside from the redacted parts, is this the original error message`https://api.github.com/repos/FringeNet/xx-xxxxxx/releases/assets/xxxxxxxx.blockmap` (particularly referring to the `blockmap` part)?
Can you reproduce the redirection and error via _Bruno_ or a similar tool? 

One thing for sure, private repo assets do and should not require authentication (via the _Authorization_ header) on the Azure backend in the event of a redirection.

FringeNet:
I figured it out. I am working on a beta branch which lacks the blockmap in release assets, resulting in 404's for that asset and hence falling back to full download.

The problem above was unique to the release channel I was using.

I have written a test file and added to the test workflow, all passing.

nogitting:
Well done. Did you re-try the end-to-end test?

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,68 @@ Please double check that your authentication token is correct. Due to security r
...
     const headers = newOptions.headers
     if (headers?.authorization) {
-      const parsedNewUrl = parseUrl(redirectUrl, options)
-      if (parsedNewUrl.hostname.endsWith(".amazonaws.com") || parsedNewUrl.searchParams.has("X-Amz-Credential")) {
+      // Parse original and redirect URLs to compare origins
+      const originalUrl = HttpExecutor.reconstructOriginalUrl(options)
+      const parsedRedirectUrl = parseUrl(redirectUrl, options)
+
+      // Strip authorization header on cross-origin redirects (different protocol, hostname, or port)
+      if (HttpExecutor.isCrossOriginRedirect(originalUrl, parsedRedirectUrl)) {
```
nogitting, line 327: One last thing, aside from my remark regarding `getEffectivePort`, if you add a debug/info log under this condition, something like `"Given the cross-origin redirect (from ${originalUrl.host} to ${parsedRedirectUrl.host}), the Authorization header will be stripped out."`, that will be of value.

FringeNet:
> Well done. Did you re-try the end-to-end test?

I created a patch using the JS output from this branch, and used it to test an update.
Only issue encountered was the blockmap file 404, which disappears when the file is present.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,71 @@ Please double check that your authentication token is correct. Due to security r
...
+
+    // Strip auth on any other protocol change
+    if (originalUrl.protocol !== redirectUrl.protocol) {
+      return true
+    }
+
+    // Strip auth on port change (accounting for default ports)
+    const originalPort = HttpExecutor.getEffectivePort(originalUrl)
+    const redirectPort = HttpExecutor.getEffectivePort(redirectUrl)
+    return originalPort !== redirectPort
```
aherrmann13: *curious*: why strip auth on port change + protocol change?  If the request URL is sending a redirect to same hostname different protocol or port I would think its probably fine to send the headers?  This is just for my own understanding the logic seems correct

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,71 @@ Please double check that your authentication token is correct. Due to security r
...
+  }
+
+  private static getEffectivePort(url: URL): number {
+    const defaultPort = url.protocol === "https:" ? 443 : 80
```
aherrmann13: I don't think you can assume non https has default port 80, two examples `wss:` would be 443 (`ws:` would be correct) and `ftp:` would be 21

nogitting: https://datatracker.ietf.org/doc/html/rfc7230#section-5.5 (from the **HTTP** spec)

packages/builder-util-runtime/src/httpExecutor.ts
nogitting: https://datatracker.ietf.org/doc/html/rfc7235#section-2.2

The _Requests_ library does not consider the special case of http-to-https (on default ports, for the same hostname) as a cross-origin redirect, only for the sake of backwards compatibility.

packages/builder-util-runtime/src/httpExecutor.ts
aherrmann13: I think this is correct in the context of an http request (RFC was for http spec) where you have http or https but not true in the context of a URI where multiple protocols apply

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: I agree with @nogitting

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: @aherrmann13 what would you propose, throw an error? return -1? Consider we only expect http/https to be used.

Example:
```
private static getEffectivePort(url: URL): number {
    if (url.port) {
      const parsed = parseInt(url.port, 10)
      if (!isNaN(parsed)) {
        return parsed
      }
    }

    if (url.protocol === "https:") {
      return 443
    }

    if (url.protocol === "http:") {
      return 80
    }

    // For other protocols, we cannot determine a default port.
    // Since this is an HttpExecutor, we only expect to see http/https protocols.
    // Returning a sentinel value to indicate "unknown default port".
    return -1
  }
  ```

Comment (aherrmann13) [reply packages/builder-util-runtime/src/httpExecutor.ts]: I would explicitly map the ones that want and throw an error to provide port number with any non http/https request.  I don't think its worth it to actually map out the edge cases?  but better to explicitly fail with a clear error message vs guessing the wrong port number and having harder to debug errors downstream

Comment (FringeNet) [reply packages/builder-util-runtime/src/httpExecutor.ts]:
```
private static getEffectivePort(url: URL): number {
    if (url.port) {
      const parsed = parseInt(url.port, 10)
      if (!isNaN(parsed)) {
        return parsed
      }
    }

    if (url.protocol === "https:") {
      return 443
    }

    if (url.protocol === "http:") {
      return 80
    }

    throw new Error(`A port must be specified for non-http/https protocols: ${url.protocol}`)
  }
  ```

nogitting: Well, as far as I am concerned, this method is unnecessarily getting cluttered. It is _**http**Executor_. The original request is expected to be a HTTP one. So is the redirected request. Now if the latter ever uses a different protocol, it will be a scheme mismatch, which will translate to a cross-origin redirect, as expected.
Even worse, with the new change, `getEffectivePort` will wrongly error out, in case of a non-http(s) protocol alongside an implicit port.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,84 @@ Please double check that your authentication token is correct. Due to security r
...
+    // Special case: allow http -> https redirect on same host with standard ports
+    // This matches the behavior of Python requests library for backward compatibility
+    if (
+      originalUrl.protocol === "http:" &&
+      HttpExecutor.getEffectivePort(originalUrl) === 80 &&
+      redirectUrl.protocol === "https:" &&
+      HttpExecutor.getEffectivePort(redirectUrl) === 443
+    ) {
+      return false
+    }
```
nogitting:
```suggestion
    // url.port returns an empty string if the port is omitted
    // or explicitly set to the default port for a given protocol. 
    if (
      originalUrl.protocol === "http:" &&
      // This can be replaced with `!originalUrl.port`, but for the sake of clarity.
      ["80", ""].includes(originalUrl.port) &&
      redirectUrl.protocol === "https:" &&
      // This can be replaced with `!redirectUrl.port`, but for the sake of clarity.
      ["443", ""].includes(redirectUrl.port)
    ) {
      return false
    }
```

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: I have to disagree. It should error out.

It is not possible for the application to handle a redirect to a non-HTTP/HTTPS protocol.

While a server could theoretically send such a redirect, the application's code will always fail in one of two ways:

Immediate Failure (Port Parsing): If the non-HTTP/HTTPS URL in the redirect does not have an explicit port, the getEffectivePort function in the base HttpExecutor class will throw an error, as it has no default port to assign.

Request Failure (Protocol Unsupported): If the URL does have a port, the request will be passed to either NodeHttpExecutor or ElectronHttpExecutor, both of which will fail because their underlying networking APIs do not support the given protocol.

Therefore, I am certain that a non-HTTP/HTTPS protocol can only be introduced via an external redirect, and that the system is not, and cannot be, a consumer of it.

packages/builder-util-runtime/src/httpExecutor.ts
```diff
@@ -319,14 +319,84 @@ Please double check that your authentication token is correct. Due to security r
...
+    // We only make an exception for HTTP -> HTTPS upgrades on standard ports for backward compatibility.
+
+    // Strip auth on any other protocol change
+    if (originalUrl.protocol !== redirectUrl.protocol) {
+      return true
+    }
+
+    // Strip auth on port change (accounting for default ports)
+    const originalPort = HttpExecutor.getEffectivePort(originalUrl)
+    const redirectPort = HttpExecutor.getEffectivePort(redirectUrl)
```
nogitting:
```suggestion
    const originalPort = originalUrl.port
    const redirectPort = redirectUrl.port
```

nogitting (commented):
`getEffectivePort` becomes unneeded.

packages/builder-util-runtime/src/httpExecutor.ts
nogitting: > It is not possible for the application to handle a redirect to a non-HTTP/HTTPS protocol.

Well, that is exactly my point. (cf. _Requests_)

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: originalPort = ""
redirectPort = "443"
"" !== "443" -> true. This is incorrect. The code would mistakenly identify this as a cross-origin redirect and strip the authorization header, which is a bug.

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: Functionally correct, but your comments are misleading.
A future developer immediately understands that it's checking the effective port, not just the raw string property.

packages/builder-util-runtime/src/httpExecutor.ts
FringeNet: That's exactly why the getEffectivePort function is necessary, and should error out.

packages/builder-util-runtime/src/httpExecutor.ts
nogitting: Do not ignore the context. When this point is reached, it means both URLs use the same protocol. `https://www.original.com` and `https://www.redirect.com:443` will both return `""`.

packages/builder-util-runtime/src/httpExecutor.ts
nogitting: Well, if you consider them misleading, we call it a day.

mmaietta:
@FringeNet just getting back from an extended vacation. Your work LGTM! Just wanted to double check though, is this ready for merge or are there any further action items you were looking to accomplish?

FringeNet:
> @FringeNet just getting back from an extended vacation. Your work LGTM! Just wanted to double check though, is this ready for merge or are there any further action items you were looking to accomplish?

I am happy with it, unless someone else has an issue they would like to bring up.

PR #9216, opened 2025-07-24
feat(github): Add `tagNamePrefix` option

taylorhadden:
This adds support for an optional `tagNamePrefix` setting in `GithubOptions`. This is joined by the a new `githubTagPrefix()` function that handles the creation of any prefix that should be prepended to the version number. When present, `tagNamePrefix` will override any `vPrefixedTagName` value.

This allows users of github updates to use any tag prefix they like. This makes the updater more adaptable and more easily used in monorepo projects that may use multiple different tag prefixes to mark the versions of different packages within the repo.

### DIFF
--- a/packages/builder-util-runtime/src/publishOptions.ts
+++ b/packages/builder-util-runtime/src/publishOptions.ts
@@ -98,8 +98,16 @@ export interface GithubOptions extends PublishConfiguration {
   /**
    * Whether to use `v`-prefixed tag name.
    * @default true
+   * @deprecated please use #tagNamePrefix instead.
    */
   readonly vPrefixedTagName?: boolean
+
+  /**
+   * If defined, sets the prefix of the tag name that comes before the semver number.
+   * e.g. "v" in "v1.2.3" or "test" of "test1.2.3".
+   * Overrides `vPrefixedTagName`
+   */
+  readonly tagNamePrefix?: string
 
   /**
    * The host (including the port if need).
@@ -141,6 +149,16 @@ export interface GithubOptions extends PublishConfiguration {
 /** @private */
 export function githubUrl(options: GithubOptions, defaultHost = "github.com") {
   return `${options.protocol || "https"}://${options.host || defaultHost}`
+}
+
+export function githubTagPrefix(options: GithubOptions) {
+  if (options.tagNamePrefix) {
+    return options.tagNamePrefix
+  }
+  if (options.vPrefixedTagName ?? true) {
+    return 'v'
+  }
+  return ''
 }
 
 /**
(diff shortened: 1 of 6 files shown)


taylorhadden:
Happy to add a changeset for this. Will hold off until I receive guidance.

packages/builder-util-runtime/src/publishOptions.ts
```diff
@@ -143,6 +150,12 @@ export function githubUrl(options: GithubOptions, defaultHost = "github.com") {
   return `${options.protocol || "https"}://${options.host || defaultHost}`
 }
 
+export function githubTagPrefix(options: GithubOptions) {
+  if (options.tagNamePrefix) return options.tagNamePrefix
+  if (options.vPrefixedTagName ?? true) return 'v'
```
mmaietta: Since we're creating a new prop that basically is duplicating `tagNamePrefix`, what are your thoughts on marking `@deprecated please use #tagNamePrefix` to the `vPrefixedTagName` prop?

Also, for if-statements, please use multi-line
```
if (options.tagNamePRefix) {
   return options.tagNamePrefix
}
```

taylorhadden: I think deprecation makes a lot of sense! I didn't want to be too heavy handed in the change.

Will update the if statements.

taylorhadden:
The build failure looks unrelated to these changes:

```
Test Files  1 failed (1)

      Tests  2 failed | 4 skipped (6)
 FAIL   node  test/src/concurrentBuildsTest.ts > win concurrent
   Start at  16:52:04
Error: EBUSY: resource busy or locked, unlink 'C:\Users\RUNNER~1\AppData\Local\Temp\t-iQzDLz\squirrel-windows-vendor-j\Squirrel-Releasify.log'
   Duration  163.78s (transform 256ms, setup 47ms, collect 1.51s, tests 161.64s, environment 0ms, prepare 174ms)
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯


 HTML  Report is generated
 FAIL   node  test/src/concurrentBuildsTest.ts > win concurrent - all targets
       You can run npx vite preview --outDir coverage to see the test results.
Error: D:\a\electron-builder\electron-builder\node_modules\.pnpm\7zip-bin@5.2.0\node_modules\7zip-bin\win\x64\7za.exe process failed 1
Exit code:
1
Output:
Exit code: 1. Command failed: D:\a\electron-builder\electron-builder\node_modules\.pnpm\7zip-bin@5.2.0\node_modules\7zip-bin\win\x64\7za.exe a -bd -mtc=off -mm=Copy -mcu C:\Users\RUNNER~1\AppData\Local\Temp\t-iQzDLz\test-project-1\dist\Test Concurrent-1.1.0-arm64.zip .

WARNING: The process cannot access the file because it is being used by another process.
.\Test Concurrent_ExecutionStub.exe
```

PR #9217, opened 2025-07-24
fix(nsis): fix file association

sabonerune:
The current File Associations that nsis writes to the registry will not appear in the "Open with" suggestions if the user or another application changes them.
This fix will allow it to appear in the "Open With" suggestions.

https://learn.microsoft.com/windows/win32/shell/fa-file-types
https://learn.microsoft.com/windows/win32/shell/fa-progids
https://learn.microsoft.com/windows/win32/shell/how-to-include-an-application-on-the-open-with-dialog-box

### DIFF
--- a/.changeset/lovely-brooms-move.md
+++ b/.changeset/lovely-brooms-move.md
@@ -1,0 +1,5 @@
+---
+"electron-builder-squirrel-windows": patch
+---
+
+fix: generate stubExecutableExe and sign it for squirrel.windows using new electron-builder-binaries asset
--- a/.changeset/lovely-geckos-chew.md
+++ b/.changeset/lovely-geckos-chew.md
@@ -1,0 +1,5 @@
+---
+"electron-builder-squirrel-windows": patch
+---
+
+fix(squirrel-windows): update loadingGif handling to prioritize user-defined option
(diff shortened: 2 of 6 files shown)


packages/app-builder-lib/templates/nsis/include/FileAssociation.nsh
```diff
@@ -103,10 +97,7 @@
 
 
 !macro APP_UNASSOCIATE EXT FILECLASS
-  ; Backup the previously associated file class
-  ReadRegStr $R0 SHELL_CONTEXT "Software\Classes\.${EXT}" `${FILECLASS}_backup`
-  WriteRegStr SHELL_CONTEXT "Software\Classes\.${EXT}" "" "$R0"
-
+  DeleteRegValue SHELL_CONTEXT "Software\Classes\.${EXT}\OpenWithProgids" "${FILECLASS}"
```
sabonerune, line 100: Please refer to the following document for the behavior when uninstalling.
https://learn.microsoft.com/windows/win32/shell/fa-file-types#deleting-registry-information-during-uninstallation

---

Maybe there are applications like electron-builder that write the ProgID to the default key only.
If you are considerate of such applications, it may be better not to write anything to the default key.
(But doing so would change previous behavior.)

PR #9223, opened 2025-07-29
refactor: replace `lodash.isequal`, `lodash.escaperegexp` with `es-toolkit/compat`

seokju-na:
This PR replaces `lodash.isequal`, `lodash.escaperegexp` dependency in `electron-updater` with `es-toolkit/compat`.

- lodash.isequal 4.5.0 is deprecated
- es-toolkit is more modern and performant than lodash.
- es-toolkit/compat passes all lodash test cases and maintains 100% interface [compatibility](https://es-toolkit.dev/compatibility.html).

### DIFF
--- a/packages/electron-updater/package.json
+++ b/packages/electron-updater/package.json
@@ -17,19 +17,16 @@
   ],
   "dependencies": {
     "builder-util-runtime": "workspace:*",
+    "es-toolkit": "^1.39.8",
     "fs-extra": "^10.1.0",
     "js-yaml": "^4.1.0",
     "lazy-val": "^1.0.5",
-    "lodash.escaperegexp": "^4.1.2",
-    "lodash.isequal": "^4.5.0",
     "semver": "^7.6.3",
     "tiny-typed-emitter": "^2.1.0"
   },
   "devDependencies": {
     "@types/fs-extra": "9.0.13",
     "@types/js-yaml": "4.0.3",
-    "@types/lodash.escaperegexp": "4.1.6",
-    "@types/lodash.isequal": "4.5.5",
     "@types/semver": "^7.3.13",
     "electron": "^31.2.1"
   },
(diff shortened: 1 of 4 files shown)


absidue:
electron-updater is shipped with applications, so pulling in a 2.49 MB dependency into every app that uses auto-updates just for two tiny utility functions really doesn't seem like a good idea.

And no es-toolkit being tree-shakable is not good enough because many Electron applications do not use a bundler, so they will end up with the entirety of es-toolkit being shipped to their users.

chyzwar:
I would like to add that whole electron-builder take 285.7 MB !! I am not sure how an extra 2MB of dependencies will affect this. Biggest offender is app-builder-bin, taking 216MB. 

Electron-builder have massive dependency tree https://npmgraph.js.org/?q=electron-builder 
My personal preference is to have the small number of well audited but bigger deps. 
As it is now, electro-builder is impossible to audit, and it sits in the middle of the supply chain of every electron app.

absidue:
@chyzwar I suggest you carefully re-read my message above, we are not talking about the size of development dependencies but about the size of runtime dependencies aka the ones packaged with applications that are installed on the users computer. With this pull request the install size of every Electron application that uses electron-builder and supports automatic updates would completely unnecessarily increase by 2.49 MB, which is unacceptable.

mmaietta:
> whole electron-builder take 285.7 MB !! I am not sure how an extra 2MB of dependencies will affect this. Biggest offender is app-builder-bin, taking 216MB.

Note: Working on removing app-builder-bin from the dependency tree, but it's a slow process since a LOT of heavy lifting is being done by the Golang binary. https://github.com/develar/app-builder There's also performance loss in migrating from Go implementations => JS.

> the size of runtime dependencies aka the ones packaged with applications that are installed on the users computer. With this pull request the install size of every Electron application that uses electron-builder and supports automatic updates would completely unnecessarily increase by 2.49 MB, which is unacceptable.

I'm aligned with this comment. 

An alternative approach is to copy the specific logic we need out (since it's only a couple functions). I took that approach for this code that pulled a trick from lodash's implementation IIRC to avoid prototype pollution
https://github.com/electron-userland/electron-builder/blob/716cb71b9f4366a07bdd95c75be83dfbdffa16de/packages/builder-util/src/mapper.ts#L3-L24

Issue #7040, opened 2022-08-03
Invalid code signing for MAS build (version 23.3.3)


thomasdao:
* **Electron-Builder Version**:  23.3.3

* **Node Version**:  16

* **Electron Version**: 19.0.10

* **Target**: MAS

From version 23.3.0 and later, signing MAS build will use `3rd Party Mac Developer Application` certificate by default, and this causes invalid build for my app. When upload to the Mac App Store, I see the below error:

`Asset validation failed (90284)
Invalid Code Signing. The executable 'com.app.name.pkg/Payload/AppName.app/Contents/Library/LoginItems/AppName Login Helper.app/Contents/MacOS/AppName Login Helper' must be signed with the certificate that is contained in the provisioning profile. (ID: c3f46e59-cbb9-4b15-841d-04ffb05cb2b7)`

I believe when the certificate doesn’t match the one embedded in provisioning profile, the build will show this error. However, when I set `identity` in `mas` key to `Apple Distribution`, I get the below error:

`  ⨯ Cannot find valid "3rd Party Mac Developer Installer" identity to sign MAS installer, please see https://electron.build/code-signing`

I believe this PR https://github.com/electron-userland/electron-builder/pull/6970 is related. When I manually patch `macPackager.js` to change `["3rd Party Mac Developer Application", "Apple Distribution"]` to `["Apple Distribution", "3rd Party Mac Developer Application"]`, the app can be signed and uploaded to the Mac App Store successfully.

Should we use `Apple Distribution` by default and move `3rd Party Mac Developer Application` to fallback? This should work for issue https://github.com/electron-userland/electron-builder/issues/6621 where `Apple Distribution` cannot be found so electron-builder will use `3rd Party Mac Developer Application` certificate.

mmaietta:
> I believe this PR https://github.com/electron-userland/electron-builder/pull/6970 is related. When I manually patch macPackager.js to change ["3rd Party Mac Developer Application", "Apple Distribution"] to ["Apple Distribution", "3rd Party Mac Developer Application"], the app can be signed and uploaded to the Mac App Store successfully.
> Should we use Apple Distribution by default and move 3rd Party Mac Developer Application to fallback? This should work for issue https://github.com/electron-userland/electron-builder/issues/6621 where Apple Distribution cannot be found so electron-builder will use 3rd Party Mac Developer Application certificate.

Can you open a PR for swapping the two values? :)

jeanfbrito:
[...]

Issue #6523, opened 2021-12-30
InvalidConfigurationError: Cannot find valid "3rd Party Mac Developer Installer" identity to sign MAS installer


goodhyun:
* **Electron-Builder Version**: 22.14.11

* **Node Version**: v17.2.0

* **Electron Version**: 14.2.3
* **Electron Type (current, beta, nightly)**:

* **Target**: mac/mas

It complains like that when you try to build a final mas pkg even if I have a valid identity in the keychain. It seems to have detected the certificate according to a debug message. 

```
 7) F561~~C9 "3rd Party Mac Developer Installer: ~~ Inc. (~~~)"
     7 valid identities found
```
just fyi, `export CSC_LINK` didn't help.

goodhyun:
As a stopgap measure, I commented out that portion of code in the `node_modules/app-builder-lib/out/macPackager.js` and just did `productbuild` manually

```
await this.doSign(signOptions);
        // https://github.com/electron-userland/electron-builder/issues/1196#issuecomment-312310209
        if (masOptions != null && !isDevelopment) {
            const certType = isDevelopment ? "Mac Developer" : "3rd Party Mac Developer Installer";
            const masInstallerIdentity = await macCodeSign_1.findIdentity(certType, masOptions.identity, keychainFile);
            // if (masInstallerIdentity == null) {
            //     throw new builder_util_1.InvalidConfigurationError(`Cannot find valid "${certType}" identity to sign MAS installer, please see https://electron.build/code-signing`);
            // }
            // // mas uploaded to AppStore, so, use "-" instead of space for name
            // const artifactName = this.expandArtifactNamePattern(masOptions, "pkg", arch);
            // const artifactPath = path.join(outDir, artifactName);
            // await this.doFlat(appPath, artifactPath, masInstallerIdentity, keychainFile);
            // await this.dispatchArtifactCreated(artifactPath, null, builder_util_1.Arch.x64, this.computeSafeArtifactName(artifactName, "pkg", arch, true, this.platformSpecificBuildOptions.defaultArch));
        }
```

robertpatrick:
@goodhyun From your original snippet of the debug message, it looks like not only is it finding the identity you want but 6 other valid identities.  Have you tried using the `CSC_NAME` environment variable to specify which of the 7 installed identities that you want it to use?

goodhyun:
[...]

PR #9226, opened 2025-07-31
feat: Support Apple Distribution Cert for MAS

hristoiankov:
Add `Apple Distribution` to the list of supported certs for MAS. This cert has replaced `3rd Party Mac Developer Installer` in the Apple Developer program. This change will allow either of these two certs to be used to sign a MAS pkg for the Apple App Store submission.

This change also expands the `findIdentity` function to take in a list of accepted certs rather than just a single cert for all cases where a cert needs to be pulled from the keystore. This also simplifies some existing code that was already trying to do this same type of thing in a more roundabout way.

Issue Identified In:
https://github.com/electron-userland/electron-builder/issues/6523
https://github.com/electron-userland/electron-builder/issues/7040

### DIFF
--- a/.changeset/lovely-brooms-move.md
+++ b/.changeset/lovely-brooms-move.md
@@ -1,0 +1,5 @@
+---
+"electron-builder-squirrel-windows": patch
+---
+
+fix: generate stubExecutableExe and sign it for squirrel.windows using new electron-builder-binaries asset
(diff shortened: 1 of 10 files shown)


mmaietta (approved):

hristoiankov:
Don't merge this one yet, I need to run a successful mas signing and submission with Apple first to have full confidence. Thanks!

mmaietta:
Sure thing, converting PR to Draft for now then so as to enforce no-merge.

Issue #9005, opened 2025-04-01
Unable to find publish configuration when publishing prebuilt artifacts with CLI


seanssel:
I'm attempting to use `electron-builder publish` to publish prebuilt artifacts to DigitalOcean Spaces. Up until this point, I have been on `electron-builder 24.6.4` and have been able to publish artifacts fine with `--publish`.

`node` v22.14.0
`npm` v10.9.2
`electron-builder` v26.0.12

Output:
I have tried all kinds of variations with this command (changing arg order, using `--config`, absolute paths, etc). 

```console
❱ npx electron-builder publish -c config/electron-builder-staging.config.js -f dist/some_artifact.dmg
(node:69695) Warning: "version" is a reserved word.
Please do one of the following:
- Disable version with `yargs.version(false)` if using "version" as an option
- Use the built-in `yargs.version` method instead (if applicable)
- Use a different option key
https://yargs.js.org/docs/#api-reference-version
(Use `node --trace-warnings ...` to show where the warning was created)
  • electron-builder  version=26.0.12 os=23.5.0
  ⨯ error publishing  message=Error: unable to find any publish configuration
    at publishPackageWithTasks (/Users/somename/dev/antarestech/ac-electron/node_modules/electron-builder/src/publish.ts:94:13)
```

Here is the config file:

```js
module.exports = function electronBuilderStagingConfig() {
  return {
    extends: './config/electron-builder-common.config.js',
    forceCodeSigning: false,
    afterPack: 'scripts/setElectronFuses.js',
    win: {
      icon: 'electron/assets/icon.ico',
      target: ['nsis'],
      signExts: ['.node', '.ps1'],
      signtoolOptions: {
        sign: './scripts/customSignWin.js',
        signingHashAlgorithms: ['sha256'],
        rfc3161TimeStampServer: 'http://timestamp.digicert.com',
      },
    },
    publish: {
      provider: 'spaces',
      name: '<edited-for-privacy>',
      region: '<edited-for-privacy>',
      path: '<edited-for-privacy>',
    },
  };
};
```

Also I assume I need to upload everything that the `--publish` version does; i.e. the `latest` .yml files, blockmaps, zip (on mac). When I do that do I pass files with `-f somefile1 -f somefile2 -f ...`? Documentation here could be improved.

seanssel:
[...]

PR #9227, opened 2025-07-31
fix: fix argument names mismatch in publish command to make it work

choegyumin:
Fixes:
- https://github.com/electron-userland/electron-builder/issues/9005 

Caused by:
- https://github.com/electron-userland/electron-builder/pull/8596 

---

**Fixes mismatched argument names:**

`config` ↔ `configurationFilePath`

**Resolves conflicts caused by the reserved word 'version' in [yargs](https://github.com/yargs/yargs), and removes warning messages (Output before fix):**

```bash
$ npx electron-builder publish -f foobar.dmg
(node:7826) Warning: "version" is a reserved word.
Please do one of the following:
- Disable version with `yargs.version(false)` if using "version" as an option
- Use the built-in `yargs.version` method instead (if applicable)
- Use a different option key
https://yargs.js.org/docs/#api-reference-version
```

### DIFF
--- a/packages/electron-builder/src/publish.ts
+++ b/packages/electron-builder/src/publish.ts
@@ -15,6 +15,7 @@ export function configurePublishCommand(yargs: yargs.Argv): yargs.Argv {
   // https://github.com/yargs/yargs/issues/760
   // demandOption is required to be set
   return yargs
+    .version(false)
     .parserConfiguration({
       "camel-case-expansion": false,
     })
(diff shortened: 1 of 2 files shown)


packages/electron-builder/src/publish.ts
```diff
@@ -31,7 +32,7 @@ export function configurePublishCommand(yargs: yargs.Argv): yargs.Argv {
       description: "The app/build version used when searching for an upload release (used by some Publishers)",
     })
     .option("config", {
-      alias: ["c"],
+      alias: ["c", "configurationFilePath"],
```
choegyumin, line 35: To ensure backward compatibility of the `publish` function, I added aliases to the command options instead of changing the function parameter names.

PR #9231, opened 2025-07-31
fix: add quotation marks around variable in AppArmor profile

### LINKED ISSUE #9230: Invalid syntax on electron-builder's AppArmor profile when app's product name contains a space
- electron-builder version: 26.0.19 (also applies to every previous version with the AppArmor profile)
- Electron version: 37.2.5 (also applies to previous versions)

On Ubuntu 24.04+, an AppArmor profile is needed to give an Electron app the permission for user namespaces so that the sandbox (and thus the whole app) can work properly (see electron/electron#41066).

However, the template for that AppArmor profile (https://github.com/electron-userland/electron-builder/blob/88cc0b06dba22139721fd1e04f6a1cf2d447edbd/packages/app-builder-lib/templates/linux/apparmor-profile.tpl, added in #8636) can have an incorrect syntax if an app's product name contains a space in it, as there are no quotation marks in the first variable for the executable:

(example with a test app having a product name with a space)
Image

This makes every app with a space in it not being able to run on Ubuntu 24.04+ without doing a workaround.

I will make a PR to fix this issue in a moment.

rastiqdev:
fixes #9230

### DIFF
--- a/.changeset/brown-boxes-know.md
+++ b/.changeset/brown-boxes-know.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: add quotation marks around variable in AppArmor profile
--- a/packages/app-builder-lib/templates/linux/apparmor-profile.tpl
+++ b/packages/app-builder-lib/templates/linux/apparmor-profile.tpl
@@ -1,7 +1,7 @@
 abi <abi/4.0>,
 include <tunables/global>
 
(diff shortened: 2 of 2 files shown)


mmaietta (approved):

peteringram0:
@rastiqdev how do i get electron-builder to generate this apparmor.d file please?

rastiqdev:
@peteringram0 It is supposed to be generated by default as long as you have a Linux target.

PR #9232, opened 2025-08-01
fix: auto-update powershell script

michalzaq12:
After the `Get-AuthenticodeSignature` command fails, in the error handler there is an additional `ConvertTo-Json test` command that should be called in the same way as `Get-AuthenticodeSignature`. Without this, `Get-AuthenticodeSignature` could fail simply due to the workarounds used (using `shell:true` `chcp 65001 >NUL` https://github.com/electron-userland/electron-builder/pull/7230 and resetting `PSModulePath` https://github.com/electron-userland/electron-builder/pull/8051), which is unintended behavior. 

After this fix, when the workarounds fail, the `ConvertTo-Json test` command will also fail (with the message `Ignoring signature validation due to unsupported PowerShell version`) and the update will be installed.

### DIFF
--- a/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
+++ b/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
@@ -1,8 +1,23 @@
 import { parseDn } from "builder-util-runtime"
-import { execFile, execFileSync } from "child_process"
+import { execFile, execFileSync, ExecFileOptions } from "child_process"
 import * as os from "os"
 import { Logger } from "./types"
 import * as path from "path"
+
+function preparePowerShellExec(command: string) {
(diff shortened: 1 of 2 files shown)


packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts
```diff
@@ -123,7 +127,7 @@ function handleError(logger: Logger, error: Error | null, stderr: string | null,
   }
 
   try {
-    execFileSync("powershell.exe", ["-NoProfile", "-NonInteractive", "-Command", "ConvertTo-Json test"], { timeout: 10 * 1000 } as any)
+    execFileSync(...preparePowerShellExec("ConvertTo-Json test"))
```
mmaietta: Shouldn't we retain the 10sec timeout here? Curious as to why it was removed.

PR #9240, opened 2025-08-06
fix: improve package manager detection for Yarn Berry and update Yarn install arguments

beyondkmp:


packages/app-builder-lib/src/util/yarn.ts
```diff
@@ -94,7 +94,7 @@ export async function installDependencies(config: Configuration, { appDir, proje
   }
 
   if (pm === PM.YARN) {
-    execArgs.push("--prefer-offline")
+    execArgs.push("--cached")
```
mmaietta, line 97: Should this only be for `=== PM.YARN_BERRY` and kept as `--prefer-offline` for `PM.YARN`?

mmaietta:
Adding this into https://github.com/electron-userland/electron-builder/pull/9309, closing this as stale

PR #9245, opened 2025-08-07
fix(nsis): undefined vars when customCheckAppRunning is defined

dlon:
Currently, if your NSIS script defines `customCheckAppRunning`, building fails with `Variable "CmdPath" not referenced or never set, wasting memory!`. PR fixes this by not always declaring the unused variables.

### DIFF
--- a/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+++ b/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
@@ -33,12 +33,12 @@
   Var /GLOBAL CmdPath
   Var /GLOBAL FindPath
   Var /GLOBAL PowerShellPath
+  StrCpy $CmdPath "$SYSDIR\cmd.exe"
+  StrCpy $FindPath "$SYSDIR\find.exe"
+  StrCpy $PowerShellPath "$SYSDIR\WindowsPowerShell\v1.0\powershell.exe"
   !ifmacrodef customCheckAppRunning
     !insertmacro customCheckAppRunning
   !else
-    StrCpy $CmdPath "$SYSDIR\cmd.exe"
-    StrCpy $FindPath "$SYSDIR\find.exe"
-    StrCpy $PowerShellPath "$SYSDIR\WindowsPowerShell\v1.0\powershell.exe"
     !insertmacro IS_POWERSHELL_AVAILABLE
     !insertmacro _CHECK_APP_RUNNING
   !endif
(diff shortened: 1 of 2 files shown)


mmaietta:
Can you please add a changeset to this using `pnpm generate-changeset` so that it's properly tagged in the release notes? It seems I can't do so from my side

dlon:
> Can you please add a changeset to this using `pnpm generate-changeset` so that it's properly tagged in the release notes? It seems I can't do so from my side

Sure. Done

packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
```diff
@@ -30,12 +30,12 @@
...
 !macro CHECK_APP_RUNNING
-  Var /GLOBAL CmdPath
-  Var /GLOBAL FindPath
-  Var /GLOBAL PowerShellPath
   !ifmacrodef customCheckAppRunning
     !insertmacro customCheckAppRunning
   !else
+    Var /GLOBAL CmdPath
+    Var /GLOBAL FindPath
+    Var /GLOBAL PowerShellPath
```
mmaietta: I'm looking further at the code and it seems to assume that `CmdPath` `FindPath` and `PowerShellPath` must exist for further functionality to succeed. I think moving the `Var`s here may actually cause a downstream bug of a var not being defined, no?
I'm looking at
`IS_POWERSHELL_AVAILABLE`
https://github.com/electron-userland/electron-builder/blob/44b28997f15314730d1bb69303a47dc26f7950d1/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh#L50-L57
`FIND_PROCESS`
https://github.com/electron-userland/electron-builder/blob/44b28997f15314730d1bb69303a47dc26f7950d1/packages/app-builder-lib/templates/nsis/include/allowOnlyOneInstallerInstance.nsh#L66-L70

dlon: Are those functions/macros not only called when not defining `customCheckAppRunning`?

I guess the `StrCpy` calls could be moved out instead?

mmaietta: Yeah, I'm thinking move it out instead, and then the `customCheckAppRunning` can override/overwrite those `Vars` if necessary?

dlon: Done

Issue #9247, opened 2025-08-13
Unhandled net::ERR_NETWORK_IO_SUSPENDED leads to app crash when resuming from sleep during download


DamonYu6:
**Version**: 
electron-updater: 6.6.7

**Target**: 
Windows, macOS

**Description of the problem**
We are monitoring our application with Sentry and have noticed a significant number of net::* related errors being reported when a new version of our app is released. These errors originate from electron-updater.

The core issue is an unhandled error that occurs when a network request is suspended, which typically happens when a user's computer goes into sleep mode. This unhandled error bubbles up as an uncaughtException in the main process, leading to an application crash if a global error handler is not in place.

**Similar Issues**:
https://github.com/electron-userland/electron-builder/issues/2398 

**Steps to reproduce**
1. Start the application and trigger an update check that begins downloading a new package.
2. While the download is in progress, close the laptop lid to put the computer into sleep mode.
3. Wait a few seconds, then re-open the lid to wake the computer.
4. The application will crash, or if you have a global handler, the following error will be caught by process.on('uncaughtException'): Error: net::ERR_NETWORK_IO_SUSPENDED

**Expected behavior**
electron-updater should internally catch network suspension errors, reject the corresponding promise, and handle the download failure gracefully instead of letting it bubble up as an unhandled exception.

**Actual behavior**
An unhandled net::ERR_NETWORK_IO_SUSPENDED error is thrown, causing the application to crash.

**Problem Analysis & Proposed Solution**
After investigating the electron-updater source code, I believe the issue lies in the MultipleRangeDownloader.

In this file:
[packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts#L92](https://github.com/electron-userland/electron-builder/blob/44b28997f15314730d1bb69303a47dc26f7950d1/packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts#L92)

The callback function passed to httpExecutor.request does not have a listener for the response's 'error' event

Adding a similar error handler to reject the promise within the MultipleRangeDownloader should fix this uncaught exception and prevent the application from crashing.

<img width="535" height="976" alt="Image" src="https://github.com/user-attachments/assets/bcfc1b91-d6d4-427e-8610-ce27fdc98b92" />

<img width="1729" height="411" alt="Image" src="https://github.com/user-attachments/assets/cdedd2cc-65c1-4948-ada9-8e783031d92b" />

beyondkmp:
Very good catch.

PR #9248, opened 2025-08-13
fix: uncaught exception during differential downloading via multipleRangDownloader

DamonYu6:
fix https://github.com/electron-userland/electron-builder/issues/9247

### DIFF
--- a/.changeset/fuzzy-kings-cry.md
+++ b/.changeset/fuzzy-kings-cry.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": major
+---
+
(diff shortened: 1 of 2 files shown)


beyondkmp:
very good catch.

.changeset/fuzzy-kings-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"electron-updater": major
```
beyondkmp: not major, it's a patch

PayneFuRC (approved):

mmaietta:
LGTM. Broken test has recently become unstable. Proceeding with merge

PR #9251, opened 2025-08-20
fix: two flaky tests

daihere1993:
You can find these failed cases in https://github.com/electron-userland/electron-builder/actions/runs/16897683885/job/48041891467?pr=9209

Fix 1: excluded `updateKBytes` so that `test/src/mac/macArchiveTest.ts > pkg scripts` passes
Fix 2: improved the sort algorithm to make sure the artifact order is deterministic

### DIFF
--- a/test/src/helpers/packTester.ts
+++ b/test/src/helpers/packTester.ts
@@ -223,6 +223,49 @@ export function getFixtureDir() {
   return path.join(__dirname, "..", "..", "fixtures")
 }
 
+/**
+ * Determines the priority of a file based on its extension for sorting.
+ * Lower numbers have higher priority in the sort order.
+ */
+function getFileTypePriority(file: string): number {
+  if (file.endsWith(".dmg")) return 1
+  if (file.endsWith(".zip")) return 2
+  if (file.endsWith(".blockmap")) return 3
+  if (file.endsWith(".yml")) return 4
+  return 5
+}
+
+/**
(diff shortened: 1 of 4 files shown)


mmaietta:
This is amazing, thank you so much for doing this!

test/src/helpers/packTester.ts
```diff
@@ -223,6 +223,74 @@ export function getFixtureDir() {
...
+  if (file.endsWith(".asar")) return 17
+
+  // Metadata and auxiliary files
+  if (file.endsWith(".blockmap")) return 18
+  if (file.endsWith(".yml")) return 19
+  if (file.endsWith(".yaml")) return 19
+
+  // Other files
+  return 20
+}
```
mmaietta, line 264: To make this more extensible in the future, and improve readability. Any time a new extension is added, no numbers need to be updated. What are your thoughts?

```
const ordering = [
".dmg",
".exe",
".msi",
".pkg",
".deb",
".rpm",
".AppImage",
".appx",
".snap",
".flatpak",

  // Archive formats
".zip",
".7z",
".tar.gz",
".tar.xz",
".tar.bz2",

  // Package formats
".nupkg",
".asar",

  // Metadata and auxiliary files
".blockmap",
".yml",
".yaml",
]

return ordering.length - ordering.findIndex(file)
```

daihere1993: Good point, this way is easy to maintain!

PR #9257, opened 2025-08-26
fix: detect wrong package manager for two package.json structure

DamonYu6:
When using yarn to manage a two package.json structure, if pnpm is installed globally (i.e., the PNPM_HOME environment variable exists), it may result in the package manager being detected incorrectly.


beyondkmp:
very good catch.

PR #9260, opened 2025-08-26
fix: Change dependency collection to parse from root

Julusian:
This resolves an issue I was seeing when using a yarn 4 monorepo, where is appeared that `npm list` disagreed with the layout on disk and produced a tree with deduplications that resulted in a bunch of packages to go missing.

For my project, the output of the `npm list` command was: https://gist.github.com/Julusian/b4f5fe20db4f211753fa0ef8292901c8.  
Looking at this output, the bits worth noting are the locations of `balanced-match` and `brace-expansion`. There are 2 of each of them, one where `brace-expansion` has a `_dependencies` containing `balanced-match` but no `dependencies`, and the other where it has `balanced-match` in both.

The one where it is in both is apparently a child of `cacache` being a dependency of the workspace root, but according to yarn and the yarn.lock that `cacache` dependency should be a few layers deep as a child of something. It is also worth noting here that npm complains `"extraneous: cacache@18.0.2 /home/julus/Projects/companion/companion/node_modules/cacache"`, so it looks to me that npm got rather confused about where this belongs and so dumped it at the root level.

Because it was at the root level, collectAllDependencies would never see the proper `brace-expansion`, only the one with deduplicated dependencies, resulting in it not knowing that `balanced-match` should be packaged and it would be missing at runtime.

I am not sure if this is the best fix, but it resolves my issue. I suspect that this approach is ok as this is simply building the tree of what dependencies there are installed within the workspace, but it isnt doing anything with them at this stage; the later step of actually extracting the production dependency graph is still using the package/project root.

-----

While debugging this, my script for testing was:
```
import { YarnNodeModulesCollector } from './out/node-module-collector/yarnNodeModulesCollector.js'
import { TmpDir } from 'builder-util'

const tmp = new TmpDir()
try {
	const collector = new YarnNodeModulesCollector('/home/julus/Projects/companion/companion/launcher', tmp)
	const res = await collector.getNodeModules()

	console.log(
		'res',
		res.length,
		res.find((r) => r.name.startsWith('brace-exp')),
		res.find((r) => r.name === 'balanced-match')
	)
} finally {
	tmp.cleanupSync()
}
```
If you want to try it yourself, it should be enough to clone and run `yarn install` for the commit https://github.com/bitfocus/companion/tree/91d92c5d6d324e86ce032bec8a92212a042267b6, then run that script against it with the corrected path, and make sure that the console.log logs that it found the packages being verified.

My testing was with electron-builder/app-builder-lib 26.0.19, node 22.18.0, yarn 4.9.2 and npm 10.9.3

### DIFF
--- a/test/snapshots/concurrentBuildsTest.js.snap
+++ b/test/snapshots/concurrentBuildsTest.js.snap
@@ -4,14 +4,12 @@ exports[`linux concurrent 1`] = `
 {
   "linux": [
     {
+      "arch": "x64",
+      "file": "Test Concurrent-1.1.0-x86_64.rpm",
+    },
+    {
       "arch": "armv7l",
-      "file": "Test Concurrent-1.1.0-armv7l.AppImage",
-      "safeArtifactName": "Test-Concurrent-1.1.0-armv7l.AppImage",
-      "updateInfo": {
-        "blockMapSize": "@blockMapSize",
-        "sha512": "@sha512",
-        "size": "@size",
-      },
+      "file": "Test Concurrent-1.1.0-armv7l.rpm",
     },
     {
       "arch": "x64",
@@ -25,11 +23,13 @@ exports[`linux concurrent 1`] = `
     },
     {
       "arch": "armv7l",
-      "file": "Test Concurrent-1.1.0-armv7l.rpm",
-    },
-    {
-      "arch": "x64",
-      "file": "Test Concurrent-1.1.0-x86_64.rpm",
+      "file": "Test Concurrent-1.1.0-armv7l.AppImage",
+      "safeArtifactName": "Test-Concurrent-1.1.0-armv7l.AppImage",
+      "updateInfo": {
+        "blockMapSize": "@blockMapSize",
+        "sha512": "@sha512",
+        "size": "@size",
+      },
     },
   ],
 }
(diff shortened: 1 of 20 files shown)


packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -23,7 +23,7 @@ export abstract class NodeModulesCollector<T extends Dependency<T, OptionalsType
   public async getNodeModules(): Promise<NodeModuleInfo[]> {
     const tree: T = await this.getDependenciesTree()
     const realTree: T = this.getTreeFromWorkspaces(tree)
-    this.collectAllDependencies(realTree)
+    this.collectAllDependencies(tree) // Parse from the root, as npm list can host and deduplicate across projects in the workspace
     this.extractProductionDependencyGraph(realTree, "." /*root project name*/)
 
     const hoisterResult: HoisterResult = hoist(this.transToHoisterTree(this.productionGraph), { check: true })
```
beyondkmp, line 29:
```
 public async getNodeModules(): Promise<NodeModuleInfo[]> {
    const tree: T = await this.getDependenciesTree()
    this.collectAllDependencies(tree)
    const realTree: T = this.getTreeFromWorkspaces(tree)
    this.extractProductionDependencyGraph(realTree, "." /*root project name*/)

    const hoisterResult: HoisterResult = hoist(this.transToHoisterTree(this.productionGraph), { check: true })
    this._getNodeModules(hoisterResult.dependencies, this.nodeModules)

    return this.nodeModules
  }

```

beyondkmp (changes requested):

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
beyondkmp: Check if this modification resolves your issue? Your approach would break the normal npm workspace functionality.

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -23,7 +23,7 @@ export abstract class NodeModulesCollector<T extends Dependency<T, OptionalsType
   public async getNodeModules(): Promise<NodeModuleInfo[]> {
     const tree: T = await this.getDependenciesTree()
     const realTree: T = this.getTreeFromWorkspaces(tree)
-    this.collectAllDependencies(realTree)
+    this.collectAllDependencies(tree) // Parse from the root, as npm list can host and deduplicate across projects in the workspace
```
beyondkmp, line 25:
```typescript
   const tree: T = await this.getDependenciesTree()
    this.collectAllDependencies(tree) // Parse from the root, as npm list can host and deduplicate across projects in the workspace
    const realTree: T = this.getTreeFromWorkspaces(tree)
````

beyondkmp:
very good catch.

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
Julusian: updated

beyondkmp (approved):

beyondkmp:
LGTM

PR #9262, opened 2025-08-29
feat: improve detecting pm and support bun

beyondkmp:
The previous logic determined it was pnpm by checking if pnpm existed in process.env. Actually, we should establish a priority: first check npm_config_user_agent, after all checks are completed, then check exec path, and finally check packageManager, which would be more accurate.


packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -42,26 +43,33 @@ export function getPackageManagerCommand(pm: PM) {
...
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf8"))
+
+    if (packageJson.packageManager) {
+      if (packageJson.packageManager.startsWith("pnpm@")) return PM.PNPM
+      if (packageJson.packageManager.startsWith("yarn@")) return PM.YARN
+      if (packageJson.packageManager.startsWith("npm@")) return PM.NPM
+    }
+  }
+
+  return null
```
mmaietta: I think we can optimize this here, but also, this new logic seems to remove `process.env.YARN_VERSION`, `process.env.YARN_REGISTRY`, and `process.env.npm_package_json` logic. `YARN_BERRY` detection was removed as well. Is that intentional?

mmaietta: To make your **new** code more easy to maintain/extend, I think this can be a drop-in improvement:
```
const pmMap = {
  pnpm: PN.PNPM,
  yarn: PM.YARN,
  npm: PM.NPM
};

const packageJsonPath = path.join(process.cwd(), "package.json");
const packageManager = fs.existsSync(packageJsonPath) ? JSON.parse(fs.readFileSync(packageJsonPath, "utf8"))?.packageManager : undefined;

const priorityChecklist = [
  (key: string) => process.env.npm_config_user_agent?.includes(key),
  (key: string) => process.env.npm_execpath?.includes(key),
  (key: string) => packageManager?.startsWith(`${key}@`)
];

for (checker of priorityChecklist) {
  for (identifier of Object.keys(pmMap)) {
    if (checker(identifier)) {
      return pmMap[identifier]
    }
  }
}
```

avarayr:
Please consider merging https://github.com/beyondkmp/electron-builder/pull/30 to this PR to re-add bun support

packages/app-builder-lib/src/node-module-collector/packageManager.ts
avarayr: @mmaietta good points on YARN BERRY detection. these PRs are getting very quick LGTMs which resulted in #9067 inadvertedly removing bun support, which slipped thru the cracks it seems like. 

I'm not well familiar with the repo structure (i'm just here because my builds are breaking haha!) but would it be feasible to add tests that reinforce behavior with different package managers, including bun? Just so that we don't rely on stuff like this happening again

beyondkmp: @avarayr Does bun support bun list? If so, I can create another PR to support native bun without setting node_modules mode.

beyondkmp: > I think we can optimize this here, but also, this new logic seems to remove `process.env.YARN_VERSION`, `process.env.YARN_REGISTRY`, and `process.env.npm_package_json` logic. `YARN_BERRY` detection was removed as well. Is that intentional?

process.env might introduce a global env that could cause issues with judgment, so it was removed. Currently, npm_config_user_agent and exec_path are sufficient to determine the package manager.

beyondkmp: > To make your **new** code more easy to maintain/extend, I think this can be a drop-in improvement:
> 
> ```
> const pmMap = {
>   pnpm: PN.PNPM,
>   yarn: PM.YARN,
>   npm: PM.NPM
> };
> 
> const packageJsonPath = path.join(process.cwd(), "package.json");
> const packageManager = fs.existsSync(packageJsonPath) ? JSON.parse(fs.readFileSync(packageJsonPath, "utf8"))?.packageManager : undefined;
> 
> const priorityChecklist = [
>   (key: string) => process.env.npm_config_user_agent?.includes(key),
>   (key: string) => process.env.npm_execpath?.includes(key),
>   (key: string) => packageManager?.startsWith(`${key}@`)
> ];
> 
> for (checker of priorityChecklist) {
>   for (identifier of Object.keys(pmMap)) {
>     if (checker(identifier)) {
>       return pmMap[identifier]
>     }
>   }
> }
> ```

Good suggestion. Fixed.

beyondkmp:
> Please consider merging [beyondkmp#30](https://github.com/beyondkmp/electron-builder/pull/30) to this PR to re-add bun support

merged

packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -32,11 +33,23 @@ export function detectPackageManager(dirs: string[]): PM {
...
   }
 
-  return detectPackageManagerByEnv("pnpm") || detectPackageManagerByEnv("yarn") || detectPackageManagerByEnv("npm") || PM.NPM
+  pm = detectPackageManagerByEnv()
+  if (pm) {
+    if (pm === PM.YARN) {
+      return detectYarnBerry()
+    }
+    return pm
+  }
```
mmaietta, line 50: This is duplicate logic of line 34-40. Can you please extract to a closure within (or outside) the function?

packages/app-builder-lib/src/node-module-collector/packageManager.ts
```diff
@@ -70,18 +72,24 @@ export function detectPackageManagerByLockfile(cwd: string): PM | null {
   const yarn = has("yarn.lock")
   const pnpm = has("pnpm-lock.yaml")
   const npm = has("package-lock.json")
+  const bun = has("bun.lock") || has("bun.lockb")
 
   const detected: PM[] = []
   if (yarn) detected.push(PM.YARN)
   if (pnpm) detected.push(PM.PNPM)
   if (npm) detected.push(PM.NPM)
+  if (bun) detected.push(PM.BUN)
```
mmaietta: Please consider prettier settings for mult-line `if` statements. This can all be simplified to the below since `has` is already pretty self-explanatory.
```
const detected: PM[] = []
if (has("yarn.lock")) {
  detected.push(PM.YARN)
}
if (has("pnpm-lock.yaml")) {
  detected.push(PM.PNPM)
}
if (has("package-lock.json")) {
  detected.push(PM.NPM)
}
if (has("bun.lock") || has("bun.lockb")) {
  detected.push(PM.BUN)
}
```

packages/app-builder-lib/src/node-module-collector/index.ts
beyondkmp: Fixed

packages/app-builder-lib/src/node-module-collector/packageManager.ts
beyondkmp: Fixed

mmaietta (approved):

mmaietta:
Nice work! LGTM

PR #9263, opened 2025-08-29
Expanding ${arch} macro properly for mac.binaries in case of native node packages

bimusiek:
We have this issue where we pack `realm.node` that needs signing. 
Our CI either builds universal or arm64 + x64 archs. 

It was not possible to say which file to sign in `mac.binaries`, as it is under `app.asar.unpacked` or `app-x64.asar.unpacked` and then for universal arch it should double the paths as we need to sign both `app-arm64.asar.unpacked` and `app-x64.asar.unpacked`

This fixes the issue, we have tested with:
```
electron-builder --mac zip:universal --universal --publish always --config electron/electron-fym.yml
```
and
```
electron-builder -lm --publish always --config electron/electron-fym.yml
```

with configuration
CleanShot 2025-08-29 at 12 22 04@2x

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -71,9 +71,6 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
   }
   
   expandArch(pattern: string, arch?: Arch | null): string[] {
-    if(!arch) {
-      return [pattern];
-    }
     if(arch === Arch.arm64) {
       return [doExpandMacro(pattern, null, this.appInfo, {}, false)]
     }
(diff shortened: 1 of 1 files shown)


packages/app-builder-lib/src/macPackager.ts
```diff
@@ -68,6 +69,19 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
...
+    if(arch === Arch.arm64) {
+      return [doExpandMacro(pattern, null, this.appInfo, {}, false)]
+    }
+    if(arch === Arch.universal) {
+      return [
+        doExpandMacro(pattern, Arch[Arch.arm64], this.appInfo, {}, false),
+        doExpandMacro(pattern, Arch[Arch.x64], this.appInfo, {}, false),
+      ]
+    }
+    return [doExpandMacro(pattern, null, this.appInfo, {}, false)];
```
mmaietta: Is this correct? It's functionally the same.
```
    if(arch === Arch.arm64) {
      return [doExpandMacro(pattern, null, this.appInfo, {}, false)]
    }
    /// ....universal specific
    return [doExpandMacro(pattern, null, this.appInfo, {}, false)];
```

bimusiek: Yes thats a mistake, for the last return it should pass `arch` as it is `asar-x64.unpacked` for x64 and `asar.unpacked` for arm64

bimusiek: Sorry, I made a mistake. It is using `asar.unpacked` for any other arch. Only universal has it with suffix.
I have fixed the code though and added comments.

bimusiek:
You can test it with:
```
    "@hernas/electron-builder": "26.0.29",
    "@hernas/app-builder-lib": "26.0.29",
```

mmaietta:
This looks solid! Thank you for the contribution!

Quick request. Can you please run the command `pnpm prettier` when you have a chance? A ton of linter warnings are in the PR. Other than that, this LGTM

bimusiek:
Hey @mmaietta, prettier issues are fixed 👍

mmaietta:
Just realized that a `changeset` is missing. Can you please add this file? Either via `pnpm generate-changeset` or just copy pasting the content below

File:
`.changeset/eighty-snails-dress.md` (or any random name within that folder)

Content:
```
---
"app-builder-lib": patch
---

feat: Expanding `${arch}` macro properly for `mac.binaries` in case of native node packages for unpacked files for macos `universal` target
```

bimusiek:
@mmaietta Changeset added.

PR #9282, opened 2025-09-23
feat: Check for Silent Flag in NSIS Uninstaller

naderhen:
Currently, there is no way to suppress the "Are you sure you want to uninstall?" dialog when running the uninstaller exe with the `\S`. It doesn't appear that there are any hooks that would allow me to check for this on my own, so I'm hoping this helps.

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -23,6 +23,7 @@ jobs:
         with:
           # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Install dependencies
         uses: ./.github/actions/pnpm
@@ -35,13 +36,12 @@ jobs:
           creds: ${{ secrets.RELEASE_BOT_APP_CREDS }}
 
       - name: Create versions PR & prepare publish
-        uses: changesets/action@e0145edc7d9d8679003495b11f87bd8ef63c0cba # v1.5.3
+        uses: changesets/action@7af85527bb2f43a404b5823a6cad47661be255c1 # v1.5.3 - @marshallofsound fork
         with:
           version: pnpm ci:version
           commit: 'chore(deploy): Release'
           title: 'chore(deploy): Release'
           publish: pnpm ci:publish
           setupGitUser: false
-          commitMode: github-api
         env:
           GITHUB_TOKEN: ${{ steps.generate-token.outputs.token }}


mmaietta:
FYI, this LGTM, but I need to fix the core `test-updater` test suite before I can merge this. Working on this currently and hope to resolve asap

naderhen:
thanks @mmaietta ! Will keep an eye out for when this goes in

packages/app-builder-lib/templates/nsis/uninstaller.nsh
```diff
@@ -8,6 +8,13 @@ Function un.onInit
   
   !insertmacro check64BitAndSetRegView
 
+  # Parse command line for /S flag and set silent mode
+  ${GetParameters} $R0
```
mmaietta, line 12: So I was thinking on this further in order to avoid this potentially being a "breaking" UX change in the uninstallation flow. I'm looking through the code and am not seeing any `/S` flag being used for an uninstaller (AFAICT?). How are you planning on leveraging this silent flag feature?

naderhen: I'm open to whatever you think would be the right way to add the functionality we're going for here (the ability to have truly silent uninstallations without needing to mess around with `perMachine` etc). My goal here was to just give any potential point in the right direction but feel free to throw it away in favor of a more robust solution.

For our complex use-cases, we require this config to support some of the (admittedly incorrect at times) requests we have from enterprise customers:

```
msiWrapped:
  wrappedInstallerArgs: /S /dept=[DEPT] /env=[ENV]
  perMachine: true
  impersonate: false
nsis:
  oneClick: true
  perMachine: false
  allowToChangeInstallationDirectory: false
  createStartMenuShortcut: false
  createDesktopShortcut: false
  include: build/installer.nsh
  ```

The IT admins at these companies want to be able to run the installer remotely, and currently cannot due to the confirmation dialog. Silent installs work great

Comment (mmaietta) [reply packages/app-builder-lib/templates/nsis/uninstaller.nsh:12]: This use case totally makes sense. So the only way you're applying the `/S` is through `wrappedInstallerArgs` currently? During auto-update for nsis, are you also explicitly setting `quitAndInstall(isSilent = true, isForceRunAfter = true)`

Wondering if an explicit property like `isSilent?: { uninstall?: boolean | null, updater?: boolean | null }` would be most appropriate here to also not change default behavior.

Comment (naderhen) [reply packages/app-builder-lib/templates/nsis/uninstaller.nsh:12]: For whatever reason (probably some requirement i can't recall quickly at this moment), we're not actually calling `quitAndInstall` at all, and relying on the users to be the ones to quit when ready so as not to disrupt their current workflow. Does that make sense?

Comment (naderhen) [reply packages/app-builder-lib/templates/nsis/uninstaller.nsh:12]: Hi @mmaietta , any updates on this from your end? Happy to provide any additional context you may need to get silent uninstallation working in the use cases i mentioned (namely oneClick: false). Thanks!

Comment (mmaietta) [reply packages/app-builder-lib/templates/nsis/uninstaller.nsh:12]: I was thinking further on this and that we could add this to the e2e updater test. Basically, Initial Install v1 -> Autoupdate to v1.0.1 -> Silent uninstall.

What are your thoughts?

https://github.com/electron-userland/electron-builder/blob/b6a34c00c35e42dc279a55d672558ea7badc7fcd/test/src/updater/blackboxUpdateTest.ts#L132-L149

I'm also just concerned that this is a breaking change, no? If `oneClick: false` in the config, then this would inadvertently be changing that behavior? Or are we only parsing it now, and a future feature could be enabling one-click uninstalls?

Comment (naderhen) [reply packages/app-builder-lib/templates/nsis/uninstaller.nsh:12]: I'm not sure if this would constitute a breaking change. Our users want to be able to call the uninstaller exe directly (outside the context of an auto-update) and have it run without the confirmation dialog. If all we're doing is adding handling for a new /S flag (new to the uninstaller specifically), is that actually breaking any existing functionality? 

Autoupdate would be unaffected AFAIK as well as any other behavior of the existing uninstaller.

Would it be helpful for me to test out whatever new uninstaller you create with the appropriate changes? I'm having a little trouble getting the monorepo dev environment setup enough to build with my changes, but if you have an artifact installer and uninstaller I'm happy to validate any workflows you're worried about.

Comment (mmaietta) [reply packages/app-builder-lib/templates/nsis/uninstaller.nsh:12]: Mind if I try pushing to this PR?

I'd like to add this to the e2e updater test. Can just easily call the Uninstaller `.exe` manually with the `/S` flag via `execSync`, correct? I can probably just add a "cleanup"/uninstall step to the updater test logic to validate that part of the flow.

Comment (naderhen) [reply packages/app-builder-lib/templates/nsis/uninstaller.nsh:12]: go for it!

File: test/src/updater/blackboxUpdateTest.ts [hunk 2]
```diff
@@ -300,6 +244,87 @@ async function doBuild(
...
+  } else if (target === "rpm") {
+    execSync(`zypper rm -y testapp`, { stdio: "inherit" })
+  } else if (target === "pacman") {
+    execSync(`pacman -R --noconfirm testapp`, { stdio: "inherit" })
+  } else if (process.platform === "win32") {
+    // access installed app's location
+    const localProgramsPath = path.join(process.env.LOCALAPPDATA || path.join(homedir(), "AppData", "Local"), "Programs", "TestApp")
+    const uninstaller = path.join(localProgramsPath, "Uninstall TestApp.exe")
+    console.log("Uninstalling", uninstaller)
+    execFileSync(uninstaller, ["/S", "/C", "exit"], { stdio: "inherit" })
```
mmaietta, line 321: @naderhen looks like the flag works correctly in unit tests! Able to clean up the previous installation with silent mode arg respected.

naderhen: amazing! thanks for your help @mmaietta . Do you typically create actual installers/uninstallers to test manually or are the tests pretty conclusive?

mmaietta: Yes, I ran the test suite also in my Windows arm64 VM and the app is cleaned after the test without any prompt. It does take like 30seconds to update in the Settings > Apps > Installed Apps list, but I can't tell if that's due to the uninstallation process still running post-test completion OR if it is just taking a little for Windows to update its registry (cache?).

naderhen: sounds great, thanks for all of your help and the quick attention on this

naderhen:
@mmaietta apologies for the re-request, github said I needed a new approval after updating the branch against main

Issue #9275, opened 2025-09-12
Problem from outdated node-gyp


crystalfp:
Seems electron-builder, through `@electron/rebuild` installs `node-gyp` version v10.2.0-electron.1 and this version fails to build native node modules with error
`/LTCG:INCREMENTAL: no such file or directory`. See https://github.com/electron/electron/issues/47613 for the complete problem.  No problems with the globally installed latest version of node-gyp (11.4.2).

Seems the solution is to update the `@electron/rebuild` package from the current version 3.7.0 to 4.0.1. This package is a dependency of `app-builder-lib` used by electron-builder.

Could you update these dependencies so to update at the end the node-gyp package? Thanks!

electron-builder 26.0.12 on Windows 11

mmaietta:
Can definitely revisit upgrading `@electron/rebuild`, the downside is that `rebuild` is an ESM-only package in v4 IIRC. So that means electron-builder needs to perform a backward-incompatible migration to ESM as well to leverage it.

It seems this may be inevitable; I was trying to close out as many bugs as possible before that breaking change though

rotu:
#9277 seems like it will be a blocker.

> So that means electron-builder needs to perform a backward-incompatible migration to ESM as well to leverage it.

I don't think this is true but I'm not sure.

100cm:
so we can't upgrade electron version 38 for now ?

rotu:
@mmaietta to my knowledge, cjs can `import()` esm packages and can even [`require` synchronous esm packages](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require). Why would `electron-builder` need to migrate to esm `@electron/rebuild`?

Edit:

Even their readme demonstrates using `@electron/rebuild` from cjs: https://github.com/electron/rebuild/blob/f254b99326c789007e5360e7eaca1532435e7e86/README.md?plain=1#L167-L179

mmaietta:
I stand corrected them, thanks for the pointer! I'll add this to my todo list

Quick note: I still have to eventually migrate to ESM though as a non-backward compatible change in order to align with the electron organization. (I'm working with electron working group and electron-builder is eventually to be migrated to `electron/` org where I'll continue supporting it)

mmaietta:
Update: Still stuck getting it working. The current electron/rebuild approach is similar to that of electron-packager, where we spawn the rebuild step in a separate process to contain the console log/noise.


[...]

Issue #9277, opened 2025-09-15
app-builder-lib references `@electron/rebuild` internal modules


rotu:
If I override @electron/rebuild to 4.0.0 and attempt to import `electron-builder`, I get an error like the following:

```
`Error: Package subpath './lib/search-module' is not defined by "exports" in D:\\SonarView\\node_modules\\@electron\\rebuild\\package.json
    at exportsNotFound (node:internal/modules/esm/resolve:313:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:660:9)
    at resolveExports (node:internal/modules/cjs/loader:650:36)
    at Function._findPath (node:internal/modules/cjs/loader:717:31)
    at Function._resolveFilename (node:internal/modules/cjs/loader:1369:27)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
    at Function._load (node:internal/modules/cjs/loader:1192:37)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:237:24)`

Package subpath './lib/search-module' is not defined by "exports" in D:\\SonarView\\node_modules\\@electron\\rebuild\\package.json`
```

This is due to a naughty import at:

https://github.com/electron-userland/electron-builder/blob/38c2085e9a344812d32611d197d1df66d6644b45/packages/app-builder-lib/src/electron/electronVersion.ts#L1

A similarly naughty import exists here too:

https://github.com/electron-userland/electron-builder/blob/38c2085e9a344812d32611d197d1df66d6644b45/packages/app-builder-lib/src/util/yarn.ts#L2

mmaietta:
`electron/rebuild` 4.0.0 is ESM only and not backward compatible. Overriding the version is causing your issue.

rotu:
> `electron/rebuild` 4.0.0 is ESM only and not backward compatible. Overriding the version is causing your issue.

Yes, I recognize that this is unsupported. Nevertheless, `app-builder-lib` should not rely on internal, unexposed features of `@electron/rebuild` that were never intended as part of its public API.

Issue #9143, opened 2025-06-04
Error while running `electron-builder install-app-deps` with `@electron/rebuild@4.0.1` with Two package.json Structure


tjcouch-sil:
- `electron-builder` version: 26.0.12 (though it does not seem to be exclusive to this version)
- `electron` version: 35.0.2 (though it does not seem to be exclusive to this version)
- `@electron/rebuild`: [4.0.1](https://github.com/electron/rebuild/releases/tag/v4.0.1) (though I think [4.0.0](https://github.com/electron/rebuild/releases/tag/v4.0.0) also has the problem. It seems 3.7.2 does not have this problem)
- Platform: Windows

Hello `electron-builder` maintainers!

When I run `npx electron-builder install-app-deps` with `@electron/rebuild@4.0.1`, I am encountering an error in which `@electron/rebuild` seems to be unable to determine the Electron version. Also, in some cases (I can't find particular rhyme or reason as to when or why it flips between these two errors), `node-gyp` seems to be unable to find a Visual Studio installation. I am using the [Two package.json Structure](https://www.electron.build/tutorials/two-package-structure.html). I can reproduce this issue easily using [Electron React Boilerplate](https://github.com/electron-react-boilerplate/electron-react-boilerplate) with some slight modifications.

This issue does not occur if I run `@electron/rebuild` directly with `electron-rebuild --parallel --types prod,dev,optional --module-dir release/app`.

To reproduce:

Either run these manual steps to reproduce on your own (which gives me the "Cannot find Electron" error below):

1. `git clone https://github.com/electron-react-boilerplate/electron-react-boilerplate` (as of when I am testing this, `main` is on commit `0c83f5faa4ba503bdfa6f75e3bc273560f938ccd`)
2. `cd electron-react-boilerplate`
3. `npm i`
4. `npm i -D @electron/rebuild@4.0.1 electron-builder@latest` (Installing `electron-builder` latest because I was sometimes getting errors that `node-abi` didn't know about the Electron version)
5. `cd release/app`
6. `npm i sqlite3`
7. `cd ../..`
8. `npx electron-builder install-app-deps` (Produces "Cannot find Electron" error below)
9. `npm i -D @electron/rebuild@3.7.1` (the version it was on before changing it)
10. `npx electron-builder install-app-deps`

Or clone my repo and reproduce:

1. `git clone https://github.com/tjcouch-sil/electron-react-boilerplate.git`
[...]

PR #9285, opened 2025-09-25
chore(deps): update electron/rebuild to resolve node-gyp dependency issue

mmaietta:
Resolves: #9275 #9277 #9143


mmaietta:
Update: This still doesn't work locally, scope is continuing to grow on this package upgrade (ESM migration may be necessary)

PR #9287, opened 2025-09-26
fix: two failed linux updater tests

daihere1993:
```
warning: cannot resolve "http-parser", a dependency of "testapp"
error: failed to prepare transaction (could not satisfy dependencies)

FAIL   node  test/src/updater/blackboxUpdateTest.ts > Electron autoupdate (fresh install & update) > linux > arch - (pacman)
FAIL   node  test/src/updater/linuxUpdaterTest.ts > test arch download and install (pacman)
```

Root cause: [Arch Linux removed the http-parser package](https://archlinux.org/todo/move-to-llhttp-from-http-parser/), but our Pacman metadata still declares it as a required dependency. Reinstall attempts during the Arch updater tests therefore fail with “unable to satisfy dependency ‘http-parser’”, causing both the blackbox pacman autoupdate test and the Linux pacman updater unit test to error out. 

Fix: I can not find our project depends on `http-parser`, so the easiest way to fix is just remove it.

@mmaietta if we do depend on `http-parser`, then we may need switch to `llhttp` instead.

### DIFF
--- a/packages/app-builder-lib/src/targets/FpmTarget.ts
+++ b/packages/app-builder-lib/src/targets/FpmTarget.ts
@@ -359,7 +359,7 @@ export default class FpmTarget extends Target {
         ]
 
       case "pacman":
-        return ["c-ares", "ffmpeg", "gtk3", "http-parser", "libevent", "libvpx", "libxslt", "libxss", "minizip", "nss", "re2", "snappy", "libnotify", "libappindicator-gtk3"]
+        return ["c-ares", "ffmpeg", "gtk3", "libevent", "libvpx", "libxslt", "libxss", "minizip", "nss", "re2", "snappy", "libnotify", "libappindicator-gtk3"]
 
       default:
         return []
--- a/packages/app-builder-lib/scheme.json
+++ b/packages/app-builder-lib/scheme.json
@@ -2290,7 +2290,7 @@
               "type": "null"
             }
           ],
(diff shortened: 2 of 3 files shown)


packages/app-builder-lib/src/targets/FpmTarget.ts
```diff
@@ -359,7 +359,7 @@ export default class FpmTarget extends Target {
         ]
 
       case "pacman":
-        return ["c-ares", "ffmpeg", "gtk3", "http-parser", "libevent", "libvpx", "libxslt", "libxss", "minizip", "nss", "re2", "snappy", "libnotify", "libappindicator-gtk3"]
+        return ["c-ares", "ffmpeg", "gtk3", "libevent", "libvpx", "libxslt", "libxss", "minizip", "nss", "re2", "snappy", "libnotify", "libappindicator-gtk3"]
```
mmaietta, line 362: This seems like a breaking change since this impacts end-user application packaging. Is there a drop-in replacement for `http-parser`?

daihere1993: I checked the history and found that these dependencies were introduced in [this commit](https://github.com/develar/app-builder/commit/f27e219314824e281719c0ba243358f00d9e0849) on June 26, 2019. At that time, `http-parser` had already been deprecated in Node.js v12 (released on April 23, 2019).

It seems we added these dependencies because some packages on Arch-based systems still relied on `http-parser` while Node.js was just beginning its transition to `llhttp`. As a workaround, we bundled additional dependencies in electron-builder. It's just my guess as there no enough description for above commit.

From my perspective, removing this package now is not a breaking change. The latest Arch Linux image no longer supports http-parser, and since Node.js has fully supported `llhttp` for many years, no drop-in replacement is required.

mmaietta:
On the premise of not changing production behavior/logic in order to fix an issue with a test case, I took a different route in https://github.com/electron-userland/electron-builder/commit/53415d3494aad997c9119e22771dd65025e27060 to resolve the test failures.

Closing this PR since `master` is building successfully

PR #9292, opened 2025-10-01
chore: remove stale issues workflow

esimkowitz:
The stale issues workflow causes a lot of pain for contributors and community members. It makes members feel unseen when issues are marked stale, simply because of inactivity. With a bursty project like this one, it is not uncommon for valid issues to go more than 30 days without a status update.


mmaietta:
I understand your reasoning, but devil's advocate:
- I'm the **only** consistent maintainer on this project, the original owner departed this project over 6 years ago. I'm the only member of Electron WG that owns this project.
- Many GH issues I have replied to and then they're never responded to by the OP user - the only way to clean this up automatically is a stale bot.
- For GH reports that don't get attention, they're either blocked by no minimum repro repo provided, are a feature request I can't dedicate time to (nor an ETA in terms of triaging), or (imo) quite simply could be debugged/resolved by the OP leveraging Search functionality or ChatGPT(or other AI tools).
- Community contributions are always welcome. Continually commenting "not stale" or "bump" does nothing on non-critical requests.

This all being said, I am open and willing to extend the time to 60 days, but I'm not planning on removing this bot entirely.

esimkowitz:
@mmaietta I totally get you're frustration with stale issues, I think it's something every repo has to deal with at a certain scale and there's no good solution. Maybe AI will help with triage at some point but for now I get why you want to keep some freshness to issues. I think extending the time to 60 days would help weed out the true stale issues from the temporarily inactive. Thank you for your commitment to this project!

PR #9293, opened 2025-10-02
Fixed the call to “download-artifact” when ELECTRON_BUILDER_CACHE is …

drivron:
There is an issue with the PR #9151 on the call to `doGetBin` in `binDownload.ts`. This function is now called with `cacheName`, which produces an invalid value when `ELECTRON_BUILDER_CACHE` is set.
For example, if `ELECTRON_BUILDER_CACHE="~cache"`, the following command is called and fails: `app-builder.exe download-artifact --name ~cachewinCodeSign`; and the following directory is created: `~cache/~cachewinCodeSign/590582919`.

### DIFF
--- a/packages/app-builder-lib/src/binDownload.ts
+++ b/packages/app-builder-lib/src/binDownload.ts
@@ -51,7 +51,7 @@ export function getBin(cacheKey: string, url?: string | null, checksum?: string 
     return promise
   }
 
-  promise = doGetBin(cacheName, url, checksum)
+  promise = doGetBin(cacheKey, url, checksum)
   versionToPromise.set(cacheName, promise)
   return promise
 }


packages/app-builder-lib/src/binDownload.ts
```diff
@@ -51,7 +51,7 @@ export function getBin(cacheKey: string, url?: string | null, checksum?: string
     return promise
   }
 
-  promise = doGetBin(cacheName, url, checksum)
+  promise = doGetBin(cacheKey, url, checksum)
```
mmaietta, line 54: I don't think this will work as intended actually.

If the flow is as following:
1) `ELECTRON_BUILDER_CACHE` is set during first build
2) rebuild another app instance without `ELECTRON_BUILDER_CACHE` set

The "bin"/toolset used in step 2 would be the same custom toolset that was configured in step 1 due to cache key reuse.

Maybe we can use `${cacheKey}${process.env.ELECTRON_BUILDER_CACHE ?? ""}` instead?

drivron: The cache managed by versionToPromise is only in memory and is therefore not reused by other instances.
I don't know exactly what its purpose is. I suppose it is used to manage parallelism of app-builder calls from the same instance.

mmaietta: > suppose it is used to manage parallelism of app-builder calls from the same instance.

This is correct.

> The cache managed by versionToPromise is only in memory and is therefore not reused by other instances.

This is incorrect. Internally, `doGetBin` calls npm upstream dep `app-builder`, which is a Go library that does actually persist the download to disk: https://github.com/develar/app-builder/blob/7004925f95d8f034fc88d7e782c9aa7583debb8e/pkg/download/artifactDownloader.go#L63
This is to reuse the tool across builds.

drivron: This persistent disk cache is entirely independent of the call promises cache. In fact, I do not understand why this promises cache key is bound to ELECTRON_BUILDER_CACHE.

mmaietta: > I do not understand why this promises cache key is bound to ELECTRON_BUILDER_CACHE.

I don't understand this. It's a reliable key to use in a map that distinguishes a custom tooling cache from the default cache.

I see no reason why we aren't taking this approach:
> we can use `const cacheName = ${cacheKey}${process.env.ELECTRON_BUILDER_CACHE ?? ""}` instead

Julusian:
I have just discovered this exact issue and came up with the same fix.
It was resulting in paths too long for windows to handle:
```
Error: C:\actions-runner-bitfocus\_work\companion-satellite\companion-satellite\.cache\electron-builder\Cactions\Cactions-runner-bitfocus_workcompanion-satellitecompanion-satellite.cacheelectron-buildernsis-3.0.4.1-nsis-3.0.4.1\Bin\makensis.exe process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
```

packages/app-builder-lib/src/binDownload.ts
Julusian: > Maybe we can use ${cacheKey}${process.env.ELECTRON_BUILDER_CACHE ?? ""} instead?

I suspect that wouldn't solve my issue, because what `app-builder-bin` is doing is forming a path of `${env.ELECTRON_BUILDER_CACHE}/${shortname}/${cacheName}`, where `shortname` is `cacheName` up to the first `-`.

The full path this is forming for me is `C:\actions-runner-bitfocus\_work\companion-satellite\companion-satellite\.cache\electron-builder\Cactions\Cactions-runner-bitfocus_workcompanion-satellitecompanion-satellite.cacheelectron-buildernsis-3.0.4.1-nsis-3.0.4.1\Bin\makensis.exe`, and it is then failing because the path is too long.  
This is with `ELECTRON_BUILDER_CACHE=C:\actions-runner-bitfocus\_work\companion-satellite\companion-satellite\.cache\electron-builder`

Notice that the full path is in here essentially twice, as the `Cactions-runner-bitfocus_workcompanion-satellitecompanion-satellite.cacheelectron-buildernsis-3.0.4.1-nsis-3.0.4.1` folder is taking the value of `cacheName` (the value this pr proposes changing), which is unnecessary because it is already inside of `ELECTRON_BUILDER_CACHE`.

mmaietta: What if we just did `path.basename(process.env.ELECTRON_BUILDER_CACHE ?? "")${cacheKey}`

I'm pretty sure that the custom cache path needs to be somehow leveraged to differentiate from the on-demand toolset versus the ELECTRON_BUILDER_CACHE toolset.

Julusian: Yeah that will probably work, but in the best case all it will do is get back the value of `cacheKey`, so why bother with the extra steps?
I am only 90% sure that will behave correctly in my case, as remember that `app-builder-bin` is ending up with a folder called `Cactions-runner-bitfocus_workcompanion-satellitecompanion-satellite.cacheelectron-buildernsis-3.0.4.1-nsis-3.0.4.1`. I expect that is a problem with escaping in the shell command, but am not 100% certain. 

To clarify, I'm advocating for this PR as is.  
I think what you are saying about the promise cache makes sense, and is a reason to keep the current `cacheName` behaviour for that. The call to `getBin` is the only thing that needs changing.

PR #9300, opened 2025-10-04
chore: updating pnpm and setting minimumReleaseAge to 1 week

mmaietta:


package.json
```diff
@@ -74,9 +76,10 @@
   },
   "engines": {
     "node": ">=14.14",
-    "pnpm": ">=9"
+    "pnpm": ">=10"
   },
-  "packageManager": "pnpm@9.4.0",
+  "minimumReleaseAge": 10080,
+  "packageManager": "pnpm@10.18.0",
```
MarshallOfSound: Can we run `corepack up` here to lock this to a SHA not just a version number?

MarshallOfSound (approved):
Left one nit RE locking to a SHA via corepack

package.json
mmaietta: Definitely!

package.json
```diff
@@ -74,9 +76,10 @@
   },
   "engines": {
     "node": ">=14.14",
-    "pnpm": ">=9"
+    "pnpm": ">=10"
   },
-  "packageManager": "pnpm@9.4.0",
+  "minimumReleaseAge": 10080,
```
MarshallOfSound: Doesn't this need to be in https://pnpm.io/settings ?

mmaietta: Yeup, thank you. Pushed the fix

PR #9301, opened 2025-10-04
chore(deps): updating and pinning semver to 7.7.2

mmaietta:


panther7:
@mmaietta why pinning is needed?

#9349

PR #9302, opened 2025-10-04
chore(deploy): migrate to `electron/npm-trusted-auth-action`

mmaietta:
Migrate to https://github.com/electron/npm-trusted-auth-action for OIDC token conversion to npm publish token

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -33,7 +33,7 @@ jobs:
       # Auto-setup of NPM_TOKEN env var for for changesets/action
       - uses: electron/npm-trusted-auth-action@v1
         with:
-          package-name: 'electron-userland/electron-builder'
+          package-name: 'electron-builder'
 
       - name: Create versions PR & prepare publish
         id: changesets
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -8,7 +8,9 @@ on:
 env:
   HUSKY: 0 # Bypass husky commit hook for CI
 
-permissions: {}
+permissions:
+  id-token: write  # Required for OIDC token
+
 jobs:
   pr-release:
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-release.yml
```diff
@@ -24,22 +24,28 @@ jobs:
...
+      - name: Install dependencies
         uses: ./.github/actions/pnpm
 
-      - name: Set up NPM credentials
-        run: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
-        env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # Auto-setup of NPM_TOKEN env var for for changesets/action
+      - name: Authenticate to npm registry
+        uses: electron/npm-trusted-auth-action@v1
```
MarshallOfSound: Pin to SHA

.github/workflows/pr-release.yml
```diff
@@ -24,22 +24,28 @@ jobs:
...
+        with:
+          package-name: 'electron-builder'
```
MarshallOfSound: Not 100% sure about package structure, is this really the only package that gets published here. I thought this was a monorepo that had a few packages in it?

.github/workflows/pr-release.yml
```diff
@@ -24,22 +24,28 @@ jobs:
...
+
+      - name: Generate GitHub App token
+        uses: electron/github-app-auth-action@v1.1.1
```
MarshallOfSound: Pin to SHA

.github/workflows/pr-release.yml
mmaietta: Hmmm, I need a single token that is able to publish the monorepo topographically via `pnpm publish -r`. Is there a way to provide a single token for a list of `package-names`?

mmaietta: Discussed via Slack to take an alternative approach of updating `npm` to 11.6.2 support automatic OIDC exchange during publishing

PR #9313, opened 2025-10-15
feat: bun package manager support for electron-builder

johnpyp:
### Additions
- support for bun isolated mode / bun workspaces proper resolution and reification to the single `node_modules` directory in the ASAR
  - works with both `--linker=isolated` and `--linker=hoisted` modes
- Improved `detectPackageManager` by adding the Git Root + "nearest parent lockfile" to the search directories for a lockfile 
  - this was previously not detecting bun workspaces where `electron-builder` was in `packages/foo` and the `bun.lock` only exists in the root
-  add a `test-bun-workspace` fixture
- add a `bun install --linker=isolated` test and `bun install --linker=hoisted` test case
  - additionally, there's a test for conflicting package versions (`lib` depends on one version of a pakage and `app` depends on another), which bun isolated mode disambguates. the `bunNodeModulesCollector` correctly handles this case and nests `node_modules` appropriately for it.
- add a few deps to `onlyBuiltDependencies`, which seems to have been missed in the pnpm v9 -> v10 upgrade.

### Notes

This PR adds support for [bun package manger](https://bun.com/docs/cli/pm) during electron `node_modules` packaging into an ASAR, particularly in the context of workspaces and bun's isolated install mode (like pnpm's default `node_modules/.pnpm` folder + symlinks, `isolated` mode in bun uses `node_modules/.bin` + symlinks).

I've confirmed on my real electron application with quite a few dependencies that this all seems to work as expected, and the new fixtures confirm that as well.

Something that seems like it could be a good idea is leveraging https://github.com/antfu-collective/package-manager-detector instead of the manual package management detection that's done currently, because there's a lot more techniques to judge it, and that library encapsulates the logic quite nicely into a configurable call. E.g I noticed that even though `bun.lock` was added in one place previously, it was missing in 2 others, leading to difficult to troubleshoot inconsistencies. If we leveraged `package-manager-detection` everywhere, that opportunity for inconsistency should hopefully vanish.

### DIFF
--- a/test/src/updater/Dockerfile-appimage
+++ b/test/src/updater/Dockerfile-appimage
@@ -45,4 +45,6 @@ RUN ln -s /usr/bin/fuse3 /usr/bin/fuse || true
 
 RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
 RUN apt-get install -y nodejs
-RUN npm --silent install --global --depth 0 pnpm
+RUN npm --silent install --global --depth 0 pnpm corepack \
+ && corepack enable
+
--- a/test/src/updater/Dockerfile-archlinux
+++ b/test/src/updater/Dockerfile-archlinux
@@ -37,6 +37,7 @@ RUN git clone https://aur.archlinux.org/http-parser.git \
 USER root
 WORKDIR /project
 
-RUN npm --silent install --global --depth 0 pnpm
+RUN npm --silent install --global --depth 0 pnpm corepack \
+ && corepack enable
 
 WORKDIR /project
--- a/test/src/updater/Dockerfile-debian
+++ b/test/src/updater/Dockerfile-debian
@@ -8,6 +8,8 @@ RUN apt-get update \
       xdg-utils libatspi2.0-0 libsecret-1-0 libgbm1 \
  && rm -rf /var/lib/apt/lists/*
 
- RUN npm --silent install --global --depth 0 pnpm
+ RUN npm --silent install --global --depth 0 pnpm corepack \
+ && corepack enable
+
 
  WORKDIR /project
--- a/test/src/updater/Dockerfile-rpm
+++ b/test/src/updater/Dockerfile-rpm
@@ -20,6 +20,7 @@ RUN dnf install -y \
 # Optional: for better font rendering or Chromium deps
 RUN dnf install -y liberation-fonts fontconfig
 
-RUN npm --silent install --global --depth 0 pnpm corepack
+RUN npm --silent install --global --depth 0 pnpm corepack \
+ && corepack enable
 
 WORKDIR /project


mmaietta:
This is awesome! Kicking off CI. Quick Q on this 👇 . What is it doing?
```
    "onlyBuiltDependencies": [
      "@parcel/watcher",
      "electron",
      "electron-winstaller",
      "esbuild"
    ]
    ```

**Comment by johnpyp** [comment]
Yeah, that's part of a breaking change in pnpm v10 (details here: https://github.com/orgs/pnpm/discussions/8945) where you're required to specify which dependencies which have `postinstall` scripts are allowed to run (for security).

It's unfortunately easy to miss it when initially migrating, because if you start from a pnpm v9 installation it won't always print out the error message... weird consistency bug in pnpm I think.

If you do `pnpm dlx npkill` to wipe out your `node_modules`, then I believe you'll see the message when you run `pnpm install` again.
image

File: packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts [hunk 1]
```diff
@@ -0,0 +1,175 @@
...
+      optionalDependencies: Object.keys(childMaps.optionalDependencies).length > 0 ? childMaps.optionalDependencies : undefined,
+    }
+
+    this.dependencyCacheByPath.set(installedPath, dependency)
+    return dependency
+  }
+
+  private findInstalledDependency(basedir: string, dependencyName: string): string | null {
+    try {
+      const packageJsonPath = resolve.sync(`${dependencyName}${path.sep}package.json`, {
```
mmaietta: Why not use `require.resolve` here?

Also, instead of manually constructing with `path.sep`, can we just use `path.join` here?

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
```diff
@@ -0,0 +1,175 @@
...
+        basedir,
+        preserveSymlinks: false,
+        includeCoreModules: false,
+      })
+      return path.dirname(packageJsonPath)
+    } catch (e: any) {
+      if (e && e.code === "MODULE_NOT_FOUND") {
```
mmaietta: Nit: This can be shortened using optionals `e?.code`

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    for (const alias of Object.keys(manifestOptionalDependencies)) {
+      loadPromises.push(
+        this.loadDependency(alias, requesterDir, true).then(dep => {
+          if (dep) optionalDependencies[alias] = dep
+        })
+      )
+    }
+
+    await Promise.all(loadPromises)
```
mmaietta: Is this serving a purpose to load all promises at once? Asking as we can approach this in two methods, already-utilized functionality in electron-builder

Use `asyncTaskPool` npm package for parallel processing, leveraging
```
import asyncPool from "tiny-async-pool"
...
 await asyncPool(MAX_FILE_REQUESTS, loadPromises, async loader => {
```
Or `AsyncTaskManager` class. In this case, we'd probably want it as a `readonly` property within the class to reuse it across node module collection methods.
```
readonly taskManager = new AsyncTaskManager(new CancellationToken())
...
// in the for-loop
taskManager.addTask(this.loadDependency(alias, requesterDir, true).then(dep => {
  if (dep) {
    optionalDependencies[alias] = dep
  }
}))
if (taskManager.tasks.length > MAX_FILE_REQUESTS) {
  await taskManager.awaitTasks()
}

// after for-loop
this.taskManager.awaitTasks()
```

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    const dependencies: string[] = []
+
+    const processDeps = [
+      { entries: tree.dependencies, manifest: tree.manifestDependencies },
+      { entries: tree.optionalDependencies, manifest: tree.manifestOptionalDependencies },
+    ]
+
+    for (const { entries, manifest } of processDeps) {
+      if (!entries) continue
```
mmaietta: Nit: Please use multi-line for all bracketable statements (such as `if`s). Please double check all other bracket statements in this PR so that I don't need to comment on each one :)
```
if (!entries) {
   continue
}
```

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
```diff
@@ -0,0 +1,175 @@
+import { log } from "builder-util"
+import * as path from "path"
+import * as resolve from "resolve"
+import { NodeModulesCollector } from "./nodeModulesCollector"
+import { PM } from "./packageManager"
+import { Dependency } from "./types"
+
+interface BunDependency extends Dependency<BunDependency, BunDependency> {
```
mmaietta: I believe this should be moved to the types.ts file for node-module-collector since the other types are located there.

Might be worth splitting this into a type too:
```
// in types.ts file

type BunManifest = {
  manifestDependencies: Record<string, string>
  manifestOptionalDependencies: Record<string, string>
}


interface BunDependency extends Dependency<BunDependency, BunDependency>, BunManifest {}
```

Leveraged in this function as both parameter and return value. This is NEEDED since `manifestDependencies` and `manifestOptionalDependencies` are the same types, which could lead to a future misuse/bug, utilizing a type is much more safe/future-proof.
```
private async resolveChildren(
    requesterDir: string,
    manifest: BunManifest
  ): Promise<BunManifest>
```

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
johnpyp: #### require.resolve
`require.resolve` doesn't resolve _from a given path_. You can half-way fix this by using `module.createRequire`, however that also requires a specific file path to resolve from. `resolve` is a cleaner way to do this imo (and many libraries agree) because it allows you to specify just a given directory to start from, plus has additional functionality.

The `resolve` library has more flexibility with regards to analyzing its results, e.g you can configure it to not follow symlinks (which would be very useful for implementing support for Yarn PnP with this same strategy, for example). It has more features too like ignoring resolutions of builtins. Comparatively, `require.resolve` just returns a string for builtins, no real distinction, requiring more special handling.

#### path.join
Yep, I'll fix.

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
johnpyp: Removed the `Promise.all` in general, as I tested it on my codebase and there's basically no performance gain or anything compared to just doing it serially. resolution is super cheap

packages/app-builder-lib/package.json
```diff
@@ -76,6 +76,7 @@
     "minimatch": "^10.0.3",
     "plist": "3.1.0",
     "resedit": "^1.7.0",
+    "resolve": "^1.22.10",
```
mmaietta: Why do we need a browserify's `resolve` module? From their repo:
> implements the [node require.resolve() algorithm](https://nodejs.org/api/modules.html#modules_all_together) such that you can require.resolve() on behalf of a file asynchronously and synchronously

Is there any benefit here to just using node's `require.resolve`?

johnpyp: node's `require` is less flexible and requires some more code... but I changed it to use it anyways. It seems like it works well enough for this usecase.

packages/app-builder-lib/package.json
```diff
@@ -108,6 +108,7 @@
     "@types/hosted-git-info": "3.0.2",
     "@types/js-yaml": "4.0.3",
     "@types/plist": "3.0.5",
+    "@types/resolve": "^1.20.6",
```
mmaietta, line 111: Is this still needed?

package.json
```diff
@@ -53,7 +53,7 @@
     "@types/node": "^22.7.4",
     "@typescript-eslint/eslint-plugin": "8.17.0",
     "@typescript-eslint/parser": "8.17.0",
-    "@vitest/ui": "3.0.4",
+    "@vitest/ui": "^3.2.2",
```
mmaietta, line 56: Can we revert this change please? Currently trying to freeze/separate dependency version changes to exclusively their own PRs unless otherwise required.

This should allow us to revert all changes to the `pnpm-lock.yaml`

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
```diff
@@ -0,0 +1,167 @@
+import { log } from "builder-util"
+import * as path from "path"
+import { NodeModulesCollector } from "./nodeModulesCollector"
+import { PM } from "./packageManager"
+import { BunDependency, BunManifest, Dependencies } from "./types"
+import { createRequire } from "module"
```
mmaietta, line 6: I'm not familiar with this `createRequire`, but 2 quick Qs. What are we using it for here as opposed to directly `require(...)`?
Can we also change the import to `from "node:module"`?

packages/app-builder-lib/src/util/appFileCopier.ts
```diff
@@ -176,12 +177,42 @@ function validateFileSet(fileSet: ResolvedFileSet): ResolvedFileSet {
...
+async function getGitRootDir(baseDir: string): Promise<string | null> {
+  try {
+    const { promisify } = require("util")
+    const { exec } = require("child_process")
+    const execAsync = promisify(exec)
+
+    const { stdout } = await execAsync("git rev-parse --show-toplevel", {
+      cwd: baseDir,
+      timeout: 5000,
+    })
```
mmaietta, line 189: Will this be called numerous times? If so, we could probably use the package `lazy-val` that is already used in this project for `Lazy<string>`

mmaietta:
I can pull this feature into my larger refactor/bug-fixes of node module collector code.

mmaietta:
Migrated code to be compatible with refactored node_module_collector code
New PR: https://github.com/electron-userland/electron-builder/pull/9357

PR #9324, opened 2025-10-31
fix: support array of publisher names in Azure signing config

lidel:
This PR:

- updates `WindowsAzureSigningConfiguration` to accept publisherName as string | Array<string> | null
- aligns JSON schema to match TypeScript type definition
- allows specifying multiple publisher names for certificate flexibility across certificate changes
- no functional changes, implementation already handles arrays via `asArray()` helper and I've confirmed that in https://github.com/ipfs/ipfs-desktop/pull/3013 where we use this patch to fix our windows release

this change makes the schema match the existing documentation which states "Several names can be provided"

### DIFF
--- a/.changeset/old-days-arrive.md
+++ b/.changeset/old-days-arrive.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+fix: support array of publisher names in Azure signing config


mmaietta:
Looks like there's some compiler errors from the logs:
```
Error: packages/app-builder-lib/src/codeSign/windowsSignAzureManager.ts(99,5): error TS2322: Type 'Promise<string | string[] | null>' is not assignable to type 'Promise<string>'.
  Type 'string | string[] | null' is not assignable to type 'string'.
    Type 'null' is not assignable to type 'string'.
Error: packages/app-builder-lib/src/options/winOptions.ts(136,12): error TS2411: Property 'publisherName' of type 'string | string[] | null' is not assignable to 'string' index type 'string | Nullish'.
```

PR #9064, opened 2025-04-27
fix: update regex for multipart content-type parsing in multipleRange

beyondkmp:
The `Content-Type` response is as follows: `Content-Type: multipart/byteranges; boundary=799ddd5a-943e-4e22-981d-e32596ca39d2`. The `boundary` can have a space before it or no space at all. 
Here's the rfc:
https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1
image

**Issue**

Currently, the check only accounts for cases where there is a space. If some servers return a response without a space, it will cause incremental downloads to fail.

**Error Info**
```
17:13:29.256 > Full: 92,770.98 KB, To download: 19,742.1 KB (21%)
17:13:29.633 > Cannot download differentially, fallback to full download: Error: Content-Type "multipart/byteranges" is expected, but got "multipart/byteranges;boundary=e888c103-165d-4ec1-9b6f-eca722ac9b10"
    at ClientRequest.<anonymous> (C:\Users\payne\AppData\Local\Programs\Cherry Studio\resources\app.asar\node_modules\electron-updater\out\differentialDownloader\multipleRangeDownloader.js:80:20)
    at ClientRequest.emit (node:events:531:35)
    at SimpleURLLoaderWrapper.<anonymous> (node:electron/js2c/browser_init:2:114720)
    at SimpleURLLoaderWrapper.emit (node:events:519:28)
17:13:34.288 > New version 1.2.9 has been downloaded to C:\Users\payne\AppData\Local\cherrystudio-updater\pending\Cherry-Studio-1.2.9-x64-setup.exe
```

**How to fix**

```
Welcome to Node.js v22.14.0.
Type ".help" for more information.
> /^multipart\/.+?\s*;\s*boundary=(?:"([^"]+)"|([^\s";]+))\s*$/i.exec("multipart/byteranges;boundary=799ddd5a-943e-4e22-981d-e32596ca39d2")
[
  'multipart/byteranges;boundary=799ddd5a-943e-4e22-981d-e32596ca39d2',
  undefined,
  '799ddd5a-943e-4e22-981d-e32596ca39d2',
  index: 0,
  input: 'multipart/byteranges;boundary=799ddd5a-943e-4e22-981d-e32596ca39d2',
  groups: undefined
]
> /^multipart\/.+?\s*;\s*boundary=(?:"([^"]+)"|([^\s";]+))\s*$/i.exec("multipart/byteranges; boundary=799ddd5a-943e-4e22-981d-e32596ca39d2")
[
[...]

PR #9327, opened 2025-11-01
fix(electron-updater): relax multipart boundary regex to allow no space

fleabagneverdie:
### Background

Currently, `electron-updater` uses a regex to detect multipart responses from the server when performing segmented downloads. The regex requires a space before the `boundary` parameter:

```regex
/^multipart\/.+?\s*;\s*boundary=(?:"([^"]+)"|([^\s";]+))\s*$/i
```

According to the HTTP standard, the space before boundary is optional. Servers that omit the space will fail detection, causing the updater to incorrectly treat the response as non-multipart, which can disable segmented downloads.

What this PR does

Updates the regex to allow the boundary parameter to appear immediately after the semicolon, removing the requirement for a space.

### DIFF
--- a/.changeset/lovely-queens-marry.md
+++ b/.changeset/lovely-queens-marry.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix(electron-updater): relax multipart boundary regex to allow no space


packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts
```diff
@@ -97,7 +97,7 @@ function doExecuteTasks(differentialDownloader: DifferentialDownloader, options:
     }
 
     const contentType = safeGetHeader(response, "content-type")
-    const m = /^multipart\/.+?\s*;\s*boundary=(?:"([^"]+)"|([^\s";]+))\s*$/i.exec(contentType)
+    const m = /^multipart\/.+?\s*;[\s]*boundary=(?:"([^"]+)"|([^\s";]+))\s*$/i.exec(contentType)
```
mmaietta, line 100: Does this need to be adjusted as well?
https://github.com/electron-userland/electron-builder/blob/08773afadfe10911ed1dff084fefe1024c5a74d8/packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts#L49

fleabagneverdie: Maybe not, but adding or omitting spaces after commas in the Range header doesn’t affect functionality. Servers like Alibaba Cloud can recognize both formats, so it's not strictly necessary to add spaces.

mmaietta: Was just checking this further again. Isn't `\s*` the same as `[\s]*`? `*` is zero or unlimited, so the former version should be the same effectively pre/post-codechange. Right? I might just be needing some coffee though

fleabagneverdie: After checking, I found that a PR was already submitted 6 months ago with the relevant changes. However, the npm package hasn’t been updated yet, so no further action is needed on this. My apologies for the confusion.

mmaietta: Can you link the relevant PR for tracking purposes here? Can this PR be closed then?

fleabagneverdie: #9064 here

Issue #9312, opened 2025-10-15
NSIS CVE-2025-43715: Which NSIS version does electron-builder use, and how can I pin/override it?


chroberino:
Hi! I’m generating a Windows installer via `nsis` and need to document the exact NSIS version in my app’s SBOM, plus make sure I’m not affected by **CVE-2025-43715** (https://nvd.nist.gov/vuln/detail/CVE-2025-43715, priv-esc in NSIS < 3.11).

A few questions:

1. **Which NSIS version does electron-builder use by default?**
   - I don’t see it in electron-builder’s `package.json`. Docs for `CustomNsisBinary` show a default download URL and version of `nsis-3.0.4.1` (https://www.electron.build/app-builder-lib.interface.customnsisbinary), which would be affected by the CVE if that’s still the bundled/default compiler. Can you confirm what version ships/gets downloaded currently?
   - As a workaround to detect the version locally, running with `DEBUG=electron-builder:*` prints the path to `makensis.exe` in the builder cache, which includes the version in the folder name (e.g. path=`C:\Users\<username>\AppData\Local\electron-builder\Cache\nsis\nsis-3.0.4.1`). Is that the recommended way to determine the NSIS version for SBOMs?
2. **If the default is < 3.11, how can I control/override the NSIS version?**
The docs mention `nsis.customNsisBinary` for supplying a custom makensis. Could you confirm that the correct approach to pin to a fixed version (e.g., 3.11+) is to host a safe `makensis` archive and set `customNsisBinary` (with url/version/checksum) in my `build.nsis` config? Is there also a supported local-path/“use this makensis on disk” option, or only URL downloads?
3. **Project ask:** Given CVE-2025-43715 affects NSIS < 3.11, would you consider bumping the default bundled NSIS to ≥ 3.11 (and noting the version in the docs/changelog), so downstreams are safe by default and can easily record the version for SBOMs?

Thanks!

mmaietta:
Hello! I noticed you already found https://github.com/electron-userland/electron-builder-binaries/pull/78 where I'm trying to get makensis updated.

The current version being used is old and from release: https://github.com/electron-userland/electron-builder-binaries/releases/tag/nsis-3.0.4.1
I believe it is 3.08, maybe older; can't recall off the top of my head or check from this laptop.

[...]

PR #9333, opened 2025-11-07
feat(app-builder-lib): Allow local nsis-resources via env var `ELECTRON_BUILDER_NSIS_RESOURCES_DIR`

chroberino:
This PR addresses the topic of allowing local nsis-resources via environment variable as discussed in https://github.com/electron-userland/electron-builder/issues/9312

In addition to that it logs the custom paths of `ELECTRON_BUILDER_NSIS_DIR` and `ELECTRON_BUILDER_NSIS_RESOURCES_DIR` to the console.

### DIFF
--- a/.changeset/modern-singers-brake.md
+++ b/.changeset/modern-singers-brake.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": minor
+---
+
+feat: Allow local nsis-resources via env var `ELECTRON_BUILDER_NSIS_RESOURCES_DIR`


mmaietta (approved):

packages/app-builder-lib/src/targets/nsis/nsisUtil.ts
```diff
@@ -22,6 +22,7 @@ export const NsisTargetOptions = (() => {
 export const NSIS_PATH = () => {
   const custom = process.env.ELECTRON_BUILDER_NSIS_DIR
   if (custom != null && custom.length > 0) {
```
dmm9, line 24: This would permit only-blanks (e.g. " \n")
custom.length > 0 --> custom.trim().length > 0

packages/app-builder-lib/src/targets/nsis/nsisUtil.ts
```diff
@@ -38,6 +39,11 @@ export const NSIS_PATH = () => {
 }
 
 export const NSIS_RESOURCES_PATH = () => {
+  const custom = process.env.ELECTRON_BUILDER_NSIS_RESOURCES_DIR
+  if (custom != null && custom.length > 0) {
```
dmm9, line 43: This would permit only-blanks (e.g. "    \n")
`custom.length > 0` --> `custom.trim().length > 0`

Issue #9320, opened 2025-10-26
Enable Wayland by default for Snaps


mitchchn:
Electron 38.x enables Wayland by default: https://www.electronjs.org/docs/latest/breaking-changes#removed-electron_ozone_platform_hint-environment-variable.

But electron-builder defaults to disabling Wayland for Snap packages only: https://www.electron.build/snap.html#allownativewayland.

So an up-to-date Electron app will not work when built as a Snap package with the default settings, even though it works in other package formats. The default value for `allowNativeWayland` should be changed to `true`, at least for current versions of Electron.

mmaietta:
Alrighty, just circling in on this. Sounds like we need a `semver` check for `> 38` to avoid a breaking change in electron-builder

Thoughts (by anyone)?

255kb:
@mitchchn Our latest snap package is not starting with the following error messages, could it be related? 

```
MESA-LOADER: failed to open iris (search paths /snap/mockoon/48/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: iris
MESA-LOADER: failed to open kms_swrast (search paths /snap/mockoon/48/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast (search paths /snap/mockoon/48/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load swrast driver
[46395:1101/103635.400797:ERROR:ui/ozone/platform/wayland/host/wayland_connection.cc:197] Failed to connect to Wayland display: No such file or directory (2)
[46395:1101/103635.400844:ERROR:ui/ozone/platform/wayland/ozone_platform_wayland.cc:282] Failed to initialize Wayland platform
[46395:1101/103635.400851:ERROR:ui/aura/env.cc:257] The platform failed to initialize.  Exiting.
Segmentation fault (core dumped)
```

PR #9337, opened 2025-11-10
fix: change default value of disable_wayland depending on electron version (in order to support `electron` >38)

mmaietta:
Fixes: https://github.com/electron-userland/electron-builder/issues/9320

### DIFF
--- a/packages/app-builder-lib/src/targets/snap.ts
+++ b/packages/app-builder-lib/src/targets/snap.ts
@@ -135,10 +135,7 @@ export default class SnapTarget extends Target {
       delete snap.plugs
     } else {
       const archTriplet = archNameToTriplet(arch)
-      // https://github.com/electron-userland/electron-builder/issues/9320
-      const electronDefault = this.isElectronVersionGreaterOrEqualThan("38.0.0") ? "" : "1"
-      appDescriptor.environment = {
-        DISABLE_WAYLAND: options.allowNativeWayland ? "" : electronDefault,
+      const environment: Record<string, string> = {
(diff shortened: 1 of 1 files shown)


packages/app-builder-lib/src/targets/snap.ts
```diff
@@ -135,8 +135,10 @@ export default class SnapTarget extends Target {
       delete snap.plugs
     } else {
       const archTriplet = archNameToTriplet(arch)
+      // https://github.com/electron-userland/electron-builder/issues/9320
+      const electronDefault = this.isElectronVersionGreaterOrEqualThan("38.0.0") ? "" : "1"
       appDescriptor.environment = {
-        DISABLE_WAYLAND: options.allowNativeWayland ? "" : "1",
+        DISABLE_WAYLAND: options.allowNativeWayland ? "" : electronDefault,
```
mitchchn: I get schema validation errors if this is set to an empty string. Ideally the `DISABLE_WAYLAND` variable wouldn't be in this object at all (even as an empty string) if `allowNativeWayland` is true.

mmaietta: Hmmm, what happens if you use `"0"` instead of the empty string?

mitchchn: I tried that and it disabled Wayland. 😆 So I think the check on the snap side is pretty clumsy and not really parsing it that carefully.

mmaietta: Noted, I'll adjust the logic accordingly 😅 
Thanks for checking

PR #9349, opened 2025-11-13
chore(deps): unpinning semver and updating it

panther7:
ref #9301

**Why?**

If one project has the version "7.7.3" and another project has "7.7.2", the bundle package (e.g., Webpack) will include both versions, 7.7.3 and 7.7.2. However, if a project uses `~` or `^`, the build will include only one version in the bundle.  
This is just one example. If all dependencies have strict versions, all modules with different versions will be included in the bundle. This creates a cascading issue.  
If it is not necessary to require a specific version (e.g., for bug fixes, security, etc.), I see no reason to do so.  
Using the tilde (`~`) is a good compromise.

@mmaietta

### DIFF
--- a/.changeset/wild-oranges-poke.md
+++ b/.changeset/wild-oranges-poke.md
@@ -1,6 +1,6 @@
 ---
 "app-builder-lib": patch
-"electron-builder": patch
+"electron-updater": patch
 ---
 
-chore(deps): unpinning semver and updating it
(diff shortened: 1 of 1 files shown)


panther7:
any update with this?

mmaietta:
Trying to merge it but every time new commits are pushed, I have to requeue all the CI tests. No more pushes and I can get this into the next release 😅

panther7:
Ok, only rebase. :-)

Let's go.

PR #9386, opened 2025-11-24
perf: optimize duplicate checking in asar packager using Set

### LINKED ISSUE #9094: Packaging speed difference v22 vs v26
Hi,

our app got hit with the issue #6865 and this forced us to roll-back `electron-builder` to version `22.14.13`. 

I don't know yet if this has solved out problem but what I noticed immediately was that packaging step is like 20-30 times faster.

It now takes just few seconds instead of few minutes on the same system and otherwise identical code base (except for different builder of course). What can be the reason for this and why newer versions of `electron-builder` are so slow with packaging?

Kindly,
Robert

beyondkmp:
Fix #9094

Replace Array.some() with Set-based lookup for checking duplicate paths in processParentDirectories, improving performance from O(n) to O(1). Also remove check option from hoister configuration.

### DIFF
--- a/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
+++ b/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
@@ -64,7 +64,9 @@ export abstract class NodeModulesCollector<ProdDepType extends Dependency<ProdDe
       throw new Error("getNodeModules cancelled after building production graph")
     }
 
-    const hoisterResult: HoisterResult = hoist(this.transformToHoisterTree(this.productionGraph, packageName), {})
+    const hoisterResult: HoisterResult = hoist(this.transformToHoisterTree(this.productionGraph, packageName), {
+      check: log.isDebugEnabled,
+    })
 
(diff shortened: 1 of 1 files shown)


packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -64,7 +64,7 @@ export abstract class NodeModulesCollector<ProdDepType extends Dependency<ProdDe
       throw new Error("getNodeModules cancelled after building production graph")
     }
 
-    const hoisterResult: HoisterResult = hoist(this.transformToHoisterTree(this.productionGraph, packageName), { check: true })
+    const hoisterResult: HoisterResult = hoist(this.transformToHoisterTree(this.productionGraph, packageName), {})
```
mmaietta: Might be worth setting it as `{ check: log.isDebugEnabled }` so that we can get full debugging console log reports from users and also receive better insight while developing. Thoughts?

beyondkmp: Good idea. Fixed.

Issue #9025, opened 2025-04-11
Unable to copy, file is symlinked outside the package


kingyue737:
After I upgraded to v26, it throws error during build:
```
  • electron-builder  version=26.0.12 os=10.0.26100
  • loaded configuration  file=D:\Solutions\rho-meter\electron-builder.json
  • writing effective config  file=release\0.0.1\builder-effective-config.yaml
  • skipped dependencies rebuild  reason=npmRebuild is set to false
  • packaging       platform=win32 arch=x64 electron=35.1.5 appOutDir=release\0.0.1\win-unpacked
  • custom electronDist provided but no zip found; assuming unpacked electron directory.  electronDist=node_modules\electron\dist expectedFile=electron-v35.1.5-win32-x64.zip
  • copying Electron  source=D:\Solutions\rho-meter\node_modules\electron\dist destination=D:\Solutions\rho-meter\release\0.0.1\win-unpacked
  ⨯ unable to copy, file is symlinked outside the package  source=node_modules\@serialport\bindings-cpp\prebuilds\win32-x64\node.napi.node realPathFile=node_modules\.pnpm\@serialport+bindings-cpp@12.0.1\node_modules\@serialport\bindings-cpp\prebuilds\win32-x64\node.napi.node
  ⨯ unable to copy, file is symlinked outside the package  source=node_modules\@libsql\win32-x64-msvc\index.node realPathFile=node_modules\.pnpm\@libsql+win32-x64-msvc@0.5.4\node_modules\@libsql\win32-x64-msvc\index.node
  ⨯ Cannot cleanup: 

Error #1 --------------------------------------------------------------------------------
Error: Cannot copy file (node.napi.node) symlinked to file (node.napi.node) outside the package as that violates asar security integrity
    at writeFileOrProcessSymlink (D:\Solutions\rho-meter\node_modules\.pnpm\app-builder-lib@26.0.12_dmg_b995bd80b8ff3a8a64cee8adb9a5f824\node_modules\app-builder-lib\src\asar\asarUtil.ts:126:15)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Promise.all (index 2)
    at AsyncTaskManager.awaitTasks (D:\Solutions\rho-meter\node_modules\.pnpm\builder-util@26.0.11\node_modules\builder-util\src\asyncTaskManager.ts:65:25)     
[...]

PR #9389, opened 2025-11-25
fix: allow home dir when it's symlinked from `/var/home`

mmaietta:
This is needed to handle specific cases such as Fedora Silverblue (an immutable distribution where `$HOME` is a symlink to `/var/home`)
PR separates logic such that there is an explicit allow list of paths: currently just `HOME` and `os.tmpdir()`
Resolves: https://github.com/electron-userland/electron-builder/issues/9025#issuecomment-3575380041

CC @Jmennius

### DIFF
--- a/packages/app-builder-lib/src/asar/asarUtil.ts
+++ b/packages/app-builder-lib/src/asar/asarUtil.ts
@@ -11,42 +11,35 @@ import { detectUnpackedDirs } from "./unpackDetector"
 import { Readable } from "stream"
 import * as os from "os"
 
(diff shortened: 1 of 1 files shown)


Jmennius:
Looking at [FHS standard for /var](https://specifications.freedesktop.org/fhs/latest/varHierarchy.html), I would say that we should also allow `/var/tmp`.

Do I understand correctly that this is purely build-time protection?

mmaietta:
Correct, this is build-time protection.

Jmennius:
❤️

PR #9409, opened 2025-12-01
fix(pkg): support arm64-only MAS submissions by setting correct hostArchitectures

### LINKED ISSUE #6740: Setting minimumSystemVersion for MAS
* **Electron-Builder Version**: 22.14.13

* **Node Version**: v16.13.2

* **Electron Version**: 16.1.0

* **Target**: `mas`

Setting `minimumSystemVersion` on `mas` has no affect. It needs to be set on `mac` which means it would apply to all targets.

According to the documentation, https://www.electron.build/configuration/mas mas inherits the `macOS Options`.

We currently use the following workaround which allows us to specify it only for `mas`.
```
"extendInfo": {
  "LSMinimumSystemVersion": "10.14.0"
}
```

davej:
**Disclosure:** Code + description were AI-assisted.

## Backwards Compatibility

Verified that passing `--product` plist with `arch: ["arm64", "x86_64"]` produces **identical output** to the previous behavior (no `--product` flag). Both generate `hostArchitectures="arm64,x86_64"` in the distribution XML. The diff between the two outputs is empty.

## Problem

On macOS Big Sur+, `productbuild --synthesize` automatically generates distribution XML with `hostArchitectures="arm64,x86_64"`, causing Apple to reject arm64-only Mac App Store submissions.

This fixes two related App Store Connect validation errors:

### 1. When minimumSystemVersion is not set:

```
Invalid bundle. The bundle supports arm64 but not Intel-based Mac
computers. Your build must include the x86_64 architecture to support
Intel-based Mac computers. To support arm64 only, your macOS
deployment target must be 12.0 or higher.
```

### 2. When minimumSystemVersion is set in Info.plist but not in the pkg:

```
Invalid bundle. The .pkg/Payload/App.app bundle supports arm64 but
not Intel-based Mac computers. Your build must include the x86_64
architecture to support Intel-based Mac computers.
```

The second error occurs because setting `LSMinimumSystemVersion` in the app's `Info.plist` is not sufficient - the distribution XML inside the pkg must also declare the architecture and OS requirements.

## Solution

This fix uses Apple's recommended `--product` plist approach to pass architecture and OS requirements to `productbuild`, which then generates the correct `hostArchitectures` and `allowed-os-versions` in the distribution XML.

## Changes

- Create a requirements plist with `arch` and `os` keys based on build target
- Pass it to `productbuild --synthesize` via `--product` flag
- `productbuild` now generates correct `hostArchitectures` (`arm64`, `x86_64`, or both) based on the target architecture
- `minimumSystemVersion` from mac/mas config is passed through to generate the `volume-check`/`allowed-os-versions` XML block

## Usage

Users building arm64-only MAS packages should set `minimumSystemVersion` to `"12.0"` or higher in their mas config, and the pkg will now correctly declare arm64-only support.

```json
{
  "mas": {
    "minimumSystemVersion": "12.0"
  }
}
```

Closes #6740

### DIFF
--- a/packages/app-builder-lib/src/targets/pkg.ts
+++ b/packages/app-builder-lib/src/targets/pkg.ts
@@ -152,7 +152,6 @@ export class PkgTarget extends Target {
       requirements.os = [minimumSystemVersion]
     }
 
-    // Write requirements plist to temp file
     const requirementsPlistFile = await this.packager.info.tempDirManager.getTempFile({ suffix: ".plist", prefix: "productbuild-requirements" })
     await savePlistFile(requirementsPlistFile, requirements)
 


packages/app-builder-lib/src/targets/pkg.ts
```diff
@@ -118,9 +134,29 @@ export class PkgTarget extends Target {
...
+    requirements.arch = archToAppleArchitectures(arch)
+
+    // Set minimum OS version - productbuild will generate allowed-os-versions in distribution XML
+    const minimumSystemVersion = this.packager.platformSpecificBuildOptions.minimumSystemVersion
+    if (minimumSystemVersion != null) {
+      requirements.os = [minimumSystemVersion]
+    }
+
+    // Write requirements plist to temp file
+    const requirementsPlistFile = path.join(this.outDir, "productbuild-requirements.plist")
```
mmaietta: If possible, I'd like to use the temp file management electron-builder already leverages. It auto-cleans up after build completes
```
const requirementsPlistFile = await this.packager.info.tempDirManager.getTempFile({ suffix: ".plist", prefix: "requirements" })
```

davej: Thanks Mike! Made requested change and pulled in latest.

Issue #8149, opened 2024-03-22
Failed to create dylib symbolic link on Windows 


kochizufan:
* **Electron-Builder Version**: 24.13.3

* **Node Version**: v18.16.1

* **Electron Version**: 29.1.5
* **Electron Type (current, beta, nightly)**: current

* **Target**: Both NSIS and portable on Windows (Both results are same)

I tried to make Windows installer of my app with following command and settings:

```
electron-builder --win --x64 --config ./build_win.js
```

build_win.js
```
const config = {
  "appId": "jp.maplat.editor",
  "directories": {
    "output": "dist"
  },
  "files": [
    "assets/win",
    "backend",
    "css",
    "frontend/api",
    "frontend/dist",
    "frontend/fonts",
    "html",
    "img",
    "locales",
    "package.json",
    "package-lock.json",
    "tms_list.json"
  ],
  "win": {
    "icon": "assets/win/icon_win.ico",
    "target": "portable"
  },
  "nsis":{
    "oneClick": false,
    "allowToChangeInstallationDirectory": true
  }
};

module.exports = config;
```

But it is failed with following errors:
```
  • downloading     url=https://github.com/electron-userland/electron-builder-binaries/releases/download/winCodeSign-2.6.0/winCodeSign-2.6.0.7z size=5.6 MB parts=1
  • downloaded      url=https://github.com/electron-userland/electron-builder-binaries/releases/download/winCodeSign-2.6.0/winCodeSign-2.6.0.7z duration=1.104s
  ⨯ cannot execute  cause=exit status 2
                    out=
    7-Zip (a) 21.07 (x64) : Copyright (c) 1999-2021 Igor Pavlov : 2021-12-26

    Scanning the drive for archives:
    1 file, 5635384 bytes (5504 KiB)

    Extracting archive: C:\Users\kochi\AppData\Local\electron-builder\Cache\winCodeSign\618674575.7z
    --
    Path = C:\Users\kochi\AppData\Local\electron-builder\Cache\winCodeSign\618674575.7z
    Type = 7z
    Physical Size = 5635384
    Headers Size = 1492
    Method = LZMA2:24m LZMA:20 BCJ2
    Solid = +
    Blocks = 2


    Sub items Errors: 2

    Archives with Errors: 1

    Sub items Errors: 2

[...]

PR #9430, opened 2025-12-07
feat: updating win-codesign resources

mmaietta:
### Summary
This PR updates the Windows code signing tooling in electron-builder to use the latest version of `win-codesign` from `electron-builder-binaries`. It introduces a backward-compatible configuration system that allows users to access both legacy and newer code signing utilities during the transition period.

Resolves issue #8149 which reported failures creating dylib symbolic links on Windows during the code signing process. The issue lies in the "combo" winCodeSign toolset that has symlinks even though it's also supposed to be extracted on Windows.

**Key Updates:**
1. **Updated win-codesign tooling** to the latest version from `electron-builder-binaries` https://github.com/electron-userland/electron-builder-binaries/releases?q=codesign&expanded=true
2. New config property for handling different versions of win-codesign 
4. Default configuration set to version `0.0.0` with option to opt-in to newer toolsets
6. Added additional CI tests for coverage of both legacy and updated signtools
7. AppxTests can now finally run again. They require the updated signtool.exe in the updated WindowsKits

### DIFF
--- a/test/vitest-scripts/_vitest-smart-cache.json
+++ b/test/vitest-scripts/_vitest-smart-cache.json
@@ -1050,7 +1050,7 @@
     "win/linux concurrent": {
       "platformRuns": {
         "win32": {
-          "runs": 2,
+          "runs": 3,
           "fails": 0,
           "avgMs": 0
         },
@@ -1070,9 +1070,9 @@
     "win concurrent": {
       "platformRuns": {
         "win32": {
-          "runs": 2,
+          "runs": 3,
           "fails": 0,
-          "avgMs": 149866.1363625
+          "avgMs": 156560.7319
         },
         "darwin": {
           "runs": 0,
@@ -1090,9 +1090,9 @@
     "win concurrent - all targets": {
       "platformRuns": {
         "win32": {
-          "runs": 2,
+          "runs": 3,
           "fails": 0,
-          "avgMs": 528714.875917
+          "avgMs": 475705.4261
         },
         "darwin": {
           "runs": 0,
@@ -1145,7 +1145,7 @@
     "files": {
       "platformRuns": {
         "win32": {
-          "runs": 3,
+          "runs": 4,
           "fails": 0,
           "avgMs": 0
         },
@@ -3335,7 +3335,7 @@
     "expand not defined env": {
       "platformRuns": {
         "win32": {
-          "runs": 1,
+          "runs": 2,
           "fails": 0,
           "avgMs": 0
         },
@@ -3355,7 +3355,7 @@
     "extraResources on Windows": {
       "platformRuns": {
         "win32": {
-          "runs": 1,
+          "runs": 2,
           "fails": 0,
           "avgMs": 0
         },
@@ -3442,7 +3442,7 @@
     "files.from asar": {
       "platformRuns": {
         "win32": {
-          "runs": 1,
+          "runs": 2,
           "fails": 0,
           "avgMs": 0
         },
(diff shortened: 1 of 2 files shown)


packages/app-builder-lib/scheme.json
```diff
@@ -7931,6 +7931,17 @@
         }
       ]
     },
+    "toolsets": {
```
beyondkmp, line 7958: toolsets is an object in configuration.ts, but the schema here uses typeof: function, which will not validate or autocomplete correctly. Suggest adding a ToolsetConfig schema definition (including winCodeSign enum) and referencing it here.

packages/app-builder-lib/src/targets/tools.ts
```diff
@@ -39,6 +71,113 @@ export async function getFpmPath() {
...
+
+  const windowsKitArch = (x86: string) => (arch === "ia32" ? x86 : arch === "arm64" ? "arm64" : "x64")
+
+  const useLegacy = winCodeSign == null || winCodeSign === "0.0.0"
+  if (useLegacy) {
+    const vendorPath = await getBin("winCodeSign")
+    return { kit: path.resolve(vendorPath, "windows-10", windowsKitArch("ia32")), appxAssets: vendorPath }
+  }
+  const file = "windows-kits-bundle-10_0_26100_0.zip"
+  const vendorPath = await getBinFromUrl("win-codesign@1.0.0", file, wincodesignChecksums[winCodeSign][file])
```
beyondkmp: This hardcodes win-codesign@1.0.0, but the checksum table includes 1.1.0. If a user sets 1.1.0, we still download 1.0.0, which can cause checksum/resource mismatches. Consider using win-codesign@${winCodeSign} (or an explicit version map). Same issue in getOsslSigncodeBundle/getRceditBundle.

packages/app-builder-lib/src/targets/AppxTarget.ts
```diff
@@ -74,7 +75,7 @@ export default class AppXTarget extends Target {
       arch,
     })
 
-    const vendorPath = await getSignVendorPath()
+    const vendorPath = await getWindowsKitsBundle({ winCodeSign: this.packager.config.toolsets?.winCodeSign, arch: process.arch })
```
beyondkmp: Passing process.arch (host arch). AppX tooling should be selected based on the target arch; use the target arch (e.g. Arch[arch]) so ia32/arm64 builds pick the correct tools.

beyondkmp (commented):
Made 3 suggestions: the toolsets type for schema, the hardcoded version download for win-codesign, and the AppX tool selection should be based on the target arch.

packages/app-builder-lib/scheme.json
mmaietta: Great callout, will double check this!

packages/app-builder-lib/src/targets/tools.ts
mmaietta: Great idea on the map! Was able to get typescript to behave nicely with this and enforce typesafe keys.
```
type CodeSignVersionKey = keyof typeof wincodesignChecksums

function _getWindowsToolsBin<V extends CodeSignVersionKey>(winCodeSign: V, file: keyof (typeof wincodesignChecksums)[V]): Promise<string> {
  return getBinFromUrl(`win-codesign@${winCodeSign}`, file as string, wincodesignChecksums[winCodeSign][file] as string)
}
```

packages/app-builder-lib/src/targets/AppxTarget.ts
mmaietta: Thanks for calling this out. Updated AppX tooling (Windows Kits) to be for the target architecture. Signtool is the host arch though based on previous logic paths from what I understand.

PR #9438, opened 2025-12-09
chore: updating all engines to node 22

### LINKED ISSUE #9483: FYI/PSA: ESM-only + Node >22 `engine` upcoming in next major semver, v27
FYI/PSA 📣

**electron-builder is migrating to ESM-only and `engine` node >22 in next major semver bump, in v27.**

This is to adopt latest updates to `electron/*` upstream packages that electron-builder leverages (osx-sign, asar, universal, fuses, etc.)

This will be an `-alpha` release state while the migration occurs.

As a consequence of the monorepo CI/CD, this will be a non-backward change and v26 support will need to be dropped as there's no way to backport fixes.

mmaietta:


luc-debug:
Can you please merge this PR?

mmaietta:
@luc-debug This is a non-backward compatible change and will be part of an `-alpha` release tag/versioning when it is merged.

I've just bumped electron-builder `latest` tag to most recent release, so I'm waiting a week or two to field any issues are reported. Then will proceed with this node 22 upgrade + also an ESM-only migration. https://github.com/electron-userland/electron-builder/issues/9483

PR #9439, opened 2025-12-10
fix: ensure correct dependency graph extraction order and yarn berry _dependencies support

beyondkmp:
Fix https://github.com/electron-userland/electron-builder/issues/9394#issuecomment-3629957529

- Change npm dependency graph extraction from parallel to sequential processing to ensure consistent ordering
- Add _dependencies field check in yarn berry isProdDependency for proper production dependency detection

### DIFF
--- a/packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
+++ b/packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
@@ -47,13 +47,16 @@ export class NpmNodeModulesCollector extends NodeModulesCollector<NpmDependency,
     this.productionGraph[dependencyId] = { dependencies: [] }
 
     const collectedDependencies: string[] = []
-    for (const [packageName, dependency] of Object.entries(resolvedDeps || {})) {
-      if (!this.isProdDependency(packageName, tree)) {
-        continue
+    if (resolvedDeps && Object.keys(resolvedDeps).length > 0) {
+      for (const packageName in resolvedDeps) {
+        if (!this.isProdDependency(packageName, tree)) {
+          continue
+        }
+        const dependency = resolvedDeps[packageName]
+        const childDependencyId = this.packageVersionString(dependency)
+        await this.extractProductionDependencyGraph(dependency, childDependencyId)
+        collectedDependencies.push(childDependencyId)
       }
-      const childDependencyId = this.packageVersionString(dependency)
-      await this.extractProductionDependencyGraph(dependency, childDependencyId)
-      collectedDependencies.push(childDependencyId)
     }
     this.productionGraph[dependencyId] = { dependencies: collectedDependencies }
   }


packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
```diff
@@ -45,17 +45,15 @@ export class NpmNodeModulesCollector extends NodeModulesCollector<NpmDependency,
...
-      .map(async ([, dependency]) => {
-        const childDependencyId = this.packageVersionString(dependency)
-        await this.extractProductionDependencyGraph(dependency, childDependencyId)
-        return childDependencyId
-      })
 
     const collectedDependencies: string[] = []
-    for (const dep of productionDeps) {
-      collectedDependencies.push(await dep)
+    for (const [packageName, dependency] of Object.entries(resolvedDeps || {})) {
```
mmaietta: I think we may have an additional issue here that could be leading to ordering?

Found this on `Object.entries`
```
Is Object.entries idempotent?

No, because object key order is not strictly insertion-ordered for all keys.

Does it retain insertion order like a Map?

Not fully — only for non-numeric string keys.
Numeric-like keys are reordered.
```
Wondering if we're hitting some edge case here during extraction

packages/app-builder-lib/src/node-module-collector/yarnBerryNodeModulesCollector.ts
```diff
@@ -36,7 +36,7 @@ export class YarnBerryNodeModulesCollector extends NpmNodeModulesCollector {
   }
 
   protected isProdDependency(packageName: string, tree: NpmDependency): boolean {
-    return tree.dependencies?.[packageName] != null || tree.optionalDependencies?.[packageName] != null
+    return tree._dependencies?.[packageName] != null || tree.dependencies?.[packageName] != null || tree.optionalDependencies?.[packageName] != null
```
mmaietta, line 39: Wondering if we should delete this function and just move the entire logic to `npmModuleCollector :: isProdDependency` base function since the npm collector does both `npm list` (`_dependencies`) and manual file hierarchy traversal (`dependencies` + `optionalDependencies`).

Thoughts?

mmaietta: Extrapolating on this. What are your thoughts on having an if-statement for `isProdDependency` that is based on a flag set in  the `npmNodeModuleColletor`?
```
protected buildNodeModulesTreeManually(baseDir: string): Promise<NpmDependency> {
    this.usedManualTraversal = true
   ....

protected isProdDependency(packageName: string, tree: NpmDependency): boolean {
    if (this.usedManualTraversal) {
      return tree.dependencies?.[packageName] != null || tree.optionalDependencies?.[packageName] != null
    }
    return tree._dependencies?.[packageName] != null // results of `npm list` has explicit dependencies under _dependencies
  }
```

If we go this route, I could potentially refactor the code a bit and extract the manual traversal logic into `nodeModuleCollector` base class as a fallback for all package managers.

packages/app-builder-lib/src/node-module-collector/npmNodeModulesCollector.ts
beyondkmp: Removed Object.entries, so there shouldn't be any order issues now.

packages/app-builder-lib/src/node-module-collector/yarnBerryNodeModulesCollector.ts
beyondkmp: I still don’t understand what the purpose of this isProdDependency is.

mmaietta: Ah. I had abstracted it for package managers that extended `NpmNodeModulesCollector` but still could leverage `extractProductionDependencyGraph` as I noticed that the only differing logic between npm/bun/yarn was just the `filter((packageName) => tree._dependencies?.[packageName] != null)`. This was for the use case where corepack is enabled, so `npm list` won't work, and it has the file traversal fallback

beyondkmp: > This was for the use case where corepack is enabled, so npm list won't work, and it has the file traversal fallback

Interesting. After corepack is enabled, does npm list become unusable? 

In several projects I’m working on, for example Cherry Studio uses yarn berry after enabling corepack, and npm list still works fine.

beyondkmp:
```
➜  cherry-studio git:(main) corepack enable
➜  cherry-studio git:(main) npm list -a --include prod --include optional --omit dev --json --long --silent > /tmp/npmlist.txt
➜  cherry-studio git:(main) head /tmp/npmlist.txt
{
  "version": "1.7.2",
  "name": "CherryStudio",
  "private": true,
  "description": "A powerful AI assistant for producer.",
  "main": "./out/main/index.js",
  "author": "support@cherry-ai.com",
  "homepage": "https://github.com/CherryHQ/cherry-studio",
  "engines": {
    "node": ">=22.0.0"
```

mmaietta: Okay, that's super odd, because the GH runners are behaving like my local setup AFAICT.
```
mikemaietta@Mikes-MBP ~/D/cherry-studio> corepack enable
mikemaietta@Mikes-MBP ~/D/cherry-studio> npm list -a --include prod --omit dev --json
This project is configured to use yarn because /Users/mikemaietta/Development/cherry-studio/package.json has a "packageManager"
```

mmaietta: Hmmm, maybe Windows behaves differently? I just noticed that `HoistedNodeModuleTest` or `packageManagerTest` aren't on the CI Windows runner

beyondkmp: image


I tested it on Windows and it works fine as well.

beyondkmp: I reproduced it; when corepack also takes over npm, it reports an error.

```
➜  cherry-studio git:(main) corepack enable npm
➜  cherry-studio git:(main) rehash
➜  cherry-studio git:(main) npm list
This project is configured to use yarn because /Users/payne/Code/cherry-studio/package.json has a "packageManager" field
```

beyondkmp: In the above situation, we don't need to write our own solution from scratch. npm list considers more cases than what we would cover ourselves.

We can use the following command to retrieve it:

```
CI=1 COREPACK_ENABLE_STRICT=0 npm list -a --include prod --omit dev --json
```

beyondkmp: Another approach is `corepack disable npm`, then `npm list`, then `corepack enable npm`

mmaietta: Oh, now that's a nifty idea! Since we're only leveraging `npm list` and a pure read-only operation, I think it's safe to not force corepack to be strictly enabled. I think we can still have our manual traversal as a fallback though.

Thinking we just edit `nodeModulesCollector :: streamCollectorCommandToFile`
```
      const child = childProcess.spawn(command, args, {
        cwd,
        env: { COREPACK_ENABLE_STRICT: "0", ...process.env },
        shell: false, // required to prevent console logs polution from shell profile loading when `true`
      })
```
`COREPACK_ENABLE_STRICT` must be before `...process.env` to allow user-overrides to still be respected.

beyondkmp: > I think we can still have our manual traversal as a fallback though.

We also don’t need to write another implementation ourselves; we can directly fall back to https://github.com/develar/app-builder. Because this app-builder will probably take a long time to be completely removed.

mmaietta: Hmmm, doesn't app-builder have incompatibilities with non-hoisted setups? But yes, I agree, that is an option we could take.

I'd prefer not to use app-builder though, as my goal is to extract this logic into a new package within the monorepo, but with a limited dependency tree/size so that other `electron/` projects can also leverage it individually

mmaietta: Wait, does `COREPACK_ENABLE_STRICT: "0"` mean that `yarn list` in `yarnNodeModulesCollector` also is unneeded since we can just always use `npm list` with manual traversal fallback?

mmaietta (approved):

packages/app-builder-lib/src/node-module-collector/yarnBerryNodeModulesCollector.ts
beyondkmp: > Wait, does `COREPACK_ENABLE_STRICT: "0"` mean that `yarn list` in `yarnNodeModulesCollector` also is unneeded since we can just always use `npm list` with manual traversal fallback?

Yes.

mmaietta: Most of the work I did to support corepack PM isolation in each unit test is effectively voided then. 

Well. PR LGTM. Let me know when okay to proceed with merge.

beyondkmp: Removing the previous code and ut is quite a bit of work; should we do it in another pr? Let's merge this one first.

PR #9440, opened 2025-12-10
chore: extract `TraversalNodeModuleCollector` for fallback project parsing

mmaietta:
- Extract fallback node collector (Traversal) to a separate class due to differing parsing logic from NPM collector object `NpmDependency`

Resolves https://github.com/electron-userland/electron-builder/issues/9394#issuecomment-3737761339


packages/app-builder-lib/src/node-module-collector/index.ts
```diff
@@ -11,21 +11,26 @@ import { Lazy } from "lazy-val"
...
-      return new YarnBerryNodeModulesCollector(rootDir, tempDirManager)
+      return new YarnBerryNodeModulesCollector(rootDir, tempDirManager, cancellationToken)
     case PM.BUN:
-      return new BunNodeModulesCollector(rootDir, tempDirManager)
+      return new BunNodeModulesCollector(rootDir, tempDirManager, cancellationToken)
     case PM.NPM:
-      return new NpmNodeModulesCollector(rootDir, tempDirManager)
+      return new NpmNodeModulesCollector(rootDir, tempDirManager, cancellationToken)
+    // should never access this case (as it's internally a fallback), but TS needs a default and we need to satisfy it
+    case PM.TRAVERSAL:
```
beyondkmp, line 30: How about throwing an error in the default case?

packages/app-builder-lib/src/node-module-collector/yarnNodeModulesCollector.ts
beyondkmp, line 1: yarn list should also be deletable, just use `npm list` directly

packages/app-builder-lib/src/node-module-collector/index.ts
mmaietta: We use TRAVERSAL enum when forcing unit tests to use Traversal collector here. The Traversal PM override is passed in here: https://github.com/electron-userland/electron-builder/pull/9440/changes#diff-7ba188ac03f1818c556b0e9a15841ac8fa5922f226e88d541fd0a467564ed024L195

New tests are located here: https://github.com/electron-userland/electron-builder/pull/9440/changes#diff-a39811e95285a0e377cb9d1c6e363e71fe1138e6acdd28a590bfed66e5e641a0

packages/app-builder-lib/src/node-module-collector/yarnNodeModulesCollector.ts
mmaietta: Let's extract that to a separate PR. I'm worried that'll convolute the purpose of this PR (extract manual Traversal) and any issues introduced in the `yarn list` removal would be difficult to bisect.

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
```diff
@@ -1,17 +1,13 @@
 import { log } from "builder-util"
-import { NpmNodeModulesCollector } from "./npmNodeModulesCollector"
 import { PM } from "./packageManager"
 import { NpmDependency } from "./types"
+import { TraversalNodeModulesCollector } from "./traversalNodeModulesCollector"
 
-export class BunNodeModulesCollector extends NpmNodeModulesCollector {
+export class BunNodeModulesCollector extends TraversalNodeModulesCollector {
```
mmaietta, line 6: @beyondkmp note: this Bun collector uses the Traversal collector as base class. It returned far more node_modules in the collection that what `npm list` returned AND the original explicit PR implementation for Bun by a community member was manual traversal as well. Decided to leave it as-is. Albeit I just noticed that the logging below needs to be updated.

packages/app-builder-lib/src/configuration.ts
```diff
@@ -134,6 +135,12 @@ export interface CommonConfiguration {
    */
   readonly nativeRebuilder?: "legacy" | "sequential" | "parallel" | null
 
+  /**
+   * The package manager to use for installing dependencies and node module collections.
+   * @default auto
+   */
+  readonly packageManager?: "auto" | PM | null
```
mmaietta: CC @beyondkmp, curious on your thoughts on adding this.

For next major semver, I'd like to reorganize `npm/node colletion` properties nested within a centralized property (maybe named `modules`?)

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
beyondkmp: Interesting, why is it so much more? It seems there's an issue with `npm list` under Bun.

packages/app-builder-lib/src/configuration.ts
beyondkmp: This is a great configuration. Sometimes PM make inaccurate judgments, but with this, users can specify it themselves.

beyondkmp (approved):

packages/app-builder-lib/src/node-module-collector/bunNodeModulesCollector.ts
mmaietta: Honestly, I couldn't figure that out but I do intend to investigate it further.

Within the scope of this PR, using TraversalNodeModulesCollector made sure none of the test snapshots changed.

I'll create a follow-up PR for `COREPACK_STRICT_ENABLE=0` and we can retest both Yarn v1 and Bun approaches in that PR

packages/app-builder-lib/src/node-module-collector/moduleManager.ts
```diff
@@ -0,0 +1,259 @@
...
+    const direct = path.join(path.resolve(parentDir), "node_modules", pkgName, "package.json")
+    if (await this.exists[direct]) {
+      const json = await this.json[direct]
+      if (json && this.semverSatisfies(json.version, requiredRange)) {
+        return { packageDir: path.dirname(direct), packageJson: json }
+      }
+    }
+
+    // 2) upward hoisted search, then 3) downward non-hoisted search
+    return (await this.upwardSearch(parentDir, pkgName, requiredRange)) || (await this.downwardSearch(parentDir, pkgName, requiredRange)) || null
```
beyondkmp, line 93: Is this only reachable during manual traversal? Because if pnpm isn't hoisted, reaching this point will definitely cause issues.

mmaietta: Yes, it's used in pnpm collector's `getProductionDependencies`. What's the issue here when pnpm isn't hoisted? We handle that usecase with the subsequent downward search.

PR #9448, opened 2025-12-15
feat: Provide progress updates for multiple range differential downloads

eliotschu:
### Summary of Changes
- Pass total number of expected download bytes and onProgress event emitter to multiple range differential download data splitter
- Update transferred and delta when processing downloaded part data
- After each chunk of data is handled in data splitter, check timer and emit progress message if 1s has passed, or download is complete
- Remove TODO note about this feature not being implemented


packages/electron-updater/src/differentialDownloader/DataSplitter.ts
```diff
@@ -69,7 +77,27 @@ export class DataSplitter extends Writable {
...
+            this.onProgress({
+              total: this.grandTotalBytes,
+              delta: this.delta,
+              transferred: this.transferred,
+              percent: (this.transferred / this.grandTotalBytes) * 100,
+              bytesPerSecond: Math.round(this.transferred / ((now - this.start) / 1000)),
+            })
+            this.delta = 0
+          }
+        }
```
mmaietta, line 97: I like the simplicity of this PR in that it's throttling/firing a callback when data is received (as opposed to a NodeJS.Timer approach), but I'm concerned with the number of properties that are being added.
```
  private start = Date.now()
  private nextUpdate = this.start + 1000
  private transferred = 0
  private delta = 0
```
Feels like we're recreating a wheel here but I can't think of a better approach. Can you at least please add a short tsdoc as to what each is doing (or expected to be used as/for)?

eliotschu: I agree that it feels overall like a very simple way of getting these updates in for single request differential downloads. It is also nice not using a timer as that would require both keeping track of the timer and passing through a cancellation token to ensure the timer is also appropriately cancelled.

This method with these four new properties is largely inspired by how progress updates are tracked for differential downloads that use individual range requests (`ProgressDifferentiaslDownloadCallbackTransform.ts`). But implemented as an addition to the data splitter as it already handles the parsing of multiple range requests and it would actually be all of the data (downloaded and copied) that would be passed through a new stream transform if that transform is placed in the same order as other transforms for progress updates.

eliotschu:
I have added comments for the new class properties and added a check that the divisors copilot review mentioned are truthy before attempting division.

eliotschu:
I see that one of the tests failed. It looks like a similar failure to a couple of the linux e2e tests that failed on the last commit to master. Let me know if you think there are actually potential issues I should look into with this branch!

Issue #9445, opened 2025-12-14
Node modules resolution in a Nx Monorepo


johnbwoodruff:
I have an electron app that is part of an Nx Monorepo. Since updating electron-builder on this project that has been working great for years I started seeing the following error when attempting to build:

```shell
• packaging       platform=darwin arch=universal electron=39.2.7 appOutDir=release/mac-universal
⨯ Detected file "Contents/Resources/app.asar.unpacked/node_modules/@nx/nx-darwin-arm64/nx.darwin-arm64.node" that's the same in both x64 and arm64 builds and not covered by the x64ArchFiles rule: "undefined"  failedTask=build stackTrace=Error: Detected file "Contents/Resources/app.asar.unpacked/node_modules/@nx/nx-darwin-arm64/nx.darwin-arm64.node" that's the same in both x64 and arm64 builds and not covered by the x64ArchFiles rule: "undefined"
    at makeUniversalApp (/Users/johnwoodruff/test/node_modules/@electron/universal/src/index.ts:181:17)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at MacPackager.doPack (/Users/johnwoodruff/test/node_modules/app-builder-lib/src/macPackager.ts:188:9)
    at MacPackager.pack (/Users/johnwoodruff/test/node_modules/app-builder-lib/src/macPackager.ts:252:9)
    at Packager.doBuild (/Users/johnwoodruff/test/node_modules/app-builder-lib/src/packager.ts:519:11)
    at executeFinally (/Users/johnwoodruff/test/node_modules/builder-util/src/promise.ts:12:14)
    at Packager.build (/Users/johnwoodruff/test/node_modules/app-builder-lib/src/packager.ts:439:31)
    at executeFinally (/Users/johnwoodruff/test/node_modules/builder-util/src/promise.ts:12:14)
```

[...]

PR #9450, opened 2025-12-15
fix: search app directory before package directory to respect two-package.json project setups

mmaietta:
Fixes: https://github.com/electron-userland/electron-builder/issues/9445

CC @johnbwoodruff


johnbwoodruff:
This change fixes my issue! 👏

PR #9470, opened 2025-12-27
fix: pnpm "Invalid string length" errors during node module collection

### LINKED ISSUE #9469: PNPM node modules collectors fails with lots of dependencies
electron-builder calls `pnpm list --prod --json --depth Infinity --long` to collect node modules with pnpm. My project has many dependencies, so this command fails with the following error:

```
{
  "error": {
    "code": "pnpm",
    "message": "Invalid string length"
  }
}
```

danteissaias:
Closes #9469 

electron-builder calls `pnpm list --prod --json --depth Infinity --long` to collect node modules with pnpm. My project has many dependencies, so this command fails with the following error:

```
{
  "error": {
    "code": "pnpm",
    "message": "Invalid string length"
  }
}
```

Removing `--long` resolves this error. This flag adds the following fields to the output:
- `description`
- `license`
- `author`
- `homepage`
- `repository`

electron-builder doesn't appear to consume any of these fields, so `--long` can safely be removed.

### DIFF
--- a/.changeset/good-lights-smell.md
+++ b/.changeset/good-lights-smell.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: pnpm "Invalid string length" errors during node module collection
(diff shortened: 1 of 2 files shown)


beyondkmp:
very good catch. 

LGTM

beyondkmp (approved):

danteissaias:
Thanks, although I'm not sure what is causing the test failure. It seems unrelated to the change I made.

mmaietta:
It's a flaky test; re-running the tests. Not sure why it's minorly flaky, but also not sure if it's worthwhile to solve for that with the complexity of install+run+serve/autoupdate+relaunch within a docker container.

Issue #5463, opened 2020-12-10
Don't publish by default on github builds


pimterry:
* **Version**: 22.9.1

When setting up a new github actions build, it attempts to publish the result automatically to github by default. I only noticed this because I received an unexpected error at the end of my build:

```
  ⨯ GitHub Personal Access Token is not set, neither programmatically, nor using env "GH_TOKEN"
```

This was a normal branch build, not tagged or anything, which I'd never normally want to publish.

This behaviour is pretty surprising, and in many cases it's a serious security risk if you don't expect it when there's github token in the environment for some other reason. Build output even in open source repos could easily contain secrets, especially when you're first setting up a new build - it would be easy for this to accidentally publish your code signing keys, for example.

Also, this surprising behaviour isn't documented - there's no default value for publish listed on the [config value](https://www.electron.build/configuration/publish) or [CLI arguments](https://www.electron.build/cli).

There's been quite a few issues discussing this in the past, all now closed due to inactivity, e.g. https://github.com/electron-userland/electron-builder/issues/4546 and https://github.com/electron-userland/electron-builder/issues/1693. I think it's pretty clear that this is causing confusion.

**Are you be open to disabling this?** I think it's just a matter of making `null` the default value for publish config (that's the current workaround for this behaviour). I'd be happy to put a PR together for this, if you're open to that.

In general, I'd personally prefer to have `electron-builder publish` as a separate action, so I can build that into my CI workflow in more complex ways (e.g. testing a built distributable before it's published) but that's a separate question, and being able to opt-in with `--publish` is fine for now.

pimterry:
This is still relevant, and AFAIK there's nothing blocking it and it's very easy to do (although it is a breaking change!). It would be great to hear thoughts from the maintainers on this issue.

adityaruplaha:
Still an issue, I don't know why this behaviour was selected as default.

Apak00:
Seems like it tries to publish when my package.json script is named "release" and it does not when the name is something else like "dist".

vraravam:
This was working correctly in version `22.11.3`, but has regressed in `22.11.7`. Can it be fixed please? We are unable to build for non-release builds where we do NOT want to publish the built artifacts.

pimterry:
This is still relevant, it has a fair few votes now, and AFAIK there's nothing blocking it and it's very easy to do (although it is a breaking change!). It would be great to hear thoughts from the maintainers on this issue.

ferdodo:
[...]

PR #9476, opened 2025-12-31
fix: disable implicit publishing by default [breaking]

biw:
Remove automatic publish detection based on CI environment, git tags, and npm lifecycle events. Publishing must now be explicitly requested via the `--publish` CLI flag or configuration.

This is a **breaking change** that addresses the security and usability concerns raised in electron-userland/electron-builder#5463.

**BREAKING CHANGE:** Publishing no longer happens automatically in CI. Use `--publish` flag explicitly (e.g., `--publish always`, `--publish onTag`).

This is designed to be merged in version 27, and #9475 is the warning for the next patch version.


mmaietta:
Thanks for prepping the multi-step set of PRs! Love the change.
Aptly timed as well, as the ESM/Node22 migration for electron-builder v27 is beginning soon.

biw:
Happy I could help!

On the test falure, is there a easy way to fix them / verify they aren't false positives?

mmaietta:
Looks like it was a transitive dependency in the node_modules collector fixtures that had a file added, so it was causing the tests to fail suddenly (no changes to `master` branch)

Just pushed the change and updated the relevant PRs to latest `master`

PR #9477, opened 2025-12-31
feat: add full pnpm support

biw:
Adds pnpm support and fixes a stream race condition in the node module collector.

### pnpm Support

- Added `resolveFromProject()` helper for pnpm-compatible module resolution (falls back to original behavior for npm/yarn)
- Fixed electron version detection for hoisted packages in pnpm workspaces
- Added support for pnpm's `link:` workspace protocol
- Deprecated the `start` command (depends on unmaintained electron-webpack)

### Bug Fixes

- Fixed stream race condition in `nodeModulesCollector.ts` causing `Unterminated string in JSON` errors on large dependency trees (~64KB+). The write stream `finish` event now properly coordinates with the child process `close` event.
- Fixed e2e test using `getTempDir()` instead of `createTempDir()`

### Backward Compatibility

- npm/yarn behavior unchanged
- All pnpm changes include fallbacks to original behavior

### Snapshot Changes

The `HoistedNodeModuleTest.js.snap` diff is large due to the `qs` package updating, which shifted asar offsets throughout the archive. No functional change.


.github/workflows/pr-netlify.yml
```diff
@@ -14,6 +14,8 @@ jobs:
   deploy:
     name: 'Deploy PR'
     runs-on: ubuntu-latest
+    # Only run on the main repository, not on forks
+    if: github.repository == 'electron-userland/electron-builder'
```
biw, line 18: this is https://github.com/electron-userland/electron-builder/pull/9474 - happy to remove from the PR if ideal

mmaietta:
Just found this PR in my history. Why was this closed?

biw:
To be honest, I thought electron-builder had no pnpm support because I was trying to use it with [vite-plugin-electron](https://github.com/electron-vite/vite-plugin-electron), and only after switching to [electron-vite](https://github.com/electron-vite/vite-plugin-electron) did I realize that electron-builder already supported pnpm 🤦 

I think there are a handful of fixes/improvements in this PR, I just didn't want to throw it over the proverbial wall and increase your maintenance burden given I misunderstood the level of pnpm support before submitting this. 

That said, I'm happy to open it back up and work to clean up anything you think would be useful if it doesn't add too much to your plate!

packages/app-builder-lib/src/util/projectModuleResolver.ts
```diff
@@ -0,0 +1,80 @@
...
+    currentDir = parentDir
+  }
+
+  // If no paths found, at least try the project directory
+  if (searchPaths.length === 0) {
+    searchPaths.push(projectDir)
+  }
+
+  try {
+    const resolved = require.resolve(moduleSpecifier, { paths: searchPaths })
```
mmaietta, line 48: electron-builder actually used `require.resolve` implementation previously, in addition to `import-meta-resolve`, and IIRC, this was incompatible with any package that did not have an entry/main in the package.json. 

`require.resolve` also doesn't guarantee that the package.json is at root (monorepo/scoped packages have it in subdirs)

Now we do a hoisted search => upward search => downward search in the following code, making sure to respect/check for semver ranges in case there's conflicting or multiple versions installed.
https://github.com/electron-userland/electron-builder/blob/c3af390ec67becfe548a8c83f0d3ba828c28b751/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts#L293-L305

And then we manually construct the dependency tree based off each package.json for pnpm
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts

packages/app-builder-lib/src/util/config/config.ts
```diff
@@ -57,13 +58,29 @@ export async function getConfig(
     if ((dependencies != null && "react-scripts" in dependencies) || (devDependencies != null && "react-scripts" in devDependencies)) {
       config.extends = "react-cra"
     } else if (devDependencies != null && "electron-webpack" in devDependencies) {
-      let file = "electron-webpack/out/electron-builder.js"
-      try {
-        file = require.resolve(file)
-      } catch (_ignore) {
-        file = require.resolve("electron-webpack/electron-builder.yml")
+      // Note: electron-webpack is deprecated. This support is maintained for legacy projects.
```
mmaietta, line 61: Wait, really? Might be time to drop support for it then in v27 for simplicity

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -21,8 +21,46 @@ export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependenc
       throw new Error(`Cannot compute production dependencies for package with empty name: ${packageName}`)
     }
 
-    const actualPath = await this.locatePackageVersion(depTree.path, packageName, depTree.version).then(it => it?.packageDir)
-    const resolvedLocalPath = await this.cache.realPath[actualPath ?? depTree.path]
+    // Handle link: protocol for pnpm workspace packages - the path is already the package directory
+    if (depTree.version?.startsWith("link:")) {
```
mmaietta, line 25: Is `link:` pnpm only? Wondering if it already works now since `file` is supported
https://github.com/electron-userland/electron-builder/blob/c3af390ec67becfe548a8c83f0d3ba828c28b751/test/src/packageManagerTest.ts#L397-L401

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
```diff
@@ -274,18 +301,17 @@ export abstract class NodeModulesCollector<ProdDepType extends Dependency<ProdDe
...
-          log.debug(null, "`npm list` returned non-zero exit code, but it MIGHT be expected (https://github.com/npm/npm/issues/17624). Check stderr for details.")
-        }
-        if (stderr.length > 0) {
-          log.debug({ stderr }, "note: there was node module collector output on stderr")
-        }
-        const shouldResolve = code === 0 || shouldIgnore
-        return shouldResolve ? resolve() : reject(new Error(`Node module collector process exited with code ${code}:\n${stderr}`))
+        exitCode = code
+        childExited = true
+        tryComplete()
```
mmaietta, line 314: This seems really applicable and useful! Could you please extract to a separate PR?

test/src/helpers/packTester.ts
```diff
@@ -179,6 +179,15 @@ export async function assertPack(expect: ExpectStatic, fixtureName: string, pack
       const { cli, prepareEntry, version } = getPackageManagerWithVersion(pm, packageManager)
       if (pm === PM.BUN) {
         log.info({ pm, version: version, projectDir }, "installing dependencies with bun; corepack does not support it currently and it must be installed separately")
+      } else if (pm === PM.NPM) {
+        // npm is already installed as part of Node.js, so we just need to prepare the specific version
+        // Skip `corepack enable npm` to avoid race conditions in parallel tests (EEXIST errors)
```
mmaietta, line 184: Where did you see these errors? I don't think I've seen them on the CI

packages/app-builder-lib/src/util/config/config.ts
biw: I guess it's no offically "deprecated" as much as the readme says:
> Project Status: maintenance mode. Consider using [Vite](https://github.com/cawa-93/vite-electron-builder) instead of Webpack.

and the project hasn't been updated in ~4.5 years. Would it be helpful for me to make this into it's own PR so other can comment before v27?

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
biw: Ya, `pnpm` has both `link:` and `file:` 

[What's the difference between pnpm link and using the file: protocol?](https://pnpm.io/cli/link#whats-the-difference-between-pnpm-link-and-using-the-file-protocol):

> When you use pnpm link, the linked package is symlinked from the source code. You can modify the source code of the linked package, and the changes will be reflected in your project. With this method pnpm will not install the dependencies of the linked package, you will have to install them manually in the source code. This may be useful when you have to use a specific package manager for the linked package, for example, if you want to use npm for the linked package, but pnpm for your project.
> 
> When you use the file: protocol in dependencies, the linked package is hard-linked to your project node_modules, you can modify the source code of the linked package, and the changes will be reflected in your project. With this method pnpm will also install the dependencies of the linked package, overriding the node_modules of the linked package.
> 
>
> Feature | pnpm link | file: Protocol
> -- | -- | --
> Symlink/Hard-link | Symlink | Hard-link
> Reflects source code modifications | Yes | Yes
> Installs dependencies of the linked package | No (manual installation required) | Yes (overrides node_modules of the linked package)
> Use different package manager for dependency | Possible (e.g., use npm for linked pkg) | No, it will use pnpm

packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
biw: Yup, will do!

test/src/helpers/packTester.ts
biw: Let me do another review on this one, and I'll get back to you.

biw:
Based on your comments, what I'm going to do is make separate PRs for some of these issues, and then we can merge them one at a time rather than this full dump of things that some were fixes for a half-working vite-plugin-electron (before I switch to electron-vite)

PR #9481, opened 2026-01-04
fix: deleting default macOS appPlist keys using extendInfo

### LINKED ISSUE #9480: Cannot unset default macOS appPlist entries via extendInfo
  - `electron-builder@26.4.0`
  - `master` (locally compiled)
* **Node Version**: Node.js v22.14.0
* **Electron Version**: 39.0.0
* **Target**:  macOS arm64 .dmg

It used to be possible to set `key: undefined` in `mac.extendInfo`, which would delete keys from the resulting `Info.plist` for macOS build.

That's very useful if we want to remove permissions that our app does not use – for example, no point in asking user for camera or microphone permissions, if our app does not need it 🙂 
Last known working state: 26.0.3, but with `electron-builder@26.4.0` it crashes during signing.

Suspected cause: https://github.com/electron-userland/electron-builder/pull/8890
Seems that the `plist` library handles nullish values differently than the previous implementation.  

:wrench:  Proposed fix in https://github.com/electron-userland/electron-builder/pull/9481

Lemonexe:
Explicitly delete nullish values from macOS `appPlist`, because it seems that the `plist` library handles nullish values differently than the previous implementation (before https://github.com/electron-userland/electron-builder/pull/8890)

Resolves https://github.com/electron-userland/electron-builder/issues/9480

I confirmed this fix with local patch on a real app built on macOS, and added test coverage.
The test passes as is, and fails if I comment out the code change :heavy_check_mark:

### DIFF
--- a/.changeset/fifty-laws-jog.md
+++ b/.changeset/fifty-laws-jog.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: delete default macOS appPlist keys using extendInfo when set Nullish
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -582,6 +582,9 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
     if (extendInfo != null) {
       Object.assign(appPlist, extendInfo)
     }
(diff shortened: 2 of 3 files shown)


mmaietta (approved):

mmaietta:
Nice work! Thank you for your contribution!

PR #9486, opened 2026-01-07
feat: add writeUpdateInfo option for zip builds

### LINKED ISSUE #9479: Add support for disabling blockmap generation for macOS ZIP files via `zip.writeUpdateInfo` option
## Problem

Currently, there's no way to disable blockmap (`.zip.blockmap`) file generation for macOS ZIP targets, even though this option exists for DMG targets via `dmg.writeUpdateInfo: false`. Blockmap generation can be time-consuming for large applications, and users should have the option to disable it.

## Use Case

I need to build both `.pkg` and `.zip` files for macOS (zip is required for auto-update), but I want to skip the `.zip.blockmap` file generation to speed up the build process. The blockmap is used for differential updates, but if build time is more important than update efficiency, users should be able to opt out.

## Current Behavior

- ZIP files always generate `.zip.blockmap` files on macOS
- DMG files support `dmg.writeUpdateInfo: false` to disable blockmap
- No equivalent option exists for ZIP files

## Desired Behavior

Add support for `zip.writeUpdateInfo: false` configuration option, similar to how `dmg.writeUpdateInfo` works.

## Code References

1. **DMG implementation** (works correctly):
   - `packages/dmg-builder/src/dmg.ts:97` - Checks `this.options.writeUpdateInfo === false`
   - `packages/app-builder-lib/src/options/macOptions.ts:315` - `DmgOptions` interface includes `writeUpdateInfo?: boolean`

2. **ZIP implementation** (missing the option):
   - `packages/app-builder-lib/src/targets/ArchiveTarget.ts:73` - Only checks `this.isWriteUpdateInfo` (constructor parameter)
   - `packages/app-builder-lib/src/macPackager.ts:121` - Hardcodes `isWriteUpdateInfo = true` when creating zip target
   - `ArchiveTarget` reads options from `(this.packager.config as any)[this.name]` but doesn't check for `writeUpdateInfo`

## Suggested Solution

Modify `ArchiveTarget.ts` to check for `writeUpdateInfo` option from config, similar to how `DmgTarget` does it:
script
// In ArchiveTarget.ts build() method
const shouldWriteUpdateInfo = this.options.writeUpdateInfo === false ? false : this.isWriteUpdateInfo
if (shouldWriteUpdateInfo && format === "zip") {
  // ... create blockmap
}This would allow users to configure:
son
{
  "build": {
    "mac": {
      "target": ["pkg", "zip"],
      "zip": {
        "writeUpdateInfo": false
      }
    }
  }
}

## Additional Notes

- This should maintain backward compatibility (default behavior unchanged)
- The `.zip` file itself would still be generated (required for auto-update)
- Only the `.zip.blockmap` file generation would be skipped
- This matches the existing pattern used for DMG targets

beyondkmp:
fix #9479

Add configurable writeUpdateInfo option to control whether update info is written for macOS zip targets.


packages/app-builder-lib/src/options/macOptions.ts
```diff
@@ -236,6 +236,12 @@ export interface MacConfiguration extends PlatformSpecificBuildOptions {
    * For security reasons it is recommended to use the first option (see https://github.com/electron-userland/electron-builder/issues/7859)
    */
   readonly notarize?: boolean
+
+  /**
+   * @private
+   * @default true
+   */
+  writeUpdateInfo?: boolean
```
mmaietta: This applies to ALL mac targets, which isn't correct since that also includes `pkg` `mas` and `dir` targets, where this option would be irrelevant. I'm not sure I'm in favor of this approach

beyondkmp: updated. added zip options.

packages/app-builder-lib/src/configuration.ts
```diff
@@ -66,6 +66,10 @@ export interface CommonConfiguration {
    * macOS PKG options.
    */
   readonly pkg?: PkgOptions | null
+  /**
+   * macOS zip options.
+   */
+  readonly zip?: ZipOptions | null
```
mmaietta, line 72: I'm definitely not a fan of this tbh for a couple reasons:
- `zip` target is more than just `macOS` specific
- ArchiveTarget supports a plethora of archive formats. Creating a `zip` target opens the door for other archive types having their own options too. That's a very slippery slope on an already-complex Configuration object

We need a different approach to this that scales and keeps complexity low for the end-user.

Before making further edits to this PR, let's please discuss the approach first so that we can limit the back n forth and time spent on your side 🙂

mmaietta:
Closing stale PRs

PR #9491, opened 2026-01-10
fix: missing transitive dependencies

iamEvanYT:
#9492 should be merged before this one - they have merge conflicts that i need to fix

## Issue

Transitive dependencies (e.g., `sax` from `electron-updater` -> `builder-util-runtime` -> `sax`) were not being bundled when using Bun as the package manager.

The parent package appeared as a hoisted duplicate in npm list output.

## Fix
- Fixed `extractProductionDependencyGraph` to use the resolved tree from `allDependencies` for both iterating over dependencies AND checking `isProdDependency` when dealing with duplicate/hoisted packages
- Previously, it was checking `isProdDependency` against the original tree (which had empty `dependencies`), causing all transitive deps to be filtered out

### DIFF
--- a/.changeset/empty-goats-cover.md
+++ b/.changeset/empty-goats-cover.md
@@ -1,0 +1,5 @@
+---
+"dmg-builder": patch
+---
+
+chore: fix dmg background by updating vendor dmgbuild
--- a/.changeset/fifty-laws-jog.md
+++ b/.changeset/fifty-laws-jog.md
@@ -1,5 +1,0 @@
----
-"app-builder-lib": patch
----
-
-fix: delete default macOS appPlist keys using extendInfo when set Nullish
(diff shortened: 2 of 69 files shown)


mmaietta:
@iamEvanYT thank you! I've merged the other PR. I believe next steps are to resolve this PR's merge conflicts?

If so, please wait until https://github.com/electron-userland/electron-builder/pull/9440 is merged. It's extracting the manual filesystem/traversal logic to a fallback collector

mmaietta:
@iamEvanYT is this needed to get into `v26` branch or can it be punted to v27-alpha releases

iamEvanYT:
> @iamEvanYT is this needed to get into `v26` branch or can it be punted to v27-alpha releases

Nope, I don't think this PR is needed anymore. Just tested, builds works fine with `v26.4.1`. I'll close this PR.

PR #9495, opened 2026-01-10
fix: peer dependencies incorrectly being bundled into the app

### LINKED ISSUE #9494: [Regression] Electron build dependencies slipping into the bundled app
I was trying to upgrade from 26.1.0 -> 26.4.0 and this issue (which wasn't a issue before the new node modules collector) popped up.

My codebase uses bun's package manager.

I have `@ghostery/adblocker-electron` installed, which has a peer dependency to `electron`. With this new module collector, this caused the `electron` package to be bundled into `app.asar`, which causes the app size to double. This also happens to other electron build tools too. Somehow, the entire set of electron build dependencies slipped into my app bundle (which wouldn't be in there in v26.1.0)

Image

I've tested this with `electron-vite`'s [react-ts](https://github.com/alex8088/quick-start/tree/master/packages/create-electron/playground/react-ts) template, and I can confirm that once I removed those packages that has a peerDependency to `electron`, it stops bundling electron inside the app.

This wasn't an issue with v26.1.0.

Image

iamEvanYT:
fixes #9494 - #9492 & #9491 should be merged before this one

## Summary
Fixes an issue where peer dependencies (like `electron`) were incorrectly being bundled into `app.asar` when using Bun as the package manager, causing app size to roughly double.

## Problem
When a package like `@ghostery/adblocker-electron` declares `electron` as a peer dependency, the entire electron package (and its build dependencies) would get bundled into the final app. This regression was introduced with the new module collector and did not occur in v26.1.0.

## Cause
`BunNodeModulesCollector` had an override of `isProdDependency` that checked `tree.dependencies` instead of `tree._dependencies`. In npm's JSON output:

- `dependencies` contains all resolved dependencies (including installed peer deps)
- `_dependencies` contains only the declared dependencies from package.json

The parent class `NpmNodeModulesCollector` correctly uses `_dependencies`, but the Bun override was using `dependencies`, causing peer deps to slip through.

## Fix
Removed the incorrect override so BunNodeModulesCollector inherits the correct behavior from its parent class.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "prettier": "prettier 'packages/**/*.{ts,js}' test/src/**/*.ts --write",
     "///": "Please see https://github.com/electron-userland/electron-builder/blob/master/CONTRIBUTING.md#run-test-using-cli how to run particular test instead full (and very slow) run",
     "test-all": "pnpm compile && pnpm pretest && pnpm ci:test",
-    "test-linux": "docker run --rm -e CI=${CI:-true} -e DEBUG=${DEBUG:-} -e UPDATE_SNAPSHOT=${UPDATE_SNAPSHOT:-false} -e TEST_FILES=\"${TEST_FILES:-HoistedNodeModuleTest}\" -w /project -v $(pwd):/project -v $(pwd)-node-modules:/project/node_modules -v $HOME/Library/Caches/electron:/root/.cache/electron -v $HOME/Library/Caches/electron-builder:/root/.cache/electron-builder ${ADDITIONAL_DOCKER_ARGS} ${TEST_RUNNER_IMAGE_TAG:-electronuserland/builder:22-wine-mono} /bin/bash -c \"bash ./docker/test-in-docker.sh\"",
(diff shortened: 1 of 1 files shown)


mmaietta:
Looks like this PR might no longer be necessary due to class inheritance + recent refactor?

I just merged in the extraction of the TraversalNodeModuleCollector so now the inheritance is:
`BunNodeModuleCollector < TraversalNodeModuleCollector < NodeModuleCollector`. The base class has this already and there's no overrides.
```
  protected isProdDependency(depName: string, pkg: ProdDepType): boolean {
    const prodDeps = { ...pkg.dependencies, ...pkg.optionalDependencies }
    return prodDeps[depName] != null
  }
  ```

Can you take a look?

Wondering if the BunNodeModuleCollector still should leverage `npm list` CLI  instead of the manual parsing. (Note: the `npm list` dramatically reduced # of node_module's discovered, and I didn't want to edit a single test snapshot as part of the refactor)

**Comment by iamEvanYT** [comment]
@mmaietta Nope, all the issues I reported seems to be fixed! However, it doesn't seem work well with yalc. Once I linked `electron-updater` with yalc,  this error was thrown:
```
  • using manual traversal of node_modules to build dependency tree
  ⨯ production dependency not found  parent=electron-updater dependency=semver version=~7.7.3
  ⨯ Production dependency semver not found for package electron-updater  failedTask=build stackTrace=Error: Production dependency semver not found for package electron-updater
    at /Users/evan/Developer/MultiboxLabs/flow-browser-vite/.yalc/app-builder-lib/src/node-module-collector/traversalNodeModulesCollector.ts:121:15
```
Should I open a issue for this?

Also, `bun pm list` exist too if you wanna check that out. However, it doesn't support JSON output and does not seem as accurate so I think the manual parsing is the most effective.

mmaietta:
@iamEvanYT  Yes, please open a new issue for that. That's new then, but at least isolated only to manual traversal in this case. I thought I had handled the fallback usecase when semver didn't match, though. Might only be a problem isolated to yalc's changes, but I want to double check

I'd like to investigate this before sending out the next release if possible

Also in the new issue, can you also share what the console output of `bun pm list` looks like? Or share a sample repo I can test with? 😁

mmaietta:
Closing per comment https://github.com/electron-userland/electron-builder/pull/9495#issuecomment-3740878772

PR #9508, opened 2026-01-16
fix(mac): use correct MAS options in signApp() for Mac App Store builds. Closes #9507

### LINKED ISSUE #9507: MAS builds/signing fail with wrong provisioning profile due to `signApp()` ignoring MAS options
_I'm not an electron pro, but I have analyzed this bug as much as I possibly can and created a small repro over at https://github.com/abergs/eb-mas-bug-repro. I've used used claude extensively in this analysis, do let me know if that's against any guidelines (I looked but didn't find any)_

When building for Mac App Store (MAS), electron-builder's `signApp()` function always passes `null` for `masOptions`, causing it to use Darwin (Developer ID) settings instead of MAS settings. This results in:

1. **Wrong provisioning profile** being requested (`mac.provisioningProfile` instead of `mas.provisioningProfile`)
2. **Wrong certificate type** being searched for ("Developer ID Application" instead of "3rd Party Mac Developer Application")
3. **Build failures** in CI when the Darwin provisioning profile doesn't exist
4. **Ad-hoc signing fallback** when certificate lookup fails

## Environment

- electron-builder version: 26.? (confirmed in 26.4.0, and confirmed not to exist in 26.0.12)
- Platform: macOS
- Target: MAS (mas, mas-dev) - both single-arch and universal

## Current Behavior

The app gets signed **twice** during MAS builds:

1. **First signing (WRONG)**: `signApp()` → `sign(path, null, null, arch)` → `platform=darwin`
2. **Second signing (CORRECT)**: `pack()` → `sign(path, outDir, masBuildOptions, arch)` → `platform=mas`

The second signing would fix the first, but **if the first signing fails, the build stops before the second signing can run**.

### CI Failure Example

```
• signing file=dist/mas-universal/Bitwarden.app platform=darwin type=distribution
         identityName=Developer ID Application: 8bit Solutions LLC
         provisioningProfile=bitwarden_desktop_developer_id.provisionprofile
⨯ Command failed: security cms -D -i bitwarden_desktop_developer_id.provisionprofile
security: unable to open "bitwarden_desktop_developer_id.provisionprofile" for reading: No such file or directory
```

Notice:
- `platform=darwin` when it should be `platform=mas` (?)
- Uses `bitwarden_desktop_developer_id.provisionprofile` (from `mac` config) instead of `bitwarden_desktop_appstore.provisionprofile` (from `mas` config)

### The Bug Location

In `packages/app-builder-lib/src/macPackager.ts`, the `signApp()` method:

```typescript
async signApp(packContext: AfterPackContext, isAsar: boolean): Promise<boolean> {
  const readDirectoryAndSign = async (sourceDirectory: string, directories: string[], shouldSign: (file: string) => boolean) => {
    await Promise.all(directories.map(async (file) => {
      if (shouldSign(file)) {
        await this.sign(path.join(sourceDirectory, file), null, null, packContext.arch);
        //                                                ^^^^  ^^^^
        //                                            Always null for masOptions!
      }
    }));
    return true;
  };
  // ...
}
```

### Why This Happens

The `sign()` method determines `isMas` based on whether `masOptions` is null:

```typescript
async sign(appPath: string, outDir: string | null, masOptions: MasConfiguration | null, arch: Arch | null) {
  const isMas = masOptions != null;  // ← null means darwin, not MAS
  const options = masOptions == null ? this.platformSpecificBuildOptions : masOptions;
  // ...
}
```

When `masOptions` is `null`:
- `isMas = false`
- Uses `this.platformSpecificBuildOptions` (the `mac` config)
- Searches for "Developer ID Application" certificates
- Uses `mac.provisioningProfile`

### The Call Flow

```
pack() for MAS:
├─> doPack({ platformName: "mas", platformSpecificBuildOptions: masBuildOptions })
│   └─> doSignAfterPack()
│       └─> signApp(packContext)  // packContext.electronPlatformName = "mas"
│           └─> sign(path, null, null, arch)  // BUG: ignores electronPlatformName!
│               └─> isMas = false → platform=darwin → WRONG PROFILE
│
└─> sign(path, outDir, masBuildOptions, arch)  // Correct MAS signing
    └─> isMas = true → platform=mas → CORRECT PROFILE
```

The `packContext` object passed to `signApp()` contains `electronPlatformName: "mas"`, but `signApp()` completely ignores it.

## Expected Behavior

`signApp()` should respect the platform context and use MAS options when `packContext.electronPlatformName === "mas"`.

Only ONE signing should occur with the correct platform settings:
```
• signing file=dist/mas-arm64/App.app platform=mas type=distribution
         identityName=3rd Party Mac Developer Application: ...
         provisioningProfile=appstore.provisionprofile
```

## Suggested Fix

Option 1: Make `signApp()` MAS-aware:

```typescript
async signApp(packContext: AfterPackContext, isAsar: boolean): Promise<boolean> {
  const isMas = packContext.electronPlatformName === "mas";
  const masOptions = isMas
    ? deepAssign({}, this.platformSpecificBuildOptions, this.config.mas)
    : null;

  const readDirectoryAndSign = async (sourceDirectory: string, directories: string[], shouldSign: (file: string) => boolean) => {
    await Promise.all(directories.map(async (file) => {
      if (shouldSign(file)) {
        await this.sign(path.join(sourceDirectory, file), packContext.outDir, masOptions, packContext.arch);
      }
    }));
    return true;
  };
  // ...
}
```

Option 2: Skip main app signing in `signApp()` when `pack()` will sign it again:

```typescript
// In signApp(), skip the main app if it will be signed by pack() anyway
const appFileName = `${this.appInfo.productFilename}.app`;
const isMas = packContext.electronPlatformName === "mas";
if (!isMas) {
  // Only sign main app for non-MAS builds (MAS handled by pack())
  await readDirectoryAndSign(packContext.appOutDir, await readdir(packContext.appOutDir), file => file === appFileName);
}
```

## Reproduction

A minimal reproduction repository is available: https://github.com/abergs/eb-mas-bug-repro

### Quick Reproduction Steps

1. Clone the repo and install dependencies:
   ```bash
   git clone <repo-url>
   cd eb-bug
   npm install
   ```

2. Run the failing build:
   ```bash
   npm run build:mas-ci-fail
   ```

3. Observe the error:
   ```bash
   grep -E "• signing|unable to open" build-fail.log
   ```

### Configuration

The reproduction uses different provisioning profiles for `mac` vs `mas`:

```json
{
  "mac": {
    "provisioningProfile": "developer_id.provisionprofile"
  },
  "mas": {
    "provisioningProfile": "appstore.provisionprofile"
  }
}
```

When building for MAS, electron-builder incorrectly tries to use `developer_id.provisionprofile`.

## Impact

- **CI builds fail** when MAS-only provisioning profiles are available

## Workaround

Set both `mac.provisioningProfile` and `mac.identity` to empty string in CI when building for CI, seems to avoid the lookup for the provisioning profile but may(?) create a bad sign? Not sure.

```bash
electron-builder --mac mas -c.mac.provisioningProfile="" -c.mac.identity=""
```

## Related

- This seems to affect both single-arch and universal MAS builds
- The `afterSign` hook may also be affected when calling `context.packager.signApp()` for re-signing

abergs:
This PR tries to fix https://github.com/electron-userland/electron-builder/issues/9507

> Note: I've also put up a [competing PR](https://github.com/electron-userland/electron-builder/pull/9509) that solves the same problem differently. The competing fix also address the fact that we're signing MAS apps twice.

I've tried this code on the repro https://github.com/abergs/eb-mas-bug-repro and it worked, with output examples below.

When building for Mac App Store (MAS), the `signApp()` method always passed `null` for `masOptions` to `sign()`, causing it to use Darwin (Developer ID) settings instead of MAS settings. This resulted in:

1. Wrong provisioning profile being used (`mac.provisioningProfile` instead of `mas.provisioningProfile`)
2. Wrong certificate type being searched for ("Developer ID Application" instead of "3rd Party Mac Developer Application")
3. Build failures in CI when the Darwin provisioning profile doesn't exist
4. Incorrect `platform=darwin` instead of `platform=mas`

## Root Cause

In `packages/app-builder-lib/src/macPackager.ts`, the `signApp()` method was calling:

```typescript
await this.sign(path.join(sourceDirectory, file), null, null, packContext.arch)
```

The `sign()` method determines `isMas` based on whether `masOptions` is null:

```typescript
const isMas = masOptions != null  // null means darwin, not MAS
```

So passing `null` always resulted in `isMas = false`, even when `packContext.electronPlatformName === "mas"`.

## The Fix

Modified `signApp()` to detect MAS builds using `packContext.electronPlatformName` and construct the appropriate MAS options:

```typescript
const isMas = packContext.electronPlatformName === "mas"
let masOptions: MasConfiguration | null = null
if (isMas) {
  masOptions = deepAssign({}, this.platformSpecificBuildOptions, this.config.mas)
  const isMasDev = packContext.targets.some(t => t.name === "mas-dev")
  if (isMasDev) {
    deepAssign(masOptions, this.config.masDev, { type: "development" })
  }
}
```

## Test Results

Tested with a [reproduction project](https://github.com/abergs/eb-mas-bug-repro) using different provisioning profiles for `mac` vs `mas` config:

```json
{
  "mac": { "provisioningProfile": "developer_id.provisionprofile" },
  "mas": { "provisioningProfile": "appstore.provisionprofile" }
}
```

### Before Fix (v26.4.0)

**arm64:**
```
• signing file=dist/mas-arm64/App.app platform=darwin type=distribution
         provisioningProfile=developer_id.provisionprofile
```

**universal:**
```
• signing file=dist/mas-universal/App.app platform=darwin type=distribution
         provisioningProfile=developer_id.provisionprofile
```

### After Fix

**arm64:**
```
• signing file=dist/mas-arm64/App.app platform=mas type=distribution
         provisioningProfile=appstore.provisionprofile
```

**universal:**
```
• signing file=dist/mas-universal/App.app platform=mas type=distribution
         provisioningProfile=appstore.provisionprofile
```

## Affected Builds

- `mas` (single-arch and universal)
- `mas-dev` (single-arch and universal)

## Breaking Changes

None **known**. But I'm not an expert, merely a tourist.

### DIFF
--- a/packages/app-builder-lib/src/macPackager.ts
+++ b/packages/app-builder-lib/src/macPackager.ts
@@ -590,11 +590,25 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
   }
 
   protected async signApp(packContext: AfterPackContext, isAsar: boolean): Promise<boolean> {
+    // Determine if this is a MAS build and construct appropriate options
+    const isMas = packContext.electronPlatformName === "mas"
+    let masOptions: MasConfiguration | null = null
+    if (isMas) {
+      masOptions = deepAssign({}, this.platformSpecificBuildOptions, this.config.mas)
+      // Check if this is a mas-dev build
+      const isMasDev = packContext.targets.some(t => t.name === "mas-dev")
+      if (isMasDev) {
+        deepAssign(masOptions, this.config.masDev, {
+          type: "development",
+        })
+      }
+    }
+
     const readDirectoryAndSign = async (sourceDirectory: string, directories: string[], shouldSign: (file: string) => boolean): Promise<boolean> => {
       await Promise.all(
         directories.map(async (file: string) => {
           if (shouldSign(file)) {
-            await this.sign(path.join(sourceDirectory, file), null, null, packContext.arch)
+            await this.sign(path.join(sourceDirectory, file), packContext.outDir, masOptions, packContext.arch)
           }
         })
       )


packages/app-builder-lib/src/macPackager.ts
```diff
@@ -590,11 +590,25 @@ export class MacPackager extends PlatformPackager<MacConfiguration> {
   }
 
   protected async signApp(packContext: AfterPackContext, isAsar: boolean): Promise<boolean> {
+    // Determine if this is a MAS build and construct appropriate options
+    const isMas = packContext.electronPlatformName === "mas"
+    let masOptions: MasConfiguration | null = null
+    if (isMas) {
+      masOptions = deepAssign({}, this.platformSpecificBuildOptions, this.config.mas)
+      // Check if this is a mas-dev build
+      const isMasDev = packContext.targets.some(t => t.name === "mas-dev")
```
mmaietta, line 599: I think this might backfire since we're using `some`.

`signApp` is supposed to be per-target, so an invocation of it could be from a `t.name: mas`, then a 2nd invocation of it could be `t.name: mas-dev`. With this current logic, `type: development` would be set on both invocations.

mmaietta:
> Breaking Changes
None known. But I'm not an expert, merely a tourist.

Thanks for being an excellent tourist 🙂

packages/app-builder-lib/src/macPackager.ts
abergs: Even though I don't necessarily like this specific PR over #9509, but would it be more appropriate to use `packContext.electronPlatformName`? Tried to grep through and I don't think `electronPlatforname` necesarily support `mas-dev`, only `mas`.

Actually. looking at the code again it seems `siSignAfterPack` also has a `platformSpecificBuildOptions` argument with a `.target`. Perhaps we could feed that through to `signApp`

mmaietta:
I'm going to pick this up in the larger logic split in https://github.com/electron-userland/electron-builder/pull/9567

Issue #9511, opened 2026-01-16
Add support for dmgbuild badge_icon


t3chguy:
As using the app icon for the DMG can be confusing when a disk mount has the same icon as the application, and overlaying an app's icon over a macOS disk doesn't age well when the OS updates.

https://github.com/electron-userland/electron-builder/blob/master/packages/dmg-builder/vendor/dmgbuild/core.py#L98

mmaietta:
@t3chguy I'm actually trying to remove that vendor dir and get its patches pushed upstream to dmgbuild

Simultaneously, I'm working on a new  portable dmgbuild+python bundle as an on-demand toolset from electron-builder-binaries. This is because the updated dmgbuild requires a much newer python than my m2 was already using. I don't want to force devs to change their build env, so I feel a portable bundle is best.
PR https://github.com/electron-userland/electron-builder-binaries/pull/118

Either way, we'll need changes pushed to one of those two sources 🙂

mmaietta:
OOPS. Looks like I forgot to hook it up to an actual config property...it's already there in the interface 😂
https://github.com/electron-userland/electron-builder/blob/6c20eeb1cf9fd10980cde3c9ce0602fa6b7c6972/packages/dmg-builder/src/dmg.ts#L13-L16

Alrighty, got a PR up for ya! https://github.com/electron-userland/electron-builder/pull/9513

t3chguy:
@mmaietta thanks for jumping on that so quickly but it doesn't seem to work.

1. icon & badge-icon are mutually exclusive in dmgbuild, annoyingly with `icon` taking priority. See https://github.com/dmgbuild/dmgbuild/blob/main/src/dmgbuild/core.py#L560-L569 - cannot skip dmg.icon due to `  ⨯ dmg.icon cannot be specified as empty string` error on e-b side.
2. I think the docstring is misleading in https://github.com/electron-userland/electron-builder/pull/9513/changes#diff-6acd25e0ee586dccfb915eb279645e3fb9186b938b131f12f34ecb830c698868R254-R256
> Defaults to the application icon (`build/icon.icns`).

I don't think this is true, by default the badge-icon system seems to be not in use, and instead the "explicit" icon defaults to `build/icon.icns`

3-ish. Untested given 1. is preventing me from trying it - https://dmgbuild.readthedocs.io/en/latest/settings.html#badge_icon states dmgbuild must be installed with the `badge_icons` extra

[...]

PR #9516, opened 2026-01-18
feat: update `dmgbuild` and migrate to portable `dmgbuild` python bundle

mmaietta:
Fixes support for https://github.com/electron-userland/electron-builder/issues/9511

Context:
The source code of `dmgbuild` pip package is stored directly in the repo within `dmg-builder` `vendor` directory. Unfortunately, this means that a minimum version of Python must already be installed. This is problematic because `dmgbuild` has been updated, and thus, its minimum Python version has also changed.

New route:
Using the same approach as `electron-builder` already does with on-demand toolsets, `dmgbuild` will be packaged into a dedicated/portable Python bundle, served from the electron-builder-binaries GH repo. (This is already the approach of `fpm` portable Ruby bundle, and AppImage runtimes w/ mksquashfs,etc). This portable `dmgbuild` will be assembled on GH runner, receive GH attestation, and then uploaded as a dedicated release during the GH CD flow. `electron-builder` then downloads from this release and validates checksums against hardcoded strings stored in electron-builder in the pursuit of closed-loop tooling.

### DIFF
--- a/.changeset/fuzzy-flowers-fix.md
+++ b/.changeset/fuzzy-flowers-fix.md
@@ -1,0 +1,6 @@
+---
+"app-builder-lib": patch
+"dmg-builder": patch
+---
+
+feat: update `dmgbuild` and migrate to portable `dmgbuild` python bundle. Fixes `badge-icon` property


packages/app-builder-lib/package.json
```diff
@@ -73,8 +74,9 @@
     "js-yaml": "^4.1.0",
     "json5": "^2.2.3",
     "lazy-val": "^1.0.5",
-    "minimatch": "^10.0.3",
+  "minimatch": "^10.0.3",
```
t3chguy: This indentation tweak seems unexpected

t3chguy:
Tested using 26.6.0 and it works just as well as it did with my patch (after removing the patch)

image

Thanks!

PR #9517, opened 2026-01-18
feat: add beta support for snap `core24`

### LINKED ISSUE #5626: build arm64 snap on x64 machine
* **Version**: 22.10.4
* **Electron Version**: 11.2.3
* **Electron Type**: current
* **Target**: snap arm64

when I try to build an arm64 snap on my laptop (ubuntu 20.04 x64) and github actions (ubuntu 18.04 x64) I get the below error
```
⨯ snapcraft does not currently support building arm64 on amd64  
  ⨯ /home/labhansh/hyper/node_modules/app-builder-bin/linux/x64/app-builder exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE  stackTrace=
Error: /home/labhansh/hyper/node_modules/app-builder-bin/linux/x64/app-builder exited with code ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
at ChildProcess.<anonymous> (/home/labhansh/hyper/node_modules/builder-util/src/util.ts:243:14)
at Object.onceWrapper (node:events:483:26)
at ChildProcess.emit (node:events:376:20)
at maybeClose (node:internal/child_process:1063:16)
at Process.ChildProcess._handle.onexit (node:internal/child_process:295:5)
error Command failed with exit code 1.
```
Is there any way I can build arm64 snap on x64 machine or ci (also x64 only)

### LINKED ISSUE #6765: Snapstore publish method deprecated
17.3.0

23.0.2

* **Node Version**: 
v14.x

* **Electron Version**:
v17.3.0
current

* **Target**: 
Linux - Snap

I am releasing my linux app to snapstore and the electron-builder command shows this deprecations:
```
DEPRECATED: The 'push' set of commands have been replaced with 'upload'.
See http://snapcraft.io/docs/deprecation-notices/dn11 for more information.
```

### LINKED ISSUE #8880: The 'snap' command is deprecated, use 'pack' instead.
app-builder-lib needs to be updated to use the `snapcraft pack` command instead of `snap` when packaging for Snap targets

### LINKED ISSUE #9452: Snap package of Electron ≥ 38 crashes at startup under GNOME on Wayland
Electron ≥ 38.2.0 enables Wayland by default (electron/electron#48001, electron/electron#48309), and a Snap package generated by `electron-builder@next` allegedly supports this (#9337), but instead it crashes at startup.

## Minimal reproduction

In an empty directory, create `package.json`:

```json
{
  "name": "electron-builder-snap-test",
  "version": "0.0.0",
  "devDependencies": {
    "electron": "^39.2.7",
    "electron-builder": "^26.3.6"
  },
  "build": {
    "files": [
      "index.js"
    ],
    "linux": {
      "target": [
        "snap"
      ]
    }
  }
}
```

and `index.js`:

```js
const { app, BrowserWindow } = require("electron");
app.whenReady().then(() => {
  new BrowserWindow().loadURL("data:text/html,<h1>Hello, world!</h1>");
});
```

Run `npm install; npx electron-builder` to build `dist/electron-builder-snap-test_0.0.0_amd64.snap`.

On a freshly booted Ubuntu 25.10 live CD (which runs GNOME on Wayland by default), run `snap install --dangerous electron-builder-snap-test_0.0.0_amd64.snap; electron-builder-snap-test`. It crashes at startup with these errors:

```console
$ electron-builder-snap-test
MESA-LOADER: failed to open virtio_gpu (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: virtio_gpu
MESA-LOADER: failed to open kms_swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load swrast driver
[12774:1216/083302.735887:ERROR:ui/gtk/gtk_ui.cc:257] Schema org.gnome.desktop.interface does not have key font-antialiasing
MESA-LOADER: failed to open virtio_gpu (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: virtio_gpu
MESA-LOADER: failed to open kms_swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load swrast driver
[12851:1216/083302.760874:ERROR:ui/gl/angle_platform_impl.cc:42] Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
ERR: Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
[12851:1216/083302.761183:ERROR:ui/gl/egl_util.cc:92] EGL Driver message (Critical) eglInitialize: Failed to get system egl display
[12851:1216/083302.761377:ERROR:ui/gl/gl_display.cc:639] eglInitialize OpenGL failed with error EGL_NOT_INITIALIZED, trying next display type
[12851:0100/000000.761644:ERROR:ui/gl/angle_platform_impl.cc:42] Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
Segmentation fault (core dumped)
ERR: Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
[12851:0100/000000.934108:ERROR:ui/gl/egl_util.cc:92] EGL Driver message (Critical) eglInitialize: Failed to get system egl display
[12851:0100/000000.934867:ERROR:ui/gl/gl_display.cc:639] eglInitialize OpenGLES failed with error EGL_NOT_INITIALIZED
[12851:0100/000000.934968:ERROR:ui/gl/gl_display.cc:674] Initialization of all EGL display types failed.
[12851:0100/000000.935010:ERROR:ui/ozone/common/gl_ozone_egl.cc:26] GLDisplayEGL::Initialize failed.
[12851:0100/000000.936843:ERROR:ui/gl/angle_platform_impl.cc:42] Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
ERR: Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
[12851:0100/000000.936966:ERROR:ui/gl/egl_util.cc:92] EGL Driver message (Critical) eglInitialize: Failed to get system egl display
[12851:0100/000000.936999:ERROR:ui/gl/gl_display.cc:639] eglInitialize OpenGL failed with error EGL_NOT_INITIALIZED, trying next display type
[12851:0100/000000.937041:ERROR:ui/gl/angle_platform_impl.cc:42] Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
ERR: Display.cpp:1093 (initialize): ANGLE Display::initialize error 12289: Failed to get system egl display
[12851:0100/000000.937065:ERROR:ui/gl/egl_util.cc:92] EGL Driver message (Critical) eglInitialize: Failed to get system egl display
[12851:0100/000000.937080:ERROR:ui/gl/gl_display.cc:639] eglInitialize OpenGLES failed with error EGL_NOT_INITIALIZED
[12851:0100/000000.937093:ERROR:ui/gl/gl_display.cc:674] Initialization of all EGL display types failed.
[12851:0100/000000.937122:ERROR:ui/ozone/common/gl_ozone_egl.cc:26] GLDisplayEGL::Initialize failed.
```

With `--disable-gpu` it prints fewer errors but still crashes:

```console
$ electron-builder-snap-test --disable-gpu
MESA-LOADER: failed to open virtio_gpu (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: virtio_gpu
MESA-LOADER: failed to open kms_swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load swrast driver
[13200:1216/083542.141433:ERROR:ui/gtk/gtk_ui.cc:257] Schema org.gnome.desktop.interface does not have key font-antialiasing
MESA-LOADER: failed to open virtio_gpu (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: virtio_gpu
MESA-LOADER: failed to open kms_swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast (search paths /snap/electron-builder-snap-test/x6/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load swrast driver
Segmentation fault (core dumped)
```

With `XDG_SESSION_TYPE=x11` or `--ozone-platform=x11` it opens successfully (but obviously without Wayland goodness such as fractional scaling).

### LINKED ISSUE #9704: Classic snap launcher script is broken on Ubuntu 26.04 LTS
Hi there

I recently upgraded to Ubuntu 26.04 LTS, and the Obsidian snap stopped working. After some investigation with colleagues, it seems this is due to a divergent behaviour in `rust-coreutils` and a (potentially incorrect?) script that is packaged in the snap by Electron builder.

The issue in this snap is `/snap/obsidian/current/desktop-common.sh`:

```
function can_open_file() {
  head -c0 "$1" &> /dev/null
}
...
if can_open_file "$REALHOME/.config/user-dirs.dirs" && can_open_file "$REALHOME/.config/user-dirs.locale"; then
...
# Create links for user-dirs.dirs
if [ $needs_xdg_links = true ]; then
  for ((i = 0; i < ${#XDG_SPECIAL_DIRS_PATHS[@]}; i++)); do
    b="$(realpath "${XDG_SPECIAL_DIRS_PATHS[$i]}" --relative-to="$HOME")"
    if [ -e "$REALHOME/$b" ]; then
      if [ -d "$HOME/$b" ]; then
        rmdir "$HOME/$b" 2> /dev/null
      fi
...
```

Basically trying to replace XDG directories for the snap by symlink to the real dir. In that case it’s a classic snap so it shouldn’t try to do that but

```
$ gnuhead -c0 ~/Documents; echo $?
0
$ head -c0 ~/Documents; echo $?
head: error reading '/home/ubuntu/Documents': Is a directory
1
```

The head behavior is different in rust-coreutils which makes the check it does to decide if it needs to do the symlink dance or not fail and try to rmdir ~/Documents and errors out

That launcher script appears to be included as part of electron builder, and hasn’t been updated for some time. 

The script seems to be in the [binary release tarball](https://github.com/electron-userland/electron-builder-binaries/releases/download/snap-template-4.0-2/snap-template-electron-4.0-2-amd64.tar.7z) but I can't find it in the source anywhere to propose a patch.

I suspect there are more snaps affected than just Obsidian

(Reference to the Obsidian forum post: https://forum.obsidian.md/t/obsidian-snap-doesnt-start-open-with-ubuntu-26-04/113569/28)

mmaietta:
Adds first-class core24 snap support through a new `snapcraft` config key, restructures the snap target into per-core classes, and modernizes the Snap Store publisher to use the native `snapcraft` CLI directly.

The old `snap` key still works exactly as before — nothing breaks for existing users. The new `snapcraft` key gives you structured access to core18/20/22 (same behavior as before) and core24 (all-new).

---

### New config shape

```jsonc
// electron-builder.json
{
  "snapcraft": {
    "base": "core24",
    "core24": {
      "useDestructiveMode": true,   // or useLXD / useMultipass / remoteBuild
      "extensions": ["gnome"],       // note: already defaults to ["gnome"] unless destructive mode
      "stagePackages": ["default", "libdrm2"],
      "forceX11": true
    }
  }
}
```

For legacy bases:

```jsonc
{
  "snapcraft": {
    "base": "core22",
    "core22": {
      "confinement": "classic",
      "stagePackages": ["default", "libsecret-1-0"]
    }
  }
}
```

Pass-through mode for teams that manage their own `snapcraft.yaml` or supply one directly:

```jsonc
{
  "snapcraft": {
    "base": "custom",
    "custom": { "yaml": "build/my-snapcraft.yaml" }
  }
}
```

---

### Snap Store publishing

The `SnapStorePublisher` has been rewritten to use `snapcraft` directly rather than the `app-builder` helper binary.

**Authentication — `SNAP_CSC_LINK` / `cscLink`**

Generate credentials once with:

```sh
snapcraft export-login -
```

Then supply the base64 output (or a path to a file containing it) via:

- Environment variable: `SNAP_CSC_LINK=<base64-or-path>`
- Config key: `snapStoreOptions.cscLink: "<base64-or-path>"`

Credentials are injected as `SNAPCRAFT_STORE_CREDENTIALS` into the `snapcraft` subprocess — they are never written to disk or exposed in the Node.js process environment. The pattern mirrors `WIN_CSC_LINK` / `CSC_LINK` used for code-signing on other platforms.

**Version requirement**

`snapcraft` ≥ 7.0.0 is required (`SNAPCRAFT_STORE_CREDENTIALS` was introduced in 7). If `snapcraft` is missing or too old, the publisher throws an actionable error.

**Command change**

| Before | After |
|--------|-------|
| `app-builder publish-snap -f <file> -c <channel>` | `snapcraft upload <file> --release <channels>` |

**`loadCscLink` utility (new, in `builder-util`)**

A new `packages/builder-util/src/cscLink.ts` module centralizes credential-link resolution so it can be shared across publishers. It handles:

- Base64 inline values (detected by `data:...;base64,` prefix, length > 2048, or trailing `=`)
- `file://` URIs
- `~/` home-directory expansion
- Relative paths resolved against `resourcesDir`
- Absolute paths

---

### core24 specifics

- **Default build uses the `gnome` extension** — handles content snaps, GTK theming, and GPU acceleration. Explicitly set `extensions: []` to opt out.
- **`gnome` + `--destructive-mode` are mutually exclusive.** Setting `useDestructiveMode: true` while `gnome` is in `extensions` throws an `InvalidConfigurationError` with a clear message explaining why and what to do instead.
- **`organize` mappings are auto-generated** at build time from the actual app directory, so helper binaries land under `app/` inside the snap regardless of what Electron ships.
- **Wayland/X11 control** — `forceX11: true` injects `--ozone-platform=x11`. Default is Wayland-allowed (replaces the legacy `allowNativeWayland` flag which had inverted semantics).
- **Remote build support** (Launchpad) for building without native hardware, via `remoteBuild: { enabled: true, buildFor: "amd64" }`. Due to electron-builder's target-to-arch-to-artifact logic paths, only one `buildFor` arch may be provided at a time.

---

### Implementation changes

The old monolithic `snap.ts` (396 lines) is gone. In its place:

| File | Purpose |
|------|---------|
| `snap/SnapTarget.ts` | `SnapTarget` (build entry point) + abstract `SnapCore<T>` base class |
| `snap/coreLegacy.ts` | core18/20/22 — identical behavior to the old `snap.ts` |
| `snap/core24.ts` | core24 — GNOME extension path, auto-generated organize mappings, content snap plugs |
| `snap/coreCustom.ts` | Pass-through for a custom `snapcraft.yaml` |
| `snap/snapcraftBuilder.ts` | snapcraft CLI invocation (pack, remote-build, LXD, Multipass, destructive mode) |
| `snap/snapcraft.d.ts` | TypeScript types for the full `snapcraft.yaml` schema |
| `builder-util/src/cscLink.ts` | Shared CSC link resolution (base64, file:// , ~/, relative, absolute) |
| `electron-publish/src/snapStorePublisher.ts` | Rewritten publisher using `snapcraft upload` + `SNAPCRAFT_STORE_CREDENTIALS` |

`LinuxTargetHelper.getSnapCore()` reads `snapcraft.base` and returns the right implementation. Record assembly in the YAML builder uses `deepAssign` and `isValidKey` throughout for safe, predictable merging.

---

### Tests

**Publisher tests** (`test/src/publisher/snap/SnapStorePublisherTest.ts`, 36+ cases):

- Version string parsing across all snapcraft output formats (`snapcraft X.Y.Z`, `snapcraft, version X.Y.Z`, edge channel)
- Channel handling — null/undefined defaults, string split, array passthrough, complex track/branch formats
- Credential injection — base64 inline, file path, `SNAP_CSC_LINK` env var priority, missing/empty errors
- Error cases — missing snapcraft binary, outdated version, unresolvable credentials

**Snapcraft build tests:**

A new `test-snap (core*)` CI matrix replaces the previous `test-linux-native` stub (which was `if: false`). Four parallel jobs run, one per core:

| Job | Image | Mode |
|-----|-------|------|
| `test-snap (core18)` | `ghcr.io/canonical/snapcraft:7_core22` | Docker, destructive mode |
| `test-snap (core20)` | `ghcr.io/canonical/snapcraft:7_core22` | Docker, destructive mode |
| `test-snap (core22)` | `ghcr.io/canonical/snapcraft:7_core22` | Docker, destructive mode |
| `test-snap (core24)` | `ghcr.io/canonical/snapcraft:8_core24` | Docker, destructive mode |

The `snapHeavyTest.ts` suite has install + launch tests (extract with `unsquashfs`, assert snap structure, run binary with `--version`) that run on native Linux CI runners with `squashfs-tools` available and are skipped inside Docker.

The existing `snapTest.ts` unit tests still run and cover all descriptor-level assertions (plugs, slots, extensions, compression, etc.) through snapshot testing.

---

### Notable changes

**None** for `snap` users. The `snapcraft` key is purely additive.

**For Snap Store publishing:** `SnapStorePublisher` now calls `snapcraft` directly instead of `app-builder`. If you were injecting credentials in a non-standard way (note: `snapcraft login` is implicitly supported), you would need to switch to `SNAP_CSC_LINK` / `cscLink` . Snapcraft ≥ 7.0.0 is required (current stable is 8.x).


kenvandine:
I am really happy to see this being worked on, thanks so much for this.

I would like to see it made clear that destructive mode should be discouraged, but available in cases where you can't use lxd or mulitpass. For example, destructive mode should never be used when building locally and in general it's difficult to trust the resulting build even in a docker container.

mmaietta:
@kenvandine Great callout.  I wonder on what's the best way to approach this then.

Practically/ideally, we should have a destructive-mode available for the functional tests that can be run from other OS's => docker container. The e2e test is run on a native linux runner, so maybe we could test a configuration matrix?

> destructive mode should never be used when building locally, and in general, it's difficult to trust the resulting build even in a docker container.

Definitely should call that out with the tsdoc guidance w.r.t. destructive mode though.  That being said, we could also theoretically mark it as `@private` to prevent it from being documented or completely remove the field/property entirely - leaving behind a skeleton to still be able to run my test suite in the Docker container.

kenvandine:
I think leaving the option is useful, just document is as not recommended. There are cases where people need it, but I think that's pretty rare.

jnsgruk:
This is looking much better as a set of defaults.

JGCarroll:
> Wayland is enabled by default for core24. Set allowNativeWayland: false to disable (sets DISABLE_WAYLAND=1).

This doesn't sound ideal IMO. I'd personally suggest removing that variable entirely leaving Wayland implicitly available at the Snap packaging level with configuring it left at the Electron level.

In Electron 39 for example which defaults to Wayland, it does so by checking the `XDG_SESSION_TYPE` variable, if set to Wayland, it will look up the Wayland socket and try use it. Should this fail, then it won't fall back to X11/XWayland, it'll simply exit unconditionally. The `DISABLE_WAYLAND` variable just breaks the Wayland socket in strict snaps by not symlinking it where the app expects, it doesn't by itself tell the app to use X11 and relies on them having a graceful fallback that Electron doesn't have. It potentially has no effect on classic snaps either (assuming they don't rewrite `$XDG_RUNTIME_DIR`)

There's very little to gain from ever using that variable as Wayland is generally considered a secure protocol anyway, I personally think it makes sense to just ignore that variable existing and rely on either the ozone platform hint flags for 38 and below, and the XDG_SESSION_TYPE var for 39 and above.

mmaietta:
Oooo great callout @JGCarroll; I'm fully open to removing that from the core24 logic path unless someone else chimes in with some different thoughts.

I'm still ironing out some kinks from my side (things went a little haywire in merge conflicts this morning), but would you be willing to review this PR when it's ready? I could use some additional coverage validating across various linux setups that differ from my own VMs setup. If so, I can ping/@ mention you when this is ready.

JGCarroll:
In theory I'm happy to help, in practise my only computer broke recently and I'm going without the ability to get much done for a few months until I can afford RAM 😂.

That said let me know anyway, I've been half getting by with Android Terminal and Google Cloud Shell on my phone, which aren't exactly conventional build environments but I'll see what I can do. It's possible with the Android 17 update next month my Pixel will be able to run snaps directly again which would certainly help in the shorter term (It briefly did on Android 16 from what I remember before some kernel config changed in the Debian VM).

What might be helpful for you is letting you know that I've a got us a strictly confined Core 24 Electron app deployed already, 'joplin-desktop', which effectively does nothing other than hijack the Electron-Builder AppImage process and then pack it, Ken will be happy knowing it doesn't come with a single repository package either 😉, absolutely everything external to Electron is provided implicitly by merely having the Gnome extension defined, which means I think it's functionally about perfect for disecting for someone like yourself to see what actually ends up in the final `snap.yaml` when it comes to layouts and etc.

(Ideally I'd rebuild it with Snapcraft 9 as I believe there's a very small tweak in the layouts on that version, otherwise the latest package on the store was built last week).

andersk:
I wrote this little test script to build, install, and run a minimal application using this PR:

https://gist.github.com/andersk/6a8968896ab24b380b3039df6e09ba4f

You can try running it on your favorite Ubuntu live CD. Warning: it needs a lot of RAM; I’m using a 32 GB VM.

On Ubuntu 26.04 I get this crash:

```
+ sudo snap install --dangerous /home/ubuntu/electron-builder-snap-test/dist/electron-builder-snap-test_0.0.0_amd64.snap
electron-builder-snap-test 0.0.0 installed
+ electron-builder-snap-test
update.go:85: cannot change mount namespace according to change mount (/run/user/1000/doc/by-app/snap.electron-builder-snap-test /run/user/1000/doc none bind,rw,x-snapd.ignore-missing 0 0): cannot inspect "/run/user/1000/doc": lstat /run/user/1000/doc: permission denied
libpxbackend-1.0.so: cannot open shared object file: No such file or directory
Failed to load module: /home/ubuntu/snap/electron-builder-snap-test/common/.cache/gio-modules/libgiolibproxy.so
[24316:0517/004921.301130:FATAL:sandbox/linux/services/credentials.cc:131] Check failed: . : Permission denied (13)
build.sh: line 92: 24316 Trace/breakpoint trap      (core dumped) electron-builder-snap-test
```

mmaietta:
Thank you @andersk ! That'll be perfect for when I test out native Linux snap builds. Currently, I was just building on Mac and dragging+dropping into a Linux VM. Native snap build on a Linux VM was to be the next e2e validation step.

I'm hitting a similar issue with hardware limitations though.... My M2 Max macbook is maxing CPU, 32gb of RAM + 20gb swap apparently used, and the problem is I have 26gb of free disk space BEFORE the swap gets taken. Parallels VMs half the time refuse to boot up or simply can't even be suspended. 😞  Trying to figure out how to resolve the hardware limitations locally, otherwise I'm stuck with purely CI builds via a GH repo - nonideal and slow for sure.

andersk:
I built the same minimal application on macOS with `"useMultipass": true`, copied the snap to an Ubuntu 26.04 VM, and got the same crash. Am I doing something different than you?

andersk:
Inspired by that [joplin-desktop snap](https://github.com/JGCarroll/joplin-desktop-snap/blob/master/snap/snapcraft.yaml), I was able to find two workarounds to get the minimal application to launch:

1. Run with `electron-builder-snap-test --disable-dev-shm-usage --no-sandbox`; or

2. Run with `electron-builder-snap-test --no-sandbox` after building with this configuration:

   ```json
   "snapcraft": {
     "base": "core24",
     "core24": {
       "plugs": [
         {
           "private-shmem": {
             "interface": "shared-memory",
             "private": true
           }
         },
         "default"
       ]
     }
   }
   ```

Maybe a `private-shmem` like that should be a default?

I tried to add an automatic `--no-sandbox` using `"core24": { "executableArgs": ["--no-sandbox"] }`, but it had no effect; is there some other way to do that?

JGCarroll:
I don't think there's much to lose by enforcing a per snap `/dev/shm/` folder by default in Electrons use case, however, the `browser-support` plug is meant to make Electron apps have sufficient access to `/dev/shm/` by itself, in theory you shouldn't have to have it be private even though I'm a fan of doing it for good measure anyway.

It would be worth running `snappy-debug` in a second terminal, installed via `sudo snap install snappy-debug` which itself tries to read the system logs in real time and give actionable advice about the sandbox. With it you can then compare the denials in that folder vs what's granted here https://github.com/canonical/snapd/blob/master/interfaces/builtin/browser_support.go

Ideally then the above interface would be expanded to to fix those denials too, but in practise it could be best to have both working, as the intention of the sandbox in general is much closer to what the private folder does than what the `browser-support` interface does (since it technically gives a small increase in sandbox access being able to read other browsers shared memory).

andersk:
With no arguments:

```console
$ electron-builder-snap-test 
[9403:0517/190649.093713:FATAL:sandbox/linux/services/credentials.cc:131] Check failed: . : Permission denied (13)
Trace/breakpoint trap      (core dumped) electron-builder-snap-test
```

```console
$ sudo snappy-debug
…
= AppArmor =
Time: 2026-05-17T19:0
Log: apparmor="DENIED" operation="userns_create" class="namespace" profile="snap.electron-builder-snap-test.electron-builder-snap-test" pid=9403 comm="electron-builde" requested="userns_create" denied="userns_create" execpath="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test"
…
```

With `--no-sandbox`:

```console
$ electron-builder-snap-test --no-sandbox
Gtk-Message: 19:08:30.494: Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.
[9729:0517/190830.666252:ERROR:net/base/address_tracker_linux.cc:224] Could not create NETLINK socket: Operation not permitted (1)
[9729:0517/190830.666496:ERROR:base/memory/platform_shared_memory_region_posix.cc:213] Creating shared memory in /dev/shm/.org.chromium.Chromium.DtpzPT failed: Permission denied (13)
Trace/breakpoint trap      (core dumped) electron-builder-snap-test --no-sandbox
```

<details>
<summary>snappy-debug output</summary>

```console
$ sudo snappy-debug
…
= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9794 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 330(pkey_alloc) compat=0 ip=0x7924dd90032b code=0x50000
Syscall: pkey_alloc

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9795 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 330(pkey_alloc) compat=0 ip=0x78b6385ec32b code=0x50000
Syscall: pkey_alloc

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9729 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7c7a355f265b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9729 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 425(io_uring_setup) compat=0 ip=0x7c7a355fc28d code=0x50000
Syscall: io_uring_setup

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9729 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 203(sched_setaffinity) compat=0 ip=0x7c7a355786f1 code=0x50000
Syscall: sched_setaffinity
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional sched_setaffinity is often just noise)

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9729 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7c7a355f265b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9729 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 425(io_uring_setup) compat=0 ip=0x7c7a355fc28d code=0x50000
Syscall: io_uring_setup

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9729 comm="DelayedTaskSche" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 425(io_uring_setup) compat=0 ip=0x7c7a355fc28d code=0x50000
Syscall: io_uring_setup

= Seccomp =
Time: 2026-05-17T19:0
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=9729 comm="ThreadPoolForeg" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 41(socket) compat=0 ip=0x7c7a3560131b code=0x50000
Syscall: socket
Suggestions:
* add account-control (if using NETLINK_AUDIT)
* add audio-playback (if using NETLINK_KOBJECT_UEVENT)
* add bluetooth-control (if using AF_{ALG,BLUETOOTH})
* add firewall-control (if using NETLINK_{FIREWALL,IP6_FW,NETFILTER,NF_LOG,ROUTE})
* add hardware-observe (if using NETLINK_{GENERIC,KOBJECT_UEVENT})
* add netlink-audit (if using NETLINK_AUDIT)
* add netlink-connector (if using NETLINK_CONNECTOR)
* add network (if using AF_INET{,6}, AF_CONN, NETLINK_ROUTE)
* add network-bind (if using AF_INET{,6}, NETLINK_ROUTE)
* add network-control (if using AF_{APPLETALK,BRIDGE,INET,INET6,IPX,PACKET,PPPOX,SNA}, NETLINK_{DNRTMSG,FIB_LOOKUP,GENERIC,INET_DIAG,ISCSI,KOBJECT_UEVENT,RDMA,ROUTE,XFRM})
* add network-observe (if using SOCK_RAW, AF_INET{,6}), NETLINK_{GENERIC,INET_DIAG,KOBJECT_UEVENT,ROUTE})
* add raw-usb (if using NETLINK_KOBJECT_UEVENT)
* add time-control (if using NETLINK_AUDIT)
* add unity7 (if using NETLINK_KOBJECT_UEVENT)
* add x11 (if using NETLINK_KOBJECT_UEVENT)

= AppArmor =
Time: 2026-05-17T19:0
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/host.conf" pid=9729 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/host.conf (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:0
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/run/systemd/resolve/stub-resolv.conf" pid=9729 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=989
File: /run/systemd/resolve/stub-resolv.conf (read)
Suggestions:
* adjust program to use $SNAP_DATA
* adjust program to use /run/shm/snap.$SNAP_NAME.*
* adjust program to use /run/snap.$SNAP_NAME.*
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add one of 'firewall-control, fwupd, network, network-bind, network-control, network-observe' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:0
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/hosts" pid=9729 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/hosts (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:0
Log: apparmor="DENIED" operation="mknod" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/dev/shm/.org.chromium.Chromium.DtpzPT" pid=9729 comm="ThreadPoolSingl" requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000
File: /dev/shm/.org.chromium.Chromium.DtpzPT (write)
Suggestions:
* adjust program to create files and directories in /dev/shm/snap.$SNAP_NAME.*
* try the snapcraft preload plugin: https://github.com/sergiusens/snapcraft-preload

= AppArmor =
Time: 2026-05-17T19:0
Log: apparmor="DENIED" operation="mknod" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/dev/shm/.org.chromium.Chromium.0UYEtE" pid=9729 comm="electron-builde" requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000
File: /dev/shm/.org.chromium.Chromium.0UYEtE (write)
Suggestions:
* adjust program to create files and directories in /dev/shm/snap.$SNAP_NAME.*
* try the snapcraft preload plugin: https://github.com/sergiusens/snapcraft-preload
```

</details>

With `--disable-dev-shm-usage --no-sandbox`, where the app launches successfully:

```console
$ electron-builder-snap-test --disable-dev-shm-usage --no-sandbox
Gtk-Message: 19:12:01.329: Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.
[10223:0517/191201.511082:ERROR:net/base/address_tracker_linux.cc:224] Could not create NETLINK socket: Operation not permitted (1)
[10223:0517/191201.513775:ERROR:content/browser/zygote_host/zygote_host_impl_linux.cc:279] Failed to adjust OOM score of renderer with pid 10355: Permission denied (13)
libva error: /snap/electron-builder-snap-test/x1/gpu-2404/usr/lib/x86_64-linux-gnu/dri//virtio_gpu_drv_video.so init failed
[10355:0517/191201.584309:ERROR:media/gpu/vaapi/vaapi_wrapper.cc:1658] vaInitialize failed: resource allocation failed
[10223:0517/191201.608317:ERROR:content/browser/zygote_host/zygote_host_impl_linux.cc:279] Failed to adjust OOM score of renderer with pid 10386: Permission denied (13)
```

<details>
<summary>snappy-debug output</summary>

```console
$ sudo snappy-debug
INFO: Following '/var/log/syslog'. If have dropped messages, use:
INFO: $ sudo journalctl --output=short --follow --all | sudo snappy-debug
kernel.printk_ratelimit = 0
= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10289 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 330(pkey_alloc) compat=0 ip=0x72e6240aa32b code=0x50000
Syscall: pkey_alloc

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10288 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 330(pkey_alloc) compat=0 ip=0x7103cfbb232b code=0x50000
Syscall: pkey_alloc

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7fe169db365b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 425(io_uring_setup) compat=0 ip=0x7fe169dbd28d code=0x50000
Syscall: io_uring_setup

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 203(sched_setaffinity) compat=0 ip=0x7fe169d396f1 code=0x50000
Syscall: sched_setaffinity
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional sched_setaffinity is often just noise)

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7fe169db365b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 425(io_uring_setup) compat=0 ip=0x7fe169dbd28d code=0x50000
Syscall: io_uring_setup

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="DelayedTaskSche" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 425(io_uring_setup) compat=0 ip=0x7fe169dbd28d code=0x50000
Syscall: io_uring_setup

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="ThreadPoolForeg" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 41(socket) compat=0 ip=0x7fe169dc231b code=0x50000
Syscall: socket
Suggestions:
* add account-control (if using NETLINK_AUDIT)
* add audio-playback (if using NETLINK_KOBJECT_UEVENT)
* add bluetooth-control (if using AF_{ALG,BLUETOOTH})
* add firewall-control (if using NETLINK_{FIREWALL,IP6_FW,NETFILTER,NF_LOG,ROUTE})
* add hardware-observe (if using NETLINK_{GENERIC,KOBJECT_UEVENT})
* add netlink-audit (if using NETLINK_AUDIT)
* add netlink-connector (if using NETLINK_CONNECTOR)
* add network (if using AF_INET{,6}, AF_CONN, NETLINK_ROUTE)
* add network-bind (if using AF_INET{,6}, NETLINK_ROUTE)
* add network-control (if using AF_{APPLETALK,BRIDGE,INET,INET6,IPX,PACKET,PPPOX,SNA}, NETLINK_{DNRTMSG,FIB_LOOKUP,GENERIC,INET_DIAG,ISCSI,KOBJECT_UEVENT,RDMA,ROUTE,XFRM})
* add network-observe (if using SOCK_RAW, AF_INET{,6}), NETLINK_{GENERIC,INET_DIAG,KOBJECT_UEVENT,ROUTE})
* add raw-usb (if using NETLINK_KOBJECT_UEVENT)
* add time-control (if using NETLINK_AUDIT)
* add unity7 (if using NETLINK_KOBJECT_UEVENT)
* add x11 (if using NETLINK_KOBJECT_UEVENT)

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/host.conf" pid=10223 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/host.conf (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/run/systemd/resolve/stub-resolv.conf" pid=10223 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=989
File: /run/systemd/resolve/stub-resolv.conf (read)
Suggestions:
* adjust program to use $SNAP_DATA
* adjust program to use /run/shm/snap.$SNAP_NAME.*
* adjust program to use /run/snap.$SNAP_NAME.*
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add one of 'firewall-control, fwupd, network, network-bind, network-control, network-observe' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/hosts" pid=10223 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/hosts (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/proc/10355/oom_score_adj" pid=10223 comm="ThreadPoolSingl" requested_mask="wc" denied_mask="wc" fsuid=1000 ouid=1000
File: /proc/10355/oom_score_adj (write)
Suggestions:
* adjust program to not access '@{PROC}/@{pid}/oom_score_adj'
* do nothing if program otherwise works properly
* add one of 'browser-support, microstack-support' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="dbus_method_call"  bus="system" path="/" interface="org.freedesktop.DBus.ObjectManager" member="GetManagedObjects" mask="send" name="org.bluez" pid=10223 label="snap.electron-builder-snap-test.electron-builder-snap-test"
DBus access

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10355 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 203(sched_setaffinity) compat=0 ip=0x7103cfb2a6f1 code=0x50000
Syscall: sched_setaffinity
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional sched_setaffinity is often just noise)

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/vulkan/implicit_layer.d/" pid=10355 comm="electron-builde" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/vulkan/implicit_layer.d/ (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/vulkan/icd.d/" pid=10355 comm="electron-builde" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/vulkan/icd.d/ (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/vulkan/implicit_layer.d/" pid=10355 comm="electron-builde" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/vulkan/implicit_layer.d/ (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/vulkan/icd.d/" pid=10355 comm="electron-builde" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/vulkan/icd.d/ (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/vulkan/implicit_layer.d/" pid=10355 comm="electron-builde" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/vulkan/implicit_layer.d/ (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/vulkan/icd.d/" pid=10355 comm="electron-builde" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/vulkan/icd.d/ (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/host.conf" pid=10359 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/host.conf (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/run/systemd/resolve/stub-resolv.conf" pid=10359 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=989
File: /run/systemd/resolve/stub-resolv.conf (read)
Suggestions:
* adjust program to use $SNAP_DATA
* adjust program to use /run/shm/snap.$SNAP_NAME.*
* adjust program to use /run/snap.$SNAP_NAME.*
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add one of 'firewall-control, fwupd, network, network-bind, network-control, network-observe' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/hosts" pid=10359 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/hosts (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10355 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 203(sched_setaffinity) compat=0 ip=0x7103cfb2a6f1 code=0x50000
Syscall: sched_setaffinity
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional sched_setaffinity is often just noise)

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10355 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7103cfba465b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/proc/10386/oom_score_adj" pid=10223 comm="ThreadPoolSingl" requested_mask="wc" denied_mask="wc" fsuid=1000 ouid=1000
File: /proc/10386/oom_score_adj (write)
Suggestions:
* adjust program to not access '@{PROC}/@{pid}/oom_score_adj'
* do nothing if program otherwise works properly
* add one of 'browser-support, microstack-support' to 'plugs'

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10386 comm="electron-builde" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x72e62409c65b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="ThreadPoolSingl" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7fe169db365b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="create" class="net" info="failed af match" error=-13 profile="snap.electron-builder-snap-test.electron-builder-snap-test" pid=10359 comm="Chrome_ChildIOT" family="inet6" sock_type="dgram" protocol=0 requested="create" denied="create"

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="ThreadPoolSingl" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7fe169db365b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/run/systemd/resolve/stub-resolv.conf" pid=10223 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=989
File: /run/systemd/resolve/stub-resolv.conf (read)
Suggestions:
* adjust program to use $SNAP_DATA
* adjust program to use /run/shm/snap.$SNAP_NAME.*
* adjust program to use /run/snap.$SNAP_NAME.*
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add one of 'firewall-control, fwupd, network, network-bind, network-control, network-observe' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/hosts" pid=10223 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/hosts (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="create" class="net" info="failed af match" error=-13 profile="snap.electron-builder-snap-test.electron-builder-snap-test" pid=10223 comm="ThreadPoolForeg" family="inet" sock_type="dgram" protocol=0 requested="create" denied="create"

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/run/systemd/resolve/stub-resolv.conf" pid=10223 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=989
File: /run/systemd/resolve/stub-resolv.conf (read)
Suggestions:
* adjust program to use $SNAP_DATA
* adjust program to use /run/shm/snap.$SNAP_NAME.*
* adjust program to use /run/snap.$SNAP_NAME.*
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add one of 'firewall-control, fwupd, network, network-bind, network-control, network-observe' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="open" class="file" profile="snap.electron-builder-snap-test.electron-builder-snap-test" name="/etc/hosts" pid=10223 comm="ThreadPoolForeg" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
File: /etc/hosts (read)
Suggestions:
* adjust program to read necessary files from $SNAP, $SNAP_DATA, $SNAP_COMMON, $SNAP_USER_DATA or $SNAP_USER_COMMON
* adjust snap to use snap layouts (https://forum.snapcraft.io/t/snap-layouts/7207)
* add 'system-files (see https://forum.snapcraft.io/t/the-system-files-interface for acceptance criteria)' to 'plugs'

= AppArmor =
Time: 2026-05-17T19:1
Log: apparmor="DENIED" operation="create" class="net" info="failed af match" error=-13 profile="snap.electron-builder-snap-test.electron-builder-snap-test" pid=10223 comm="ThreadPoolForeg" family="inet" sock_type="dgram" protocol=0 requested="create" denied="create"

= Seccomp =
Time: 2026-05-17T19:1
Log: auid=1000 uid=1000 gid=1000 ses=2 subj=snap.electron-builder-snap-test.electron-builder-snap-test pid=10223 comm="ThreadPoolSingl" exe="/snap/electron-builder-snap-test/x1/app/electron-builder-snap-test" sig=0 arch=c000003e 141(setpriority) compat=0 ip=0x7fe169db365b code=0x50000
Syscall: setpriority
Suggestion:
* ignore the denial if the program otherwise works correctly (unconditional setpriority is often just noise)
```

</details>

andersk:
Hmm, the app launches successfully with no arguments if I first manually run `snap connect electron-builder-snap-test:browser-support`. I’m not sure if it’s reasonable to expect users to do that. The [browser-support interface documentation](https://snapcraft.io/docs/reference/interfaces/browser-support-interface/) says

> Auto-connect: no when `allow-sandbox: true`, yes otherwise
> …
> > ⓘ A browser’s internal sandbox requires numerous privileged security policy rules to work and is typically considered trusted outside of snapd. For this reason, `--allow-sandbox=true` is limited to trusted publishers only.

andersk:
If I configure with

```json
"core24": {
  "plugs": [
    "default",
    {
      "browser-support": {
        "allow-sandbox": false
      }
    }
  ]
}
```

or just

```json
"core24": {
  "plugs": [
    "default",
    "browser-support"
  ]
}
```

then indeed `:browser-support` gets auto-connected and `electron-builder-snap-test --no-sandbox` works.

(I’m not excited about the reduced security of `--no-sandbox` but it seems like it will be necessary for passing Snap store reviews; it’s unfortunate that Snap seemingly requires the developer to make that choice in advance.)

Then I’m just back to the question of how to pass `--no-sandbox` automatically.

mmaietta:
Let me investigate further for you on that one. I'm curious if we can avoid the `--no-sandbox` aspect, but it seems that it needs to be conditionally applied.

Re: the `plugs` w/ user-overrides, that will deactivate the auto-injection of browser-support
Current core 24 logic path utilizes:
```
    // Always add browser-support with allow-sandbox so Chromium's internal sandbox
    // can create user namespaces under strict confinement.  Without allow-sandbox: true
    // the app crashes immediately with "FATAL: Permission denied (13)" in credentials.cc.
    // Skip the injection only when the user has explicitly provided their own plugs
    // (they are responsible for including browser-support in that case).
    if (!options.plugs) {
      rootPlugs = { ...rootPlugs, "browser-support": { interface: "browser-support", "allow-sandbox": true } }
      if (!appPlugs?.includes("browser-support")) {
        appPlugs = [...(appPlugs ?? []), "browser-support"]
      }
    }

    const { root: rootSlots, app: appSlots } = options.slots ? this.processPlugOrSlots(options.slots) : { root: undefined, app: undefined }

    // Build the effective arg list for the snap command.
    // Start with any user-supplied executableArgs, then conditionally add --no-sandbox
    // if browser-support with allow-sandbox:true is not present in the resolved plugs
    // (mirrors the same logic in SnapCoreLegacy.buildSnap).
    const extraArgs: string[] = [...(this.options.executableArgs ?? [])]
    if (this.options.allowNativeWayland === false) {
      if (!extraArgs.includes("--ozone-platform=x11")) {
        extraArgs.push("--ozone-platform=x11")
      }
    }
    if (this.helper.isElectronVersionGreaterOrEqualThan("5.0.0") && !this.isBrowserSandboxAllowed(rootPlugs)) {
      if (!extraArgs.includes("--no-sandbox")) {
        extraArgs.push("--no-sandbox")
      }
    }
    const commandSuffix = extraArgs.length > 0 ? ` ${extraArgs.join(" ")}` : ""

// and seperately

  private isBrowserSandboxAllowed(plugs: Record<string, any> | undefined): boolean {
    if (!plugs) {
      return false
    }
    for (const plug of Object.values(plugs)) {
      if (plug?.interface === "browser-support" && plug["allow-sandbox"] === true) {
        return true
      }
    }
    return false
  }
```

Latest changes seem to launch the snap core24 correctly for me (built on arm64 mac, installed on ubuntu 24.04 parallels VM)
Config:
```
  snapcraft: {
    base: "core24",
    core24: {
      useMultipass: true,
    },
  },
```
image

There's a lot of config permutations to test still, same with native linux build+install.

Wondering if I can use a linux Parallels VM setup for local e2e tests (similar to the approach used for windows e2e tests through a Parallels VM sync)
https://github.com/electron-userland/electron-builder/blob/8a98647f79c19fff7e7f5347d0a1ff80db18520e/test/src/updater/blackboxUpdateHelpers.ts#L278-L332

andersk:
If I configure with `"linux": { "artifactName": "${productName}-${version}-${arch}.${ext}" }` (Zulip Desktop has this for historical reasons), the build fails because `snapcraft pack` outputs a filename with underscores but it tries to copy the result using a source filename with hyphens:

```
Packed electron-builder-snap-test_0.0.0_amd64.snap
  ⨯ snap build failed  error=ENOENT: no such file or directory, copyfile '/home/ubuntu/electron-builder-snap-test/dist/__snap-amd64/electron-builder-snap-test-0.0.0-amd64.snap' -> '/home/ubuntu/electron-builder-snap-test/dist/electron-builder-snap-test-0.0.0-amd64.snap'
  ⨯ ENOENT: no such file or directory, copyfile '/home/ubuntu/electron-builder-snap-test/dist/__snap-amd64/electron-builder-snap-test-0.0.0-amd64.snap' -> '/home/ubuntu/electron-builder-snap-test/dist/electron-builder-snap-test-0.0.0-amd64.snap'  failedTask=build stackTrace=Error: ENOENT: no such file or directory, copyfile '/home/ubuntu/electron-builder-snap-test/dist/__snap-amd64/electron-builder-snap-test-0.0.0-amd64.snap' -> '/home/ubuntu/electron-builder-snap-test/dist/electron-builder-snap-test-0.0.0-amd64.snap'
```

andersk:
Another little issue is that `linux.category` and `linux.description` no longer make their way into the `.desktop` file. I get this warning:

```
  • application Linux category is set to default "Utility"  reason=linux.category is not set and cannot map from macOS docs=https://www.electron.build/linux
```

even when `linux.category` is set. It seems I have to set `core24.category` and `core24.description` instead.

mmaietta:
@andersk thank you so much for the beta testing on your side! I really do appreciate it!

Update here: I've pushed additional updates (and tests) to verify the cascading linux root + specific core configurations.

Going to go through some more test cases of build+install+run on my linux VM in between hospital visits. Stay tuned.
(Also, please do continue reporting any issues you encounter. Happy to chat live via discord as well -> you can find me in the electronjs discord server in the electron-builder channel [or can ping me 1:1])

jnsgruk:
Amazing, thanks @mmaietta!

vladimiry:
@mmaietta, I've just tested the `base: core24` feature using v26.15.3, with the following config (tested as a local build and via GitHub CI too):
```yaml
snapcraft:
  base: core24
  core24:
    confinement: strict
    plugs:
      - browser-support
      - desktop
      - desktop-legacy
      - network
      - password-manager-service
    useLXD: true
```
A few minor issues I had to overcome:
- The `chrome-sandbox` file doesn't get automatically removed. So I did that in the `afterPack` hook. The `--no-sandbox` arg being automatically added - nice.
- There is `snapcraft expand-extensions`-based validation of the `snapcraft.yaml` file, which is not supposed to break the build (based on a brief look into the code). But in fact it breaks the build if it detects some issues in `snapcraft.yaml`.
  In my case, the validation issue was about using forbidden characters in `apps.<app-name>.command` as listed in https://documentation.ubuntu.com/snapcraft/stable/reference/snapcraft-yaml/#apps.%3Capp-name%3E.command For example, one can set `executableArgs=['--js-flags="--max-old-space-size=12288"']` which automatically goes into `apps.<app-name>.command` (yes, I know this arg doesn't make much sense anymore, since rolled out V8 pointer compression puts a 4GB maximum on the heap size, but this is not the point here), but `snapcraft expand-extensions` validation doesn't like some characters here. To please the validation, I just wrapped the command with a basic sh-script.

PS I haven't tried to publish the package, yet. Update: manual publishing, via `snapcraft upload`, worked out.

mmaietta:
@vladimiry thanks for the follow-up, really appreciate it!
I've opened PR https://github.com/electron-userland/electron-builder/pull/9897 to address your feedback.

Btw, for further feedback, please consider opening as a new GH issue (easier for tracking 🙂)

vladimiry:
> I've opened PR https://github.com/electron-userland/electron-builder/pull/9897 to address your feedback.

@mmaietta Great, thanks! Going to test the changes in a new release.

> Btw, for further feedback, please consider opening as a new GH issue (easier for tracking 🙂)

Yeah, right. Just noticed the feature is in a "beta" state, so I posted the report here.

PR #9543, opened 2026-01-27
feat: add `size` option to DmgOptions for explicit DMG filesystem size

### LINKED ISSUE #8223: No space left on device (sparse files)
* **Electron-Builder Version**: 23.0.3

* **Node Version**: v16.20.2

* **Electron Version**: 22.3.4

* **Target**: mac

```
Copying…
could not access /Volumes/thing/Contents/Resource/sparsefile - No space left on device
Finishing…
```

This is due to an upstream issue with `hdiutil` interaction with sparse files (apparently on macOS 14) that can be observed thus:
```
➜ mkdir sparse
➜ dd if=/dev/zero of=sparse/big bs=1 count=0 seek=10000000
0+0 records in
0+0 records out
0 bytes transferred in 0.000014 secs (0 bytes/sec)
➜  hdiutil create -srcFolder sparse sparse.dmg
....could not access /Volumes/sparse/big - No space left on device

hdiutil: create failed - No space left on device
➜  mkdir notsparse
➜  cp -S sparse/big notsparse
➜  hdiutil create -srcFolder notsparse notsparse.dmg
............................................................
created: /Users/camhorn/sparsetest/notsparse.dmg
➜  ls -ls sparse notsparse .
.:
total 56
 0 drwxr-xr-x  3 camhorn  staff     96 May 17 11:48 notsparse
56 -rw-r--r--@ 1 camhorn  staff  24914 May 17 11:48 notsparse.dmg
 0 drwxr-xr-x  3 camhorn  staff     96 May 17 11:47 sparse

notsparse:
total 19536
19536 -rw-r--r--  1 camhorn  staff  10000000 May 17 11:48 big

sparse:
total 0
0 -rw-r--r--  1 camhorn  staff  10000000 May 17 11:47 big
```

It would be nice if electron-builder could override with the correct  `-size` for the initial `hdiutil create`:
https://github.com/electron-userland/electron-builder/blob/088523652934b87419c15c068459627dcf9a0535/packages/dmg-builder/src/dmg.ts#L200

Or at least allow `-size` to be manually set.

BarakXYZ:
## Summary

Add `size` option to `DmgOptions` to allow specifying the DMG filesystem size explicitly. This fixes "No space left on device" errors when building DMGs for large applications (>2GB) or applications with sparse files.

The size option accepts the same syntax as hdiutil's `-size` argument (e.g., `"150m"`, `"4g"`). If not specified, the size is calculated automatically (existing behavior).

## Changes

- Added `size` property to `DmgOptions` interface in `macOptions.ts`
- Added `size` to JSON schema validation in `scheme.json`
- Added `size` to `DmgBuildConfig` interface in `dmg.ts`
- Pass `size` through to dmgbuild settings in `dmgUtil.ts`
- Added test for the new option

## Usage

```typescript
dmg: {
  size: "4g", // Explicit 4GB filesystem size
}
```

## Dependency

This PR requires a corresponding change in dmgbuild to read the `size` option from JSON configuration files:
- **dmgbuild PR**: https://github.com/dmgbuild/dmgbuild/pull/279

Once the dmgbuild PR is merged and electron-builder-binaries updates its bundled dmgbuild, this feature will be fully functional.

Fixes #8223

### DIFF
--- a/packages/dmg-builder/src/dmgUtil.ts
+++ b/packages/dmg-builder/src/dmgUtil.ts
@@ -20,16 +20,16 @@ async function getDmgVendorPath(): Promise<string> {
     return path.resolve(customDmgbuildPath)
   }
 
-  // https://github.com/electron-userland/electron-builder-binaries/releases/tag/dmg-builder%401.1.0
-  const releaseVersion = "9614277"
+  // https://github.com/electron-userland/electron-builder-binaries/releases/tag/dmg-builder%401.2.0
+  const releaseVersion = "75c8a6c"
   const arch = process.arch === "arm64" ? "arm64" : "x86_64"
   const config = {
-    "dmgbuild-bundle-arm64-9614277.tar.gz": "28e11550cf990f78180a2d82090f35a24588beda3d9165098837714f90ee47ce",
-    "dmgbuild-bundle-x86_64-9614277.tar.gz": "4dbf1cc186af62921f8b6f4a5956b28d8622d211797a8b05eb75a260ee9c3fdb",
+    "dmgbuild-bundle-arm64-75c8a6c.tar.gz": "a785f2a385c8c31996a089ef8e26361904b40c772d5ea65a36001212f1fc25e0",
+    "dmgbuild-bundle-x86_64-75c8a6c.tar.gz": "87b3bb72148b11451ee90ede79cc8d59305c9173b68b0f2b50a3bea51fc4a4e2",
   }
   const filename: keyof typeof config = `dmgbuild-bundle-${arch}-${releaseVersion}.tar.gz`
   const file = await downloadArtifact({
-    releaseName: "dmg-builder@1.1.0",
+    releaseName: "dmg-builder@1.2.0",
     filenameWithExt: filename,
     checksums: config,
     githubOrgRepo: "electron-userland/electron-builder-binaries",


test/src/mac/dmgTest.ts
```diff
@@ -81,6 +81,27 @@ describe("dmg", { concurrent: true }, () => {
...
+          size: "500m",
+          // speed-up test
+          writeUpdateInfo: false,
+          title: "Explicit Size",
+        },
+      },
+      effectiveOptionComputed: async it => {
+        expect(it.specification.size).toEqual("500m")
+        return Promise.resolve(true)
+      },
```
mmaietta: Any way we can check the size post-packaging?

Something like: (off the top of my head / untested)
```
packed: async context => {
	const size = await exec("du", ["-sh", context.getAppPath(...])
	expect(size).toEqual("500M")
}
```

mmaietta:
Thanks for opening a PR! Can you please also run `pnpm generate-changeset` so this is picked up by the CD pipeline?

BarakXYZ:
Thanks for the review! I've made the following updates:

**1. Added changeset** as requested.

**2. Added `shrink` option and enhanced the test**

To properly verify that the explicit filesystem size is retained in the final DMG, I needed to add a `shrink` option. Here's why:

By default, dmgbuild runs `hdiutil resize -sectors min` after populating the image, which shrinks the filesystem to the minimum size regardless of what `size` was specified. The `size` option still works correctly in this case - it ensures enough space during the build phase - but the final filesystem ends up at minimum size.

To test that the filesystem actually retains a specific size (e.g., 500MB), we need to disable this shrinking behavior. So I created a follow-up PR to dmgbuild:
- **dmgbuild PR #280**: https://github.com/dmgbuild/dmgbuild/pull/280

The `shrink` option (defaults to `true`) is independent and additive - users can use `size` alone, `shrink` alone, or both together depending on their needs.

The test now uses `size: "500m"` with `shrink: false` and verifies the mounted filesystem is in the expected range (450-600MB).

Let me know if you'd like any adjustments!

mmaietta:
Nicely done! I really like this.

Next steps is to open a PR to pull in latest dmgbuild into our portable/closed-loop toolsets is to update the commit hash here to latest
https://github.com/electron-userland/electron-builder-binaries/blob/5fde804b8911fb43e8a9b41fcc668125c25cd6ec/packages/dmg-builder/build.sh#L8
After that PR is merged + deployed, we can simply bump the toolset versions in
https://github.com/electron-userland/electron-builder/blob/e394e0c48be965bd8a02c593ecda7e57b84b4f15/packages/dmg-builder/src/dmgUtil.ts#L27-L28

It's a few-but-simple hoops to jump through ever since I recently got that build pipeline set up. :D

packages/app-builder-lib/src/options/macOptions.ts
```diff
@@ -293,6 +293,20 @@ export interface DmgOptions extends TargetSpecificOptions {
    */
   format?: "UDRW" | "UDRO" | "UDCO" | "UDZO" | "UDBZ" | "ULFO"
 
+  /**
+   * The size of the DMG filesystem. Accepts the same syntax as the `-size` argument to `hdiutil`, e.g. `"150m"`, `"4g"`.
```
mmaietta: Maybe can phrase as `"The *initial* size of the DMG filesystem"` so that it distinguishes itself as dmg-building phase from the later dmg-shrinking phase.
Anything similar works too.

BarakXYZ:
Thanks for the feedback! Updated the `size` description to clarify it refers to the initial filesystem size.

Also created the PR to update dmgbuild in electron-builder-binaries:
- https://github.com/electron-userland/electron-builder-binaries/pull/135

Once that's merged and deployed with new binaries, I'll bump the `releaseVersion` and checksums in `dmgUtil.ts`.

mmaietta:
Boomshakalaka
[dmg-builder@1.2.0](https://github.com/electron-userland/electron-builder-binaries/releases/tag/dmg-builder%401.2.0)
`digest256` are the hashes you want to use

BarakXYZ:
pewpewpew! Bumped to dmg-builder@1.2.0 with the new checksums. Thank you!!

PR #9544, opened 2026-01-28
chore: v27 migration to node22 ESM package

### LINKED ISSUE #9483: FYI/PSA: ESM-only + Node >22 `engine` upcoming in next major semver, v27

mmaietta:
This pull request introduces major breaking changes as part of the v27 release, migrating the entire electron-builder ecosystem to native ES modules (ESM) and requiring Node.js >=22.12.0. It removes legacy dependencies, updates package configurations, and improves developer tooling and documentation to reflect these changes. The migration is designed to be as smooth as possible, with continued support for CommonJS (`require()`) on supported Node versions, but projects using deprecated features will need to update.

**Breaking changes and ESM migration:**

- All packages now use native ES modules (`"type": "module"`), and Node.js >=22.12.0 is required. The `electronCompile` config option and the `electron-compile` dependency are fully removed. The four `electron-forge-maker-*` plugins are now ESM. A comprehensive migration guide is added in `MIGRATION.md`, and the README is updated with requirements and usage examples for both ESM and CJS.

- The minimum Node.js version is enforced in `package.json` (`>=22.12.0`) and documentation; CI and workflow files are updated to target Node 22. 

**Package and build system updates:**

- The build output is moved from `out` to `dist`, and package exports are restructured to provide both public and internal entry points. A new `@electron-builder/app-builder-lib-helpers` package is added for workspace-internal sharing. 

- Scripts are updated to use `tsx` instead of `ts-node`, add new build/watch/typecheck scripts, and handle errors more gracefully in schema generation. ESLint and Prettier ignore patterns are updated to match new output locations. 

**API and schema changes:**

- The `electronCompile` configuration option is removed from the JSON schema, and the schema is updated for ESM compatibility. 

- Source files are updated to use explicit ESM imports (with `.js` extensions), and helper scripts are migrated to use dynamic `import()` instead of `require()`. 

**Release and versioning changes:**

- Changeset configuration and pre-release metadata are updated to include new packages and mark the release as `alpha`. 

**Documentation improvements:**

- The migration guide and README are updated to clearly explain the breaking changes, migration steps, and provide new usage examples for both ESM and CJS.


packages/app-builder-lib/src/targets/AppImageTarget.ts
skmedix, line 1: This file is unused

packages/electron-updater/package.json
```diff
@@ -33,8 +37,14 @@
     "@types/semver": "7.7.1",
     "electron": "35.7.5"
   },
-  "typings": "./out/main.d.ts",
+  "types": "./out/main.d.ts",
```
skmedix: Is this still correct since `main.js` is gone?

packages/app-builder-lib/package.json
```diff
@@ -106,6 +109,7 @@
     "@babel/preset-env": "7.24.8",
     "@babel/preset-react": "7.24.7",
     "@types/debug": "4.1.7",
+    "@types/fs-extra": "9.0.13",
```
skmedix: You can probably bump it to v11 In all `package.json`s

mmaietta:
For visibility on code line diff count
|Directory|Insertions|Deletions|
|--------|--------|--------|
|packages/| +1860|-1379|
|test/|+825|-1368|
```
git diff master...HEAD --numstat | awk '
  /^[0-9]/ && $3 ~ /^packages\// { pkg_add+=$1; pkg_del+=$2 }
  /^[0-9]/ && $3 ~ /^test\//     { tst_add+=$1; tst_del+=$2 }
  END {
    print "packages/  +" pkg_add " -" pkg_del
    print "test/      +" tst_add " -" tst_del
  }
'
```

Issue #8536, opened 2024-09-25
NSIS Installer doesn't launch on Windows 11 when perUser is set by default


pnedkovp:
* **Electron-Builder Version**: 24.13.3
* **Electron Version**: 31.3.1
* **Target**:  nsis

We have an issue with nsis installer on Windows 11. After the application is build the installer doesn't launch on some machines.
We have tried building the application with different attributes and it seems that it has something to do with the perMachine property.
These are our production build attributes:
"nsis": {
      "oneClick": false,
      "allowElevation": true,
      "perMchine": false,
      "artifactName": "${productName} ${version}.${ext}"
    }

If we change to oneClick: true or selectPerMachineByDefault: true then the installer starts successfully but with our current setup (onClick: false, perMachine: false) sometimes it crashes.

The error in the event viewer is this:
Faulting module name: System.dll, version: 0.0.0.0, time stamp: 0x5c157efa
Exception code: 0xc0000005
Fault offset: 0x00001581
Faulting process id: 0x0xF30
Faulting application start time: 0x0x1DB05CD5454D5C0
Faulting application path: ***
Faulting module path: C:\Users***\AppData\Local\Temp\nsmD439.tmp\System.dll
Report Id: 0e92fa50-e642-47bd-8aac-4b2d8e89f8f8
Faulting package full name:
Faulting package-relative application ID:

mmaietta:
What sticks out to me is the lack of any name here:
```
Faulting package full name:
Faulting package-relative application ID:
```

I found this regarding DEP: https://www.tenforums.com/software-apps/206789-application-stopped-running-i-dont-know-what-else-try-2.html and this regarding the Exception Code: https://answers.microsoft.com/en-us/windows/forum/all/how-do-i-fix-this-exception/33382768-19f5-473f-98a2-07e08194272b
>Application Error 0xc0000005 (Access Violation) error is usually caused by your computer not being able to correctly process the files and settings required to run a particular program or installation.

I'm not sure if there's much support I can provide here from a code standpoint though. I'm pretty unfamiliar with windows exceptions.

pnedkovp:
The reason I turn to you is because the installer fails only with a specific set of attributes and not every time. For me it seems it has something to do with the nsis installer script in the case when the default option is to install for the current user only.

pnedkovp:
I have already tried turning off security features including DEP but it didn't do the trick.
[...]

Issue #2237, opened 2017-10-29
Entry left in Task manager startup apps even after uninstalling electron app


GorvGoyl:
* **Version**:  19.42.2

* **Target**:  Windows 10

I've set my electron app to auto-start at bootup.
Here's the setting for that:
```
app.setLoginItemSettings({
    openAtLogin: true,
    path: process.execPaths
  })
```

package.json settings:
```
 "nsis": {
      "runAfterFinish": true,
      "createDesktopShortcut": true,
      "deleteAppDataOnUninstall": true
    }
```
The issue is when i uninstall the app i can still see the entry in task manager -> startup apps tab.
see the image below (note that 'open file location' is disabled for this entry):
![image](https://user-images.githubusercontent.com/7106086/32142830-1e883df6-bcc5-11e7-8052-4761a0c974ac.png)
On digging I found that electron adds the registry entry @ `Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run` which doesn't get cleared on app uninstall.

I'm not sure how to get it removed maybe some additional parameters to include in the uninstaller script or any other way? 
And also what would be the process in MAC?

VictorLeach96:
This issue still persists

BTARPAR:
This issue still persists

neekai:
This issue still persists

eltonchan:
This issue still persists

beyondkmp:
This issue still persists

jaime-rivas:
Same here.

mmaietta:
Not much electron-builder can do if electron is adding that itself. Makes sense that it can't cleanup during uninstall though, as the app wouldn't be aware of it.
Try using a `customUnInstall` hook in your https://www.electron.build/configuration/nsis#custom-nsis-script to remove the entry

Mac doesn't have a hook for uninstalls since it's just a file (.app) being deleted from the filesystem

Mero-Plaform:
This is still an issue.

mmaietta:
That's nice, thanks for chiming in.

>Try using a customUnInstall hook in your https://www.electron.build/configuration/nsis#custom-nsis-script to remove the entry

I'm not familiar with how electron itself inserts a startup app (i.e. LoginItem) nor with Windows registry, but if **you** can identify a generic way to remove the login item from the registry, then I'm happy to guide you on submitting a PR for resolving an electron-created issue

JPCrounse:
[...]

PR #9564, opened 2026-02-05
feat(nsis): detect windows version and exclude win 7 compatibility logic when not needed

JPCrounse:
Fix https://github.com/electron-userland/electron-builder/issues/8536

The multiUser.nsh contains compatibility logic for windows 7 which introduces a dependency on System.dll via System::store()
This can cause a race condition in the NSIS installer when oneClick is set to false and the application is being installed in perUser mode.

In the specific use case of requiring a assisted installer which can also optionally be run in silent mode while supporting both perUser and perMachine installs via command line flags the only remaining workarounds appear to be overriding the multiUser.nsh using a before-build script or manually writing the $INSTALLDIR registry keys to avoid this dependency from being triggered.

Added auto version detection using the AtLeastWin8 macro from WinVer.nsh to prevent unecessary usage / dependency on the System.dll for windows versions higher than Windows 7


mmaietta:
Thinking something like:
```nsis
  # Check if running on Windows 7 (version 6.1)
  ${If} ${IsWin7}
    System::Store S
    # Win7 has a per-user programfiles known folder and this can be a non-default location
    System::Call 'SHELL32::SHGetKnownFolderPath(g "${FOLDERID_UserProgramFiles}", i ${KF_FLAG_CREATE}, p 0, *p .r2)i.r1'
    ${If} $1 == 0
      System::Call '*$2(&w${NSIS_MAX_STRLEN} .s)'
      StrCpy $0 $1
      System::Call 'OLE32::CoTaskMemFree(p r2)'
    ${EndIf}
    System::Store L
  ${EndIf}
```

JPCrounse:
Indeed, the autoscaling approach using WinVer.nsh seems preferable as opposed to introducing a new NSIS config option.
I will update it accordingly.

I believe it may also be useful in potentially fixing https://github.com/electron-userland/electron-builder/issues/2237 
I could include it as a separate commit on this PR or a separate PR if preferred?

mmaietta:
Interesting!
Hmmm, separate PR just in case would probably be best.

mmaietta:
LGTM. Just to pulse check, have you also tested the new changes out locally?

JPCrounse:
I've tested locally in windows 10 and 11 and confirmed it works and also prevents the sporadic crash of the installer due to the System.dll dependency.

Tbh I haven't tested it in windows 7 yet but I might have time to spin up a VM tomorrow just to be sure.

mmaietta:
@JPCrounse quick follow-up here. No further changes or testing/validations pending from your side? PR LGTM otherwise!

JPCrounse:
> @JPCrounse quick follow-up here. No further changes or testing/validations pending from your side? PR LGTM otherwise!

@mmaietta No more code changes or testing/validations pending, good to go :)

mmaietta:
@JPCrounse can you please address the two copilot review comments? I think I'm in line with the comments/actions requested.

PR #9577, opened 2026-02-09
fix: use projectDir instead of workspaceRoot for rebuild path

johnpyp:
WIP - some edgecases

---

Using `workspaceRoot` by default for rebuild path causes issues, as demonstrated in the added test fixtures where multiple monorepos are trying to rebuild `classic-level` with `electron-builder install-app-deps`, but clobber eachother due to the incorrect scoping.

(If you revert the fix I added in `yarn.ts` you'll see how those tests now fail, with both bun/pnpm using their "isolated installation" strategies)

Test cases are added a little oddly as integration tests using the cli directly to try to keep the environment as realistic and similar to the error case as I currently face in my real repo.

This change doesn't seem to regress #9376, as rebuilderTest still passes.

### DIFF
--- a/.changeset/lazy-turkeys-change.md
+++ b/.changeset/lazy-turkeys-change.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: use projectDir instead of workspaceRoot for rebuild path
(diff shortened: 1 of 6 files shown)


johnpyp:
Realized I was accidentally "fixing" an issue covering up a deeper issue with the multi-package workspace... closing this and will reopen a different PR :)

PR #9581, opened 2026-02-12
feat: pass release notes to GitHub/GitLab release body

### LINKED ISSUE #9235: Release notes for GitHub publisher
Hi there!

How to pass the release notes to GitHub publisher? I've tried different ways:

electron-builder.json5:
```
  publish: [
    {
      provider: "github",
      releaseNotes: "Test"
    }
  ],
```

and

```
  publish: [
    {
      provider: "github",
      releaseNotesFile: "dist/release-notes.md"
    }
  ],
```

and 

```
  publish: [
    {
      provider: "github",
    }
  ],
  releaseInfo: {
      releaseNotes: "Test"
  }
```

Nothing is working. I generate a markdown string in my GitHub workflow but I don't see a way how to pass it to electron-builder 🤨

asamofal:
Closes #9235

Hey! This wires the existing `releaseInfo` infrastructure (which already resolves `releaseNotes`, `releaseNotesFile`, and auto-detects `release-notes.md`) through to the GitHub and GitLab publishers. Until now that data only ended up in auto-update YAML files — the actual release drafts were always created with an empty body.

## What changed

- Added `releaseBody` and `releaseName` to `GithubOptions` and `GitlabOptions`
- `PublishManager.createPublisher()` now resolves release info from config and injects it into publisher options (with explicit values taking precedence)
- `GitHubPublisher.createRelease()` passes `body` and `name` to the GitHub API
- `GitlabPublisher.createRelease()` uses `releaseBody` as `description` instead of the hardcoded `"Release vX.Y.Z"`
- Release body is truncated at **100k characters** with a warning log if exceeded (addresses the size concern from the issue)
- 4 unit tests covering body, name, omission, and truncation

## Resolution order for release body

1. Explicit `releaseBody` on publish config (new, takes precedence)
2. `releaseInfo.releaseNotes` from config
3. `releaseInfo.releaseNotesFile` read from disk
4. `release-notes.md` in project root (auto-detected fallback)

## Config examples

```json5
// inline
{ "releaseInfo": { "releaseNotes": "## Changes\n- stuff" } }

// file reference
{ "releaseInfo": { "releaseNotesFile": "CHANGELOG.md" } }

// just drop release-notes.md in project root — auto-detected

// or override directly on publish config
{ "publish": [{ "provider": "github", "releaseBody": "custom body" }] }
```

Also works via CLI dot-notation: `electron-builder --publish always -c.releaseInfo.releaseNotesFile=notes.md`

## Tested

- Unit tests pass
- Verified end-to-end against a real GitHub repo — draft release created with correct body and name


packages/app-builder-lib/src/publish/PublishManager.ts
```diff
@@ -308,6 +308,26 @@ export async function getPublishConfigsForUpdateInfo(
...
 }
 
+async function resolveReleaseBody(packager: Packager): Promise<string | null> {
+  const releaseInfo = packager.config.releaseInfo
+  if (releaseInfo?.releaseNotes) {
+    return releaseInfo.releaseNotes
+  }
+  if (releaseInfo?.releaseNotesFile) {
+    try {
+      return await readFile(path.resolve(packager.projectDir, releaseInfo.releaseNotesFile), "utf-8")
```
mmaietta, line 318: Can use `packager.getResource(releaseInfo.releaseNotesFile)` here IIRC to handle resolving the changelog from multiple directories (inc. projectDir)

packages/builder-util-runtime/src/publishOptions.ts
```diff
@@ -144,6 +144,25 @@ export interface GithubOptions extends PublishConfiguration {
...
+   * Truncated to 100000 characters with a warning if exceeded.
+   */
+  readonly releaseBody?: string | null
+
+  /**
+   * The release name/title. If not set, defaults to the version string (e.g. `1.0.0`).
+   *
+   * If not set explicitly, resolved from `releaseInfo.releaseName`.
+   */
+  readonly releaseName?: string | null
```
mmaietta: Instead of duplicating code between Github and Gitlab options, let's create a separate interface called PublishReleaseNotes (or similar) to achieve
```
interface GithubOptions extends PublishConfiguration, PublishReleaseNotes { ... }
```

packages/electron-publish/src/gitHubPublisher.ts
```diff
@@ -230,12 +230,22 @@ export class GitHubPublisher extends HttpPublisher {
...
+    }
+    if (this.info.releaseBody) {
+      const maxLength = 100000
+      if (this.info.releaseBody.length > maxLength) {
+        log.warn({ length: this.info.releaseBody.length, maxLength }, "release body exceeds GitHub API limit, truncating")
+        data.body = this.info.releaseBody.substring(0, maxLength)
+      } else {
+        data.body = this.info.releaseBody
+      }
+    }
```
mmaietta, line 244: Can probably extract this common function between Github and Gitlab to a separate/new `util.ts` that auto-trims a string based on 2nd param and reuse between the publishers. Thoughts?
```
const trimStringWithWarn(str, maxLength, warnMessage) { ... }
```

asamofal: Don't think `trimStringWithWarn` will be reused somewhere else, but fixing DRY violations just in these two cases sounds good for me 🙂

packages/app-builder-lib/src/publish/PublishManager.ts
mmaietta: Please address this comment and we should be good to go! 🙃

asamofal: omg, I wrote the comment but not published the review, so it was in pending 🤦‍♂️

---

Good call on using `getResource`-style resolution instead of just `projectDir`. One concern though: `createPublisher` currently receives `Packager` (root), not `PlatformPackager`, and that seems intentional since a GitHub/GitLab release is created once for the project, not per-platform.

Threading `PlatformPackager` through would mean the release body could resolve differently depending on which platform runs first (e.g. picking up `release-notes-mac.md`), which doesn't feel right for a provider-level concept.

Would you prefer:

1. Thread `PlatformPackager` through anyway (and just skip platform-specific filenames), or
2. Replicate the `getResource` lookup logic (check `buildResourcesDir` first, then `projectDir`) directly in `resolveReleaseBody` using the root `Packager`?

I think option 2 keeps the semantics clean and still gets the multi-directory resolution you suggested. What do you think?

asamofal: @mmaietta Your call :)

mmaietta: So I'm not fully comfortable with this PR in the current state tbh, and multiple personal events have taken away from time I'd spend on electron-builder.

Let me check out this PR locally and see if I can add more safeguards and/or consolidate logic
For instance, I'm not sure I understand this TSdoc.
```
   * If not set explicitly, it is resolved from `releaseInfo` configuration:
   * - `releaseInfo.releaseNotes` — inline release notes string
   * - `releaseInfo.releaseNotesFile` — path to a file containing release notes
```
Is the user setting `releaseInfo`? Since this is a property for the root config, I'm trying to understand how this tsdoc helps (won't doesn't confuse) the developer.

For the interim, please stop rebasing this PR via the GH site; I can take this from here 🙂 . Still am curious on the `releaseInfo` though

Looking at this, it seems like we're duplicating interfaces on a property that is already accessible via config and via command-line and on a per-platform basis (or not if explicitly specified without any macros)
https://github.com/electron-userland/electron-builder/blob/5d4651dbb29d0730d0fb3be96de79d9348c0a99c/packages/app-builder-lib/src/options/PlatformSpecificBuildOptions.ts#L206-L219
>Until now that data only ended up in auto-update YAML files — the actual release drafts were always created with an empty body.

Can we just pipe that info from the auto-update yaml files? Or is that requiring a much larger logic-flow refactor for v27? I'm worried about creating multiple sources for release notes being set up (being it's also already supporting on a per `PlatformSpecificConfiguration` basis)

asamofal: Actually, you're right. I should've just used `config.releaseInfo` from the start instead of introducing new props on publisher options, I just didn't connect the dots.

Since a GitHub/GitLab release is one per version (not per-platform), root `releaseInfo` is the right source. No need for `PublishReleaseInfo` and the new props for `GithubOptions`/`GitlabOptions`, and instead we can pass the resolved values to publishers via constructor params internally.

Re the TSDoc question, that goes away entirely once we drop the props from publisher options. The resolution logic becomes internal, no user-facing docs needed.

Let me know if you want to handle this or if I should push the commit.

mmaietta: Yes! I think we're aligned on the proposed route. If you have time to handle these changes, it would be immensely appreciated and we can get this in for a v26.9 release.

asamofal: @mmaietta Done! I tested it, the real release draft was successfully published in my repo 👌

Hopefully I don't miss anything this time, but feel free to share extra feedback if you have any ;)

Issue #4469, opened 2019-11-26
`--publish onTag` only works in specific CIs


thisismydesign:
* **Version**: 21.2.0 & master

* **Target**: all

https://github.com/electron-userland/electron-builder/blob/master/packages/electron-publish/src/publisher.ts#L98-L101

This gave us some headaches using GitHub Actions and wouldn't work locally either.

As is: `--publish onTag` only supports a set of specific CIs. This is [not documented](https://www.electron.build/configuration/publish).

Expected: if this remains the case the exact behavior should be documented. However, the correct behavior could be achieved by relying on `git` instead.

For example the following pseudo-code returns whether there's a tag checked out:
`!(git symbolic-ref -q HEAD) && (git describe --tags --exact-match)`

See also: https://stackoverflow.com/questions/18659425/get-git-current-branch-tag-name

tgxn:
Still seems problematic that this only supports GitHub and Bintray.
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/publish/PublishManager.ts#L162

thisismydesign:
The link above using fixed lines: https://github.com/electron-userland/electron-builder/blob/e7179b57bdba192acfdb439c03099e6629e98f6a/packages/app-builder-lib/src/publish/PublishManager.ts#L165-L168

ScubaDrew:
Does not seem to work correctly on GitLab. Tags result in `current build is not for a git tag`

PR #9590, opened 2026-02-13
feat(appimage): If necessary, start with `--no-sandbox`

### LINKED ISSUE #8440: How to run and/or install AppImage on Ubuntu
My versions:
    "electron": "^31.0.2",
    "electron-builder": "^24.13.3",
    "electron-vite": "^2.3.0",

Node version:
v18.19.1

Target:
Linux x64

ISSUE
When I run the AppImage (Ubuntu 24.04 KDE Plasma 5.27) I get the following error.  I gave my user account rwx persmissions to the "appForLinux.AppImage" file.

[30783:0831/200557.279232:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /tmp/.mount_deskto9XlUHg/chrome-sandbox is owned by root and has mode 4755.
Trace/breakpoint trap (core dumped)

Is this an error with my (limited) understanding of how to run an AppImage on Ubuntu or does this error indicate the packaging is bad/corrupt with Electron Builder?

I'm more of a Windows/Mac developer.  Should I install the Snap or .deb version of the test app built I with Electron Builder?  I thought it was straightforward to just run any AppImage on Ubuntu.  But, maybe I should install the electron app some other way?  If so, what would that command look like to install a Linux app built with Electron Builder?

I use "apt install" a lot...  But, I'm new to Linux dev so I'm not sure what my other options are to test an Electron app on Linux.  I can run the app just fine from VSCode on Ubuntu.  And it works but how do I actually install it and run it?

Thanks

sabonerune:
fix #8440

The method in #4496 has the following drawbacks:
- Doesn't work without desktop integration
The AppImage can be launched without desktop integration.
It also doesn't work when started from the command line.
- Sandbox is always disabled
user namespace should now be available on many Linux systems. 
This means that `--no-sandbox` is only needed in limited environments.
Even in Ubuntu 24.04 or later, starting from file manager bypasses AppArmor restrictions, so `--no-sandbox` is not required.
https://discourse.ubuntu.com/t/understanding-apparmor-user-namespace-restriction/58007#p-148026-bypassing-apparmor-security-restriction
It's also possible that the user has disabled AppArmor.

This solves these problems by automatically detecting when the user namespace is unavailable and adding `--no-sandbox` at runtime.

### DIFF
--- a/packages/app-builder-lib/src/targets/appimage/appImageUtil.ts
+++ b/packages/app-builder-lib/src/targets/appimage/appImageUtil.ts
@@ -242,15 +242,15 @@ for arg in "\${args[@]}" ; do
   fi
 done
 NO_SANDBOX=
-# Use `unshare -Ur true` as a heuristic to detect whether user namespaces are available.
+# Use 'unshare -Ur true' as a heuristic to detect whether user namespaces are available.
 # Notes:
 #   - When running as root, this check will always succeed even if the sandbox configuration
 #     actually relies on unprivileged user namespaces. In practice, Chrome/Electron usually
 #     disables or adjusts the sandbox separately when running as root, so this probe is mostly
 #     a no-op in that scenario.
-#   - On minimal systems (e.g. Alpine or stripped-down containers) `unshare` may not exist.
+#   - On minimal systems (e.g. Alpine or stripped-down containers) 'unshare' may not exist.
 #     In that case the shell will return exit code 127 ("command not found"), which will cause
-#     us to add `--no-sandbox`. This is an intentional fail-safe: we prefer the app to start
+#     us to add '--no-sandbox'. This is an intentional fail-safe: we prefer the app to start
 #     without sandboxing rather than crash on startup.
 if [ $HAVE_NO_SANDBOX -eq 0 ] && ! unshare -Ur true 2>/dev/null ; then
   NO_SANDBOX=--no-sandbox


absidue:
I would suggest changing the pull request title and changelog entry, as it sounds like you are adding the `--no-sandbox` flag in **more** cases, but the goal of this pull request is to add that flag in **less** cases. Something along the lines of: "feat: Only start AppImage with `--no-sandbox` when necessary" would be more clear.

packages/app-builder-lib/src/targets/appimage/AppImageTarget.ts
```diff
@@ -32,7 +32,7 @@ export default class AppImageTarget extends Target {
     super("appImage")
 
     this.desktopEntry = new Lazy<string>(() => {
-      const args = this.options.executableArgs?.join(" ") || "--no-sandbox"
```
mmaietta, line 35: This is a breaking change for appimage toolset `0.0.0` (legacy) as it's AppRun script is hardcoded in upstream dependency `app-builder-bin`

packages/app-builder-lib/src/targets/appimage/AppImageTarget.ts
```diff
@@ -32,7 +32,9 @@ export default class AppImageTarget extends Target {
...
+      const appimageTool = packager.config.toolsets?.appimage
+      const defaultArgs = appimageTool == null || appimageTool === "0.0.0" ? "--no-sandbox" : ""
+      const args = this.options.executableArgs?.join(" ") || defaultArgs
       return helper.computeDesktopEntry(this.options, `${APP_RUN_ENTRYPOINT} ${args} %U`, {
```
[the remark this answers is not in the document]
sabonerune: I didn't really like the Copilot code so I fixed it manually.

packages/app-builder-lib/src/targets/appimage/AppImageTarget.ts
```diff
@@ -32,8 +32,11 @@ export default class AppImageTarget extends Target {
     super("appImage")
 
     this.desktopEntry = new Lazy<string>(() => {
-      const args = this.options.executableArgs?.join(" ") || "--no-sandbox"
-      return helper.computeDesktopEntry(this.options, `${APP_RUN_ENTRYPOINT} ${args} %U`, {
+      const appimageTool = packager.config.toolsets?.appimage
+      const defaultArgs = appimageTool == null || appimageTool === "0.0.0" ? ["--no-sandbox"] : []
+      const args = this.options.executableArgs ?? defaultArgs
+      const exec = [APP_RUN_ENTRYPOINT, ...args, "%U"].join(" ")
```
mmaietta, line 38: Thoughts on also checking whether `executableArgs` already includes `%U` before we append it?

(Looks like this logic flow is missing in the original implementation)

mmaietta:
Looks like we have some tsc errors. Can you please TAL when you have a chance

packages/app-builder-lib/src/targets/appimage/AppImageTarget.ts
sabonerune: I think it would be good to implement it.
I feel like `executableArgs` should also be validated and escaped at the same time.
But I think that fix should be in a separate PR.

mmaietta (approved):

Nantris:
> starting from file manager bypasses AppArmor restrictions, so --no-sandbox is not required.

I have not found this to be the case in Kubuntu. I don't know if this is Ubuntu specific but I've used every recent version of Kubuntu and this doesn't match my experience in any of them.

----

Would you happen to have a moment to elaborate this point?

> user namespace should now be available on many Linux systems.
> This means that --no-sandbox is only needed in limited environments.

Nantris:
Just to add, on Kubuntu in Dolphin if you try to launch an `AppImage` it fails with: 

> _"For security reasons, launching executables is not allowed in this context."_

My understanding is Ubuntu will close this route soon in Nautilus.

More importantly, I'm skeptical that using `--no-sandbox` is as trivial and free from risk as the initial comment suggests. `--no-sandbox` doesn't just skip namespace creation - it disables all renderer sandboxing. This means any renderer exploit becomes direct code execution as the user, whereas with the sandbox, a renderer exploit is contained to the sandbox barring a second exploit to escape it.

sabonerune:
@Nantris 
>Would you happen to have a moment to elaborate this point?

To be honest, I'm not very familiar with Linux distributions.
However, I thought that most Linux distributions, excluding those based on Ubuntu 24.04 or later, could be booted without `--no-sandbox`, unlike in #4469.

---

Prior to this pull request, it wouldn't work if desktop integration wasn't enabled, and if it was enabled, `--no-sandbox` was added even if it wasn't necessary.

I simply made this behavior more as intended.

If you believe the application should not automatically add `--no-sandbox` when it starts, it would be better to create a new issue.

Nantris:
Thanks for your reply @sabonerune.

Before I make another issue, can you clarify if this behavior is on by default where `kernel.apparmor_restrict_unprivileged_userns === 1`? It seems to be, and if so I will create another issue regarding concerns. Thank you.

sabonerune:
@Nantris
If `kernel.apparmor_restrict_unprivileged_userns === 1`, then in most cases, executing `unshare -Ur true` will fail.
Therefore, `--no-sandbox` is added.
However, if AppArmor is bypassed or permitted by the profile, it will launch without `--no-sandbox` as it will successfully execute `unshare -Ur true`.

mmaietta:
Hi folks, attempting to catch up here. What was the resolution and decision? (and/or created issues? CC @Nantris)

Need to understand if this PR is an acceptable change before any release is pushed out.

Nantris:
@mmaietta I've not gotten around to making a new issue but did plan to.

I'm skeptical of this being activated by default. I see far less issue with it being opt-in.

I suspect most developers using `electron-builder` will not be aware their app is running with `--no-sandbox`, much less their end-users. It's debatable whether users need to know that detail, but certainly the developer responsible for the security model needs to know, at least in my view.

absidue:
@Nantris Before this pull request `--no-sandbox` was always passed to Electron, with this pull request it is only passed when needed, which is an improvement.

Nantris:
@absidue thank you for the correction - I overlooked that. With that in mind the PR seems unquestionably an improvement over the current code. 

However if I'm reading the `git blame` properly the inclusion of `--no-sandbox` at all is a development from around a month ago which still makes it a recent change worth real consideration, and I don't think `electron-builder` should ever configure the app with `--no-sandbox` unless the developer who is using `electron-builder` specifically opts in.

Am I interpreting the code properly that this is only for desktop shortcuts? I don't think it's good security practice to be bypassing it even selectively without being explicitly opt-in.

Nantris:
Lazily created #9659 so further discussion of the use of `--no-sandbox` itself can happen there.

sabonerune:
@Nantris `AppRun` is the entry point for AppImage.
Therefore, this code is always executed.
https://docs.appimage.org/reference/appdir.html#general-description

PR #9610, opened 2026-02-19
fix(appimage): resolve toolset binaries from linux/<arch> directory

f-io:
I tried to use the new configuration option:

```yaml
toolsets:
  appimage: "1.0.2"
```

On a default GitLab runner, the AppImage build failed with:

spawn .../mksquashfs ENOENT

The issue is that `getAppImageTools()` can return paths based on a relative `artifactPath` (e.g. `.cache/...`).  
Since `mksquashfs` is executed from a different `cwd`, the relative path can break and result in ENOENT.

### Fix

- resolve `artifactPath` using `path.resolve(...)`
- use the correct `linux/<arch>` subdirectory for the binaries
- keep runtime and library paths unchanged

This ensures `mksquashfs` is referenced via an absolute path and prevents ENOENT during AppImage builds.

Related: https://github.com/electron-userland/electron-builder/pull/9558

### DIFF
--- a/.changeset/wild-zebras-cheat.md
+++ b/.changeset/wild-zebras-cheat.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(win): regenerate app exe if header of the asar has changed so that asar integrity properly matches
--- a/packages/app-builder-lib/src/winPackager.ts
+++ b/packages/app-builder-lib/src/winPackager.ts
@@ -1,10 +1,11 @@
-import { Arch, CopyFileTransformer, executeAppBuilder, FileTransformer, InvalidConfigurationError, log, use, walk } from "builder-util"
+import { Arch, CopyFileTransformer, executeAppBuilder, exists, FileTransformer, InvalidConfigurationError, log, use, walk } from "builder-util"
 import { Nullish } from "builder-util-runtime"
+import { isCI } from "ci-info"
 import { createHash } from "crypto"
 import { readdir } from "fs/promises"
-import { isCI } from "ci-info"
 import { Lazy } from "lazy-val"
 import * as path from "path"
+import { readAsarHeader } from "./asar/asar"
 import { SignManager } from "./codeSign/signManager"
 import { signWindows, WindowsSignOptions } from "./codeSign/windowsCodeSign"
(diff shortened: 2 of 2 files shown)


packages/app-builder-lib/src/toolsets/linux.ts
```diff
@@ -57,11 +57,22 @@ export async function getAppImageTools(targetArch: Arch) {
...
+    targetArch === Arch.armv7l
+      ? "arm32"
+      : targetArch === Arch.arm64
+      ? "arm64"
+      : targetArch === Arch.ia32
+      ? "ia32"
+      : "x64"
+
+  const root = path.resolve(artifactPath)
+  const binDir = path.join(root, "linux", runtimeArch)
```
mmaietta: `linux` should be `process.platform` as `darwin` also is included in the package, but it seems like we only need `artifactPath = path.resolve(artifactPath)`.

Reason being is that the `mksquashfs` at root (and desktop-file-validate, obj_decompress as well) is supposed to autoroute to the correct platform+arch combination. Is this not occurring?
https://github.com/electron-userland/electron-builder-binaries/blob/96a936e905ad104340e16e579a85d0cfaefbefdc/packages/appimage/assets/bundle-and-compress.sh#L141-L165

Toolset contents:
```
tree ./appimage-tools-runtime-20251108 -L2
├── appimage-tool
├── darwin
│   ├── arm64
│   └── x64
├── desktop-file-validate -> appimage-tool
├── lib
│   ├── arm32
│   ├── arm64
│   ├── ia32
│   └── x64
├── linux
│   ├── arm32
│   ├── arm64
│   ├── ia32
│   └── x64
├── mksquashfs -> appimage-tool
├── opj_decompress -> appimage-tool
├── runtimes
│   ├── runtime-arm32
│   ├── runtime-arm64
│   ├── runtime-ia32
│   ├── runtime-x64
│   └── VERSION.txt
└── select-runtime.env

15 directories, 10 files
```

f-io:
Tested on Linux x86_64 and arm64 — both builds successful.
https://github.com/f-io/LIVI/actions/runs/22222648514

mmaietta (approved):

PR #9616, opened 2026-02-23
feat(dmg): add `filesystem` option to support APFS dmg format

### LINKED ISSUE #9615: [Bug] DMG build fails on macOS 26 Tahoe beta due to HFS+ mounting issue
## Electron-Builder Version

26.4.0

## Target

macOS (arm64)

---
## Description
DMG builds fail on **macOS 26.4 Tahoe (beta)** with the error `"Unable to attach disk image"` followed by `"hdiutil: attach failed - no
mountable file systems"`.
This is caused by a known issue in the macOS Tahoe beta affecting HFS+ mounting. Apple has acknowledged this in their official release
notes:
> **From [macOS Tahoe 26.4 Beta Release 
Notes](https://developer.apple.com/documentation/macos-release-notes/macos-26_4-release-notes):**
>
> *"External Media - Known Issues: HFS external media might fail to mount automatically. (168672160)"*
The `dmgbuild` library bundled in `dmg-builder` defaults to creating HFS+ sparse images, which triggers this bug.
---
## Error Output

```
⨯ /usr/bin/python3 process failed 1
Exit code: 1
dmgbuild.core.DMGError: Unable to attach disk image:
hdiutil: attach failed - no mountable file systems
```
---

## Root Cause Analysis
The DMG build process in `dmgbuild/core.py`:
1. Creates a writable HFS+ sparse image
2. Attaches the sparse image via `hdiutil attach` ← **fails on macOS Tahoe beta**
3. Populates the mounted volume
4. Detaches and converts to read-only DMG
**Verification:**

HFS+ sparse image - FAILS on macOS 26 Tahoe

```
hdiutil create -size 200m -fs HFS+ -volname "Test" -type SPARSE /tmp/test_hfs.sparseimage
hdiutil attach -nobrowse -owners off /tmp/test_hfs.sparseimage
```

Error: hdiutil: attach failed - no mountable file systems

APFS sparse image - WORKS

```
hdiutil create -size 200m -fs APFS -volname "Test" -type SPARSE /tmp/test_apfs.sparseimage
hdiutil attach -nobrowse -owners off /tmp/test_apfs.sparseimage
```

---
## Proposed Fix
Change the default filesystem from `HFS+` to `APFS` in `dmg-builder/vendor/dmgbuild/core.py`.
### Why APFS is the better default
| Aspect | HFS+ | APFS |
|--------|------|------|
| **Introduced** | 1998 | 2017 (macOS High Sierra) |
| **Default on new Macs** | No (since 2017) | Yes |
| **macOS minimum** | All versions | 10.13+ (High Sierra) |
| **Optimized for** | HDDs | SSDs & Flash storage |
| **Status** | Legacy | Actively developed |
**Additional benefits of APFS:**
- **Future-proof**: Apple has been moving away from HFS+ since 2017. All new Macs ship with APFS.
- **Better performance**: APFS is optimized for modern storage (SSDs, NVMe) which all Apple Silicon Macs use.
- **Wide compatibility**: macOS 10.13+ covers 8+ years of macOS releases (2017-2025), which is sufficient for most Electron app
deployment targets.
- **Avoids beta bugs**: Works around the current macOS Tahoe beta issue and any future HFS+ deprecation Apple may implement.
- **No user impact**: The filesystem is only used during the build process for the temporary sparse image - the final DMG output format
remains unchanged. Electron itself has minimum macOS requirements - Electron 22+ requires macOS 10.13+, so your users already need High Sierra or later

---
## Environment
- **macOS**: 26.4 Tahoe (beta)
- **Architecture**: Apple Silicon (arm64)
- **electron-builder**: 26.4.0
- **dmg-builder**: 26.4.0

---
## Related Issues
- #8415 - hdiutil issues (different error - `hdiutil convert`)
- #5850 - M1 deprecating HFS+
- #5465 - HFS+ issues on Big Sur

### LINKED ISSUE #9706: DMG silently drops files when packaging large apps on HFS+ (dmgbuild size calc + unchecked ditto exit code)
## Summary

`dmg-builder@26.8.1` produces DMGs that are **missing files** for apps in the ~1.9 GB+ range, with no error reported during the build. The resulting DMG installs cleanly but the app crashes on launch because critical Mach-O binaries (in our case the `Electron Framework` binary itself) were silently dropped.

The bug is reproducible with raw shell commands and traces to two specific issues in the vendored `dmgbuild` Python tool.

## Versions

- `electron-builder@26.8.1`
- `dmg-builder@26.8.1` 
- `app-builder-bin@5.0.0-alpha.12`
- Bundled `dmgbuild` from `electron-userland/electron-builder-binaries` release `dmg-builder@1.2.0` (commit `75c8a6c`), running on Python 3.14
- macOS 26.2 (build 25C56), Apple Silicon

## Symptom

A universal Electron app at ~2.1 GB unpacked is built into a DMG that compresses to ~1 GB. Running `hdiutil verify` on the DMG reports valid. Mounting the DMG and listing the .app contents shows **234 files vs 235 in the source** — exactly one missing, the 178 MB / 375 MB `Electron Framework` Mach-O binary at:

```
Keel.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework
```

The directory and symlinks exist; only the binary is gone. App crashes on launch with:

```
Library not loaded: @rpath/Electron Framework.framework/Electron Framework
Reason: tried: '...Electron Framework.framework/Electron Framework' (no such file)
```

## Root cause

Two interacting bugs in the bundled `dmgbuild` (`al45tair/dmgbuild`):

### 1. Size calculation is too small for HFS+

`dmgbuild/core.py:474`:

```python
total_size = str(max(total_size / 1000, 1024)) + "K"
```

`total_size` is in bytes; dividing by 1000 then formatting as `K` (which `hdiutil` interprets as 1024 bytes) gives a requested image size of `bytes × 1.024` — only **~2.4% headroom** over the raw payload. HFS+ catalog overhead for a deeply nested app bundle (Electron .app has 200+ small `lproj` files plus large frameworks) easily exceeds that, leaving no room.

### 2. `ditto` exit code is not checked

`dmgbuild/core.py:693`:

```python
subprocess.call(["/usr/bin/ditto", f, f_in_image])
```

When `ditto` fails partway through with `No space left on device` (which happens reliably given #1), the script proceeds to the next stages — finalizing the DMG, signing, etc. — without any error. Files that didn't make it into the RW image are silently absent from the final DMG.

## Reproduction without electron-builder

Confirms it's the dmgbuild path, not raw `hdiutil`:

```bash
# This succeeds — raw hdiutil is fine
hdiutil create -volname "Test" -srcfolder /path/to/Keel.app -ov -format UDZO /tmp/raw.dmg

# This reproduces the silent ditto failure on HFS+
hdiutil create -size 2g -fs HFS+ -volname "Test HFS" -ov /tmp/test.dmg
hdiutil attach /tmp/test.dmg -mountpoint /tmp/mount -nobrowse
/usr/bin/ditto /path/to/Keel.app /tmp/mount/Keel.app
echo "exit: $?"  # 1, with dozens of "No space left on device" messages
ls -la "/tmp/mount/Keel.app/Contents/Frameworks/Electron Framework.framework/Versions/A/" 
# Framework binary missing or truncated
```

Same image as `-fs APFS`: `ditto` exits 0, copy is complete.

## Why this isn't always visible

- Apps under ~1.5 GB unpacked have enough headroom that the 2.4% margin doesn't get eaten by HFS+ overhead. They build fine.
- The bug ships silent; we discovered it because the resulting DMG crashed on launch with a missing dyld error. Anyone whose app grows past ~1.5–2 GB unpacked will trip this without warning.

## Suggested fixes

In rough order of effort:

1. **Check `ditto`'s exit code** in `core.py:693`. Switch `subprocess.call` to `subprocess.check_call`, or check the return value and raise. This alone would surface the failure loudly instead of producing a silently broken DMG.
2. **Fix the size formula** in `core.py:474`. `bytes / 1000` should be `bytes / 1024`, or better, multiply by a real safety factor (e.g. 1.10) for HFS+ catalog overhead. Or default to `-fs APFS` when unspecified — APFS has dramatically lower overhead and the same image size succeeds where HFS+ fails.

## Workaround for downstream users

Set `dmg.size` explicitly in `electron-builder` config to force a larger image:

```js
dmg: {
  size: '3g',  // or larger; 'shrink: true' (default) shrinks the final DMG anyway
  // ...
}
```

This buys headroom but doesn't fix the underlying unchecked-exit-code or size-formula bugs — a future growth spurt would trip the same silent failure.

## Related

While investigating, I also noticed `dmg-builder/out/dmgUtil.js:122` has the line that would expose the `filesystem` option to user config commented out:

```js
// filesystem: specification.filesystem || "HFS+",
```

If that's intentional, fine; if accidental, it would also be a useful workaround (set `filesystem: 'APFS'` and the size issue largely goes away).

Happy to send a PR for #1 (`check_call`) if useful — that one is uncontroversial and would catch silent failures from any source, not just this size mismatch.

RobbieMinderhoud:
### Problem

Two related issues:

1. **APFS DMGs failed to mount in electron-builder.** `attachAndExecute` located the mounted volume by running `hdiutil info` and filtering lines matching the container device name (e.g. `disk8*`). When an APFS disk image is attached, macOS synthesizes a *separate* disk device (e.g. `disk9`) for the APFS volume — a device that `disk8*` never matches — causing the "Cannot find volume mount path" error at runtime.

2. **The `filesystem` config option was never wired up.** The `filesystem` property was commented out in `customizeDmg`, so the setting had no effect regardless of what users specified.

### Changes

**`packages/dmg-builder/src/dmgUtil.ts`**
- Fixed `attachAndExecute` to find the mounted volume path by scanning the `hdiutil attach` output directly for any `/Volumes/…` path. This works for both HFS+ (volume on `disk8s2`) and APFS (volume on a synthesized `disk9s1`) without needing a separate `hdiutil info` call.
- Removed the now-dead `findMountPath` helper (which relied on the broken device-name approach).
- Activated the previously commented-out `filesystem` line in `customizeDmg`, wiring the user's config through to the `dmgbuild` settings JSON.

**`packages/app-builder-lib/src/options/macOptions.ts` + `scheme.json`**
- Added `filesystem?: string | null` to `DmgOptions`.
- Default is `"HFS+"` — matching dmgbuild.py's own default — so existing builds are unaffected.
- JSDoc notes that the default will change to `"APFS"` in the next major release, encouraging users on macOS 10.13+ to opt in explicitly.

**`test/src/mac/dmgTest.ts`**
- Replaced the single "explicit size" test (which used a very wide 450 MB–1 GB tolerance) with two focused tests:
  - **`explicit size - HFS+`**: builds with `filesystem: "HFS+"`, verifies mounted size is 490–540 MiB (500 MiB target ± small HFS+ metadata overhead).
  - **`explicit size - APFS`**: builds with `filesystem: "APFS"`, verifies mounted size is 490–560 MiB (slightly wider upper bound for APFS container metadata).
- Each test asserts that `specification.filesystem` matches the configured value.

### Why not default to APFS now?

dmgbuild.py defaults to `HFS+`. Matching that default means no behaviour change for existing users, while still making APFS fully supported and usable today. The documentation explicitly signals the upcoming default change so users can prepare.

Resolves #9615


packages/dmg-builder/src/dmgUtil.ts
```diff
@@ -132,7 +132,7 @@ export async function customizeDmg({ appPath, artifactPath, volumeName, specific
     "text-size": iconTextSize,
 
     "compression-level": Number(process.env.ELECTRON_BUILDER_COMPRESSION_LEVEL || "9"),
-    // filesystem: specification.filesystem || "HFS+",
```
mmaietta, line 150: Technically, this is a breaking change. From what I understand though, we do need this in order to resolve dmg builds on latest macos, so I'm in favor of proceeding.

Just to play it safe, I looked up the release+deprecation dates of HFS and APFS. APFS is almost a decade old. I think we can just pull the trigger and transition to APFS
> APFS (Apple File System) cannot natively run or be read by Windows, Linux, or macOS versions older than 10.13 High Sierra

I don't think we need to worry about High Sierra

RobbieMinderhoud: I thought i'd be a breaking change too. But this version of the electron builder does not support High Sierra on its self. So all possible users are already compatible with this file system.

```
Electron itself has minimum macOS requirements - Electron 22+ requires macOS 10.13+
```

mmaietta: Thanks for the very quick reply! Sounds good, I'm aligned.

Kicking off CI, will get this merged soon (probably today if all pass!)

mmaietta (changes requested):
Please resolve tsc errors when you have a chance and we can quickly move forward with this PR. Thanks!

RobbieMinderhoud:
@mmaietta Sorry, it's been a while. Was working with a local patch. This is fixed now. Let's hope the tests pass. I'll check tomorrow.

RobbieMinderhoud:
Currently making a fix for the last test

RobbieMinderhoud:
@mmaietta Would you happen to have time to look at it again?

mmaietta (dismissed):

RobbieMinderhoud:
THank you!

Fixing the tests currently

RobbieMinderhoud:
@mmaietta Increased the size of DMGTest.js, APFS can be larger.

RobbieMinderhoud:
@mmaietta Increasing the size, just a little more... Let's hope this will do it.

test/src/mac/dmgTest.ts
```diff
@@ -127,9 +127,10 @@ describe.heavy.ifMac("dmg", { sequential: true }, () => {
             const totalBytes = stats.bsize * stats.blocks
 
             // 500m should give ~524,288,000 bytes (500 * 1024 * 1024)
-            // Allow margin for filesystem overhead (450MB to 600MB)
+            // Allow margin for filesystem overhead. APFS can report larger sizes than HFS+ due to
+            // different block sizing and allocation, so use a wide range (450MB–1GB).
             const minBytes = 450 * 1024 * 1024
-            const maxBytes = 600 * 1024 * 1024
+            const maxBytes = 1024 * 1024 * 1024
```
mmaietta: This is a huuuge new range for just changing filesystem. Is the dmg size being minimized?
Maybe we can do 2 tests, one for HFS and another for APFS

RobbieMinderhoud: Could be a good one. Let's try

RobbieMinderhoud: Ur right, lemme see what I can do.

mmaietta:
Hi @RobbieMinderhoud , have you had a chance to investigate on this further?

mmaietta:
Update here, going to pick this PR up and carry it across the finish line 🙃

PR #9618, opened 2026-02-24
fix: handle pnpm deduped dependencies correctly in node module collector

### LINKED ISSUE #9603: Builds broken after upgrading pnpm 10.28.2->10.29.3/10.30.0
Same symptoms as https://github.com/electron-userland/electron-builder/issues/9394

May be related to https://github.com/pnpm/pnpm/pull/10586
May be fixed by https://github.com/pnpm/pnpm/releases/tag/v10.30.0

Image

Debug build log: https://github.com/element-hq/element-desktop/actions/runs/22184882165/job/64155774962?pr=2866#step:15:164

beyondkmp:
fix #9603
Newer pnpm versions mark deduped dependencies with `dedupedDependenciesCount` instead of including nested dependency subtrees. Fix several bugs in the dedup handling:

- Use `tree.from` instead of `tree.name` for dedup lookup key, since child dependency objects don't have a `name` field in pnpm's JSON output
- Skip deduped entries before calling `locatePackageVersion` to avoid wasted filesystem lookups
- Add dedup skip check to optionalDependencies loop (was only in dependencies)
- Remove dead `dedupedDependenciesCount: pkg?.dedupedDependenciesCount` that always evaluated to undefined and overwrote the correct spread value
- Make `dedupedDependenciesCount` optional in PnpmDependency type since older pnpm versions and non-deduped entries don't have this field


packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -19,6 +19,17 @@ export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependenc
     }
     this.productionGraph[dependencyId] = { dependencies: [] }
 
+    if (tree.dedupedDependenciesCount && tree.dedupedDependenciesCount > 0) {
+      const depKey = tree.from || tree.name
+      const realDep = this.allDependencies.get(`${depKey}@${tree.version}`)
+      if (realDep) {
+        tree = realDep
+      } else {
+        log.error(`Could not find real dependency for ${depKey}@${tree.version}`)
```
mmaietta: I added a "summary" tracking/logger to the `moduleManager.ts` so that we can aggregate missed/non-installed/unresolved dependencies into a few lines logged after node_module collection completes.

Please consider utilizing that; feel free to create a new enum if you feel the current one's aren't adequate. (I may need to consolidate them anyhow though)

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -19,6 +19,17 @@ export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependenc
     }
     this.productionGraph[dependencyId] = { dependencies: [] }
 
+    if (tree.dedupedDependenciesCount && tree.dedupedDependenciesCount > 0) {
+      const depKey = tree.from || tree.name
+      const realDep = this.allDependencies.get(`${depKey}@${tree.version}`)
```
mmaietta: I recall previous issues occurring from using `tree.from`. Won't this pull the wrong package name that doesn't correlate with what is in the package.json `key-value` pair?

packages/app-builder-lib/src/node-module-collector/types.ts
```diff
@@ -28,6 +28,7 @@ export type ParsedDependencyTree = {
 export interface PnpmDependency extends Dependency<PnpmDependency, PnpmDependency> {
   readonly from: string
   readonly resolved: string
+  readonly dedupedDependenciesCount?: number
```
mmaietta, line 33: Might be a convenient time to also add a comment as to this property being added in `10.29.3`

Also, curious question. Why are we using `dedupedDependenciesCount` instead of the `true/false` `deduped` property?

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
beyondkmp: Replaced inline pnpm deduped unresolved logging with ModuleManager summary aggregation.
Added a dedicated summary key:

PKG_DUPLICATE_REF_UNRESOLVED = "unresolved duplicate dependency references"
mapped to warn level.

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
beyondkmp: Good Catch. Updated pnpm deduped resolution lookup to use dependencyId directly (instead of rebuilding key from name/from + version) to avoid alias/key mismatch.

packages/app-builder-lib/src/node-module-collector/types.ts
beyondkmp: Great suggestion — I’ve added a comment in [types.ts](app://-/index.html#) noting this field appears in pnpm [10.29.3+](app://-/index.html#), and linked the upstream context: https://github.com/pnpm/pnpm/issues/10601.
We’re currently using dedupedDependenciesCount because our logic is threshold-based (> 0) and the count also provides better diagnostic context than a pure boolean flag.
If you prefer, I can also include deduped?: boolean as a companion field and treat either signal as deduped.

mmaietta: Thank you!

Nah, this is good as-is, was just curious as to what the intent was of using one measurement/flag versus the other.

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -59,13 +70,19 @@ export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependenc
   protected async collectAllDependencies(tree: PnpmDependency) {
     // Collect regular dependencies
     for (const [key, value] of Object.entries(tree.dependencies || {})) {
+      if (value?.dedupedDependenciesCount && value.dedupedDependenciesCount > 0) {
```
diogotcorreia:
```suggestion
      if ((value?.dedupedDependenciesCount ?? 0) > 0) {
```

Nitpick: make this consistent with line 22

packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -59,13 +70,19 @@ export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependenc
...
+        continue
+      }
       const pkg = await this.cache.locatePackageVersion({ pkgName: key, parentDir: this.rootDir, requiredRange: value.version })
       this.allDependencies.set(`${key}@${value.version}`, { ...value, path: pkg?.packageDir ?? value.path })
       await this.collectAllDependencies(value)
     }
 
     // Collect optional dependencies if they exist
     for (const [key, value] of Object.entries(tree.optionalDependencies || {})) {
+      if (value?.dedupedDependenciesCount && value.dedupedDependenciesCount > 0) {
```
diogotcorreia:
```suggestion
      if ((value?.dedupedDependenciesCount ?? 0) > 0) {
```

ditto

diogotcorreia (commented):
Small nitpick, feel free to ignore

t3chguy:
Tested using https://github.com/element-hq/element-desktop/pull/2866/ and tests are passing :shipit: https://github.com/element-hq/element-desktop/actions/runs/22487430030/job/65141033093

PR #9626, opened 2026-03-02
feat(electron-publish): add targetCommitish support for GitHub releases

EYadroshnikov:
### Summary

Adds support for specifying which branch or commit GitHub releases are created from. By default, GitHub uses the default branch (e.g. `main`); this change allows targeting a specific branch (e.g. `development`) or a commit SHA.

### Motivation

Useful when:
- Releasing from non-default branches (e.g. `development`, `release/1.0`)
- Creating releases from the exact commit built in CI (e.g. `GITHUB_SHA` in GitHub Actions)

### Changes

- **`GithubOptions`** (`builder-util-runtime`): Added optional `targetCommitish` field
- **`GitHubPublisher`** (`electron-publish`): Resolves `targetCommitish` from config, `EP_TARGET_COMMITISH`, or `GITHUB_SHA` env vars and passes it as `target_commitish` in the GitHub API request
- **Schema** (`app-builder-lib/scheme.json`): Added `targetCommitish` property
- **Documentation** (`pages/publish.md`): Added usage examples and env var notes
- **Tests**: New `GitHubPublisherTest.ts` covering config, env vars, and precedence

### Usage

**Config:**

```
{
  "build": {
    "publish": {
      "provider": "github",
      "owner": "my-org",
      "repo": "my-app",
      "targetCommitish": "development"
    }
  }
}

```

### DIFF
--- a/packages/electron-publish/src/gitHubPublisher.ts
+++ b/packages/electron-publish/src/gitHubPublisher.ts
@@ -27,6 +27,7 @@ interface Asset {
 
 export class GitHubPublisher extends HttpPublisher {
   private readonly tag: string
+  private readonly targetCommitish: string | undefined
   readonly _release = new Lazy(() => (this.token === "__test__" ? Promise.resolve(null as any) : this.getOrCreateRelease()))
 
   private readonly token: string
@@ -66,6 +67,11 @@ export class GitHubPublisher extends HttpPublisher {
     }
 
     this.tag = githubTagPrefix(info) + version
+
+    this.targetCommitish = info.targetCommitish ?? process.env.EP_TARGET_COMMITISH ?? process.env.GITHUB_SHA ?? undefined
+    if (this.targetCommitish != null && this.targetCommitish.length > 0) {
+      log.info({ targetCommitish: this.targetCommitish }, "GitHub release target_commitish is set")
+    }
 
     if (isEnvTrue(process.env.EP_DRAFT)) {
       this.releaseType = "draft"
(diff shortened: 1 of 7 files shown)


EYadroshnikov:
@mmaietta, this is ready for review whenever you have some time. Thanks!

packages/electron-publish/src/gitHubPublisher.ts
```diff
@@ -67,6 +68,11 @@ export class GitHubPublisher extends HttpPublisher {
 
     this.tag = githubTagPrefix(info) + version
 
+    this.targetCommitish = info.targetCommitish ?? process.env.EP_TARGET_COMMITISH ?? process.env.GITHUB_SHA ?? undefined
```
mmaietta, line 71: I'm hesitant to add the default of `process.env.EP_TARGET_COMMITISH ?? process.env.GITHUB_SHA` since this is completely net-new functionality and likely not expected by another dev.
I consider new functionality that is extracted from env vars to be breaking changes.

I'd suggest only using `info.targetCommitish` and then in the tsdoc, specify an example (or recommended) value.

packages/electron-publish/src/gitHubPublisher.ts
```diff
@@ -230,12 +236,16 @@ export class GitHubPublisher extends HttpPublisher {
...
   private createRelease() {
-    return this.githubRequest<Release>(`/repos/${this.info.owner}/${this.info.repo}/releases`, this.token, {
+    const body: Record<string, any> = {
       tag_name: this.tag,
       name: this.version,
       draft: this.releaseType === "draft",
       prerelease: this.releaseType === "prerelease",
-    })
+    }
+    if (this.targetCommitish != null && this.targetCommitish.length > 0) {
```
mmaietta, line 245: Nit: We can simplify this to `!!this.targetCommitish?.length` (or use `isEmptyOrSpaces` from builder-util)

mmaietta (changes requested):
Thanks for adding docs/example! I left a comment re: this being a breaking change since we're adding conditional functionality that impacts prod tagging based on environment variables.
In order to proceed with this PR, I'd advocate for reading only from the config property and then allow it to be opt-in for v26.

Happy to discuss making it the default logic for v27 though!

PR #9628, opened 2026-03-04
test: stabilize HoistedNodeModuleTest lockfile fixtures and snapshots

beyondkmp:
## Summary
- make lockfile fixture lookup resilient to Vitest test names that include suite prefixes
- apply frozen/immutable install flags when a lockfile fixture is used
- allow UPDATE_LOCKFILE_FIXTURES=true to overwrite existing fixture files
- refresh HoistedNodeModuleTest lockfile fixtures and update related snapshots

## Why
- fix snapshot drift caused by lockfile fixture misses and transitive dependency movement
- keep dependency installs deterministic when fixture lockfiles are present

## Notes
- remaining failing cases are environment-specific (wine64 binary compatibility on local macOS arm64) and are not changed by this PR

### DIFF
--- a/test/snapshots/BuildTest.js.snap
+++ b/test/snapshots/BuildTest.js.snap
@@ -1362,6 +1362,9 @@ exports[`posix smart unpack 3`] = `
             "SECURITY.md": {
               "size": "<size>",
             },
+            "empty.js": {
+              "size": "<size>",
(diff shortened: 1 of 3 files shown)


mmaietta:
Many thanks @beyondkmp !!

PR #9651, opened 2026-03-23
feat(appimage): add GPG signing support for AppImage builds

### LINKED ISSUE #9642: Support for signing AppImages
Hey all,

We sign our Windows, MacOS, DEB and RPM builds, but we aren't able to sign our app images.

Signing AppImage files is supported by the official `appimagetool`:
https://docs.appimage.org/packaging-guide/optional/signatures.html

It should theoretically be possible to implement the same logic.

rathboma:
## Summary

- Adds GPG signing support for AppImage builds, following the [AppImage signing specification](https://docs.appimage.org/packaging-guide/optional/signatures.html)
- Embeds an ASCII-armored GPG detached signature into the `.sha256_sig` ELF section of the AppImage runtime
- Supports configuration via `appImage.sign` config object or environment variables (`APPIMAGE_SIGN`, `APPIMAGE_GPG_KEY_ID`, `APPIMAGE_GPG_PATH`, `APPIMAGE_GPG_PASSPHRASE`)
- Optionally produces a detached `.sig` file alongside the AppImage

Closes #9642

## Changes

| File | Description |
|------|-------------|
| `linuxOptions.ts` | New `AppImageSignOptions` interface and `sign` property on `AppImageOptions` |
| `elfSigSection.ts` | Pure Node.js ELF header parser to locate the `.sha256_sig` section |
| `appImageSign.ts` | Signing orchestrator: zeros section, GPG signs, writes signature back |
| `AppImageTarget.ts` | Integration into build flow after `stageDir.cleanup()` |
| `index.ts` | Export `AppImageSignOptions` |
| `scheme.json` | Regenerated to include new config options |
| `appImageSignTest.ts` | Unit tests for ELF parser + integration test that builds and verifies a signed AppImage |

## Usage

```json
{
  "appImage": {
    "sign": {
      "gpgKeyId": "ABCD1234",
      "detachedSigFile": true
    }
  }
}
```

Or via environment variables:
```bash
APPIMAGE_SIGN=true APPIMAGE_GPG_KEY_ID=ABCD1234 electron-builder --linux appimage
```

## Test plan

- [x] Unit tests for ELF section parser (find section, missing section, non-ELF file)
- [x] Unit tests for `zeroSigSection` and `writeSigSection`
- [x] Unit tests for `signAppImage` skip behavior
- [x] Integration test: builds AppImage with signing, verifies `.sha256_sig` section contains PGP signature, verifies `gpg --verify` succeeds, verifies `--appimage-signature` outputs the signature
- [x] Existing `linuxPackagerTest` suite passes without regressions (26 tests)
- [x] Manual validation: built AppImage, confirmed `readelf -x .sha256_sig` shows signature data, `gpg --verify` returns "Good signature"

### DIFF
--- a/packages/app-builder-lib/src/targets/appimage/appImageSign.ts
+++ b/packages/app-builder-lib/src/targets/appimage/appImageSign.ts
@@ -60,6 +60,17 @@ function resolveSignConfig(options: AppImageOptions): ResolvedSignConfig | null 
 /**
  * Signs an AppImage by embedding a GPG detached signature into its `.sha256_sig` ELF section,
  * following the AppImage signing specification.
+ *
+ * Specification: https://docs.appimage.org/packaging-guide/optional/signatures.html
+ * ELF section layout: https://github.com/AppImage/AppImageSpec/blob/master/draft.md#signatures
+ *
+ * The process:
+ * 1. Locate the `.sha256_sig` section (pre-allocated in the AppImage runtime, typically 1024 bytes)
+ * 2. Zero the section so the signature covers the file without a previous signature
+ * 3. Create an ASCII-armored GPG detached signature of the file
+ * 4. Write the signature back into the `.sha256_sig` section
+ *
+ * The resulting AppImage can be verified with: ./<name>.AppImage --appimage-signature
  */
 export async function signAppImage(appImagePath: string, options: AppImageOptions): Promise<void> {
   const config = resolveSignConfig(options)
--- a/packages/app-builder-lib/src/targets/appimage/appImageSign.ts
+++ b/packages/app-builder-lib/src/targets/appimage/appImageSign.ts
@@ -1,0 +1,129 @@
+import { exec, log } from "builder-util"
+import * as fs from "fs-extra"
+import * as os from "os"
+import * as path from "path"
+import { AppImageOptions, AppImageSignOptions } from "../../options/linuxOptions"
+import { findSha256SigSection, zeroSigSection, writeSigSection } from "./elfSigSection"
+
+interface ResolvedSignConfig {
+  gpgPath: string
+  gpgKeyId: string | null
+  passphrase: string | null
+  detachedSigFile: boolean
+}
+
+function isSigningRequested(options: AppImageOptions): boolean {
+  if (process.env.APPIMAGE_SIGN === "true") {
+    return true
+  }
+  return options.sign != null && options.sign !== false
+}
+
+async function findGpgBinary(configured: string | null | undefined): Promise<string> {
+  if (configured) {
+    return configured
+  }
+
+  const envPath = process.env.APPIMAGE_GPG_PATH
+  if (envPath) {
+    return envPath
+  }
+
+  // Try gpg2 first, then gpg
+  for (const candidate of ["gpg2", "gpg"]) {
+    try {
+      await exec("which", [candidate])
+      return candidate
+    } catch {
+      // not found, try next
+    }
+  }
+
+  throw new Error("GPG binary not found. Install gpg2 or gpg, or set gpgPath / APPIMAGE_GPG_PATH.")
+}
+
+function resolveSignConfig(options: AppImageOptions): ResolvedSignConfig | null {
+  if (!isSigningRequested(options)) {
+    return null
+  }
+
+  const signOpts: AppImageSignOptions = typeof options.sign === "object" && options.sign != null ? options.sign : {}
+
+  return {
+    gpgPath: "", // resolved async later
+    gpgKeyId: signOpts.gpgKeyId ?? process.env.APPIMAGE_GPG_KEY_ID ?? null,
+    passphrase: signOpts.passphrase ?? process.env.APPIMAGE_GPG_PASSPHRASE ?? null,
+    detachedSigFile: signOpts.detachedSigFile ?? false,
+  }
+}
+
+/**
+ * Signs an AppImage by embedding a GPG detached signature into its `.sha256_sig` ELF section,
+ * following the AppImage signing specification.
+ */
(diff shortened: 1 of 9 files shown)


packages/app-builder-lib/src/targets/appimage/appImageSign.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  try {
+    const gpgArgs: string[] = ["--batch", "--yes", "--armor", "--detach-sign"]
+
+    if (config.gpgKeyId) {
+      gpgArgs.push("--default-key", config.gpgKeyId)
+    }
+
+    if (config.passphrase) {
+      gpgArgs.push("--passphrase", config.passphrase, "--pinentry-mode", "loopback")
```
mmaietta, line 114: Can you verify that the passphrase doesn't leak in console logs, `proc/<pid>/cmdline`, or other console history with this logic flow?

Alternatively, we could use `--passphrase-file <tmpfile>` with a temporary file?

packages/app-builder-lib/src/targets/appimage/appImageSign.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  // Create detached signature using GPG
+  const tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), "appimage-sig-"))
+  const sigFile = path.join(tmpDir, "signature.sig")
+
+  try {
+    const gpgArgs: string[] = ["--batch", "--yes", "--armor", "--detach-sign"]
+
+    if (config.gpgKeyId) {
+      gpgArgs.push("--default-key", config.gpgKeyId)
```
mmaietta, line 110: From what I understand, `--default-key` sets the default for the session but can still be overridden by `gpg-agent` defaults.

I think this should be `--local-user` for avoid a build silently being signed with the wrong key

packages/app-builder-lib/src/targets/appimage/elfSigSection.ts
```diff
@@ -0,0 +1,168 @@
...
+      shstrndx = readUint16(identBuf, 50, le)
+    }
+
+    if (shoff === 0 || shnum === 0 || shstrndx >= shnum) {
+      return null
+    }
+
+    // Read all section headers
+    const shTableSize = shnum * shentsize
+    const shTableBuf = Buffer.alloc(shTableSize)
```
mmaietta, line 94: Can we add a quick sanity check/guard before allocating the buffer to make sure the values are valid?
```
if (shentsize < 40 || shnum > 4096 || shTableSize > 16 * 1024 * 1024) {
  return null
}
```

packages/app-builder-lib/src/targets/appimage/appImageSign.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  const envPath = process.env.APPIMAGE_GPG_PATH
+  if (envPath) {
+    return envPath
+  }
+
+  // Try gpg2 first, then gpg
+  for (const candidate of ["gpg2", "gpg"]) {
+    try {
+      await exec("which", [candidate])
```
mmaietta, line 35: I ran into a previous issue where `which` wasn't supported on a dev's machine.

Please adjust this to utilize `command -v` (or the `which` npm pkg that should already be in the app-builder-lib deps)

packages/app-builder-lib/src/targets/appimage/appImageSign.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  // Find .sha256_sig section
+  const section = await findSha256SigSection(appImagePath)
+  if (section == null) {
+    log.warn({ file: path.basename(appImagePath) }, "AppImage runtime does not contain a .sha256_sig section — skipping signing")
+    return
+  }
+
+  // Zero the signature section before signing (spec requirement)
+  const fd = await fs.open(appImagePath, "r+")
```
mmaietta, line 95: Instead of writing to the appimage file in-place, could you please take this atomic approach to avoid the appimage becoming corrupted (if signing were to error/fail/cancel at any point).

Copy appimage to the temp dir -> open fd on temp appimage -> zeroSig -> gpg -> writeSig -> copy/overwrite temp appimage back to final `appImagePath`

test/snapshots/linux/appImageSignTest.js.snap
```diff
@@ -0,0 +1,24 @@
...
+      "updateInfo": {
+        "blockMapSize": "@blockMapSize",
+        "sha512": "@sha512",
+        "size": "@size",
+      },
+    },
+    {
+      "arch": "x64",
+      "file": "Test App ßW-1.1.0.AppImage.sig",
+      "safeArtifactName": "TestApp-1.1.0-x86_64.AppImage",
```
mmaietta, line 19: Can you double-check this?

I noticed that the `file` is `Test App ßW-1.1.0.AppImage.sig` with `.sig` suffix/ext. But `safeArtifactName` is missing the `.sig` suffix.

packages/app-builder-lib/src/targets/appimage/appImageSign.ts
```diff
@@ -0,0 +1,140 @@
...
+
+interface ResolvedSignConfig {
+  gpgPath: string
+  gpgKeyId: string | null
+  passphrase: string | null
+  detachedSigFile: boolean
+}
+
+function isSigningRequested(options: AppImageOptions): boolean {
+  if (process.env.APPIMAGE_SIGN === "true") {
```
mmaietta, line 16: Could you please remove this env var override?

mmaietta (changes requested):
Awesome work! Thank you for your contribution.

Apologies on the delayed review; left a few comments

PR #9671, opened 2026-04-07
chore(builder-util): add engines field to package.json

gameroman:
This makes it easier for consumers to know what to expect as far as compatibility goes. Both human consumers and static analysis tools.

### DIFF
--- a/packages/builder-util/package.json
+++ b/packages/builder-util/package.json
@@ -41,6 +41,6 @@
     "@types/tiny-async-pool": "^1.0.5"
   },
   "engines": {
-    "node": ">=22"
+    "node": ">=16"
   }
 }
--- a/packages/builder-util/package.json
+++ b/packages/builder-util/package.json
@@ -39,5 +39,8 @@
     "@types/js-yaml": "4.0.3",
     "@types/source-map-support": "0.5.4",
     "@types/tiny-async-pool": "^1.0.5"
+  },
+  "engines": {
+    "node": ">=22"
   }
 }
--- a/packages/builder-util/package.json
+++ b/packages/builder-util/package.json
@@ -41,6 +41,6 @@
     "@types/tiny-async-pool": "^1.0.5"
   },
   "engines": {
-    "node": ">=16"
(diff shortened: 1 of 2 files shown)


packages/builder-util/package.json
```diff
@@ -39,5 +39,8 @@
     "@types/js-yaml": "4.0.3",
     "@types/source-map-support": "0.5.4",
     "@types/tiny-async-pool": "^1.0.5"
+  },
+  "engines": {
+    "node": ">=22"
```
mmaietta: This looks to be a breaking change as I think builder-util still can support older versions of node (or at least, no compiler errors when I set to node `>=16`.

I can state that the bump to node 22 minimum will happen in electron-builder v27 (development in progress as it also requires migrating this project to ESM)

mmaietta (changes requested):

packages/builder-util/package.json
gameroman:
```suggestion
    "node": ">=16"
```

gameroman: Updated 👍

gameroman: > (development in progress as it also requires migrating this project to ESM)

Let me know if I could help with that in any way

packages/builder-util/package.json
```diff
@@ -39,5 +39,8 @@
     "@types/js-yaml": "4.0.3",
     "@types/source-map-support": "0.5.4",
     "@types/tiny-async-pool": "^1.0.5"
+  },
+  "engines": {
+    "node": ">=16"
```
mmaietta: Was doing some additional checks, app-builder-lib and electron-builder both are `node>=14.0.0` right now. We should mirror the same values here. Could you please make that edit?

gameroman:
```suggestion
    "node": ">=14.0.0"
```

gameroman: Done 👍

mmaietta (dismissed):

mmaietta (approved):

PR #9680, opened 2026-04-14
feat(electron-publish): add support for a read/write token to publish to a private gitlab repository

andreamarelli:
Using the same approach followed by [`GitHubPublisher`](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-publish/src/gitHubPublisher.ts#L52), this PR add support for `GITLAB_RELEASE_TOKEN` token to be used during publishing to a private repository, where the token requires write permissions.

### DIFF
--- a/packages/electron-publish/src/gitlabPublisher.ts
+++ b/packages/electron-publish/src/gitlabPublisher.ts
@@ -36,9 +36,9 @@ export class GitlabPublisher extends HttpPublisher {
   ) {
     super(context, true)
 
-    let token = info.token || null
-    if (isEmptyOrSpaces(token)) {
-      token = process.env.GITLAB_TOKEN || null
+    let token = info.token
+    if (isEmptyOrSpaces(token) || process.env.GITLAB_RELEASE_TOKEN) {
+      token =  process.env.GITLAB_RELEASE_TOKEN ?  process.env.GITLAB_RELEASE_TOKEN : process.env.GITLAB_TOKEN || null
       if (isEmptyOrSpaces(token)) {
         throw new InvalidConfigurationError(`GitLab Personal Access Token is not set, neither programmatically, nor using env "GITLAB_TOKEN"`)
       }


packages/electron-publish/src/gitlabPublisher.ts
```diff
@@ -36,9 +36,9 @@ export class GitlabPublisher extends HttpPublisher {
   ) {
     super(context, true)
 
-    let token = info.token || null
-    if (isEmptyOrSpaces(token)) {
-      token = process.env.GITLAB_TOKEN || null
+    let token = info.token
+    if (isEmptyOrSpaces(token) || process.env.GITLAB_RELEASE_TOKEN) {
+      token =  process.env.GITLAB_RELEASE_TOKEN ?  process.env.GITLAB_RELEASE_TOKEN : process.env.GITLAB_TOKEN || null
```
mmaietta, line 41: Could you please update the GitlabOptions description about this new env var?
https://github.com/electron-userland/electron-builder/blob/96d782af6172a3be519785251e3c13e1873e4fee/packages/builder-util-runtime/src/publishOptions.ts#L168-L170
Also, is this intended to always override the GITLAB_TOKEN env var even if the RELEASE_TOKEN is present? This logic seems odd to me but at least should be described in the docs.

After updating the `publishOptions.ts` tsdocs, please run `pnpm generate-all` for the schema and site docs.

packages/electron-publish/src/gitlabPublisher.ts
```diff
@@ -36,9 +36,9 @@ export class GitlabPublisher extends HttpPublisher {
...
       if (isEmptyOrSpaces(token)) {
         throw new InvalidConfigurationError(`GitLab Personal Access Token is not set, neither programmatically, nor using env "GITLAB_TOKEN"`)
```
mmaietta, line 43: This error message should be updated to mention the `GITLAB_RELEASE_TOKEN` as well.

mmaietta (changes requested):
Please also generate a changeset file via `pnpm generate-changeset` so this is properly picked up by the CI/CD pipeline

PR #9683, opened 2026-04-20
fix: bundling source files when using extraMetadata.name in Yarn Berry

### LINKED ISSUE #9619: [regression] electron-builder 26.3.1 pulls entire source code to `app.asar.unpacked` when using `extraMetadata.name`
We use monorepo + yarn 4. Additionally, we override the app name using `extraMetadata.name` as suggested in https://github.com/electron-userland/electron-builder/issues/4727#issuecomment-642625022.

When upgrading to electron-builder 26.3.1, that was released in response to #9350 by @mmaietta, I've noticed the entire app folder is pulled into `app.asar.unpacked`.

I've created a reproduction repository with additional information for reproduction: https://github.com/DCzajkowski/electron-builder-monorepo-codesign-reproduction.

DCzajkowski:
Fixes #9619

I've added a test that fails on `master` and succeeds with the fix.

cc: @mmaietta @gtluszcz


test/fixtures/test-app-yarn-workspace/packages/test-app/package.json
```diff
@@ -19,7 +19,7 @@
     }
   },
   "dependencies": {
```
DCzajkowski, line 21: Yarn Berry automatically sorts dependencies, so running the test caused this to be sorted.

DCzajkowski:
AFAIK the failed tests are also failing on `master` and are not related to this PR.

packages/app-builder-lib/src/packager.ts
```diff
@@ -109,6 +109,12 @@ export class Packager {
     return this._metadata!
   }
 
+  /** Stores original "name" value from package.json before merging with extraMetadata. */
+  private _originalPackageJsonName: string | undefined
+  get originalPackageJsonName(): string | undefined {
+    return this._originalPackageJsonName
+  }
```
mmaietta, line 122: I think this might be too specific to this usecase and could warrant a different approach.

What if we stored the original metadata and the extra metadata separately? Psuedocode-ish / untested:

```
// First we remove :
this._originalPackageJsonName = this.metadata.name
deepAssign(this.metadata, configuration.extraMetadata)
// Replace with:
this._originalMetadata = this.metadata


// Accessed via:
private _originalMetadata: Metadata | null = null
get originalMetadata() : Metadata {
	return this._originalMetadata!
}
get extraMetadata() : Metadata | undefined {
	return this.configuration.extraMetadata
}
get metadata(): Metadata {
	return deepAssign({}, this.metadata, this.extraMetadata ?? {})
}
get nodePackageName() {
	return this.originalMetadata.name || this.extraMetadata?.name!
}

// Used like:

deps = await collector.getNodeModules({
	packageName: packager.nodePackageName
})
```

mmaietta (commented):
Apologies on the delayed review. Added a comment as I think we can optimize this for more scalability/flexibility/simplicity.

packages/app-builder-lib/src/packager.ts
DCzajkowski: 1. 

```ts
get metadata(): Metadata {
	return deepAssign({}, this.metadata, this.extraMetadata ?? {})
}
```

I think computing metadata every time it's accessed can be slow and memory intensive. I'd keep the computation to occur only the first time.

2.

```ts
get nodePackageName() {
	return this.originalMetadata.name || this.extraMetadata?.name!
}
```

I think `originalMetadata.name` should be always there (I don't imagine there not being a "name" property in `package.json`). So I think this should be enough:

```ts
get nodePackageName() {
	return this.originalMetadata.name! // here we could throw instead of using the non-null assertion operator
}
```

---

What do you think?

mmaietta: This all sounds reasonable. Please proceed with the proposed changes you've mentioned and then I think we should be good to go.

packages/app-builder-lib/src/packager.ts
```diff
@@ -388,7 +400,8 @@ export class Packager {
     } else {
       this._metadata = await this.readProjectMetadataIfTwoPackageStructureOrPrepacked(appPackageFile)
     }
-    deepAssign(this.metadata, configuration.extraMetadata)
+    this._originalMetadata = deepAssign({}, this._metadata)
```
DCzajkowski, line 403: I have to clone here, as the metadata will be mutated in the next line

packages/app-builder-lib/src/packager.ts
```diff
@@ -388,7 +400,8 @@ export class Packager {
...
+    deepAssign(this._metadata, configuration.extraMetadata)
```
DCzajkowski, line 404: I've changed the lvalue to be the property itself, as it's weird to mutate a value returned from a getter.

packages/app-builder-lib/src/packager.ts
```diff
@@ -104,11 +104,23 @@ export class Packager {
     return (await (await this._packageManager.value).workspaceRoot) || this.projectDir
   }
 
+  /** Stores original metadata merged with extraMetadata from configuration. */
   private _metadata: Metadata | null = null
   get metadata(): Metadata {
     return this._metadata!
   }
 
+  /** Stores original metadata from package.json before merging with extraMetadata. */
```
DCzajkowski, line 113: I've added doc blocks to explain what each property holds. Please let me know if I should remove them for consistency with other properties in this project.

packages/app-builder-lib/src/packager.ts
DCzajkowski: I've applied the changes in e13fe30

mmaietta (approved):

Issue #3598, opened 2019-01-10
Add the ability to configure the installPrefix for linux


jasunbourne:
* **Version**: 20.38.4

* **Target**: linux rpm

I know that the fpm configuration allows for adding prefixes, but the install prefix of "/opt" is still added to the end of the prefix from fpm. I believe adding a configurable option to set the install prefix from electron-builder would make this a much simpler process.

PR #9685, opened 2026-04-21
feat(linux): add installDirectory config option

rdn32:
Allow the installation directory used in Linux builds - which still defaults to being /opt/${appInfo.sanitizedProductName} - to be configured.

### DIFF
--- a/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
+++ b/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
@@ -6,8 +6,6 @@ import { LinuxPackager } from "../linuxPackager"
 import { LinuxTargetSpecificOptions } from "../options/linuxOptions"
 import { IconInfo } from "../platformPackager"
 
-export const installPrefix = "/opt"
-
 export class LinuxTargetHelper {
   private readonly iconPromise = new Lazy(() => this.computeDesktopIcons())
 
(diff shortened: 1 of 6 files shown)


rdn32:
The motivating case here would be something like a product being called "My App", but wanting it to be installed to `/opt/myco/my-app` rather than `/opt/My App`.

Installing under a `/opt/<provider/` directory is definitely a thing: I've seen chome and Edge installed under `/opt/google/chome` and `/opt/microsoft/msedge` respectively, the Filesystem Hierarchy Standard allows for this sort of thing.

I'm guessing this is what might have motivated https://github.com/electron-userland/electron-builder/issues/3598.

packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
```diff
@@ -25,6 +23,11 @@ export class LinuxTargetHelper {
     return this.mimeTypeFilesPromise.value
   }
 
+  getInstallDirectory(options: LinuxTargetSpecificOptions): string {
+    const installDirectory = this.packager.expandMacro(options.installDirectory ?? "/opt/${productName}", null)
```
mmaietta, line 27: Should this be `/opt/${sanitizedProductName}`?

Also, if so, should we sanitize the `installDirectory` as well?

rdn32: > Should this be `/opt/${sanitizedProductName}`?

I don't think so. I believe that in this case, `expandMacro` is being called via `PlatformPackager` which means that `isProductNameSanitized` defaults to `true` - which means when `${productName}`  is expanded `appInfo.santizedProductName` is used.

> Also, if so, should we sanitize the `installDirectory` as well?

Sanitizing would strip out directory separators.

PR #9719, opened 2026-05-09
fix: select matching pnpm dependency tree

### LINKED ISSUE #9711: Packaged app dependencies are missing when using pnpm v11
**Electron-Builder Version:** 26.9.1
**Node Version:** 24.15.0
**Electron Version:** 41.1.1
**pnpm Version:** 11.0.8

After upgrading pnpm from v10.31.1 to v11.0.8, our packaged Electron app fails to start because a runtime dependency is not included in the packaged app:

```text
Uncaught Exception:
Error: Cannot find module '<abc>'
```
Our project uses pnpm workspaces, with the Electron app located in one workspace package.
From debugging, it looks like `electron-builder` collects production dependencies by running:
```
pnpm list --prod --json --depth Infinity --silent
```
The JSON output of this command appears to have changed between pnpm v10 and v11 for workspaces.

With pnpm v10, running the command from the workspace root and from a workspace package directory produced different results. A dependency declared only by a workspace package was listed only when the command was run from that package directory.

With pnpm v11, running the command from either the workspace root or the workspace package directory returns the same workspace-wide output, including all workspace packages and their dependencies.

I believe this can cause `electron-builder` to fail to collect app dependencies.

beyondkmp:
fix #9711

## Summary
- pnpm 11 changed the workspace pnpm list --json output shape so running it from the workspace root or from a package directory can return the same dependency tree array.
- Select the dependency tree whose name matches the app package name before falling back to the first tree.
- Keep compatibility with pnpm 10 and earlier because their single/first-tree output still works through the fallback path.
- Add unit coverage for matching, fallback, omitted package name, and empty pnpm output.
- Add a changeset for app-builder-lib.

## Tests
- TEST_FILES=parseNameVersionTest pnpm ci:test
- pnpm exec eslint packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts test/src/parseNameVersionTest.ts
- git diff --check

### DIFF
--- a/.changeset/all-dancers-yell.md
+++ b/.changeset/all-dancers-yell.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: select matching pnpm dependency tree for workspace packages
--- a/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
+++ b/packages/app-builder-lib/src/node-module-collector/nodeModulesCollector.ts
@@ -51,7 +51,7 @@ export abstract class NodeModulesCollector<ProdDepType extends Dependency<ProdDe
     nodeModules: NodeModuleInfo[]
     logSummary: ModuleManager["logSummary"]
   }> {
-    const tree: ProdDepType = await this.getDependenciesTree(this.installOptions.manager)
+    const tree: ProdDepType = await this.getDependenciesTree(this.installOptions.manager, packageName)
 
     await this.collectAllDependencies(tree, packageName)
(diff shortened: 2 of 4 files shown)


mmaietta:
@beyondkmp I don't think this actually solves the underlying issue. Additionally, the tests added are stubbing input value, not actually functional tests validating the differences (and support) for both pnpm < 11 and pnpm >= 11.

Please TAL at my PR https://github.com/electron-userland/electron-builder/pull/9720. Curious on your thoughts on that approach.

PR #9720, opened 2026-05-09
fix: add support for pnpm 11

### LINKED ISSUE #9711: Packaged app dependencies are missing when using pnpm v11

mmaietta:
- Fixes pnpm v11 workspace builds producing empty bundles (`Error: Cannot find module`) due to a breaking change in `pnpm list --json` output format
- pnpm v11 returns one array entry **per workspace package**; v10 returned the root entry with workspace packages nested as dependencies — electron-builder assumed the old layout and collected zero dependencies
- Detects pnpm major version via a `Lazy<number>` and exposes a version-aware `allWorkspacePackages` getter that returns the full multi-entry array for v11 and single-tree behavior for v10/non-workspace
- Adds a regression test using `pnpm@11.0.9` in a workspace fixture, including the required `allowBuilds: { electron: true }` in `pnpm-workspace.yaml` (pnpm v11 replaced `onlyBuiltDependencies` with this)

### DIFF
--- a/packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
+++ b/packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
@@ -1,12 +1,36 @@
+import { Lazy } from "lazy-val"
 import { LogMessageByKey } from "./moduleManager"
 import { NodeModulesCollector } from "./nodeModulesCollector"
-import { PM } from "./packageManager"
+import { getPackageManagerCommand, PM } from "./packageManager"
 import { PnpmDependency } from "./types"
 
 export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependency, PnpmDependency> {
   public readonly installOptions = {
     manager: PM.PNPM,
     lockfile: "pnpm-lock.yaml",
+  }
+
+  // Raw backing field — all entries from `pnpm list --json`
+  private _allWorkspacePackages: PnpmDependency[] = []
+  // Cached after parseDependenciesTree resolves the Lazy; 0 = safe default (treated as < v11)
+  private _pnpmMajorVersion = 0
+  // Runs `pnpm --version` once and caches the major version number
+  private readonly pnpmVersion = new Lazy<number>(async () => {
+    const result = await this.asyncExec(getPackageManagerCommand(PM.PNPM), ["--version"])
+    const major = parseInt((result.stdout ?? "0").split(".")[0], 10)
+    return isNaN(major) ? 0 : major
+  })
+
+  /**
+   * Returns the workspace packages to iterate over, gated by detected pnpm version:
+   * - pnpm v11+: multi-entry workspace output → return the full parsed array
+   * - pnpm < v11 / non-workspace / detection failure: single-tree behavior → return only [0]
+   */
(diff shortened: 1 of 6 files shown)


packages/app-builder-lib/src/node-module-collector/pnpmNodeModulesCollector.ts
```diff
@@ -9,6 +10,29 @@ export class PnpmNodeModulesCollector extends NodeModulesCollector<PnpmDependenc
...
+  })
+
+  /**
+   * Returns the workspace packages to iterate over, gated by detected pnpm version:
+   * - pnpm v11+: multi-entry workspace output → return the full parsed array
+   * - pnpm < v11 / non-workspace / detection failure: single-tree behavior → return only [0]
+   */
+  private get allWorkspacePackages(): PnpmDependency[] {
+    if (this._pnpmMajorVersion >= 11) {
+      return this._allWorkspacePackages
```
beyondkmp, line 31: We likely don't need to return all packages. It should be sufficient to just return the first workspace package and the app package whose names match. If this covers all use cases, we won't need to differentiate between PNPM v10 and v11 anymore.

mmaietta: Thanks for the check! I don't think we can safely reduce to just the first + matching package though

pnpm v11 uses `dedupedDependenciesCount` to mark packages whose subtree has already been expanded *elsewhere in the same JSON output*. Our `collectDepsRecursively` skips any entry with `dedupedDependenciesCount > 0` (it can't recurse into a stub). `extractProductionDependencyGraph` then resolves those stubs by looking up the real entry in `allDependencies`.

As an example, if `app` and `lib` both depend on `lodash`, pnpm expands `lodash` fully only in whichever workspace package comes first in the array (which could be, `lib`) and marks it as deduped in the other package (`app`). If we only process the `app` entry, `lodash` is encountered as a deduped stub and skipped — it never enters `allDependencies`. When the production graph later tries to resolve it, it hits `PKG_DUPLICATE_REF_UNRESOLVED` and silently drops it. No error, just a missing module in the bundle.

We also can't fully eliminate the v10/v11 differentiation: `getTreeFromWorkspaces` still needs to know whether to look for the app nested in `[0].dependencies` (v10) or as a top-level entry (v11).

PR #9732, opened 2026-05-13
perf(pnpm): memoize package resolution and skip downward search in flat virtual store

IAmJSD:
This patch was generated by AI but tested in https://github.com/Infrawrench/Infrawrench/actions/workflows/desktop-build.yml to save 3 hours on the build!

Memoize `locateFromDepOrRoot` by `name@version` and add a visited set to `collectDepsRecursively` to avoid re-walking shared subtrees in the pnpm list output. Add a `skipDownwardSearch` opt-out to `locatePackageVersion` since pnpm's `.pnpm` virtual store is flat, so the BFS would burn thousands of `readdir`/`lstat` calls finding nothing.

### DIFF
--- a/.changeset/tough-flies-yawn.md
+++ b/.changeset/tough-flies-yawn.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+perf(pnpm): memoize package resolution and skip downward search in flat virtual store.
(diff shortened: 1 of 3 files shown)


mmaietta:
Nice find @IAmJSD! Thanks for your contribution

mmaietta (approved):

IAmJSD:
I think windows shard 4 there is a test flake?

IAmJSD:
Test raced again it seems 😭

IAmJSD:
WOOOOO 🎉

PR #9764, opened 2026-05-22
feat(updater): robust PowerShell invocation for Windows code-signature verification

### LINKED ISSUE #8162: Certificates with special characters are not accepted anymore since electron-updater v1.6.9
* **Electron-Builder Version**: 24.13.3

* **Node Version**: v20.11.1

* **Electron Version**: 29.1.6

* **Electron Updater**: 6.2.1 (last working version is 6.1.8)

* **Target**: Windows

I am using electron-builder to build macOS and Windows builds of our application. I was on electron-updater 6.1.8 and updated to 6.1.9 a couple of weeks ago and everything worked fine. Now we wanted to release another update for Windows, and then we discovered that the electron-updater does not accept updates using our Sectigo Code Signing certificate (which did not change!) as it contains a German special character (ä) due to our legal name (Company name UG (haftungsbeschränkt). So all customers who have the version using electron-updater 6.1.9 won't be able to automatically update our application.

I tried multiple different things and figured out that in 6.1.8 the certificate was accepted as usual and starting from 6.1.9 until the latest version, 6.2.1, it does not accept the certificate anymore.

When the auto-updater is run, I get the following log outputs:
2024-03-31 13-05-09@2x
2024-03-31 13-07-25@2x

I tried everything I could imagine. The Windows Version does not matter, the OS language is not a problem, I tried the PowerShell commands which return the certificate information from the electron-updater source code, but everything seems correct.

Now I have to figure out how to notify our customers to manually update the application on all devices...

### LINKED ISSUE #9688: Powershell Get-AuthenticodeSignature fails
Hello, we are using `electron-updater` in Trezor Suite project, and ever since the update from 6.7.3 → 6.8.3, we are getting errors in Sentry:
```
Error: Command failed: set "PSModulePath=" & chcp 65001 >NUL & powershell.exe -NoProfile -NonInteractive -InputFormat None -Command "Get-AuthenticodeSignature -LiteralPath 'C:\Users\[*]\AppData\Local\@trezorsuite-desktop-updater\pending\temp-Trezor-Suite-26.4.2-win-x64.exe' | ConvertTo-Json -Compress"
```
The path seems correct, it's the same that `electron-updater` used in the past.
Any thoughts what change could have caused the regression, or how to debug it, how to reproduce it? 🤔 
Thank you very much for any advice or solution 🙏 

### Related issues

Seems related – same failed command: #2589, #7127, #8162, #8822
We redact user data in Sentry events, so it is unknown if special characters in user name cause this, as was the case previously.

### Context
All error events we have are at Electron 40.1.0, node 24.11.1.
Windows builds: 67% Windows 10.0.19045, 22 % Windows 10.0.26200 (the stable releases), 11 % various other builds

mmaietta:
_Requested by **Mike Maietta** · [Slack thread](https://electronhq.slack.com/archives/C0BGNPTEH8A/p1788282351553289)_

Fixes #9688 — `Get-AuthenticodeSignature` failures on Windows 10 / Electron 40 / Node 24 configurations.

## Summary

- **Core fix — `shell: false` + `-EncodedCommand`:** PowerShell is now invoked directly via `execFile` instead of being chained through `cmd.exe`. The command is passed as a Base64-encoded UTF-16LE string via `-EncodedCommand`, removing the cmd.exe invocation and the `chcp 65001 >NUL &` pipeline entirely.

- **UTF-8 output encoding:** Moved from the `chcp 65001` shell command to `$OutputEncoding = [Console]::OutputEncoding = [Text.Encoding]::UTF8` configured inside the PowerShell script. Both variables must be set together for `ConvertTo-Json` to emit UTF-8 when stdout is captured by Node (fixes #8162 class of errors — non-ASCII characters in the update path).

- **Module loading hardened:** `Import-Module` uses `"$PSHOME\Modules\Microsoft.PowerShell.Security"` (PowerShell's own trusted install directory) before `$env:PSModulePath` is cleared, so the correct module is loaded even on Windows Server 2025 where auto-loading requires PSModulePath at import time. PSModulePath is also deleted from the Node-side `env` option as belt-and-suspenders.

- **Progress-stream suppression:** `$ProgressPreference = 'SilentlyContinue'` is set first in the script so the "Preparing modules for first use." CLIXML record is never written to stderr, eliminating false-positive error detection on first-run environments.

- **Path escaping covers Unicode single quotes:** PowerShell treats the smart-quote code points U+2018–U+201B (‘ ’ ‚ ‛) as single-quote string delimiters, so a path like `C:\Users\D’Andre` used to terminate the `-LiteralPath` string early ("The string is missing the terminator"). All four variants are now doubled exactly like the ASCII `'`. Other PS metacharacters (`$`, `` ` ``, `\`) are literal inside single-quoted strings, so no further escaping is needed (documented in a short comment instead of the old multi-line example block).

- **Bug fix — double-settlement:** Two paths could settle the Promise twice: the `if (error) { reject() } if (stderr) { reject() }` pattern in `handleError` (corrected to `else if`), and a LiteralPath mismatch, which called `reject()` inside a helper and then let the caller `resolve(null)`. The helper now throws a dedicated `LiteralPathMismatchError` that `verifySignature` rejects with exactly once — and rejects directly, without running the spurious ConvertTo-Json availability probe, since the signature data was already parsed successfully.

- **Typo fix:** `normlaizedUpdateFilePath` → `normalizedUpdateFilePath`.

- **Reduced cyclomatic complexity:** Extracted three private helpers to flatten `verifySignature` (CC 6 → 3):
  - `checkLiteralPath` — validates `data.Path` against the expected update path; throws `LiteralPathMismatchError` on mismatch, returns normally when the check passes or is skipped (missing `data.Path`, fail-open with the v28 deprecation warning).
  - `matchPublisher` — DN matching loop.
  - `evaluateSignatureResult` — parses the PS JSON output and coordinates the LiteralPath + publisher checks.

## Tests

- **The e2e tests now actually run.** They were written with the legacy `test(name, fn, timeout)` signature, which the repo's vitest wrapper cannot parse: the timeout argument was popped as the test body and the real function discarded, so every e2e test was silently registered as `todo` and never executed on Windows CI. All e2e tests now use `test(name, { timeout }, fn)`, and runtime prerequisite guards use `context.skip()` so unmet prerequisites report as skipped rather than passed.

- **No hook-fed shared state.** Both suites consume the per-test `tmpDir` context fixture (auto-created and auto-cleaned per test) instead of suite-level `TmpDir` instances, and every test derives its own logger/fixtures from a stateless `setup()` helper inside the test body — no `beforeEach`/`beforeAll`-populated `let` variables — so tests stay independent for parallel execution.

- **48 unit tests (platform-agnostic, mocked `child_process`):** invocation parameters (`shell: false`, encoded-command round-trip, script ordering, PSModulePath stripping, 20 s timeout), publisher/DN matching incl. case-sensitivity and partial DNs, LiteralPath validation incl. the single-rejection behavior, `parseOut` field stripping, the full `handleError` matrix (old Win 6.x fail-open, ConvertTo-Json probe, both-error-and-stderr single settlement, malformed/empty stdout), and path-injection prevention for `'`, all four smart-quote variants, `$`, and backtick.

- **18 e2e tests (Windows-only, real PowerShell):** unsigned files, paths with spaces / quotes / smart quotes / non-ASCII characters (#8162), directory names with `&`, `%VAR%`, `$`, and backtick — all dangerous under the old cmd.exe invocation — and a symlink LiteralPath-mismatch guard.

- **Real-certificate e2e suite (8 of those 18):** a new `test/src/helpers/windowsSignedFixture.ts` helper generates an ephemeral self-signed code-signing identity with a multi-RDN subject (CN/O/C), installs its public cert into `Cert:\LocalMachine\Root` (skipping dynamically when the runner is not elevated), and Authenticode-signs a per-test copy of a real system PE with `Set-AuthenticodeSignature`. This drives genuine `Status = Valid` results through the whole pipeline deterministically, replacing the previous 4 notepad.exe tests whose signer DN had to be discovered at runtime and varied across Windows versions. Covered: full-DN match (as reported and RDN-order-swapped), CN-only match incl. deprecation warning, subset-DN match, wrong-publisher mismatch, a signed exe inside a smart-quote + spaces directory (a real positive through the escaping), a symlink to a validly-signed exe, and a tampered (byte-appended) signed exe. Cleanup removes the installed cert by thumbprint in `afterAll`. `selfSignedIdentity` gained an optional `additionalDnFields` option (existing consumers unaffected).

### Notes for future readers — two Windows CI gotchas hit along the way

- **PSModulePath vs. module autoload:** any script run with PSModulePath stripped (as both the production verifier and the test fixture do) must `Import-Module "$PSHOME\Modules\Microsoft.PowerShell.Security"` explicitly — on Windows Server 2025, autoload of `Set-AuthenticodeSignature`/`Get-AuthenticodeSignature` fails without PSModulePath at import time (`CouldNotAutoloadMatchingModule`).
- **Catalog-signature precedence:** a byte-identical copy of an OS binary still hash-matches the Windows security catalog (the Authenticode hash excludes the embedded certificate table), and `Get-AuthenticodeSignature` reports the catalog signer (`CN=Microsoft Windows`) in preference to a newly embedded signature. The fixture therefore appends marker bytes before signing so no catalog matches — the embedded signature covers the appended data, like a signed NSIS installer's archive — and round-trips the signature after signing, failing loudly on the fixture's own thumbprint rather than as a confusing per-test publisher mismatch.

## Recent changes (since the last review round)

- Escape the Unicode single-quote variants U+2018–U+201B in the `-LiteralPath` argument (unit + e2e coverage).
- Reject exactly once on LiteralPath mismatch via `LiteralPathMismatchError` (no follow-up `resolve(null)`, no ConvertTo-Json probe).
- Fix e2e test registration — the legacy timeout signature had silently turned every e2e test into `todo`; they now execute on Windows CI.
- Create temp dirs before writing e2e fixtures (`TmpDir.getTempDir` only reserves a path; three tests hit ENOENT).
- Move both suites to the per-test `tmpDir` context fixture, then remove all remaining describe-level mutable state fed by hooks (requested in review).
- Add the really-signed-executable e2e suite via `windowsSignedFixture` (self-signed cert in `LocalMachine\Root`, per-test signed copies, dynamic skip when not elevated) — 8 deterministic tests replacing the 4 runtime-discovered notepad DN tests.
- Fixture hardening for Windows CI: explicit `$PSHOME` Security-module import (autoload fails with PSModulePath stripped), and de-cataloging the signed copy + post-sign thumbprint round-trip (catalog-signature precedence). Details under "Notes for future readers" above.

## Changed Files

| File | Change |
|------|--------|
| `packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts` | Core fix: `shell: false`, `-EncodedCommand`, smart-quote escaping, helper extraction, single-rejection LiteralPath handling, bug fixes |
| `test/src/updater/windowsExecutableCodeSignatureVerifierTest.ts` | Suite overhaul: registration fix, per-test `tmpDir` context, stateless per-test setup, 48 unit + 18 e2e tests incl. the real-certificate suite |
| `test/src/helpers/windowsSignedFixture.ts` | New: self-signed cert install to `LocalMachine\Root`, per-test Authenticode-signed executables, de-catalog + round-trip verification, thumbprint cleanup |
| `test/src/helpers/selfSignedIdentity.ts` | Optional `additionalDnFields` for multi-RDN subjects |
| `test/snapshots/updater/windowsExecutableCodeSignatureVerifierTest.js.snap` | Snapshot of encoded script (path placeholder replaces dynamic tmp path) |
| `.changeset/late-results-grab.md` | Patch changeset for `electron-updater` |


mayfield:
Is this going to get merged?  I have users reporting errors upgrading and I'm hoping this fixes it.

mmaietta:
I do plan to revisit this PR, but I had to pause work due to personal events. Hoping to reapply focus here soon. There was additional testing I was looking to do.

michalzaq12:
Will this PR also solve the issue with `’` in user path?

```
Error: Error: Command failed: set "PSModulePath=" & chcp 65001 >NUL & powershell.exe -NoProfile -NonInteractive -InputFormat None -Command "Get-AuthenticodeSignature -LiteralPath 'C:\Users\D’Andre\AppData\Local\xxx-updater\pending\temp-XXX.exe' | ConvertTo-Json -Compress"
The string is missing the terminator: '.
```

test/src/updater/windowsExecutableCodeSignatureVerifierTest.ts
```diff
@@ -0,0 +1,795 @@
...
+  let defaultFile = ""
+  let logger: Logger
+
+  // Injects the per-test defaultFile so callers don't repeat it everywhere.
+  // An explicit filePath in opts still wins (used by path-injection tests).
+  const makeJson = (opts: SignatureJsonOpts = {}) => makeSignatureJson({ filePath: defaultFile, ...opts })
+
+  // `tmpDir` comes from the per-test context fixture in vitest-tmpdir.ts (auto-cleaned after each test).
+  beforeEach(async ({ tmpDir }) => {
+    defaultFile = path.join(await tmpDir.getTempDir(), "test-update-1.0.1.exe")
```
mmaietta: @claude move this logic into **each test**, not the beforeEach. We want to eventually return to parallel testing, which means having "global" describe-level variables to cause issues in the future.

It is supposed to be used as:
```
test("sample name", async ({ tmpDir }) => {
// test
})
```

PR #9766, opened 2026-05-22
chore: add engines field to package.json

gameroman:
Follow-up on #9671

Adds `"engines"` where it is missing


mmaietta:
Most of the engines are incorrect and would introduce breaking changes - please do your research first. Closing as the engines requirement will be taken care of in v27 ESM upgrade; until then, these will remain as they currently are.

PR #9769, opened 2026-05-23
fix(nsis): safely copy UserProgramFiles path

adamgauthier:
## What

Fix the NSIS per-user install path lookup for `FOLDERID_UserProgramFiles`.

The installer template asks Windows for the current per-user Programs known folder, then uses that path as the default per-user install root. This change keeps that behavior but copies the returned `PWSTR` safely.

## Why

`SHGetKnownFolderPath(FOLDERID_UserProgramFiles)` returns a NUL-terminated `PWSTR` allocated by the Shell. The previous NSIS code read that pointer with `&w${NSIS_MAX_STRLEN}`, which makes `System.dll` copy a fixed number of WCHARs from a buffer that is only sized for the actual path. On some heap layouts this can overread into unmapped memory and crash the setup EXE before extraction.

This replaces the fixed-size read with `lstrcpynW`, bounded by `NSIS_MAX_STRLEN`, and preserves scratch registers with `Push` / `Pop` instead of `System::Store S/L` so the copied path is not restored away.

This also removes the Win8+ guard added in #9564. `FOLDERID_UserProgramFiles` was introduced in Windows 7, but it remains the per-user Programs known folder on later Windows versions and is redirectable. With the unsafe copy fixed, the installer can ask Windows for the actual per-user install root on all supported Windows versions and continue falling back to `$LocalAppData\Programs` if the API call fails.

## Validation

I validated this locally with deterministic NSIS repro harnesses:

- unsafe fixed-size read: exits `-1073741819` / `0xc0000005`
- bounded copy, Unicode NSIS: exits `0`
- bounded copy, ANSI NSIS: exits `0`
- real `SHGetKnownFolderPath(FOLDERID_UserProgramFiles)` path with bounded copy: exits `0`

I will post the PowerShell repro script and sample output in a follow-up comment so the PR description stays focused.

Refs #8536
Refs #9564

### DIFF
--- a/packages/app-builder-lib/templates/nsis/multiUser.nsh
+++ b/packages/app-builder-lib/templates/nsis/multiUser.nsh
@@ -1,6 +1,5 @@
 !include FileFunc.nsh
 !include UAC.nsh
-!include WinVer.nsh
 
 !define FOLDERID_UserProgramFiles {5CD7AEE2-2219-4A67-B85D-6C9CE15660CB}
 !define KF_FLAG_CREATE 0x00008000
@@ -30,17 +29,20 @@ Var installMode
     ${else}
       StrCpy $0 "$LocalAppData\Programs"
 
-      ${IfNot} ${AtLeastWin8}
-        System::Store S
-        # Win7 has a per-user programfiles known folder and this can be a non-default location
-        System::Call 'SHELL32::SHGetKnownFolderPath(g "${FOLDERID_UserProgramFiles}", i ${KF_FLAG_CREATE}, p 0, *p .r2)i.r1'
-        ${If} $1 == 0
-          System::Call '*$2(&w${NSIS_MAX_STRLEN} .s)'
-          StrCpy $0 $1
-          System::Call 'OLE32::CoTaskMemFree(p r2)'
-        ${endif}
-        System::Store L
-      ${EndIf}
+      Push $1
+      Push $2
+      # UserProgramFiles is the per-user install root and can be a non-default location
+      StrCpy $2 0
+      System::Call 'SHELL32::SHGetKnownFolderPath(g "${FOLDERID_UserProgramFiles}", i ${KF_FLAG_CREATE}, p 0, *p .r2)i.r1'
+      ${If} $1 == 0
+        System::Call 'KERNEL32::lstrcpynW(w .r0, p r2, i ${NSIS_MAX_STRLEN})p'
+      ${endif}
+      # SHGetKnownFolderPath may return allocated memory even on failure
+      ${If} $2 != 0
+        System::Call 'OLE32::CoTaskMemFree(p r2)'
+      ${endif}
+      Pop $2
+      Pop $1
 
       StrCpy $INSTDIR "$0\${APP_FILENAME}"
     ${endif}
--- a/.changeset/safe-buckets-push.md
+++ b/.changeset/safe-buckets-push.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(nsis): safely copy UserProgramFiles path


adamgauthier:
Maintainer repro script and sample output. I would not propose committing this as a repo test as-is; it is a quick Windows/NSIS harness to demonstrate the unsafe fixed-size WCHAR read and the bounded-copy replacement.

Run from PowerShell on Windows after an electron-builder NSIS build has populated the local NSIS cache, or pass `-MakensisPath` explicitly:

```powershell
.\Repro-NsisKnownFolderCopy.ps1
```

Sample local output with small attempt counts:

```text
unsafe-fixed-read:         -1073741819:2
safe-bounded-copy-unicode: 0:2
safe-bounded-copy-ansi:    0:2
safe-known-folder-api:     0:2
```

Expected behavior:

- `unsafe-fixed-read` exits with `-1073741819` (`0xc0000005`)
- `safe-bounded-copy-unicode` exits `0`
- `safe-bounded-copy-ansi` exits `0`
- `safe-known-folder-api` exits `0`

<details>
<summary>Full repro script</summary>

```powershell
param(
  [string]$MakensisPath,
  [int]$UnsafeAttempts = 3,
  [int]$SafeAttempts = 5,
  [switch]$KeepArtifacts
)

$ErrorActionPreference = 'Stop'

function Resolve-Makensis {
  if ($MakensisPath) {
    return (Resolve-Path $MakensisPath).Path
  }

  $fromPath = Get-Command makensis.exe -ErrorAction SilentlyContinue
  if ($fromPath) {
    return $fromPath.Source
  }

  $cacheRoot = Join-Path $env:LOCALAPPDATA 'electron-builder\Cache\nsis'
  if (Test-Path $cacheRoot) {
    $cached = Get-ChildItem $cacheRoot -Recurse -Filter makensis.exe -ErrorAction SilentlyContinue |
      Where-Object { $_.FullName -like '*\Bin\makensis.exe' } |
      Sort-Object LastWriteTime -Descending |
      Select-Object -First 1
    if ($cached) {
      return $cached.FullName
    }
  }

  throw 'makensis.exe was not found. Run an electron-builder NSIS build once, install NSIS, or pass -MakensisPath.'
}

function Write-NsisScript([string]$Path, [string]$Content, [string]$OutFile) {
  $Content.Replace('__OUTFILE__', $OutFile) | Set-Content -Path $Path -Encoding ascii
}

function Invoke-Makensis([string]$Makensis, [string]$Script) {
  & $Makensis $Script | Select-Object -Last 8
  if ($LASTEXITCODE -ne 0) {
    throw "makensis failed for $Script with exit code $LASTEXITCODE"
  }
}

function Invoke-Attempts([string]$Exe, [int]$Attempts) {
  $counts = @{}
  for ($i = 1; $i -le $Attempts; $i++) {
    $p = Start-Process -FilePath $Exe -ArgumentList '/S' -Wait -PassThru -WindowStyle Hidden
    $code = [int]$p.ExitCode
    if ($counts.ContainsKey($code)) {
      $counts[$code]++
    } else {
      $counts[$code] = 1
    }
  }

  return (($counts.GetEnumerator() | Sort-Object Name | ForEach-Object { "$($_.Key):$($_.Value)" }) -join '; ')
}

$unsafeTemplate = @'
Unicode true
Name "unsafe-fixed-wide-read"
OutFile "__OUTFILE__"
RequestExecutionLevel user
SilentInstall silent
Function .onInit
  System::Call 'KERNEL32::VirtualAlloc(p 0, i 8192, i 0x2000, i 0x01)p.r4'
  System::Call 'KERNEL32::VirtualAlloc(p r4, i 4096, i 0x1000, i 0x04)p.r5'
  IntOp $2 $4 + 4080
  System::Call '*$2(&w8 "abc")'
  System::Store S
  System::Call '*$2(&w${NSIS_MAX_STRLEN} .s)'
  StrCpy $0 $1
  System::Store L
  System::Call 'KERNEL32::VirtualFree(p r4, i 0, i 0x8000)'
FunctionEnd
Section "Main"
SectionEnd
'@

$safeTemplate = @'
Unicode __UNICODE__
Name "safe-bounded-copy-__KIND__"
OutFile "__OUTFILE__"
RequestExecutionLevel user
SilentInstall silent
Function .onInit
  StrCpy $0 "$LocalAppData\Programs"
  StrCpy $1 "sentinel-one"
  StrCpy $2 "sentinel-two"
  System::Call 'KERNEL32::VirtualAlloc(p 0, i 8192, i 0x2000, i 0x01)p.r4'
  System::Call 'KERNEL32::VirtualAlloc(p r4, i 4096, i 0x1000, i 0x04)p.r5'
  IntOp $3 $4 + 4080
  System::Call '*$3(&w8 "abc")'
  Push $1
  Push $2
  StrCpy $2 $3
  System::Call 'KERNEL32::lstrcpynW(w .r0, p r2, i ${NSIS_MAX_STRLEN})p'
  Pop $2
  Pop $1
  StrCmp $0 "abc" 0 bad
  StrCmp $1 "sentinel-one" 0 bad
  StrCmp $2 "sentinel-two" 0 bad
  System::Call 'KERNEL32::VirtualFree(p r4, i 0, i 0x8000)'
  Return
bad:
  System::Call 'KERNEL32::VirtualFree(p r4, i 0, i 0x8000)'
  Abort "bounded copy or register restore failed: $0 $1 $2"
FunctionEnd
Section "Main"
SectionEnd
'@

$knownFolderTemplate = @'
Unicode true
Name "safe-known-folder-copy"
OutFile "__OUTFILE__"
RequestExecutionLevel user
SilentInstall silent
!define FOLDERID_UserProgramFiles {5CD7AEE2-2219-4A67-B85D-6C9CE15660CB}
!define KF_FLAG_CREATE 0x00008000
Function .onInit
  StrCpy $0 "$LocalAppData\Programs"
  StrCpy $1 "sentinel-one"
  StrCpy $2 "sentinel-two"
  Push $1
  Push $2
  StrCpy $2 0
  System::Call 'SHELL32::SHGetKnownFolderPath(g "${FOLDERID_UserProgramFiles}", i ${KF_FLAG_CREATE}, p 0, *p .r2)i.r1'
  StrCmp $1 0 doCopy afterCopy
doCopy:
  System::Call 'KERNEL32::lstrcpynW(w .r0, p r2, i ${NSIS_MAX_STRLEN})p'
afterCopy:
  StrCmp $2 0 noFree
  System::Call 'OLE32::CoTaskMemFree(p r2)'
noFree:
  Pop $2
  Pop $1
  StrCmp $1 "sentinel-one" 0 bad
  StrCmp $2 "sentinel-two" 0 bad
  StrCmp $0 "" bad
  Return
bad:
  Abort "known-folder copy or register restore failed: $0 $1 $2"
FunctionEnd
Section "Main"
SectionEnd
'@

$makensis = Resolve-Makensis
$workDir = Join-Path $env:TEMP "nsis-known-folder-copy-repro-$([Guid]::NewGuid().ToString('N'))"
New-Item -ItemType Directory -Path $workDir -Force | Out-Null

try {
  $cases = @(
    @{
      Name = 'unsafe-fixed-read'
      Attempts = $UnsafeAttempts
      Script = Join-Path $workDir 'unsafe-fixed-read.nsi'
      Exe = Join-Path $workDir 'unsafe-fixed-read.exe'
      Content = $unsafeTemplate
      Expect = 'crash'
    },
    @{
      Name = 'safe-bounded-copy-unicode'
      Attempts = $SafeAttempts
      Script = Join-Path $workDir 'safe-bounded-copy-unicode.nsi'
      Exe = Join-Path $workDir 'safe-bounded-copy-unicode.exe'
      Content = $safeTemplate.Replace('__UNICODE__', 'true').Replace('__KIND__', 'unicode')
      Expect = 'success'
    },
    @{
      Name = 'safe-bounded-copy-ansi'
      Attempts = $SafeAttempts
      Script = Join-Path $workDir 'safe-bounded-copy-ansi.nsi'
      Exe = Join-Path $workDir 'safe-bounded-copy-ansi.exe'
      Content = $safeTemplate.Replace('__UNICODE__', 'false').Replace('__KIND__', 'ansi')
      Expect = 'success'
    },
    @{
      Name = 'safe-known-folder-api'
      Attempts = $SafeAttempts
      Script = Join-Path $workDir 'safe-known-folder-api.nsi'
      Exe = Join-Path $workDir 'safe-known-folder-api.exe'
      Content = $knownFolderTemplate
      Expect = 'success'
    }
  )

  $summary = @()
  foreach ($case in $cases) {
    Write-Host "## $($case.Name)"
    Write-NsisScript -Path $case.Script -Content $case.Content -OutFile $case.Exe
    Invoke-Makensis -Makensis $makensis -Script $case.Script
    $exitCodes = Invoke-Attempts -Exe $case.Exe -Attempts $case.Attempts
    Write-Host "exit codes: $exitCodes"
    $summary += [pscustomobject]@{
      Case = $case.Name
      Attempts = $case.Attempts
      Expected = $case.Expect
      ExitCodes = $exitCodes
    }
  }

  $summary | Format-Table -AutoSize

  $unsafe = $summary | Where-Object Case -eq 'unsafe-fixed-read'
  $safeFailures = @($summary | Where-Object { $_.Expected -eq 'success' -and $_.ExitCodes -ne "0:$($_.Attempts)" })
  if ($unsafe.ExitCodes -notlike '*-1073741819:*') {
    throw "Unsafe repro did not produce the expected access violation. Exit codes: $($unsafe.ExitCodes)"
  }
  if ($safeFailures.Count -gt 0) {
    throw "One or more safe repros failed: $($safeFailures | ConvertTo-Json -Compress)"
  }
} finally {
  if ($KeepArtifacts) {
    Write-Host "Artifacts kept at: $workDir"
  } else {
    Remove-Item $workDir -Recurse -Force -ErrorAction SilentlyContinue
  }
}
```

</details>

packages/app-builder-lib/templates/nsis/multiUser.nsh
```diff
@@ -30,17 +29,19 @@ Var installMode
     ${else}
       StrCpy $0 "$LocalAppData\Programs"
 
-      ${IfNot} ${AtLeastWin8}
```
adamgauthier, line 33: Note: I'm removing the guard added in #9564 because this logic is not Win7-only. `FOLDERID_UserProgramFiles` was introduced in Windows 7, but it remains the per-user Programs known folder on later Windows versions and can be redirected afaiu.

The original issue #8536 appears to have been hitting the same fault this PR fixes: the report shows the installer crashing in the temporary NSIS `System.dll` with `0xc0000005` at offset `0x00001581`, and the reporter later narrowed it to this `multiUser.nsh` per-user install path branch. #9564 mitigated that by avoiding this unsafe code path on Win8+, but the underlying issue was the fixed-size `&w${NSIS_MAX_STRLEN}` read from the Shell-allocated `PWSTR`. With that copy fixed, the Win8+ guard should no longer be necessary.

packages/app-builder-lib/templates/nsis/multiUser.nsh
```diff
@@ -30,17 +29,19 @@ Var installMode
...
+      Push $2
+      # UserProgramFiles is the per-user install root and can be a non-default location
+      StrCpy $2 0
+      System::Call 'SHELL32::SHGetKnownFolderPath(g "${FOLDERID_UserProgramFiles}", i ${KF_FLAG_CREATE}, p 0, *p .r2)i.r1'
+      ${If} $1 == 0
+        System::Call 'KERNEL32::lstrcpynW(w .r0, p r2, i ${NSIS_MAX_STRLEN})p'
+      ${endif}
+      ${If} $2 != 0
+        System::Call 'OLE32::CoTaskMemFree(p r2)'
+      ${endif}
```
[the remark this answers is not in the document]
adamgauthier: Thanks for the careful look. I think this free is intentional for `SHGetKnownFolderPath`'s ownership contract: the docs say the caller is responsible for freeing `ppszPath` with `CoTaskMemFree` "whether `SHGetKnownFolderPath` succeeds or not": https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath
> When this method returns, contains the address of a pointer to a null-terminated Unicode string that specifies the path of the known folder. The calling process is responsible for freeing this resource once it is no longer needed by calling [CoTaskMemFree](https://learn.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-cotaskmemfree), whether SHGetKnownFolderPath succeeds or not. The returned path does not include a trailing backslash. For example, "C:\Users" is returned rather than "C:\Users\".

`$2` is initialized to `0` before the call, so `$2 != 0` is the ownership check here. Adding `$1 == 0` to the free condition could leak a non-null pointer returned with a failure HRESULT. I added a short inline comment to make that contract clearer.

adamgauthier:
@mmaietta Thanks for requesting a review from Copilot. I think its comment is wrong, I replied to it if you want to take another look at the PR

mmaietta:
Nice work @adamgauthier, really appreciate this contribution! I had to do some research on my side and am aligned with your assessment re: copilot's comment. Kicking off the CI now. If any tests fail, I'll take a look first and can ping you if needed (some of the OS-level tests can act flaky when run massively in parallel atm)

packages/app-builder-lib/templates/nsis/multiUser.nsh
mmaietta: Makes sense, thanks for explicitly calling/commenting on this.

mmaietta (approved):

PR #9781, opened 2026-05-29
chore: document non-existent swiftshader directory in newer electron versions

Lemonexe:
Tiny nit: in https://github.com/electron-userland/electron-builder/pull/6682, the `swiftshader` directory was explicitely added to be crawled for files eligible for signing.

But I found out this may be outdated. I do not know the exact electron version where it changed, but I confirmed that the folder `swiftshader`  was there in a build with electron 15, but it is gone at electron 20 (and latest versions).

Note: the dlls used to be duplicated, now only the root ones remain: 
```
./libEGL.dll
./libGLESv2.dll
./swiftshader/libEGL.dll
./swiftshader/libGLESv2.dll
```


packages/app-builder-lib/src/winPackager.ts
```diff
@@ -305,6 +305,7 @@ export class WinPackager extends PlatformPackager<WindowsConfiguration> {
       const outDir = path.join(packContext.appOutDir, ...filepath)
       return walk(outDir, (file, stat) => stat.isDirectory() || this.shouldSignFile(file))
     }
+    // Note that the `swiftshader` directory is absent in modern electron versions, and is listed here only for backwards compatibility.
```
mmaietta:
```suggestion
    // Note: The `swiftshader` directory is absent in modern electron versions. `swiftshader/` held Chromium's legacy SwiftShader GL fallback (libEGL.dll / libGLESv2.dll), removed in Chromium 102 (Electron 19+) in favor of SwANGLE (ANGLE + SwiftShader Vulkan). This is kept here only for backwards compat with older Electron; `walk` no-ops on a missing dir (readdir ENOENT is swallowed), so this is harmless when the directory is absent.
```

I ran a quick Claude prompt on this to see if we can make this more descriptive.

PR #9792, opened 2026-05-30
chore: prepare upgrade pathway for updating to latest `electron/____` npm packages that are ESM-only

mmaietta:
## Problem

Several `@electron/*` packages (`@electron/notarize`, `@electron/osx-sign`, `@electron/fuses`, `@electron/universal`, `@electron/asar`, `@electron/get`, and `@electron/rebuild`) are dropping CommonJS support and shipping as ESM-only. Because electron-builder's output is CJS (TypeScript compiles with `"module": "CommonJS"`), these packages can no longer be statically `require()`'d without throwing at runtime.

There is an additional subtlety: TypeScript with `"module": "CommonJS"` silently transforms `await import("pkg")` into `await Promise.resolve().then(() => require("pkg"))` in compiled output — meaning inline dynamic imports in TypeScript source are **not** a viable workaround.

## Solution

Route all `@electron/*` imports through the existing `helpers/dynamic-import.js` helper, which is a plain `.js` file that TypeScript never transforms. Its native `import()` call is preserved as-is in the output and works correctly with both CJS and ESM-only packages at runtime (Node.js has supported `import()` in CJS modules since v12).

### New utility

`src/util/dynamicImport.ts` — a thin TypeScript wrapper over `helpers/dynamic-import.js` that provides typed access and a clear explanation of why this indirection exists.

### Per-package changes

| Package | Files changed | Change type |
|---------|--------------|-------------|
| `@electron/notarize` | `MacTargetHelper.ts` | Static import → `dynamicImport()` at call site |
| `@electron/osx-sign` | `MacTargetHelper.ts`, `macCodeSign.ts`, `macPackager.ts` | Value imports → `dynamicImport()`; type-only imports → `import type` |
| `@electron/fuses` | `platformPackager.ts` | Enum + function imports → `dynamicImport()`; `generateFuseConfig` made async |
| `@electron/universal` | `macPackager.ts` | Static import → `dynamicImport()` at call site |
| `@electron/asar` | `asarUtil.ts`, `asarFileChecker.ts` | Static import → `dynamicImport()`; `checkFileInArchive` made async |
| `@electron/get` | `binDownload.ts` | Static import → `dynamicImport()` at call site |
| `@electron/rebuild` | `rebuild.ts`, `yarn.ts` | Already handled via subprocess; type import → `import type` |

### What is not changed

- The build system (`tsc --build`, `tsconfig-base.json`) is unchanged.
- No new runtime dependencies.
- The public API surface is unchanged. `addElectronFuses` was already effectively async (it returned a Promise); it is now explicitly `async`.
- Downstream consumers using `require('electron-builder')` are unaffected.

## How it works at runtime

```
TypeScript call site
  └─ dynamicImport("@electron/notarize")      ← compiled: require("../util/dynamicImport").dynamicImport(...)
       └─ helpers/dynamic-import.js            ← plain JS, TypeScript never touches it
            └─ await import("@electron/notarize")  ← real native import(), works with ESM-only packages
```

The Node.js module cache ensures repeated calls to `dynamicImport("@electron/pkg")` for the same specifier are free after the first load.

## Upgrading a package

When any `@electron/*` package publishes an ESM-only version:

1. Bump the version in `packages/app-builder-lib/package.json`.
2. If the package restructures its exports (e.g., `@electron/osx-sign/dist/cjs/util-identities` → a new path), update the specifier string in the relevant call site and the `import type` path at the top of the file.
3. Run `pnpm compile && TEST_FILES=<affected> pnpm ci:test` to verify.


Sytten:
@mmaietta This PR is breaking stuff, see the issue https://github.com/electron-userland/electron-builder/issues/9816

PR #9826, opened 2026-06-02
fix(mac): add compat symlinks after helper renaming to ensure Electron helper discovery

### LINKED ISSUE #9771: [macOS 26] Packaged arm64 app crashes on launch (exit 133 / SIGTRAP): helper apps renamed but Electron binary not patched
## Environment

| | |
|---|---|
| electron-builder | 26.x |
| electron | 42.2.0 |
| macOS | 26.5 (Darwin 25.5.0, Apple Silicon arm64) |
| Node.js | 22.x |
| Package manager | pnpm |

## Describe the bug

A macOS arm64 app packaged with `electron-builder --mac` crashes immediately on launch with **exit code 133 (SIGTRAP)** and produces no output. The crash happens before any JavaScript runs.

Running the app binary directly from the terminal confirms the fatal error:

```
[FATAL:electron/shell/app/electron_main_delegate_mac.mm:66] Unable to find helper app
```

The app works correctly in development mode (`electron .`).

## Root cause

electron-builder renames the helper `.app` bundles inside `Contents/Frameworks/` to match the product name (e.g. `Electron Helper.app` → `MyApp Helper.app`) but **does not patch the main Electron binary** to reflect the new name.

The packaged binary has an **identical SHA-256 hash** to the original binary from the npm package, confirming it was never modified:

```bash
shasum -a 256 "dist/mac-arm64/MyApp.app/Contents/MacOS/MyApp"
# c86c2701f15ca66ba460f4cdcd7ebcab0b5133a4cc299295f02108b3754e9eed

shasum -a 256 "node_modules/.pnpm/electron@42.2.0/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron"
# c86c2701f15ca66ba460f4cdcd7ebcab0b5133a4cc299295f02108b3754e9eed
```

The unmodified binary still looks for `Electron Helper.app` at runtime, which no longer exists after renaming.

This was verified by renaming the helpers back to their original names via an `afterPack` hook. The app launches successfully with that workaround in place.

## Steps to reproduce

1. Create a standard Electron project with `productName: "MyApp"` in the electron-builder config
2. Run `electron-builder --mac --arm64` on macOS 26 (arm64 machine)
3. Remove the quarantine attribute: `xattr -dr com.apple.quarantine "dist/mac-arm64/MyApp.app"`
4. Launch the app or run the binary directly from terminal
5. App exits immediately with code 133, no output

## Expected behavior

The packaged app launches correctly. electron-builder should either:

- patch the Electron binary to replace the hardcoded `Electron Helper` string with the product name (as it apparently did in earlier versions), or
- not rename the helper bundles at all and leave that responsibility to Electron's runtime resolution.

## Workaround

Add an `afterPack` hook that renames the helpers back to their original `Electron Helper` names:

**package.json**
```json
{
  "build": {
    "afterPack": "./scripts/afterPack.js"
  }
}
```

**scripts/afterPack.js**
```js
const fs   = require('fs')
const path = require('path')

exports.default = async function ({ appOutDir, packager }) {
  if (packager.platform.name !== 'mac') return

  const frameworksDir = path.join(
    appOutDir,
    `${packager.appInfo.productFilename}.app`,
    'Contents',
    'Frameworks'
  )

  const product = packager.appInfo.productFilename
  const suffixes = ['', ' (GPU)', ' (Plugin)', ' (Renderer)']

  for (const suffix of suffixes) {
    const src = path.join(frameworksDir, `${product} Helper${suffix}.app`)
    const dst = path.join(frameworksDir, `Electron Helper${suffix}.app`)
    if (!fs.existsSync(src) || fs.existsSync(dst)) continue

    fs.renameSync(src, dst)

    const macOSDir = path.join(dst, 'Contents', 'MacOS')
    const oldBin   = path.join(macOSDir, `${product} Helper${suffix}`)
    const newBin   = path.join(macOSDir, `Electron Helper${suffix}`)
    if (fs.existsSync(oldBin)) fs.renameSync(oldBin, newBin)
  }
}
```

## Additional context

- The issue is reproducible with both `hardenedRuntime: true` and `hardenedRuntime: false`.
- Setting `identity: null` (skip code signing) does not affect the outcome — the binary is never patched regardless of signing configuration.
- The `node_modules` Electron binary (linker-signed, never re-signed by electron-builder) launches correctly with the app's `app.asar`.
- All FUSE flags (including `EnableEmbeddedAsarIntegrityValidation`) are identical between the working and failing binaries.
- This may be a macOS 26 regression if previous versions were more lenient in resolving helper app names — testing on macOS 15 would be welcome.

See also: companion issue filed against the Electron repository regarding making he

mmaietta:
Fixes #9771

## Summary

- **Root cause investigated**: Electron's `OverrideChildProcessPath()` uses a two-step helper lookup: first via the compile-time `ELECTRON_PRODUCT_NAME` constant ("Electron"), then via `GetApplicationName()` (reads `CFBundleName` from `Info.plist`). On macOS 26 (Darwin 25.x), the `CFBundleName`-based fallback fails to resolve the renamed bundles, causing a `LOG(FATAL)` crash at startup with exit 133 / SIGTRAP.

- **Fix – `electronMac.ts`**: After renaming helpers from `Electron Helper.app` → `MyApp Helper.app`, call the new `addHelperCompatSymlinks()` to create two relative symlinks per helper variant:
  1. `Frameworks/Electron Helper.app` → `./MyApp Helper.app` (directory symlink — primary lookup path)
  2. `MyApp Helper.app/Contents/MacOS/Electron Helper` → `./MyApp Helper` (executable symlink — resolves the full path check)

  `stat()` / `base::PathExists()` follows symlinks, so Electron's primary `ELECTRON_PRODUCT_NAME` lookup now resolves to the correctly renamed bundle on all macOS versions. Branding in Activity Monitor is preserved (the resolved bundle name is `MyApp Helper`).

- **Fix – `electronMacUtils.ts`** (new file): Extracted the three pure, side-effect-free helpers (`assertSafeHelperName`, `getAvailableHelperSuffixes`, `addHelperCompatSymlinks`) into a lightweight module with minimal dependencies to enable direct unit testing without triggering the full packager module graph.

- **Input validation**: `assertSafeHelperName()` added to `moveHelpers()` — rejects product names or branding names that contain `/`, `\`, or null bytes, preventing path traversal into unexpected directories. Symlink creation is additionally guarded by `sanitizeDirPath()` containment checks.

- **Tests – `test/src/mac/electronMacTest.ts`** (new file, 28 tests):
  - `assertSafeHelperName`: validates acceptance of normal names and rejection of path-traversal inputs
  - `getAvailableHelperSuffixes`: covers all plist presence combinations, ordering, and the always-present `" Helper"` base
  - `addHelperCompatSymlinks`: no-op when names are equal; bundle-level directory symlink created; executable symlink created inside bundle; transitive `stat()` resolves to real binary; all modern helper variants (`(Renderer)`, `(GPU)`, `(Plugin)`); names with spaces; custom `electronBranding.productName`; rejection of path-separator inputs

## Changed Files

| File | Change |
|------|--------|
| `packages/app-builder-lib/src/electron/electronMac.ts` | Import utils from new module; add `assertSafeHelperName` validation to `moveHelpers`; call `addHelperCompatSymlinks` after rename; remove duplicated local `getAvailableHelperSuffixes` |
| `packages/app-builder-lib/src/electron/electronMacUtils.ts` | **New** — pure helpers: `assertSafeHelperName`, `getAvailableHelperSuffixes`, `addHelperCompatSymlinks` |
| `test/src/mac/electronMacTest.ts` | **New** — 28-test suite for the above helpers |
| `.changeset/fix-mac-helper-symlinks-macos26.md` | **New** — patch changeset |

## References

- Closes https://github.com/electron-userland/electron-builder/issues/9771
- Electron v42 source: `shell/app/electron_main_delegate_mac.mm` — `OverrideChildProcessPath()` and `GetHelperAppPath()`
- Electron v42 source: `shell/common/application_info_mac.mm` — `GetApplicationName()` reads `CFBundleName`

### DIFF
--- a/.changeset/fix-mac-helper-symlinks-macos26.md
+++ b/.changeset/fix-mac-helper-symlinks-macos26.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(mac): add compat symlinks after helper renaming to ensure Electron's primary helper-discovery path resolves correctly on all macOS versions
--- a/packages/app-builder-lib/src/electron/electronMac.ts
+++ b/packages/app-builder-lib/src/electron/electronMac.ts
@@ -8,44 +8,22 @@ import { MacPackager } from "../macPackager"
 import { normalizeExt } from "../platformPackager"
 import { savePlistFile, parsePlistFile, PlistObject, PlistValue } from "../util/plist"
 import { createBrandingOpts } from "./ElectronFramework"
+import { addHelperCompatSymlinks, assertSafeHelperName, getAvailableHelperSuffixes } from "./electronMacUtils"
+export { addHelperCompatSymlinks, assertSafeHelperName, getAvailableHelperSuffixes }
 
 function doRename(basePath: string, oldName: string, newName: string) {
   return rename(path.join(basePath, oldName), path.join(basePath, newName))
 }
 
 function moveHelpers(helperSuffixes: Array<string>, frameworksPath: string, appName: string, prefix: string): Promise<any> {
+  assertSafeHelperName(appName, "Product name")
+  assertSafeHelperName(prefix, "Electron branding name")
   return Promise.all(
     helperSuffixes.map(suffix => {
       const executableBasePath = path.join(frameworksPath, `${prefix}${suffix}.app`, "Contents", "MacOS")
       return doRename(executableBasePath, `${prefix}${suffix}`, appName + suffix).then(() => doRename(frameworksPath, `${prefix}${suffix}.app`, `${appName}${suffix}.app`))
     })
   )
-}
-
-function getAvailableHelperSuffixes(
-  helperEHPlist: PlistObject | null,
-  helperNPPlist: PlistObject | null,
-  helperRendererPlist: PlistObject | null,
-  helperPluginPlist: PlistObject | null,
-  helperGPUPlist: PlistObject | null
-) {
-  const result = [" Helper"]
-  if (helperEHPlist != null) {
-    result.push(" Helper EH")
-  }
-  if (helperNPPlist != null) {
-    result.push(" Helper NP")
-  }
-  if (helperRendererPlist != null) {
-    result.push(" Helper (Renderer)")
-  }
-  if (helperPluginPlist != null) {
-    result.push(" Helper (Plugin)")
-  }
-  if (helperGPUPlist != null) {
-    result.push(" Helper (GPU)")
-  }
-  return result
 }
 
 /** @internal */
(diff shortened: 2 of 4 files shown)


mmaietta:
Discussed further with electron devs. Symlinking is the incorrect approach; the CFBundleName and the helper must be the same name. Closing and will remove NFD sanitization of productName in v27 major semver bump

PR #9877, opened 2026-06-11
feat: signed update manifests (Ed25519)

mmaietta:
_Requested by **Mike Maietta** · [Slack thread](https://electronhq.slack.com/archives/C0BGNPTEH8A/p1788058282556049)_

## Summary

Auto-update integrity previously rested on TLS plus a self-referential checksum: `latest*.yml` carries each artifact's `sha512`, and the updater validates the download against it — but nothing authenticated the manifest itself. Anyone able to serve the manifest (compromised host/CDN, stolen publish token, MITM) could ship a malicious artifact with a matching hash.

This PR closes that gap with **Ed25519-signed update manifests**: optionally sign `latest*.yml` at publish time and verify the signature in `electron-updater` **before any download** (fail-closed). This authenticates the manifest, which in turn makes the existing `sha512` artifact check cryptographically meaningful end-to-end. Installs trust a **list** of public keys and a manifest may carry **several signatures**, so keys can be rotated without a flag day.

> **Note:** the Linux unsigned-package install gating originally bundled here landed separately on master as `AppUpdater.allowUnverifiedLinuxPackages` ([#9990](https://github.com/electron-userland/electron-builder/pull/9990)). This PR's docs (new Signed Update Manifests and Key Rotation guides, auto-update, security pages) cross-reference it — manifest signing protects the update *metadata*; that flag governs the package manager's own GPG checks at install time.

## Design

- **Signature scheme:** Ed25519 via Node's built-in `crypto` (no new dependencies). Signatures are embedded directly in `latest*.yml` — one file, works with every provider/CDN, immune to YAML formatting drift:
  - `signatures: [{ keyId, signature }]` — one base64 signature per signing key, tagged with the key's id (`keyId` = lowercase hex SHA-256 of the public key's SPKI DER, `computeUpdateManifestKeyId`).
  - `signature` — legacy single field, always the **first** signing key's signature, kept so a single-signature verifier keeps working. Both fields are always written, even with one key.
- **What is signed:** a deterministic canonical serialization (`canonicalizeForSigning`, prefixed `EBUM1` as a versioned wire contract shared byte-identically by signer and verifier) of the integrity- and rollout-critical fields only:
  - `version`, `stagingPercentage`, `minimumSystemVersion` (its absence is signed too, so an OS-version gate can't be added or removed after signing);
  - each `files[]` entry's `url` / `sha512` / `size` (sorted, so file order doesn't matter);
  - each NSIS web-installer `packages[arch]` entry's arch key and `path` / `sha512` / `size` / `blockMapSize` / `isAdminRightsRequired`.
  - Cosmetic fields (`releaseDate`, `releaseNotes`, `releaseName`) are excluded so they can change post-signing. `signature`/`signatures` themselves are not covered, so signatures can be added/removed independently.
  - **The encoding is injective:** every value goes through `encodeField` (strings JSON-quoted, numbers bare, absent = empty), so no value can contain an unescaped newline/tab or impersonate another record — two manifests produce the same bytes only if every signed field is identical. On top of that, `validateSignedManifestShape` runs in both the verifier and the signer and rejects a manifest with an empty/missing `files`, wrongly typed fields, or control characters in any signed string. The deprecated top-level `path`/`sha512` mirror is unsigned and is therefore **never** used to download a verified update.
- **Keys:** one secret to manage in the common case. The private key(s) are supplied at publish time; the matching public key(s) are **derived automatically** and embedded into `app-update.yml` as the trust list (the same path `publisherName` already travels).

## Build side

- Signing keys resolved once per platform on `PlatformPackager.updateSigningKeys` (a `MemoLazy` keyed on the effective `updateManifest` config, yielding `Array<KeyObject>`) — the **single source of truth** shared by `updateInfoBuilder` (signing `latest*.yml`) and `PublishManager` (embedding the derived `updateManifestPublicKey` into `app-update.yml`), so the two sides can never disagree about whether signing is on or which keys are in play.
- Resolution precedence: platform-specific `updateManifest` → root config, then within it `signingKey` → `signingKeyFile` → `ELECTRON_BUILDER_UPDATE_SIGN_KEY` → `ELECTRON_BUILDER_UPDATE_SIGN_KEY_FILE` env vars (preferred for CI secrets — env-var-only signing needs no config block at all). The first source that is set wins, but each source may hold **several keys**: config values accept a string or an array, a PEM value may contain several concatenated `-----BEGIN PRIVATE KEY-----` blocks, and `ELECTRON_BUILDER_UPDATE_SIGN_KEY_FILE` accepts several paths joined with `path.delimiter`. Every key must be a distinct Ed25519 key (duplicates / other key types fail the build).
- Relative `signingKeyFile` paths resolve against the **project directory** like every other config path; a relative `ELECTRON_BUILDER_UPDATE_SIGN_KEY_FILE` keeps env-var semantics (cwd).
- `writeUpdateInfoFiles` signs each manifest **after** all multi-arch/zip `files` merges so the signed array is final, resolving the keys per task so each platform's manifest is signed iff that platform's config requires it. Every configured key signs (`createUpdateManifestSignatures`); the signer refuses to sign a manifest the updater would reject (`validateSignedManifestShape`).
- Trust list embedding (`getAppUpdatePublishConfiguration`): an explicit `updateManifest.publicKey` (string or array) is embedded as-is; otherwise the public half of every signing key is derived. One key is written as a plain string (byte-identical to the single-key format), several as a YAML list. A manually set `publish.updateManifestPublicKey` is rejected with `InvalidConfigurationError` (the field is managed by electron-builder). A build-time warning flags an explicit `publicKey` list that contains none of the signing keys (only sensible for a deliberate bridge release).

```jsonc
// electron-builder config (top-level or per platform, e.g. under "win" / "linux")
"updateManifest": {
  "signingKey": "-----BEGIN PRIVATE KEY-----…",   // string | string[]; or "signingKeyFile": "./update-private-key.pem" (string | string[], project-relative)
  // "publicKey": ["<current>", "<next>"]          // optional string | string[] — derived from the signing key(s) when omitted
}
```

## Runtime side (`electron-updater`)

- `AppUpdater.updateManifestPublicKey: string | string[] | null` — the trust list; when set (non-empty) it overrides the value in `app-update.yml`.
- Verification (`verifyManifestSignature`) runs in `getUpdateInfoAndProvider` immediately after the manifest is fetched — **provider-agnostic** (Generic, GitHub, GitLab, Keygen, Bitbucket, S3, Spaces all covered) and on the raw manifest, before the provider resolves `files[].url` / `packages[arch].path`:
  - No key configured → verification skipped, one-time warning (opt-in phase).
  - Key(s) configured, manifest carries no `signature` and no `signatures` → throws `ERR_UPDATER_MANIFEST_NOT_SIGNED`.
  - Key(s) configured, manifest malformed (fails `validateSignedManifestShape`) or no trusted key validates any signature → throws `ERR_UPDATER_MANIFEST_SIGNATURE_INVALID` with the reason in the message.
  - Accepted when **any** trusted key validates **any** signature (`verifyManifestSignatures`): entries tagged with that key's `keyId` are tried first, then the untagged legacy `signature`, so a stale/foreign `keyId` never blocks a legitimately signed manifest.
- A throw propagates before any download starts (fail-closed).

## Key rotation

Because an install trusts a list and a manifest can carry several signatures, rotation is a transition rather than a flag day:

1. **Bridge release(s):** sign with `[old, new]` (old first so it also fills the legacy `signature`), e.g. `ELECTRON_BUILDER_UPDATE_SIGN_KEY_FILE=old.pem:new.pem`. The derived trust list embeds both; installs in the field verify via the old signature and afterwards trust `[old, new]`.
2. **Keep dual-signing** for the support window.
3. **Drop the old key.** Manifests are signed by the new key only; installs that skipped every bridge release fail with `ERR_UPDATER_MANIFEST_SIGNATURE_INVALID`.

Recommended steady state: embed the *next* key ahead of time via `updateManifest.publicKey: [current, next]` so a compromise can be answered by switching keys immediately.

## Tooling & docs

- New CLI: `electron-builder create-update-key [--out <path>]` — generates an Ed25519 keypair, writes the private key `0600` (created exclusively; refuses to overwrite an existing file), and prints the public key and its key id.
- New docs pages: *Signed Update Manifests* (quick setup, config reference, what is written/signed, verification behavior, key rotation) and *Key Rotation* (runbook for the Ed25519 manifest key plus Windows/macOS/Linux code-signing rotation), added to the sidebar and linked from the auto-update, security and code-signing pages.

## Tests

- **Unit:** `updateManifestSignatureTest` (canonical form incl. exact `EBUM1` wire format, tamper cases for every signed field incl. `minimumSystemVersion`/`packages`, injectivity — the empty-`files` + `minimumSystemVersion` forgery and newline/tab smuggling — `validateSignedManifestShape`, key ids, multi-PEM splitting, trust-list / multi-signature verification), `manifestSignatureUpdaterTest` (`AppUpdater` error codes, dual-signed / trust-list acceptance, fail-closed rejection, malformed-manifest rejection before any legacy-path fallback), `updateInfoBuilderTest` (per-task signing, dual-signing, key resolution from config/env/files incl. project-relative `signingKeyFile`, `app-update.yml` trust-list embedding and `publish.updateManifestPublicKey` rejection), `createUpdateKeyTest` (0600 file, refuses to overwrite).
- **Blackbox e2e:** `signed update manifest` (two builds signed with a runtime-generated key; the installed app verifies before updating) and `key rotation` (A → [A, B] → B over three builds, including served manifests the installed app must refuse with `ERR_UPDATER_MANIFEST_SIGNATURE_INVALID` / `ERR_UPDATER_MANIFEST_NOT_SIGNED`) on macOS (x64 zip; separate `blackboxUpdateSignedManifest.e2e.ts` / `blackboxUpdateKeyRotation.e2e.ts` shards), Windows (NSIS, latest toolset), and Linux (AppImage, deb/debian, rpm/fedora, pacman/arch).

## Rollout & compatibility

- **Opt-in everywhere.** Signing is off unless a key is configured; verification is off unless a public key is present. Zero breakage for existing apps.
- **Forward/backward compatible.** An older updater ignores the unknown `signature`/`signatures` fields; a newer updater against an unsigned manifest with no configured key behaves exactly as before. `updateManifestPublicKey` stays a plain string for a single key, so single-key `app-update.yml` is byte-identical to the pre-trust-list format.
- **Changeset:** `minor` for `builder-util-runtime`, `builder-util`, `app-builder-lib`, `electron-updater`, `electron-builder`.


packages/app-builder-lib/src/publish/updateInfoBuilder.ts
```diff
@@ -216,6 +217,33 @@ export async function writeUpdateInfoFiles(updateInfoFileTasks: Array<UpdateInfo
...
+      loggedSigning = true
+      log.info(null, "signing update manifests with Ed25519 key")
+    }
+    let signingKey = signingKeyCache.get(signingKeyPem)
+    if (signingKey == null) {
+      signingKey = parsePrivateKey(signingKeyPem)
+      signingKeyCache.set(signingKeyPem, signingKey)
+    }
+    return signingKey
+  }
```
mmaietta: @claude please use `MemoLazy` here so that we don't need to maintain 3 properties. MemoLazy will:
- only regenerate value on provided a new `updateManifest` argument and log only during the _setting_ of the Lazy value. https://github.com/electron-userland/electron-builder/blob/master/packages/builder-util-runtime/src/MemoLazy.ts

PR #9889, opened 2026-06-12
feat: modernize macOS build config API and upgrade all `electron/____` dep packages (BREAKING)

### LINKED ISSUE #9770: Update @electron/notarize to v3
Actual version is 2 years old.

https://github.com/electron/notarize/releases

mmaietta:
## Summary

This PR upgrades every `@electron/*` build-tooling dependency to its latest major and reshapes the macOS build-config API to pass through to those libraries directly instead of via hand-maintained translation layers.

### Dependency upgrades

| Package | Before | After | Drives |
|---|---|---|---|
| `@electron/osx-sign` | 1.3.3 | **2.4.0** | `mac.sign` → `ElectronSignOptions` pass-through |
| `@electron/universal` | 2.0.3 | **3.0.4** | `mac.universal` → `ElectronUniversalOptions` pass-through |
| `@electron/get` | ^3.0.0 | **^5.0.0** | `electronDownload` → `electronGet`; `got` → `fetch` downloader |
| `@electron/asar` | 3.4.1 | **4.1.1** | `asarFileChecker` static import + sync API (held below `4.1.2` — see *Bug fix — embedded ASAR integrity*) |
| `@electron/notarize` | 2.5.0 | **3.1.1** | static import; notarytool-only path |
| `@electron/fuses` | ^1.8.0 | **^2.1.1** | `FuseOptionsV1` typed against `FuseV1Options` |

### Breaking config changes

- **[BREAKING] `mac.sign` consolidation** — ~15 scattered signing fields removed from `MacConfiguration` root and replaced with a single `sign?: CustomMacSign | ElectronSignOptions | string | null`. `ElectronSignOptions` is a typed pass-through to `@electron/osx-sign` v2, declared as `interface ElectronSignOptions extends Omit<OnlySignOptions, "optionsForFile" | "version" | "type">` plus convenience shorthand fields (`identity`, `entitlements`, `hardenedRuntime`, `requirements`, `timestamp`, …). It is an **`interface … extends Omit<>`** rather than a `type = Omit<> & { … }` intersection because `typescript-json-schema` silently drops the inline `& { … }` member of an intersection during schema generation — the interface form emits all own properties into `scheme.json` so AJV can validate/coerce them. **`mac.identity` moves into `sign.identity`**, mirroring the Windows `win.sign` grouping so all signing config lives in one bag (certificate still selectable via `CSC_LINK`/`CSC_NAME`). `signIgnore` is renamed to `sign.ignore` (the `@electron/osx-sign` canonical name). Same shape applies to `mas`/`masDev`.

- **[BREAKING] `mac.universal` consolidation** — `mergeASARs`, `singleArchFiles`, `x64ArchFiles` removed from root and replaced with `universal?: ElectronUniversalOptions | null`, declared as `interface ElectronUniversalOptions extends Omit<MakeUniversalOpts, "x64AppPath" | "arm64AppPath" | "outAppPath" | "force">`. No effect unless `arch: "universal"`.

- **[BREAKING] `electronDownload` → `electronGet`** — `Configuration.electronDownload` renamed to `electronGet`, typed directly as `ElectronGetOptions` (a thin `Omit` of `@electron/get`'s `ElectronPlatformArtifactDetails`). The legacy `ElectronDownloadOptions` type and its runtime translation bridge are removed.

- **[BREAKING] `MasConfiguration` flattened** — now an empty interface extending `MacConfiguration`; the duplicated `entitlements`/`entitlementsInherit`/`binaries` fields are gone (use `mas.sign.*`).

### Bug fixes — PlatformType threading

`PlatformType` (`"mac" | "mas" | "mas-dev"`) is now carried as `platformType?` through `DoPackOptions` and `BeforeCopyExtraFilesOptions` (internal), and threaded directly as a parameter to `signApp()` — kept **out** of the public `PackContext`/`AfterPackContext`. This closes the point where the 3-way flavor collapsed to the lossy 2-way `ElectronPlatformName = "darwin" | "mas"`, fixing four defects:

- **Defect A** — `masDev` cascade config (e.g. `bundleVersion`) was silently dropped for `mas-dev` targets (`getPlatformConfig("mas")` instead of `getPlatformConfig("mas-dev")`).
- **Defect B** — `isMasFlavor` excluded `mas-dev`, so `mas-dev` builds got the non-MAS localhost ATS config injected into `Info.plist`.
- **Defect C** — `mas-dev` signed with Apple **Distribution** certs instead of Apple **Development** certs because `isDevelopment` was never threaded to signing.
- **Defect D** — `createMasInstaller` (a distribution-only step) ran for `mas-dev` builds.

### Bug fix — embedded ASAR integrity (`@electron/asar` held at 4.1.1)

`@electron/asar` `4.1.2`–`4.2.0` compute a packed file's embedded integrity from `fs.readFileSync(<destination path>)` instead of the stream being packed. electron-builder packs from streams (to support content transforms), so the destination path resolves against the build **CWD**; when the CWD holds a same-named relative path (e.g. `package.json`, `node_modules/<dep>/…` — the norm when building from a project root) the header records the **wrong** file's hash while the archive stores the correct bytes. Apps built with `enableEmbeddedAsarIntegrityValidation` then fail to launch — Electron rejects `app.asar` before any application JS runs, the process exits immediately with no stdout.

This is what broke the Windows **`blackboxWin`** e2e suite (`result.version` `undefined` → `.toMatch()` on `undefined`); macOS/Linux masked it because their blackbox **launch** test is in `skipPerOSTests`. Pinned to **`4.1.1`**, the last release that hashes the stream content (`getFileIntegrity(streamGenerator())`). Regression guard added in [`test/src/asarIntegrityTest.ts`](test/src/asarIntegrityTest.ts) (verified: fails on `4.2.0` with a `package.json` integrity mismatch, passes on `4.1.1`); upstream write-up in [`ELECTRON_ASAR_INTEGRITY_BUG_REPORT.md`](ELECTRON_ASAR_INTEGRITY_BUG_REPORT.md).

### Internal modernization (driven by the dep majors)

- **Import modernization** — Deep `@electron/osx-sign/dist/cjs/*` imports replaced with top-level `@electron/osx-sign` imports; `Identity` re-exported from `macCodeSign.ts`. Dynamic `import()` of `@electron/notarize`, `@electron/universal`, and `@electron/asar` replaced with static imports. (This is distinct from the `nodenext` switch already on `origin/master`.)
- **`@electron/get` v5 adapter** — Removed the `got`-based proxy path (`GotDownloaderOptions`, `buildGotProxyAgent()`); v5 downloads via `fetch`, with proxy wired through `initializeProxy()`/`reinitializeProxy()` (reads `HTTP(S)_PROXY`/`NO_PROXY`). Timeout switched to `AbortSignal.timeout(...)`. `ElectronArtifactDetails` replaces the inferred `Parameters<typeof get.downloadArtifact>[0]` type.
- **`@electron/asar` v4 adapter** — `asarFileChecker.ts` switches to a static `import * as asar`; v4 no longer re-exports filesystem entry types, so they're derived from `statFile`'s return type, and `checkFileInArchive` becomes synchronous.
- **`FuseOptionsV1` extracted** — Moved from `configuration.ts` to its own file and typed against `FuseV1Options` from `@electron/fuses` v2 to enforce compile-time completeness when upstream adds fuse enums.
- **`MacTargetHelper` API cleanup** — `getCertificateTypes`, `isHardenedRuntimeEnabledForSigning`, `getOptionsForFile`, `buildSignOptions`, `findSigningIdentity`, `createMasInstaller` now take `PlatformType` (not `isMas`/`isDevelopment` boolean pairs) or `ElectronSignOptions | null` (not the full `MacConfiguration`); added `isMasDevelopment()`.
- **`MacPackager` cleanup** — Removed the `_activePackConfig` workaround (superseded by threaded `platformType`); `sign()` takes `targetPlatform: PlatformType`.
- **`electronMac.ts` cleanup** — `createMacApp` receives `ElectronPlatformName` not `isMas`; bundle/helper IDs cascade via `getOptionsForTarget(targetPlatform)`.
- **`dmg.ts` / `pkg.ts`** — `signDmg` reads identity from `sign.identity`; `pkg.ts` uses the pkg target's own `pkg.identity` (dropped the `|| mac.identity` fallback).
- **CLI nested coercion (`builder.ts`)** — `normalizeOptions` coerces `mac.sign` and `mac.universal` (these live behind an `anyOf` union AJV can't coerce into); restores `-c.mac.sign.identity=null` / `-c.mac.sign.hardenedRuntime=false` / `-c.mac.sign=null`.
- **`scripts/fix-schema.js`** — Removed the obsolete `ElectronDownloadOptions.checksums` schema patch (the type no longer exists).

### Tooling & docs

- **`migrate-schema`** — Added v26→v27 transforms: mac/mas/masDev signing fields → `sign.*` (with `signIgnore` → `sign.ignore` and `sign: null` semantic-preservation), universal fields → `universal.*`, and `electronDownload` → `electronGet` (with warnings for the dropped `cache`/`customDir`/`customFilename`/`strictSSL` fields). Custom-signer strings are warned about, not clobbered.
- **Migration docs** — [`MIGRATION.md`](MIGRATION.md) and [`website/docs/migration/v26-to-v27.md`](website/docs/migration/v26-to-v27.md) updated with the new auto-migrated rows, restructured-config sections, checklist items, and design notes.
- **API-reference rendering of pass-through options** — `ElectronSignOptions` and `ElectronUniversalOptions` are now re-exported from `app-builder-lib`'s public entry point; `ElectronUniversalOptions` and `ElectronGetOptions` were converted from `type = Omit<…>` aliases to `interface … extends Omit<…>`. TypeDoc only expands and lists the inherited members of an `Omit<ExternalType>` when the symbol is **(a)** exported from a package entry point and **(b)** declared as an `interface extends Omit<>` (a `type =` alias renders as the bare `Omit<…>` expression with no member list). With both in place, every forwarded field from `@electron/osx-sign`, `@electron/universal`, and `@electron/get` — e.g. `batchCodesignCalls`, `binaries`, `provisioningProfile`, `mergeASARs`, `singleArchFiles`, `checksums` — now renders on the docs site with its upstream TSDoc, and `MacConfiguration`'s `sign`/`universal`/`electronGet` `@see {@link …}` references resolve to real pages. `website/docs/api/*` is git-ignored and regenerated by the docs build, so no generated files are committed.

### `mac.sign.*` (all signing options, incl. `identity`)

```jsonc
// Before
{
  "mac": {
    "identity": "Developer ID Application: ...",
    "entitlements": "build/entitlements.mac.plist",
    "hardenedRuntime": true,
    "gatekeeperAssess": false,
    "signIgnore": ["**/*.txt"]
  }
}

// After — all signing options move into sign.* (mirrors win.sign)
{
  "mac": {
    "sign": {
      "identity": "Developer ID Application: ...",
      "entitlements": "build/entitlements.mac.plist",
      "hardenedRuntime": true,
      "gatekeeperAssess": false,
      "ignore": ["**/*.txt"]
    }
  }
}
```

### `mac.universal.*`

```jsonc
// Before
{ "mac": { "mergeASARs": true, "singleArchFiles": "*.node" } }

// After
{ "mac": { "universal": { "mergeASARs": true, "singleArchFiles": "*.node" } } }
```

### `electronDownload` → `electronGet`

```jsonc
// Before
{ "electronDownload": { "mirror": "https://my-mirror/" } }

// After — ElectronGetOptions shape (maps directly to @electron/get v5)
{ "electronGet": { "mirrorOptions": { "mirror": "https://my-mirror/" } } }
```

> Run `electron-builder migrate-schema` to apply all three automatically.


packages/app-builder-lib/src/targets/mac/MacTargetHelper.ts
```diff
@@ -41,24 +40,16 @@ export class MacTargetHelper {
     }
 
     if (identity == null) {
-      if (!isMas && !isDevelopment && config.type !== "distribution") {
-        identity = await findIdentity("Mac Developer", qualifier, keychainFile)
-        if (identity != null) {
-          log.warn("Mac Developer is used to sign app — it is only for development and testing, not for production")
-        }
-      }
```
liamcmitchell, line 49: @mmaietta removing this path was an undocumented breaking change for me.

My fix:
```diff
   mac: {
     target: ["dmg"],
-    type: "development",
+    sign: {
+      identity: "-",
+    },
```

Using `identity: "-"` is well  documented: https://www.electron.build/docs/mac#certificate-identity

Not sure if it was always that way or how I ended up with `type: "development"` instead.

mmaietta: Thanks for the ping @liamcmitchell ! That should have been transfered internal to `sign.type` as `identity: -` is  different and represents an adhoc signing identity.

Let me get that fixed up asap.

liamcmitchell: PR for fix here: https://github.com/electron-userland/electron-builder/pull/10080

PR #9890, opened 2026-06-12
chore: resolve all zizmor GitHub Actions security findings

mmaietta:
## Summary

Ran `zizmor .` audit on all workflows in the repo.

- **Added `persist-credentials: false`** to every `actions/checkout` step across all workflows — prevents the default Git credential helper from persisting tokens into the checkout, limiting exposure to downstream steps and sub-processes ([artipacked](https://docs.zizmor.sh/audits/#artipacked) audit)

- **Moved inline `${{ context }}` expressions out of shell scripts** — replaced with `env:` mappings so GitHub Actions context values are never interpolated directly into shell source code ([template-injection](https://docs.zizmor.sh/audits/#template-injection) audit). Affects `pr-release.yml`, `docker-build.yml`, `test.yaml`, and `pretest/action.yml`

- **`VITEST_CACHE_REF`: switched from `$GITHUB_ENV` write to `$GITHUB_OUTPUT`** — the computed cache ref is now written to `$GITHUB_OUTPUT` (via `id: compute-cache-ref`) and passed as an explicit `env:` entry only to the `pnpm ci:test:fetch-cache` step that needs it. The calling workflow's upload step already recomputes the ref independently, so no calling-side changes are needed ([github-env](https://docs.zizmor.sh/audits/#github-env) audit)

- **`RESET_VITEST_SHARD_CACHE`: switched to conditional static write** — the action input (boolean) now controls a branch that writes the literal string `"true"` or `"false"`, so no tainted value flows to `$GITHUB_ENV` ([github-env](https://docs.zizmor.sh/audits/#github-env) audit)

## Changed Files

| File | Change |
|------|--------|
| [.github/actions/pretest/action.yml](.github/actions/pretest/action.yml) | `VITEST_CACHE_REF` → `$GITHUB_OUTPUT` + step `env:`; `RESET_VITEST_SHARD_CACHE` → conditional static write; move `${{ github.event.pull_request.number }}` to `env:` |
| [.github/workflows/deploy-docker.yml](.github/workflows/deploy-docker.yml) | `persist-credentials: false` on checkout |
| [.github/workflows/deploy-netlify.yml](.github/workflows/deploy-netlify.yml) | `persist-credentials: false` on checkout |
| [.github/workflows/docker-build.yml](.github/workflows/docker-build.yml) | `persist-credentials: false`; `${NODE_TAG}` env var instead of `${{ env.NODE_TAG }}` inline in shell |
| [.github/workflows/pr-labeler.yml](.github/workflows/pr-labeler.yml) | `persist-credentials: false` on checkout |
| [.github/workflows/pr-netlify.yml](.github/workflows/pr-netlify.yml) | `persist-credentials: false` on checkout |
| [.github/workflows/pr-release.yml](.github/workflows/pr-release.yml) | `${{ github.ref_name }}`, `${{ steps.*.outputs.* }}` expressions moved to `env:` blocks in shell steps |
| [.github/workflows/test.yaml](.github/workflows/test.yaml) | `persist-credentials: false` on all 9 checkout steps; `${GITHUB_REF_NAME}` env var in cache step |

## Design notes

**Why `$GITHUB_OUTPUT` instead of `$GITHUB_ENV` for `VITEST_CACHE_REF`?**
Writing a tainted value (branch name, PR number) to `$GITHUB_ENV` using the single-line `echo "VAR=$VALUE"` form is dangerous: if `$VALUE` contains `\n`, it can inject additional env vars for all subsequent steps in the job. The delimiter form partially mitigates this but zizmor still traces the taint. `$GITHUB_OUTPUT` is evaluated by the runner, not the shell, so it carries a fundamentally different (and narrower) trust model. The value is then bound to a single step via `env:` at the YAML level, where it is set by the runner before the shell starts.

**Why a conditional static write for `RESET_VITEST_SHARD_CACHE`?**
The input is boolean — it will always be `"true"` or `"false"`. Writing the literal string directly (not the variable) breaks the taint chain: `$GITHUB_ENV` receives only strings that could never have been injected by an attacker, regardless of what the input contained.


.github/workflows/test.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
       force-docker-build: ${{ steps.changes.outputs.docker == 'true' || steps.changes.outputs.workflow == 'true' }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with: { fetch-depth: 0 }
+        with: { fetch-depth: 0, persist-credentials: false }
```
dsanders11: nit:
```suggestion
        with:
          fetch-depth: 0
          persist-credentials: false
```

That's unusual syntax for a YAML file, so let's make it consistent with everything else.

dsanders11 (dismissed):

dsanders11 (approved):

PR #9918, opened 2026-06-16
fix(linux): write `executableArgs` verbatim into the `.desktop` Exec

### LINKED ISSUE #9914: Desktop file produced for Linux packages is broken if `executableArgs` param was filled
The https://github.com/electron-userland/electron-builder/pull/9778 PR made `executableArgs` wrapped into `""` which made the desktop file invalid. The package gets installed, but you can't start it from the apps menu. If you remove `""` from the desktop file `Exec` args and update the desktop file cache, the app appears in the menu, and you can start it from there then as usual.

For example, the config has this (real app used to have this arg for many years):
```yaml
  executableArgs:
    - '--js-flags="--max-old-space-size=12288"'
```
So now you get `Exec=/opt/myapp/myapp-binary "--js-flags=\"--max-old-space-size=12288\"" %U` in the desktop file which is invalid on recent Ubuntu at least - it refuses to list this desktop file in the menu. Versus `Exec=/opt/myapp/myapp-binary --js-flags="--max-old-space-size=12288" %U` used to be there for a long time, and `Exec=/opt/myapp/myapp-binary --js-flags=\"--max-old-space-size=12288\" %U` also works (with removed outer wrapping with `""`).

So far, I've tested it on the DEB package only on recent Ubuntu, and this looks like a very breaking change there, at least to non-advanced users.

mmaietta:
Fixes #9914

### Summary

- Linux `.desktop` files are once again generated with `executableArgs` written **verbatim** into the `Exec=` line, restoring the long-standing behavior. Previously each argument was wrapped in an outer pair of double quotes whenever it contained certain characters, which produced an `Exec` line that recent Ubuntu/GNOME refuses to list in the applications menu.

  For example, with:

  ```yaml
  linux:
    executableArgs:
      - '--js-flags="--max-old-space-size=12288"'
  ```

  the generated entry is now:

  ```
  Exec=/opt/MyApp/myapp --js-flags="--max-old-space-size=12288" %U
  ```

  instead of the previous menu-breaking form:

  ```
  Exec=/opt/MyApp/myapp "--js-flags=\"--max-old-space-size=12288\"" %U
  ```

- Extracted a small shared helper `buildExecArgs(executableArgs)` in the Linux target helper that joins the arguments verbatim and strips only carriage-return / newline characters (the `Exec` key is a single line, so an embedded line break would otherwise split it into additional entries).

- The DEB/RPM (`fpm`) path and the AppImage path now both run their arguments through this single helper, so the two targets produce identical `Exec` argument formatting. Previously, only the AppImage path joined arguments verbatim, while the `fpm` path applied the extra quoting — the source of the regression.

- Field codes (`%U`, `%f`, etc.) and the executable-path quoting are unchanged. String fields such as `Name`, `Comment`, and `StartupWMClass` continue to be escaped per the freedesktop string rules.

### Notes

- Argument grouping remains the developer's responsibility, matching the historical behavior: an argument intended to contain spaces should be provided as a single `executableArgs` entry, exactly as it was before.
- No snapshot output changes for existing default configurations (e.g. the default `--no-sandbox`), confirmed by the existing AppImage suite passing unchanged.

### DIFF
--- a/.changeset/linux-exec-args-verbatim.md
+++ b/.changeset/linux-exec-args-verbatim.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(linux): write `executableArgs` verbatim into the `.desktop` Exec line so packaged apps (deb/rpm) still appear in the application menu
--- a/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
+++ b/packages/app-builder-lib/src/targets/LinuxTargetHelper.ts
@@ -35,39 +35,18 @@ function desktopStringEscape(value: string): string {
 }
 
 /**
- * Characters that require an Exec argument to be double-quoted per the
- * freedesktop Desktop Entry Specification.  Plain alphanumeric args and
- * field codes must NOT be wrapped in quotes.
+ * Build the argument portion of a .desktop Exec line.
+ *
+ * executableArgs are written verbatim — matching AppImage and the behavior that
+ * shipped for years — so an arg like `--js-flags="--max-old-space-size=12288"`
+ * reaches the launcher exactly as the developer specified. Only carriage-return
+ * and newline characters are removed: the Exec key is a single line, and an
+ * embedded line break would split it into additional .desktop entries.
  *
  * @see https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables
  */
-const EXEC_RESERVED_RE = /[\s"'`\\<>~|&;$*?#()]/
-
-/**
- * Quote a single argument for use in a .desktop file Exec key.
- *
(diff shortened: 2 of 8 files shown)


vladimiry:
> The DEB/RPM (fpm) path and the AppImage path now both run their arguments through this single helper

Why not have it unified to all Linux targets? I mean, if `executableArgs` applies to all Linux targets, why not process it in the same way to all involved targets?

mmaietta:
@vladimiry Great question. Snap and Flatpak already use an entrypoint.
- Snap (legacy) passes args into a command.sh wrapper for execution. Snap (core24) in v27-alpha does the same as well.
- Flatpak has a wrapper launcher script

All other targets fold into the FpmTarget here

Issue #2517, opened 2018-01-23
Windows NSIS installer defaults to German if system language is not supported by electron-builder


agritchie:
I'm using electron-builder version 19.37.2 to build my Electron project into a Windows installer via NSIS. This issue was seen on both Windows 10 and 7.

If Windows is set to a language that is not supported by electron-builder, rather than defaulting to English, the NSIS installer ends up displaying German. This was seen when Windows was set to Thai, Turkish, and Icelandic.

Wello6143:
Yeah, it also seen on Vietnamese too.

PR #9927, opened 2026-06-16
feat(nsis): add Hebrew (he) translations and bundle he_IL for installer messages

kdroidFilter:
## Summary

Completes the Hebrew (`he`) localization for the NSIS installer message templates **and ensures those strings actually reach the built installer**.

- **`messages.yml`** — adds `he` for the 5 keys that were still missing it: `win7Required`, `x64WinRequired`, `installing`, `areYouSureToUninstall`, `appClosing`. (`appRunning`, `appCannotBeClosed`, `decompressionFailed` and `uninstallFailed` already had Hebrew.)
- **`assistedMessages.yml`** — adds `he` for all 17 keys (the file previously had no Hebrew at all).
- **`langs.ts`** — adds `he_IL` to `bundledLanguages`, fixing a pre-existing bug where Hebrew was never included in the default multi-language set. Without this, `toLangWithRegion("he")` did not map to `he_IL` and none of the Hebrew strings (new or pre-existing) reached the built installer.

The `${PRODUCT_NAME}` placeholders and `\n` line breaks are preserved, and the new entries are placed right after the existing `fr` line, matching the convention already used for the `he` entries in `messages.yml`.

## Notes

- These are the only two YAML files in the repo that carry per-language installer strings, so Hebrew installer coverage is now complete.
- Both files were validated as parseable YAML and every top-level key now has a `he` value.
- `en` remains first in every block (per #2517).

### DIFF
--- a/.changeset/he-nsis-translations.md
+++ b/.changeset/he-nsis-translations.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+feat(nsis): add Hebrew (he) translations for one-click and assisted installer messages
--- a/.changeset/he-nsis-translations.md
+++ b/.changeset/he-nsis-translations.md
@@ -3,3 +3,5 @@
 ---
 
 feat(nsis): add Hebrew (he) translations for one-click and assisted installer messages
+
+Also add `he_IL` to `bundledLanguages` so Hebrew strings actually reach the built installer (fixes a pre-existing bug where Hebrew was never included in the default multi-language set).
--- a/packages/app-builder-lib/src/util/langs.ts
+++ b/packages/app-builder-lib/src/util/langs.ts
@@ -22,6 +22,7 @@ export const bundledLanguages = [
   "sk_SK",
   "hu_HU",
   "ar_SA",
+  "he_IL",
   "tr_TR",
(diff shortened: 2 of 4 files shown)


mmaietta:
Thanks for contributing! I had a quick analysis ran on this PR and discovered that that none of the Hebrew strings (these new ones or the 4 that already existed) actually reach a built installer today. This is a pre-existing bug, not introduced by this PR, but I feel could be looped into this PR while we're at it.

One-line fix is to add "he_IL" to `bundledLanguages` in `langs.ts`. Then `toLangWithRegion("he")` maps to he_IL (the map is built from `bundledLanguages`), and Hebrew joins the default multi-language set.

Let me know if you'd prefer to have it in a separate PR and we can easily do that instead.

kdroidFilter:
Thanks for the thorough analysis! That makes total sense — no point shipping the translations if they never reach the installer. I've folded the fix into this PR: added `he_IL` to `bundledLanguages` in `langs.ts`, so `toLangWithRegion("he")` now maps to `he_IL` and Hebrew joins the default multi-language set. Updated the changeset accordingly. Let me know if you'd like anything else!

PR #9988, opened 2026-06-25
fix(nsis): reliably install the main executable and native binaries on x64 and arm64

### LINKED ISSUE #9983: NSIS installer silently skips all PE files (missing main exe + native binaries) — bundled 7za 24.09 BCJ2 filter not decodable by install-time Nsis7z (26.15 archiver regression, sub-2GB)
## Summary

On Windows, the NSIS installer **silently** drops every PE file (the main `.exe`, all `.dll`s, and bundled `.exe`/`.node` binaries) at install time. The installed directory ends up containing only the **data** files (`app.asar`, `*.pak`, `icudtl.dat`) plus the NSIS-generated `Uninstall <App>.exe`. The installer exits `0` with no error or warning, so the app is simply unrunnable (no main executable on disk).

The root cause is a **producer/decoder mismatch on the `BCJ2` 7z filter**, not the well-known 2 GB NSIS size limit. The app `.7z` is packed by the modern bundled `7za` (24.09), which applies the `BCJ2` filter to executable content. The 7z decoder embedded in the install-time `Nsis7z::Extract` plugin cannot decode that `BCJ2` stream, so it writes the plain-`LZMA2` blocks (data) and silently skips the `BCJ2` blocks (every executable).

This is reproducible and **100% correlated** with the filter (evidence below), and is fully worked around by forcing the simpler `BCJ` filter via `ELECTRON_BUILDER_7Z_FILTER=BCJ`.

## Environment

- electron-builder: **26.15.5** (also reproduced on 26.15.3)
- electron: 42.4.0
- Bundled packer: **7-Zip 24.09** (`Cache/7zip@1.0.0/.../7za.exe`, downloaded by `app-builder-lib/out/toolsets/7zip.js`)
- Target: `nsis` (oneClick, perMachine: false)
- OS: Windows 11 (reproduced on both **x64** and **arm64**)
- App payload: **~778 MB uncompressed** (well under the 2 GB NSIS limit)

## Symptom

After a clean install, the install dir (`%LOCALAPPDATA%\Programs\<app>`) contains:

- ✅ `app.asar`, `locales\*.pak`, `resources.pak`, `icudtl.dat`, `Uninstall <App>.exe`
- ❌ **missing**: `<App>.exe`, every Chromium `*.dll` (`ffmpeg.dll`, `libGLESv2.dll`, `vk_swiftshader.dll`, `d3dcompiler_47.dll`, …), bundled `*.exe` (`gk.exe`, agent CLIs, `OpenConsole.exe`, `elevate.exe`), and native `*.node`

Installer exit code is `0`. No log error. A manual `7z x` of the same embedded payload extracts **everything** correctly, so the payload is intact — only the install-time NSIS extraction drops files.

## Evidence — the skipped files are exactly the `BCJ2`-filtered ones

Listing the embedded `$PLUGINSDIR\app-64.7z` with `7z l -slt`:

| Entry | 7z `Method` | Extracted at install time? |
|---|---|---|
| `icudtl.dat`, `locales\*.pak`, `app.asar`, `resources.pak` (data) | `LZMA2:20` | ✅ yes |
| `<App>.exe`, `ffmpeg.dll`, `libGLESv2.dll`, `vk_swiftshader.dll`, `d3dcompiler_47.dll`, `gk.exe`, codex/agent `*.exe`, `*.node` (PE) | **`BCJ2` LZMA2:20 LZMA:20 …** | ❌ silently skipped |

Every skipped file is `BCJ2`-coded; every extracted file is plain `LZMA2`. The archive header reports `Method = LZMA2:20 LZMA:20 BCJ2`.

## Root cause

electron-builder 26.15 reworked archiving to use the downloaded modern `7za` for all formats (`app-builder-lib/out/targets/archive.js`, see the "matches pre-26.15 behavior" / "Use 7za for all formats" comments; `toolsets/7zip.js` downloads 7-Zip 24.09). At `-mx=9` (the default for the app `.7z`), `7za` auto-applies the `BCJ2` filter to executables.

The install-time extractor is the `Nsis7z::Extract` plugin (`templates/nsis/include/extractAppPackage.nsh`). Its embedded 7z decoder is old and does not handle the `BCJ2` stream produced by 7-Zip 24.09, so those blocks are skipped without raising an error.

This is **distinct from #8399** (the 2 GB `makensis` limit): the payload here is only ~778 MB, and the skip is filter-specific (PE-only), not a size truncation.

## Confirmed workaround

Forcing the single-stream `BCJ` filter (which the `Nsis7z` decoder *does* support) instead of `BCJ2`:

```
ELECTRON_BUILDER_7Z_FILTER=BCJ   # validated value, set before electron-builder runs
```

With this set, the embedded `app-*.7z` packs every PE as `BCJ LZMA2:20` (zero `BCJ2` entries), and a clean silent install lays down `<App>.exe` and all native binaries correctly. Empirically confirmed on both x64 and arm64. `compression: store` (`-mx=0`, no filter) also works but balloons the payload; `-mf=off` (plain `LZMA2`, no filter) would be ideal but isn't reachable since `ELECTRON_BUILDER_7Z_FILTER` only accepts the filter names in `ALLOWED_7Z_FILTERS` (no `off`).

Empirical filter behavior (`7z a <opt>` on a sample PE, then `7z l -slt`):

| args | resulting `Method` |
|---|---|
| `-mx=9` (default) | `LZMA2 + BCJ2` ← breaks NSIS extraction |
| `-mx=9 -mf=BCJ` | `LZMA2 + BCJ` ← works |
| `-mx=9 -mf=off` | `LZMA2` only (no filter) ← would be ideal |
| `-mx=0` (`compression: store`) | `Copy` (no filter) |

## Questions / suggested fix

1. Is `BCJ2` expected to be decodable by the bundled `Nsis7z` plugin? If not, should the app `.7z` for the **NSIS target** default to `BCJ` (or `-mf=off`) instead of `BCJ2`, given the install-time decoder can't read `BCJ2`?
2. Could the bundled `Nsis7z` decoder be updated to a 7-Zip that supports the `BCJ2` variant emitted by 7za 24.09?
3. At minimum, this fails **silently** (exit 0, no warning) and is easily misattributed to #8399's 2 GB limit. A post-pack verification that every entry actually extracts (or a warning when `BCJ2` is used with the NSIS target) would prevent shipping broken installers.

Happy to provide the full `7z l -slt` dump or a minimal repro.

mmaietta:
Fixes #9983.

### Summary

- The NSIS app payload (`app-*.7z`) is now pinned to a filter the install-time extractor can decode, so every entry — the main `.exe`, all `.dll`s, native `.node`, and bundled executables — is laid down on disk. Previously the modern bundled `7za` (24.09) auto-applied an architecture-specific branch/exec converter to executable content (`BCJ2` on x64/x86, `ARM64` on arm64) that the self-vendored `Nsis7z` install-time extractor does not understand, so those entries were skipped and the installed app was missing its binaries. Data files (`app.asar`, `*.pak`, `icudtl.dat`) were unaffected, which made the symptom arch- and content-specific.
- The chosen filter is the single-stream `BCJ` filter when compressing (the best-ratio filter the vendored extractor can read) and plain `Copy` when storing. `BCJ` keeps the archive within ~0.5–1.4% of `7za`'s default (undecodable) size in mixed payloads — far better than disabling filters entirely — while remaining fully decodable at install time.
- Applies to both the regular `nsis` target and `nsis-web`, to differential-aware builds, and to the separately-appended `elevate.exe`, since they all flow through the same packing path.
- Because the vendored extractor is fixed and known, the NSIS payload's filter is chosen by electron-builder and is intentionally not overridable via `ELECTRON_BUILDER_7Z_FILTER` — that override can never select a filter the bundled extractor can't read, so it can't reintroduce a broken installer. The override still applies to non-NSIS `7z` archives (e.g. the standalone `7z` distributable), which keep `7za`'s default behavior.

### DIFF
--- a/.changeset/nsis-app-archive-decodable-filter.md
+++ b/.changeset/nsis-app-archive-decodable-filter.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix(nsis): pack the app archive with a filter the install-time extractor can decode so the main executable and native binaries are reliably installed on x64 and arm64 (#9983)
--- a/packages/app-builder-lib/src/targets/archive.ts
+++ b/packages/app-builder-lib/src/targets/archive.ts
@@ -10,7 +10,10 @@ import { getPath7za } from "../toolsets/7zip.js"
 import _fsExtra from "fs-extra"
 const { move } = _fsExtra
 
-const ALLOWED_7Z_FILTERS = new Set(["BCJ", "BCJ2", "ARM", "ARMT", "IA64", "PPC", "SPARC", "DELTA"])
+// Values accepted by ELECTRON_BUILDER_7Z_FILTER (passed through as `-mf=<value>`). "OFF" disables the
+// branch/exec filter entirely (plain LZMA2); the rest are 7-Zip branch converters. 7za matches these
+// case-insensitively, so the value is canonicalized to uppercase before use.
+const ALLOWED_7Z_FILTERS = new Set(["OFF", "BCJ", "BCJ2", "ARM", "ARMT", "IA64", "PPC", "SPARC", "DELTA"])
 
 function validateCompressionLevel(level: string): void {
   if (!/^[0-9]$/.test(level)) {
@@ -97,6 +100,18 @@ export interface ArchiveOptions {
    * @default false
    */
   preserveSymlinks?: boolean
+
+  /**
+   * Restrict the 7z filter to one the install-time extractor (the self-vendored Nsis7z plugin) can
+   * decode. Modern 7za (24.09) auto-applies a CPU branch converter to executable content at
+   * `-mx>=1` — `BCJ2` on x86/x64 and `ARM64` on arm64 — which that decoder silently skips, dropping
+   * every executable from the install. When set, the archive is pinned to the single-stream `BCJ`
+   * filter while compressing (decodable, and the best-ratio filter that decoder supports) and to
+   * plain `Copy` while storing. Deliberately NOT overridable by `ELECTRON_BUILDER_7Z_FILTER`, which
+   * could otherwise reintroduce an unreadable archive. Only affects the 7z format. See #9983.
+   * @default false
+   */
+  installTimeDecodable?: boolean
 }
 
 /**
@@ -107,6 +122,22 @@ export interface ArchiveOptions {
  */
 export function shouldPreserveSymlinks(platform: Platform): boolean {
   return platform !== Platform.WINDOWS
+}
+
+/**
+ * Builds the exclude switches for the given masks, rejecting any pattern that contains a path
+ * traversal sequence. `prefix` is the tool-specific flag (`-xr!` for 7za, `-x` for native zip).
+ */
+function buildExcludeArgs(excluded: Array<string> | null | undefined, prefix: string): Array<string> {
+  if (excluded == null) {
+    return []
+  }
+  return excluded.map(mask => {
+    if (mask.includes("..")) {
+      throw new Error(`Excluded archive pattern contains path traversal sequence: "${mask}"`)
+    }
+    return `${prefix}${mask}`
+  })
 }
 
 export function compute7zCompressArgs(format: string, options: ArchiveOptions = {}) {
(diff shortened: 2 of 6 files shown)


julianpoemp:
thank you very much for the fix! It solves an issue I had for some days! 🎉

PR #9991, opened 2026-06-27
fix: strip credential env vars from child-process environment on Windows

mmaietta:
- `getProcessEnv` previously returned the Windows environment unchanged, so every spawned tool (signtool, makeappx, package managers) inherited credential vars such as `CSC_KEY_PASSWORD`, `WIN_CSC_KEY_PASSWORD`, and publish tokens. It now applies the same credential deny-list (`stripSensitiveEnvVars`) on Windows that the other platforms already use, for both the explicit-env and inherited-env paths. Required Windows system vars (`PATH`, `SYSTEMROOT`, `TEMP`, etc.) are preserved, and signing tools already receive the certificate password as an explicit CLI argument rather than via the environment, so signing is unaffected.

### DIFF
--- a/packages/builder-util/src/util.ts
+++ b/packages/builder-util/src/util.ts
@@ -104,10 +104,14 @@ export function filterSensitiveEnv(env: Record<string, string | undefined>): Rec
 }
 
 function getProcessEnv(env: Record<string, string | undefined> | Nullish): NodeJS.ProcessEnv | undefined {
-  // Windows: passing a filtered env to execFile drops critical system vars (PATH, SYSTEMROOT, TEMP)
-  // that many tools require. Credential stripping is therefore not applied on Windows.
+  // Windows does not need the LC_* locale normalisation below, but child processes (signtool, makeappx,
+  // package managers) MUST still have credential env vars (CSC_KEY_PASSWORD, tokens, etc.) stripped before they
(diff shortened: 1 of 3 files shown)


packages/builder-util/src/util.ts
```diff
@@ -104,10 +104,14 @@ export function filterSensitiveEnv(env: Record<string, string | undefined>): Rec
...
-  // that many tools require. Credential stripping is therefore not applied on Windows.
+  // Windows does not need the LC_* locale normalisation below, but child processes (signtool, makeappx,
+  // package managers) MUST still have credential env vars (CSC_KEY_PASSWORD, tokens, etc.) stripped before they
+  // inherit them. Signing tools receive the certificate password as an explicit CLI argument (e.g. signtool
+  // `/p <password>`), never via the environment, so stripping these keys does not break signing.
+  // stripSensitiveEnvVars only removes credential-named keys and preserves required Windows system vars
+  // (PATH, SYSTEMROOT, TEMP, etc.).
   if (process.platform === "win32") {
-    return env == null ? undefined : env
+    return stripSensitiveEnvVars(env == null ? process.env : env)
```
erikian, line 114: question: on non-Windows platforms, `stripSensitiveEnvVars` is not called when a explicit `env` is present, is the different behavior on Windows expected or is the non-Windows code just incorrect?

PR #9998, opened 2026-07-03
fix(app-builder-lib): expand lower channels when channel name has a suffix

ylkhayat:
We publish a separate update feed per architecture using `channel: "${channel}-${arch}"`, so our channels resolve to names like `beta-arm64` and `latest-x64`. We need the arch in the channel name because that is the only way we found to keep the arm64 and x64 feeds apart. Without the suffix both architectures write to the same `latest.yml` / `beta.yml` and overwrite each other, so the per-arch suffix is what makes separate feeds possible in the first place. We also have `generateUpdatesFilesForAllChannels` turned on so that beta users get promoted to stable and alpha users get promoted to beta.

That promotion silently stopped working once the arch suffix was in play. `computeChannelNames` only expands the lower channels when the channel is exactly `alpha`, `beta`, or `latest`. Anything else falls through to the default branch and returns just the one channel, so `beta-arm64` never produced an `alpha-arm64.yml` and our arm64 alpha users were stuck.

This PR reads the base channel off the front of the name and reattaches the suffix to the expanded channels:

- `beta-arm64` produces `beta-arm64` and `alpha-arm64`
- `latest-arm64` produces `latest-arm64`, `alpha-arm64`, and `beta-arm64`
- `alpha-arm64` stays on its own, same as bare `alpha`
- bare `alpha` / `beta` / `latest` behave exactly as before

We have been running this as a patch against `app-builder-lib` in our desktop app for a while and it fixed the promotion problem for us.

Added tests to `updateInfoBuilderTest.ts` covering the suffixed beta, latest, and alpha cases. The full `updateInfoBuilderTest.ts` suite passes locally (21 tests).

### DIFF
--- a/.changeset/channel-arch-suffix-expansion.md
+++ b/.changeset/channel-arch-suffix-expansion.md
@@ -1,0 +1,7 @@
+---
+"app-builder-lib": patch
+---
+
+fix(app-builder-lib): expand lower channels when the channel name has a suffix
+
+With `generateUpdatesFilesForAllChannels` enabled, the extra `alpha`/`beta` update files were only written when the channel was exactly `alpha`, `beta`, or `latest`. A per-arch feed configured as `channel: "${channel}-${arch}"` resolves to something like `beta-arm64`, which hit the default branch, so `alpha-arm64.yml` was never written and arm64 pre-release users stopped getting promoted. The base channel is now read off the front of the name and the suffix is reattached to the expanded channels, so `beta-arm64` yields `beta-arm64` and `alpha-arm64`, and `latest-arm64` yields all three.
--- a/.changeset/channel-arch-suffix-expansion.md
+++ b/.changeset/channel-arch-suffix-expansion.md
@@ -1,7 +1,9 @@
 ---
-"app-builder-lib": patch
+"app-builder-lib": major
 ---
 
 fix(app-builder-lib): expand lower channels when the channel name has a suffix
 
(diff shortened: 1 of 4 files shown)


.changeset/channel-arch-suffix-expansion.md
```diff
@@ -0,0 +1,7 @@
+---
+"app-builder-lib": patch
```
mmaietta: I think this needs to be a `major` semver bump

With generateUpdatesFilesForAllChannels on and a non-github provider, any existing `beta-*`/`latest-*` channel silently goes from publishing 1 yml to 2–3 ([updateInfoBuilder.ts:48-60](https://github.com/electron-userland/electron-builder/blob/ed00f38/packages/app-builder-lib/src/publish/updateInfoBuilder.ts#L48-L60)), with no opt-out. A stable latest-x64 publish would now overwrite beta-x64.yml/alpha-x64.yml in the same bucket, changing what live beta users are offered.

Note: I'm okay with this change, but just want to iterate that it needs to be in `major` and in the Breaking Changes documentation https://github.com/electron-userland/electron-builder/blob/master/website/docs/migration/v27-breaking-changes.md

ylkhayat: Moved from `patch` to a `major` and documented in the v27 breaking changes markdown. Have a look at your convenience

mmaietta (approved):

PR #10075, opened 2026-08-10
fix: Ensure arch is passed to blockmap path macro expansion

### LINKED ISSUE #9582: ${arch} file macro isn't correctly passed to NSIS blockmap
* **Electron-Builder Version**: 26.7.0

* **Electron Updater (as-needed)**:

* **Node Version**: 20.20.0

* **Electron Version**: 40.3.0

* **Platform & Target**: Windows x64

* **Debug Logs**: No errors, just unexpected behavior described below

When publishing NSIS installer to S3, the installer (`${productName}-Setup-${arch}-${version}.${ext}`) has the correct name and is uploaded to the correct path (`releases/${os}/${arch}`), but the associated blockmap is uploaded to the path without the architecture in the file macro (`releases/${os}`). It is always architecture that is not included in the path, meaning if I change the path to `releases/${arch}/${os}`, it ends up still uploaded to `releases/${os}`. I noticed that the macroExpander specifically removes arch if it is null, so it probably isn't being passed correctly for the blockmap creation.

Below is a minimal example of my config for reference:
```
  win: {
    target: 'nsis',
    publish: [
      {
        provider: 's3',
        bucket: 'my-bucket',
        path: 'releases/${os}/${arch}',
      },
    ],
  },
  nsis: {
    artifactName: '${productName}-Setup-${arch}-${version}.${ext}',
  },
```

eliotschu:
Fixes #9582

### Summary of Changes
- Add new default null parameter to createBlockmap that passes arch so it is eventually passed to path macro expansion when publishing
- Update usages of createBlockmap to pass the relevant arch
- Update test snapshots to expect arch for blockmaps


eliotschu:
In my bit of checking it looks like a bunch of these failures are actually the new expected behavior? Is that a correct impression that we just need to start working through the tests and updating the relevant ones?

eliotschu:
@mmaietta I added arch to I think all of the blockmap snapshots that were failing. Let me know if there was a different way you wanted this done.

eliotschu:
Looks like I missed a couple of tests and some of the snapshot objects needed to be reordered so that the installer artifact and blockmap order is corresponding for each arch respectively.

It also seems that some credentials for mac code signing or notarization might be out of date.

Hopefully this is getting close on having the tests updated.

mmaietta:
@eliotschu let me see if I can push additional test updates to this PR and if able to, I can take the PR from there :)

eliotschu:
@mmaietta You should have access to make edits, feel free to go ahead and update the remaining tests!

mmaietta (approved):

PR #10079, opened 2026-08-11
fix: preserve Yarn Berry registry credentials

### LINKED ISSUE #9999: Stripping environment variables prevents app-deps from being installed
## Problem

After updating electron-builder to v26.15.6 in my project, I received the following error when running `electron-builder install-app-deps`:

```
  • electron-builder  version=26.15.6
  • loaded configuration  file=package.json ("build" field)
  • installing dependencies  pm=yarn-berry platform=linux arch=x64 projectDir=xxx appDir=xxx/release/app workspaceRoot=xxx
  ⨯ yarn process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
Exit code:
1
Output:
Usage Error: Environment variable not found (XXX_NPM_TOKEN) in xxx/.yarnrc.yml (in xxx/.yarnrc.yml)

```

I backed up versions to see where this started breaking.  It works with version 26.12.0 but fails in v26.12.1.  It possibly started breaking with commit 4866737e4fe1f0cca210d318b9ce440f8e978a63.

The problem is we use a private npm registry in our project and have the `npmAuthToken` for this registry set using an environment variable in our `.yarnrc.yml` config.  That environment variable is being stripped out of the environment by `stripSensitiveEnvVars` because it includes `TOKEN` in its name.  We can avoid the error by changing the name of this environment variable but it doesn't seem like we should have to have to do that.  This stripping of environment variables seems overly aggressive, breaking normal functionality.

## Environment

* electron-builder: 26.12.1
* electron: 39.8.10
* OS: ubuntu 22.04.5 LTS amd64
* yarn: 4.17.0

atirna:
Fixes #9999

Yarn Berry resolves private-registry credentials from `.yarnrc.yml`, but electron-builder’s sensitive environment filter removed custom credential variables before Yarn ran.

This keeps the existing deny-list and restores only variables explicitly referenced by `npmAuthToken` or `npmAuthIdent` in the project’s Yarn config. Other sensitive variables remain filtered.

Tests:
- `TEST_FILES=yarnRegistryEnvTest pnpm ci:test`
- `pnpm compile`
- `pnpm ci:validate`


test/src/yarnRegistryEnvTest.ts
```diff
@@ -0,0 +1,73 @@
...
+    await rm(projectDir, { recursive: true, force: true })
+  }
+})
+
+test("passes only Yarn npmAuthToken variables to the Yarn Berry install child", async ({ expect }) => {
+  const projectDir = await mkdtemp(path.join(tmpdir(), "electron-builder-yarn-registry-env-"))
+  try {
+    await writeFile(path.join(projectDir, ".yarnrc.yml"), ['npmAuthToken: "${XXX_NPM_TOKEN}"', 'npmRegistryServer: "${GITHUB_TOKEN}"'].join("\n"))
+    process.env.XXX_NPM_TOKEN = "registry-token"
+    process.env.GITHUB_TOKEN = "publish-token"
```
mmaietta: Pretty sure that setting `process.env` variables will contaminate other tests' environment running. 
I think we need to use `stubEnv` here like:
```vi.stubEnv('VITE_ENV', 'staging')```

File: test/src/yarnRegistryEnvTest.ts [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+        GITHUB_TOKEN: "publish-token",
+      })
+    ).toEqual({ ROOT_NPM_TOKEN: "root-token", SCOPED_NPM_TOKEN: "scoped-token" })
+  } finally {
+    await rm(projectDir, { recursive: true, force: true })
+  }
+})
+
+test("passes only Yarn npmAuthToken variables to the Yarn Berry install child", async ({ expect }) => {
+  const projectDir = await mkdtemp(path.join(tmpdir(), "electron-builder-yarn-registry-env-"))
```
mmaietta: To avoid using `mkdtemp`, the tests already have a test-context-unique creator. Please add to `async ({ expect, tmpDir }) =>` and use `tmpDir` directly. You won't need to do a `try-finally` w/ `rm` operation at the end - it's automatically cleaned up within the text context

Same comment apply to the other tests temporary fixture file/directory creations.

Accessed via
`tmpDir.createTempDir{{ prefix: "electron-builder-yarn-registry-env-" })`

test/src/yarnRegistryEnvTest.ts
atirna: switched that test to `vi.stubEnv` + `vi.unstubAllEnvs()` in afterEach so it doesnt leak into other tests

test/src/yarnRegistryEnvTest.ts
atirna: yep, all 3 tests now use `tmpDir.createTempDir` and i dropped the mkdtemp/rm cleanup

mmaietta:
@atirna could you please pull in some changes/improvements into your branch that I've/Claude suggested in review.  Feel free to review the suggested changes before pushing to your branch
```
git fetch https://github.com/electron-userland/electron-builder pr-10079-additions
git push git@github.com:atirna/electron-builder.git FETCH_HEAD:agent/yarn-berry-registry-token-env
```

atirna:
found a bug in the regex while re-reading this, have pushed a fix.

the `${VAR}` matcher from memory instead of checking yarn. read `miscUtils.replaceEnvVariables` in yarnpkg-core:

- `${PRIMARY:-${FALLBACK}}` only gave me `PRIMARY`, so a nested default still fails to install
- `$BARE`, `\${ESCAPED}` and `${_LEADING}` all got preserved, and yarn treats none of those as refs. so i was putting vars into the child env that yarn never reads, which is the thing `stripSensitiveEnvVars` stops

now using yarns grammar for the name extraction. added a test for those cases, it fails on the previous commit (preserves `BARE_TOKEN` and `ESCAPED_TOKEN`, misses `FALLBACK_TOKEN`) and passes now. `TEST_FILES=yarnRegistryEnv pnpm ci:test` 4/4, lint and prettier clean.

packages/app-builder-lib/src/util/installOrRebuild.ts
```diff
@@ -88,6 +89,56 @@ export function getGypEnv(frameworkInfo: DesktopFrameworkInfo, platform: NodeJS.
...
+
+function getYarnNpmAuthTokenValues(config: unknown): Array<string> {
+  if (!isRecord(config)) {
+    return []
+  }
+
+  const values = [config.npmAuthToken]
+  for (const scopes of [config.npmScopes, config.npmRegistries]) {
+    if (isRecord(scopes)) {
+      values.push(...Object.values(scopes).map(scope => (isRecord(scope) ? scope.npmAuthToken : undefined)))
```
mmaietta: Should we change this to include yarn's credential alternative `npmAuthIdent`? e.g. `? [scope.npmAuthToken, scope.npmAuthIdent] : []`

atirna: yes right, good catch, added `npmAuthIdent` alongside `npmAuthToken` at the root, `npmScopes`, and `npmRegistries` levels. tests are good: 
`TEST_FILES=yarnRegistryEnv pnpm ci:test` is 4/4 and `pnpm ci:validate` passes.

kyrofa:
Thank you for this, @atirna!

mmaietta (approved):

PR #10082, opened 2026-08-12
fix: Only logs that we've signed a file if we actually did

DiefBell:
Currently `signWindows` shows "signing with..." before signing actually happens, so if it fails then the user never knows.

Uses the existing return value of `signWithRetry" to check whether it signed successfully before logging anything, and also alerts the user if the signing failed. Also logs before signing to let the user know we are in fact doing something

### DIFF
--- a/packages/app-builder-lib/package.json
+++ b/packages/app-builder-lib/package.json
@@ -1,7 +1,7 @@
 {
   "name": "app-builder-lib",
   "description": "electron-builder lib",
-  "version": "27.0.0-alpha.6",
+  "version": "27.0.1-alpha.1",
   "type": "module",
   "main": "./dist/index.js",
   "exports": {
--- a/packages/app-builder-lib/package.json
+++ b/packages/app-builder-lib/package.json
@@ -1,7 +1,7 @@
 {
   "name": "app-builder-lib",
   "description": "electron-builder lib",
-  "version": "27.0.1-alpha.1",
+  "version": "27.0.0-alpha.6",
   "type": "module",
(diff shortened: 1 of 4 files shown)


packages/app-builder-lib/src/codeSign/win/windowsCodeSign.ts
```diff
@@ -9,17 +9,25 @@ export interface WindowsSignOptions {
 
 export async function signWindows(options: WindowsSignOptions, packager: WinPackager): Promise<boolean> {
   const signing = resolveWindowsSigningConfiguration(options.options)
-  if (signing?.type === "azure") {
-    log.info({ path: log.filePath(options.path) }, "signing with Azure Trusted Signing")
+  const packageManager = await packager.signingManager.value
+
+  const path = log.filePath(options.path)
+  console.info(`Signing ${path}...`)
```
mmaietta: Please update to `log.info`

packages/app-builder-lib/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "app-builder-lib",
   "description": "electron-builder lib",
-  "version": "27.0.0-alpha.6",
+  "version": "27.0.1-alpha.1",
```
mmaietta: Please revert this change and the CHANGELOG.md changes. You can utilize `pnpm generate:changeset` to generate a changeset that the CI/CD will automatically pick up for the next release.

mmaietta (changes requested):

mmaietta (approved):

Issue #7161, opened 2022-09-27
Cannot download differentially, fallback to full download: Error: Content-Type "multipart/byteranges" is expected, but got "application/x-msdownload"


xzy-96:
"electron-builder": "^23.1.0",

* **Electron-Builder Version**: 

* **Node Version**: 
16.16.0

 "electron": "^19.0.6",

* **Electron Version**:
* **Electron Type (current, beta, nightly)**:

"electron-updater": "^5.2.1",

* **Target**: 
{
  "name": "electron-vite-vue",
  "private": true,
  "version": "1.0.6",
  "description": "electron",
  "author": "",
  "keywords": [
    "electron",
    "vue",
    "vite",
    "pinia",
    "antdv"
  ],
  "main": "electron/main.js",
  "repository": "",
  "scripts": {
    "electron:serve": "concurrently -k \"pnpm run dev\" \"pnpm run electron\"",
    "electron:build": "yarn run build && electron-builder",
    "electron:build-test": "yarn run build:test && electron-builder",
    "pack": "electron-builder",
    "dev": "vite --mode serve-dev --host",
    "build": "vite build --mode build",
    "build:test": "vite build --mode  build-test",
    "preview": "vite preview",
    "build-icon": "electron-icon-builder --input=./public/icon.png --output=dist_electron --flatten",
    "electron": "wait-on tcp:3008 && cross-env NODE_ENV=development electron .",
    "lint": "eslint . --ext .vue,.js,.ts,.jsx,.tsx --fix",
    "format": "prettier --write \"./**/*.{html,vue,ts,js,json,md}\"",
    "lint:style": "stylelint \"./**/*.{css,less,vue,html}\" --fix",
    "prepare": "husky install"
  },
  "build": {
    "productName": "*",
    "copyright": "",
    "mac": {
      "category": "public.app-category.utilities"
    },
    "appId": "iden.client.xldf",
    "asar": true,
    "win": {
      "target": [
        "nsis"
      ],
      "icon": "dist_electron/icons/icon.ico",
      "asarUnpack": [
        "src/**",
        "README.md"
      ]
    },
    "publish": [
      {
        "provider": "generic",
        "url": "https://aliyuncs.com"
      }
    ],
    "nsis": {
      "oneClick": false,
      "allowElevation": true,
      "language": "2052",
      "allowToChangeInstallationDirectory": true
    },
    "files": [
      "dist/**/*",
      "electron/**/*"
    ],
    "directories": {
[...]

Issue #3742, opened 2019-03-12
Electron Updater doesn't like server response headers for multipart byterange


obsius:
* **Version**: 20.39.0
* **Updater Version**: 4.0.6
* **Target**: NSIS

Doing a delta request to Digital Ocean Spaces.  Updater is configured properly and works with the delta updates off.

I was getting this error:
`Content-Type "multipart/byteranges" is expected, but got "application/octet-stream"`

so I changed my server header to match and then got this error:

`Content-Type "multipart/byteranges" is expected, but got "multipart/byteranges"`

which makes no sense.

So I see that `multipleRangeDownloader.ts` has this block of code:
```ts
    const contentType = safeGetHeader(response, "content-type")
    const m = /^multipart\/.+?(?:; boundary=(?:(?:"(.+)")|(?:([^\s]+))))$/i.exec(contentType)
    if (m == null) {
      reject(new Error(`Content-Type "multipart/byteranges" is expected, but got "${contentType}"`))
      return
    }
```

The server is returning a `Content-Range`, and I think that's to spec, so I don't know why this code is only checking `Content-Type` when `Content-Range` is a valid response.

I think this server response should be acceptable, but Electron Updater doesn't agree:
```
HTTP/1.1 206 Partial Content
Content-Length: 57891
Content-Range: bytes 154514-212404/50262321
Accept-Ranges: bytes
Last-Modified: Tue, 12 Mar 2019 11:14:40 GMT
Content-Type: application/octet-stream
Date: Tue, 12 Mar 2019 11:35:35 GMT
```

obsius:
Closing, I see now that there's an option to send out single requests: useMultipleRangeRequest = false

SinusPi:
Still, the multiple-range downloader shouldn't fall back to a full download if multiple ranges aren't supported, or a single range was requested and served.

PR #10085, opened 2026-08-14
NSIS: skip unnecessary elevation on silent per-machine updates when already writable

imlucas:
## What

For a `perMachine: false` NSIS installer redirected outside its default per-user location (e.g. to `C:\Program Files` via `allowToChangeInstallationDirectory`), **every subsequent silent update apply unconditionally re-elevates via `UAC_RunElevated`** — driven purely by the persisted `hasPerMachineInstallation` registry flag (`installer.nsi`'s `Section "install"` [before this patch](https://github.com/electron-userland/electron-builder/blob/c0b8235d7f86d90ffe7218765115b6948b180739/packages/app-builder-lib/templates/nsis/installer.nsi#L94-L124), set in `initMultiUser` from [`HKLM\Software\<APP_GUID>\InstallLocation`](https://github.com/electron-userland/electron-builder/blob/c0b8235d7f86d90ffe7218765115b6948b180739/packages/app-builder-lib/templates/nsis/assistedInstaller.nsh#L107-L120)), never by a live check of whether elevation is actually still needed.

This PR adds two capability probes (`IsDirWritable`, `IsRegKeyWritable`, in `common.nsh`) and only elevates when at least one actually fails a live check, instead of trusting the persisted flag alone. This is useful for an app whose `customInstall` hook has deliberately loosened its own install directory's (and registry key's) ACL after the one unavoidable admin-elevated initial install, so a standard user can write there on later updates without a UAC prompt every time.

## How Chrome and Firefox actually avoid this (corrected from an earlier draft of this PR)

An earlier version of this description and the in-code comments called this "the Chrome/Firefox pattern." That was an inaccurate simplification and I want to correct it rather than leave a wrong citation standing:

- **Firefox** ships a Windows service, the **Mozilla Maintenance Service**, installed once with elevation. Its *service* ACL (not the install directory's file ACL) is modified so a non-elevated process can start/stop it (`SERVICE_START` granted to authenticated users); the service itself runs with the rights needed to update files in `Program Files`, and performs the write on the user's behalf. Sources: [Mozilla Support — "What is the Mozilla Maintenance Service?"](https://support.mozilla.org/en-US/kb/what-mozilla-maintenance-service), [MozillaWiki — Windows Service Silent Update](https://wiki.mozilla.org/Windows_Service_Silent_Update#Service_installation), [Bug 481815 — original service proposal](https://bugzilla.mozilla.org/show_bug.cgi?id=481815).
- **Chrome** uses **Google Update (Omaha)**. For a system-level install, the actual update task runs under the **LocalSystem** account, not the interactive user's token, so there's nothing for the interactive user to elevate. Sources: [google/omaha](https://github.com/google/omaha), [Omaha 3 walkthrough](https://github.com/google/omaha/blob/main/doc/Omaha3Walkthrough.md).

Both are a **standing privileged service** pattern — a separate, persistently-elevated process does the write — not a loosened file/registry ACL on the interactive user's own token, which is what this PR's checks actually support. They're a heavier but more thorough answer to the same problem (they also cover cases this PR doesn't, like a registry key the app never loosens). I've corrected the in-code comment accordingly; this PR is a narrower, service-free mechanism for apps that would rather not ship one.

## Why both checks are necessary (found the hard way)

I initially wrote this with *only* a directory-writability check. That looked like a complete fix — it correctly skipped elevation, no UAC prompt, apply succeeded. But it silently corrupts installer state: `installSection.nsh` also writes the per-machine registration ([`registryAddInstallInfo`](https://github.com/electron-userland/electron-builder/blob/c0b8235d7f86d90ffe7218765115b6948b180739/packages/app-builder-lib/templates/nsis/include/installer.nsh#L103-L105)) to `HKLM` via plain `WriteRegStr`, which **fails silently** (no error surfaced, no abort) when the process isn't elevated. Skipping elevation on the directory check alone lets the file copy "succeed" while quietly wiping the `InstallLocation` value that the *next* launch's `initMultiUser` depends on — I reproduced this directly: `Get-ItemProperty HKLM:\Software\<APP_GUID>` went from a correct path to empty after one unelevated apply, and the following update defaulted to a brand-new per-user install at the wrong path instead of finding the existing Program Files install.

Adding the registry-writability probe alongside the directory one fixes this: elevation is only skipped when *both* checks pass, so there's no path where this can leave the registration in a half-written state.

## Verification

Tested by hand on a real Windows 11 admin machine, building this branch locally and installing/updating a minimal Electron fixture app through several version hops. Full writeup and a scripted, from-scratch reproduction: [imlucas/nsis-update-elevation](https://github.com/imlucas/nsis-update-elevation).

- **Baseline (upstream, unpatched):** every silent apply elevates. Confirmed with a temporary debug build logging `hasPerMachineInstallation` / `UAC_IsAdmin` at the `Section "install"` entry point.
- **Directory-writable-only check (not what this PR ships):** elevation skipped, apply succeeds, but `HKLM\Software\<APP_GUID>\InstallLocation` is wiped on the next `initMultiUser` read — confirmed via direct registry inspection before/after.
- **This patch (directory + registry check):** with only the directory ACL loosened (not the registry key) — the realistic case for any app that hasn't *also* loosened its registry key ACL — the registry probe correctly fails and the patch falls back to elevating exactly like upstream. No regression, no corruption.
- With *both* ACLs loosened by a `customInstall` hook, the patch skips elevation and the apply completes cleanly with no corruption, by construction of the check (not separately re-verified live in this pass, since the registry-ACL-loosening half wasn't exercised in the same run — flagging this explicitly rather than claiming more than was directly observed).

## Related prior art on this repo

The differential-download server requirement this same investigation ran into (`Content-Type "multipart/byteranges" is expected, but got ...`, per [`multipleRangeDownloader.ts`](https://github.com/electron-userland/electron-builder/blob/c0b8235d7f86d90ffe7218765115b6948b180739/packages/electron-updater/src/differentialDownloader/multipleRangeDownloader.ts#L106) and specified in [RFC 7233 §4.1](https://www.rfc-editor.org/rfc/rfc7233.html)) is a well-known, previously-reported gap, unrelated to this PR but hit in the same test pass: [#5758](https://github.com/electron-userland/electron-builder/issues/5758), [#3742](https://github.com/electron-userland/electron-builder/issues/3742), [#7161](https://github.com/electron-userland/electron-builder/issues/7161).

## Scope / non-goals

- **No behavior change for the common case.** Every existing app's registry key ACL is untouched by default, so `IsRegKeyWritable` will keep failing and this falls back to the current elevate-every-time behavior identically. This is purely additive for apps that opt in via their own `customInstall` hook.
- This does **not** loosen any ACLs itself — that remains the app author's own `customInstall` responsibility (as electron-builder already supports via `nsis.include`), most plausibly using the [NSIS `UAC` plugin](https://github.com/tpn/nsis-uac) already vendored here, or Windows' own `icacls`. This PR only makes the *elevation decision* live instead of registry-flag-only.
- I'd love feedback on whether `common.nsh` is the right home for these two macros, and whether the interactive (non-silent, wizard) elevation path deserves the same treatment — this PR only touches the silent-apply branch in `installer.nsi`, which is what electron-updater's `quitAndInstall` actually exercises for unattended updates.


imlucas:
## Self-review: gaps in this PR's elevation-skip logic

I went back over this after rebasing onto current `master`, adversarially — trying to find the ways the two writability probes could be wrong or incomplete rather than just confirming they work for the case I tested by hand. Two things came out of it: one I've already fixed, and one that's still open and I think needs your input before this lands.

### 1. Fixed — the probes didn't cover everything `installSection.nsh` writes

The original version only checked `$INSTDIR` and `INSTALL_REGISTRY_KEY`. But `registryAddInstallInfo` also writes the Programs-and-Features metadata (`DisplayVersion`, `UninstallString`, `DisplayIcon`, `Publisher`, `EstimatedSize`, ...) to a *different* key, `UNINSTALL_REGISTRY_KEY` (`Software\Microsoft\Windows\CurrentVersion\Uninstall\<guid>`) — and `installSection.nsh` also (re)creates the Start Menu/desktop shortcuts under the same all-users `SHELL_CONTEXT` via `CreateShortCut`. Both fail silently when unelevated, same as the `InstallLocation` corruption this PR already documented — just on keys/paths the original two probes didn't check. A `customInstall` hook that loosens only `$INSTDIR` and its own `INSTALL_REGISTRY_KEY` (exactly the scenario this PR describes) would pass both original probes, skip elevation, and then silently leave the Programs-and-Features entry stale and/or shortcuts unrefreshed.

Fixed by extending the same live-check pattern to `UNINSTALL_REGISTRY_KEY` and (guarded by the existing `DO_NOT_CREATE_START_MENU_SHORTCUT`/`DO_NOT_CREATE_DESKTOP_SHORTCUT` `!ifdef`s) `$SMPROGRAMS`/`$DESKTOP`:
https://github.com/imlucas/electron-builder/blob/b6ead16967fcbd1368f39d6f05d04a2ddfef093f/packages/app-builder-lib/templates/nsis/installer.nsi#L94-L156

What the original two probes covered vs. what `installSection.nsh` actually writes on every run:

| Location | Checked by the probes? | Actually written by `installSection.nsh`? |
|---|---|---|
| `$INSTDIR` | ✅ | ✅ file copy |
| `INSTALL_REGISTRY_KEY` | ✅ | ✅ `InstallLocation`, `KeepShortcuts`, `ShortcutName`, ... |
| `UNINSTALL_REGISTRY_KEY` | ❌ | ✅ `DisplayVersion`, `UninstallString`, `DisplayIcon`, ... |
| `$SMPROGRAMS` (Start Menu) | ❌ | ✅ `CreateShortCut` |
| `$DESKTOP` | ❌ | ✅ `CreateShortCut` |

The three ❌ rows are where elevation could get skipped while a write further down still silently fails.

Verified with a real `makensis` build through the repo's own test suite (not just a syntax read) — `assistedInstaller__wcs-1.2.1__nsis-1.2.1__wine-0.0.0__Test`, including the `allowToChangeInstallationDirectory`, `only perMachine`, and `only perMachine and elevated` cases — 7/8 pass (1 unrelated skip; the one transient failure on first run was a McAfee hit on a freshly-built `requireAdministrator` .exe, confirmed by rerun).

Also added the changeset this needs (`app-builder-lib`, patch) — the changeset-bot flag on the PR should clear.

### 2. Open — `uninstallOldVersion` re-elevates independently, defeating the point of this PR for the exact users it targets

This one I want to flag rather than quietly patch, because fixing it properly is a design question, not a probe gap.

**The mechanism:** `installSection.nsh` calls `uninstallOldVersion SHELL_CONTEXT` unconditionally, with no elevation guard before it:
https://github.com/imlucas/electron-builder/blob/b6ead16967fcbd1368f39d6f05d04a2ddfef093f/packages/app-builder-lib/templates/nsis/installSection.nsh#L52

When a per-machine install is already registered (exactly the `hasPerMachineInstallation == "1"` condition this PR's own check is gated on), that function finds a non-empty `UninstallString` and `ExecWait`s the **previous version's own uninstaller.exe**, passing `/allusers`:
https://github.com/imlucas/electron-builder/blob/b6ead16967fcbd1368f39d6f05d04a2ddfef093f/packages/app-builder-lib/templates/nsis/include/installUtil.nsh#L224

That old binary's `un.onInit` → `initMultiUser` → `setInstallModePerAllUsers` (compiled with `BUILD_UNINSTALLER`) unconditionally elevates if it isn't already admin — no ACL check, no `${Silent}` check, and it's baked into every already-shipped uninstaller.exe, so nothing in this PR (or any future one) can change what that binary does:
https://github.com/imlucas/electron-builder/blob/b6ead16967fcbd1368f39d6f05d04a2ddfef093f/packages/app-builder-lib/templates/nsis/multiUser.nsh#L66-L72

**Why the current (pre-this-PR) code elevates unconditionally at that point in the first place:** it isn't incidental — it's the deliberate fix from #6450 ("per-machine installs must properly elevate during silent install/updates"), which closed #5468 and #6425. #5468's reported symptom is almost exactly this mechanism, worst case:

> 1. user closes the app, gets a UAC popup for `old-uninstaller.exe`
> 2. gives permission
> 3. **the old version is deleted**
> 4. **nothing else happens** — new version never installs

#2363 has an independent report diagnosing the same root cause in almost the same terms: elevation is requested for the `ExecWait old-uninstaller.exe` step, but the subsequent file extraction "will just not extract anything into it without any complaints" when unelevated.

**Why this PR reopens that gap specifically for the users it's meant to help:** #6450's fix works by getting the *outer* installer process elevated *before* it reaches `uninstallOldVersion`, so the nested `ExecWait` inherits an already-elevated token and the old uninstaller's own `UAC_IsAdmin` check passes for free — no second prompt, no failure. This PR's whole purpose is to let the outer process skip that elevation. Once it does, the nested old-uninstaller.exe spawns unelevated, detects that itself, and fires its *own* independent `UAC_RunElevated` — a second-order elevation request the outer (still-unelevated) process doesn't see or control.

Same starting condition in both diagrams below: `hasPerMachineInstallation == "1"`, silent update, process not already admin.

**BEFORE this PR — #6450's fix, this is why it works today:**

```mermaid
flowchart TD
    A["new-installer.exe<br/><b>unelevated</b>"] --> B{"Section install:<br/>UAC_IsAdmin?"}
    B -- "NO" --> C["UAC_RunElevated<br/>➜ <b>UAC PROMPT #1</b><br/>admin consents"]
    C --> D["Quit<br/>(relaunches itself, elevated)"]
    D --> E["new-installer.exe<br/><b>ELEVATED</b> relaunch"]
    E --> F{"Section install:<br/>UAC_IsAdmin?"}
    F -- "YES" --> G["continue, no prompt"]
    G --> H["installSection.nsh — elevated"]
    H --> I["uninstallOldVersion SHELL_CONTEXT"]
    I --> J["ExecWait old-uninstaller.exe<br/>child inherits the ELEVATED token"]
    J --> K{"old-uninstaller.exe onInit:<br/>UAC_IsAdmin?"}
    K -- "YES (inherited)" --> L["deletes old files + registry<br/>exits 0"]
    L --> M["installApplicationFiles /<br/>registryAddInstallInfo / shortcuts<br/>all succeed — whole tree is elevated"]
    M --> N["✅ RESULT: exactly ONE prompt, total"]
```

**THIS PR — probes pass, elevation is skipped, and the problem surfaces one level down:**

```mermaid
flowchart TD
    A2["new-installer.exe<br/><b>unelevated</b>"] --> B2{"IsDirWritable / IsRegKeyWritable x2 /<br/>shortcut-dir checks"}
    B2 -- "ALL PASS (ACLs loosened)" --> C2["skip UAC_RunElevated entirely<br/>process stays UNELEVATED"]
    C2 --> D2["installSection.nsh — still unelevated"]
    D2 --> E2["uninstallOldVersion SHELL_CONTEXT"]
    E2 --> F2["ExecWait old-uninstaller.exe<br/>child inherits the UNELEVATED token"]
    F2 --> G2{"old-uninstaller.exe onInit:<br/>UAC_IsAdmin?"}
    G2 -- "NO" --> H2["UAC_RunElevated<br/>➜ <b>UAC PROMPT</b> — unexpected,<br/>second-order, fired from inside a<br/>subprocess the outer script never sees"]
    H2 --> I2{"who's running it?"}
    I2 -- "admin account" --> J2["consents, update limps through"]
    I2 -- "real standard user" --> K2["no admin creds available<br/>➜ HANG or fails outright"]
    J2 --> L2["Quit<br/>(old-uninstaller relaunches itself, elevated)"]
    L2 --> M2["elevated copy deletes old<br/>files + registry, exits 0"]
    M2 --> N2["ExecWait returns 0 to the outer<br/>process — <b>STILL UNELEVATED</b>"]
    N2 --> O2["installApplicationFiles /<br/>registryAddInstallInfo / shortcuts<br/>succeed only because probes<br/>verified writability up front"]
    O2 --> P2["⚠️ RESULT (admin): extra, unexplained<br/>prompt — easy to miss or click through"]
    K2 --> Q2["❌ RESULT (standard user): #5468 again<br/>('deletes old version, nothing else<br/>happens'), or a silent hang"]
```

I also checked the electron-updater side to see whether anything there catches this: `NsisUpdater.ts` only pre-elevates the whole installer launch via a separate `elevate.exe` helper when the build-time `isAdminRightsRequired` flag is set (i.e. `perMachine: true`). This PR's own scenario is `perMachine: false` + `allowToChangeInstallationDirectory`, so that flag is false — electron-updater launches the new installer directly, unelevated, and only retries via `elevate.exe` on a bare `EACCES` spawn error. A hang or non-`EACCES` failure inside the nested old-uninstaller UAC step isn't caught by that retry path.

Concretely: my own by-hand verification for this PR was done on an admin account, where the nested prompt (if it appeared at all) resolves with a quick consent click and might not even be noticed as separate from the intended flow. For a genuine non-admin standard user — the audience the PR description explicitly invites ("a standard user can write there... without a UAC prompt") — that nested prompt asks for **admin credentials they don't have**, in what's supposed to be a silent, unattended update. That's not an extra click; it's a hang or hard failure, i.e. #5468 again, or worse.

**Options, none of them a quick patch:**
1. Apply the same live-writability treatment to `multiUser.nsh`'s `BUILD_UNINSTALLER` branch. This only helps *future* update chains where both the old and new binaries carry the fix — it does nothing for the first transition away from any already-shipped uninstaller, which will always carry the unconditional elevate.
2. Have the outer installer explicitly and deliberately elevate just for the `uninstallOldVersion` sub-step, rather than relying on the old binary's own `onInit` to do it implicitly. More predictable, but doesn't reduce prompts for a real standard user — just moves the surprise.
3. Be explicit that this feature's benefit only materializes some number of versions after adoption (once both sides of an update carry the fix), and document that rather than presenting it as an immediate no-prompt guarantee.

I don't think this should block cleaning up and merging the two probes I already extended (registry key + shortcuts), but I do think it's worth resolving — or at minimum calling out in the PR description as a known limitation — before this ships, since right now the description reads as "no UAC prompt on update" for exactly the case where I now think that isn't reliably true.

**References:** #6450, #5468, #6425, #2363

imlucas:
## Update: applied option 1 from finding #2 above

Extended the same live-writability check to `multiUser.nsh`'s `setInstallModePerAllUsers`, gated to the silent case only:
https://github.com/imlucas/electron-builder/blob/7a479bb1a3c6945d3ae11521042490b7ccc372df/packages/app-builder-lib/templates/nsis/multiUser.nsh#L59-L132

For a **silent** uninstall (the only kind `uninstallOldVersion` drives), elevation is now skipped only when `$INSTDIR`, `INSTALL_REGISTRY_KEY`, `UNINSTALL_REGISTRY_KEY[_2]`, and the shortcut directories — everything the uninstall section (`uninstaller.nsh`'s `RMDir`/`DeleteRegKey` calls) actually deletes under `SHELL_CONTEXT` — are all already writable. Interactive (non-silent) uninstalls are untouched and keep the original unconditional elevate, since that's a visible, user-initiated action this PR was never trying to change.

To be clear about what this does and doesn't fix, worth restating from above: this only helps *future* update chains, once both the old and new uninstaller carry this change. It can't retroactively alter what an already-shipped `uninstaller.exe` does — nothing can. So the very first update after adopting this feature can still hit the nested-elevation issue if the previously-installed version predates this fix; it's from the *next* update onward (old binary also fixed) that the chain stays elevation-free end to end.

Verified with real `makensis` builds through two of the repo's own test suites — `assistedInstaller__wcs-1.2.1__nsis-1.2.1__wine-0.0.0__Test` (7/8 pass, 1 unrelated skip) and `oneClickInstallerTest` (7/7 pass, 8 unrelated skips) — covering `only perMachine`, `only perMachine and elevated`, `allowToChangeInstallationDirectory`, and standard one-click flows, all of which exercise both the installer and uninstaller compile passes of these shared templates.

Updated the changeset to describe both sides of the fix.

Issue #10093, opened 2026-08-18
Linux: the elevation dialog shows /bin/bash -c … instead of the install command


joevin-slq-docto:
* **Electron-Builder Version**: 26.15.3 — the code below is quoted from `master` (27.0.0-alpha.7), where it is unchanged
* **Electron Updater (as-needed)**: 6.8.9 (and 7.0.0-alpha.6)
* **Node Version**: the one bundled with Electron 42.5.0
* **Electron Version**: 42.5.0
* **Platform & Target**: Linux (Ubuntu 24.04, GNOME 46, Wayland), `deb` target — the same code path serves `rpm` and `pacman`
* **Debug Logs**: see the dialog text and the Node warning below

### What happens

The privileged command is wrapped in `/bin/bash -c '…'`
([LinuxUpdater.ts#L54](https://github.com/electron-userland/electron-builder/blob/0a0c683d81993a63d4c64dfa67e42fa6fc41267d/packages/electron-updater/src/LinuxUpdater.ts#L54)), so the authentication dialog shows the wrapper instead of the
command the user is being asked to authorize:

```
Authentication is needed to run '/bin/bash -c dpkg -i /home/…/pending/App-1.2.3.deb' as the super user
```

polkit elides the middle of long strings, so on a real updater cache path a good part of what remains legible
is the wrapper rather than the package being installed.

Two consequences beyond readability:

- the wrapper is what forces the manual metacharacter escaping in `installerPath`
  ([LinuxUpdater.ts#L20-L34](https://github.com/electron-userland/electron-builder/blob/0a0c683d81993a63d4c64dfa67e42fa6fc41267d/packages/electron-updater/src/LinuxUpdater.ts#L20-L34)), which is documented as not supporting paths that
  contain a single quote
- `spawnSyncLog` passes an args array together with `shell: true`
  ([BaseUpdater.ts#L299-L306](https://github.com/electron-userland/electron-builder/blob/0a0c683d81993a63d4c64dfa67e42fa6fc41267d/packages/electron-updater/src/BaseUpdater.ts#L299-L306)), so recent Node versions warn:

```
(node:538599) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead
to security vulnerabilities, as the arguments are not escaped, only concatenated.
```

### Suggested direction

`pkexec` and `sudo` accept an argv array, so the command can be spawned without a shell at all: the dialog
[...]

PR #10094, opened 2026-08-18
fix(updater): install Linux packages without freezing the app

### LINKED ISSUE #10092: Linux: the app freezes while the elevation dialog is open (spawnSync install)
* **Electron-Builder Version**: 26.15.3 — the code below is quoted from `master` (27.0.0-alpha.7), where it is unchanged
* **Electron Updater (as-needed)**: 6.8.9 (and 7.0.0-alpha.6)
* **Node Version**: the one bundled with Electron 42.5.0
* **Electron Version**: 42.5.0
* **Platform & Target**: Linux (Ubuntu 24.04, GNOME 46, Wayland), `deb` target — the same code path serves `rpm` and `pacman`
* **Debug Logs**: nothing is logged; the symptom is the desktop marking the app as not responding while the elevation dialog is open

### What happens

`LinuxUpdater.runCommandWithSudoIfNeeded` installs through `spawnSyncLog`
([LinuxUpdater.ts#L37-L54](https://github.com/electron-userland/electron-builder/blob/0a0c683d81993a63d4c64dfa67e42fa6fc41267d/packages/electron-updater/src/LinuxUpdater.ts#L37-L54)), so the Electron main process is blocked for as long
as the polkit dialog is open — that is, while the user is typing their password. The desktop notices the
unresponsive process and offers to force-quit the app:

> "App" is not responding — Force Quit / Wait

Timers cannot run either, so any app-side safety net around the install is frozen too.

### Steps to reproduce

1. Package an app with the `deb` target and install it (`sudo dpkg -i app.deb`).
2. Publish a newer version and let the app download it.
3. Call `quitAndInstall()` (or `installPendingUpdateIfAvailable()` on `master`).
4. When the polkit dialog appears, wait ~15 s before typing the password.

### Why it is not simply "make it async"

`doInstall` and `install` are documented as *must be sync (because quit even handler is not async)*
([BaseUpdater.ts#L80-L83](https://github.com/electron-userland/electron-builder/blob/0a0c683d81993a63d4c64dfa67e42fa6fc41267d/packages/electron-updater/src/BaseUpdater.ts#L80-L83)) — the constraint comes from installing inside the
`app.on("quit")` handler.

That constraint does not cover the paths that matter here. On `master`, targets whose install requires
elevation are already excluded from the automatic install at launch and go through an explicit
`installPendingUpdateIfAvailable()` ([BaseUpdater.ts#L52-L61](https://github.com/electron-userland/electron-builder/blob/0a0c683d81993a63d4c64dfa67e42fa6fc41267d/packages/electron-updater/src/BaseUpdater.ts#L52-L61)), and
`quitAndInstall()` is called by the application, not by a quit handler. Both can await an install. Only the
on-quit path cannot — and there the app is exiting anyway, so blocking is far less harmful.

I have a patch for this and will open a PR shortly.

joevin-slq-docto:
Fixes #10092.

## Problem

`LinuxUpdater.runCommandWithSudoIfNeeded` installs through `spawnSyncLog`, so the Electron main process is
blocked for as long as the elevation dialog is open — that is, while the user is typing their password. The
desktop notices the unresponsive process and offers to force-quit the app:

> "App" is not responding — Force Quit / Wait

Timers cannot run either, so any app-side safety net is frozen too.

## Change

`quitAndInstall()` and `installPendingUpdateIfAvailable()` install through a new asynchronous path:

- `BaseUpdater.installAsync()` mirrors `install()` and calls `doInstallAsync()`, which **defaults to the
  synchronous `doInstall`** — Windows and macOS keep their exact behaviour and timing, since `doInstall` is
  still evaluated synchronously before the promise settles.
- `spawnAsyncLog` mirrors `spawnSyncLog`: same command, same environment, same `shell: true`, awaited instead
  of blocking. **What gets executed does not change in this PR**, only when the main process is free.
- the on-quit install keeps the synchronous path: a quit handler cannot await, and the app is exiting anyway.

The `must be sync` constraint on `doInstall` comes from that quit handler. It does not apply to the two paths
above: `quitAndInstall()` is called by the app, and elevation targets are already excluded from the automatic
install at launch in favour of an explicit `installPendingUpdateIfAvailable()`.

To avoid two divergent implementations, each package manager now declares its commands once as an
`InstallPlan` — a list of alternatives, the next one attempted only if the previous failed — executed by both
`runInstallPlan` (sync) and `runInstallPlanWithSudoIfNeededAsync`. `DebUpdater.installWithCommandRunner`,
`RpmUpdater.installWithCommandRunner` and `PacmanUpdater.installWithCommandRunner` keep their signatures and
behaviour; they now run the plan.

## Tests

`pnpm pretest` → lint-deps, lint, typecheck, typecheck:test all clean.
`TEST_FILES="linuxUpdaterUnit,baseUpdaterUnit,installOnNextLaunch" pnpm ci:test` → 79 passed.

Added to `linuxUpdaterUnitTest.ts`: the async install issues the same command as the synchronous path, never
reaches the blocking `spawnSync` (only the unprivileged `command -v` probes may be sync), still repairs
dependencies with apt-get when dpkg fails, and reports the error instead of relaunching when everything fails.

The existing `installWithCommandRunner` tests are untouched and still pass, which is the check that the sync
on-quit path did not change. `installOnNextLaunchTest.ts` needed one helper: its assertions spy on
`doInstall`, so the async path is routed back to that spy for the Linux fixtures.

## Notes

`quitAndInstall()` returns before the install finishes on the Linux path. It already returned `void` and quit
from a `setImmediate`, and the quit now happens once the install resolved, but it is a timing change worth
knowing about for that platform.

Follow-up in #10093: now that the command is spawned asynchronously, it can also be spawned
without a shell, which makes the authentication dialog show the actual command.

### DIFF
--- a/packages/electron-updater/src/LinuxUpdater.ts
+++ b/packages/electron-updater/src/LinuxUpdater.ts
@@ -1,10 +1,33 @@
 import { AllPublishOptions } from "builder-util-runtime"
 import { AppAdapter } from "./AppAdapter.js"
 import { BaseUpdater } from "./BaseUpdater.js"
+import { Logger } from "./types.js"
 
 // Matches safe package manager names: alphanumeric, hyphens, underscores only.
 // Rejects names with shell metacharacters that could cause command injection.
 const SAFE_PM_REGEX = /^[a-zA-Z0-9_-]+$/
+
+/**
+ * An install plan: a list of alternatives, each a sequence of commands. The first alternative that succeeds
+ * wins; the next one is attempted only when the previous failed. It exists so the commands of a package
+ * manager are declared once and executed by both the synchronous on-quit path and the asynchronous one.
+ */
+export type InstallPlan = string[][][]
+
+/** Runs an {@link InstallPlan} synchronously, for the on-quit path where an async install cannot be awaited. */
+export function runInstallPlan(plan: InstallPlan, commandRunner: (commandWithArgs: string[]) => void, logger: Logger): void {
+  for (let i = 0; i < plan.length; i++) {
+    try {
+      plan[i].forEach(commandRunner)
+      return
+    } catch (error: any) {
+      if (i === plan.length - 1) {
+        throw error
+      }
+      logger.warn(`${error.message ?? error} — trying the next install command`)
+    }
+  }
+}
 
 export abstract class LinuxUpdater extends BaseUpdater {
   constructor(options?: AllPublishOptions | null, app?: AppAdapter) {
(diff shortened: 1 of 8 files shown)


packages/electron-updater/src/LinuxUpdater.ts
```diff
@@ -1,11 +1,34 @@
...
 // Matches safe package manager names: alphanumeric, hyphens, underscores only.
 // Rejects names with shell metacharacters that could cause command injection.
 const SAFE_PM_REGEX = /^[a-zA-Z0-9_-]+$/
 
+/**
+ * An install plan: a list of alternatives, each a sequence of commands. The first alternative that succeeds
+ * wins; the next one is attempted only when the previous failed. It exists so the commands of a package
+ * manager are declared once and executed by both the synchronous on-quit path and the asynchronous one.
+ */
+export type InstallPlan = string[][][]
```
mmaietta, line 15: Nit: Can we structure this as a different type? An array of arrays of arrays seems like a bug waiting to happen when later code changes touch this area of the codebase.

Spitballing here; curious if you could expand on this further? (open to proposals as well)
```
export type InstallPlan = {
   execCommand: string[]
   recoveryChain?: InstallPlan[] 
}
```

packages/electron-updater/src/LinuxUpdater.ts
```diff
@@ -40,18 +63,55 @@ export abstract class LinuxUpdater extends BaseUpdater {
...
+    this._logger.info(`Running as non-root user, using sudo to install: ${sudo}`)
+    const wrapper = this.commandWrapperFor(sudo)
+    await this.spawnAsyncLog(sudo[0], [...(sudo.length > 1 ? sudo.slice(1) : []), `${wrapper}/bin/bash`, "-c", `'${commandWithArgs.join(" ")}'${wrapper}`])
+  }
+
+  /**
+   * Runs an install plan: the first sequence of commands that succeeds wins, the next one is only attempted
+   * when the previous failed. Both install paths share it so the commands themselves live in one place.
+   */
+  protected async runInstallPlanWithSudoIfNeededAsync(plan: InstallPlan): Promise<void> {
```
mmaietta, line 101: Nit: Instead of duplicating logic almost verbatim from `runInstallPlan`, can we refactor this function or `runInstallPlan` to leverage a common utility function/wrapper?

PR #10123, opened 2026-08-31
fix(runtime): stop oversized buffer downloads

OskarEichler:
## Summary

- settle oversized in-memory downloads once
- destroy the response immediately after crossing the 500 MB limit
- ignore subsequent data/end events
- add a synthetic regression that does not allocate a 500 MB buffer

## Why

The limit rejected the public promise but kept reading and appending every later response chunk. A malicious or misconfigured server could therefore continue growing memory after the size guard had already fired.

## Verification

- `TEST_FILES=downloadToBufferTest corepack pnpm ci:test`
- `corepack pnpm compile`
- `corepack pnpm ci:validate`
- `git diff --check`

## Breaking changes

None.

### DIFF
--- a/.changeset/stop-oversized-buffer-downloads.md
+++ b/.changeset/stop-oversized-buffer-downloads.md
@@ -1,0 +1,5 @@
+---
+"builder-util-runtime": patch
+---
+
+fix: stop oversized in-memory downloads at the configured limit
--- a/packages/builder-util-runtime/src/httpExecutor.ts
+++ b/packages/builder-util-runtime/src/httpExecutor.ts
@@ -289,16 +289,28 @@ Please double check that your authentication token is correct. Due to security r
           },
           responseHandler: (response, callback) => {
             let receivedLength = 0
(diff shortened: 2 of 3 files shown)


mmaietta:
Stating, just for posterity, that the PR descripton/conclusion/"why" here is incorrect - claiming memory keeps growing after the limit. It doesn't — the data handler returns before pushing once the limit is crossed. The actual defects are an unterminated transfer and a promise settled repeatedly.

Why you might ask?
The size guard rejected the promise but did not stop the transfer. After the limit was crossed, the response kept streaming to completion (socket held open, bandwidth wasted), the error callback fires again on every further chunk, and the `end` handler still runs `Buffer.concat` over the ~500 MB of retained chunks on an already-rejected promise. Chunks past the limit are never retained, so memory does not grow unboundedly.

That being said, this PR change is correct in that it terminates the transfer at the limit and settles exactly once.

mmaietta (approved):

PR #10127, opened 2026-08-31
fix(updater): validate update cache filenames

OskarEichler:
## Summary

- extract update cache filename resolution into a focused helper
- reject empty, dot-segment, null-byte, and non-portable separator results
- preserve safe basenames for nested manifest paths
- add traversal and valid-path regressions

## Why

`path.basename()` neutralizes ordinary nested paths, but `path.basename("..")` remains `".."`. Joining that result to the pending cache directory resolves to its parent, so a hostile update manifest could place the downloaded file outside the intended directory.

## Verification

- `TEST_FILES=updateCacheFilenameTest corepack pnpm ci:test`
- `corepack pnpm compile`
- `corepack pnpm ci:validate`
- `git diff --check`

## Breaking changes

Manifests whose resolved artifact filename is empty, `.`/`..`, contains a null byte, or remains path-like after basename normalization now fail with `ERR_UPDATER_INVALID_FILE_NAME`.

### DIFF
--- a/.changeset/validate-update-cache-filename.md
+++ b/.changeset/validate-update-cache-filename.md
@@ -1,0 +1,5 @@
+---
+"electron-updater": patch
+---
+
+fix: reject traversal segments as update cache filenames
--- a/packages/electron-updater/src/AppUpdater.ts
+++ b/packages/electron-updater/src/AppUpdater.ts
@@ -809,22 +809,10 @@ export abstract class AppUpdater extends (EventEmitter as new () => TypedEmitter
     const version = updateInfo.version
     const packageInfo = fileInfo.packageInfo
 
-    function getCacheUpdateFileName(): string {
-      // NodeJS URL doesn't decode automatically
-      const urlPath = decodeURIComponent(taskOptions.fileInfo.url.pathname)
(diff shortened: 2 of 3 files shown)


mmaietta:
I appreciate the contribution, but the description/conclusion/"why" is incorrect: a hostile manifest "could place the downloaded file outside the intended directory." It can't: `basename` yields one segment at most, so `..` maps the update path onto the cache dir's parent, an existing directory, and the final rename fails. It's a confusing+late error, but not an attacker-chosen write.
`path.basename()` collapses nested paths to a single segment, but `basename("..")` and `basename(".")` stay `".."` and `"."`, so the existing comment claiming `basename` alone neutralizes traversal is not quite true. With such a manifest, the update file path resolves to the cache directory itself or its parent, and the download fails late with a confusing rename error instead of a clear validation error. Because `basename` can only produce one segment, this **cannot write to an attacker-chosen location**
That being said, this PR change does apply a defense-in-depth that fails fast with a more appropriate error:  `ERR_UPDATER_INVALID_FILE_NAME`.

mmaietta (approved):

PR #10176, opened 2026-09-04
fix(snap): stop emitting a duplicate desktop entry for core24 snaps

### LINKED ISSUE #10077: [Snap Core24] Unnecessary .desktop file created
We noticed an unnecessary (duplicate or wrong) .desktop file created while building Obsidian Snap package.

Our setup is electron builder 26.15.7, core24, classic confinement, `syncDesktopName: true`, and `desktopName: md.obsidian.Obsidian`

Two .desktop files were generated (so there are two items in the user's app launcher).

`md.obsidian.Obsidian.desktop`
`obsidian.desktop`

We were expecting only the first one, not the second.
They have the same content.

```
[Desktop Entry]
Name=Obsidian
Exec=obsidian %U
Terminal=false
Type=Application
Icon=${SNAP}/meta/gui/icon.png
StartupWMClass=md.obsidian.Obsidian
Comment=Obsidian
MimeType=x-scheme-handler/obsidian;
Categories=Office;
```

atirna:
Fixes #10077

Core24 snaps get two launcher entries when `desktopName` differs from the Snap app name. I removed the `apps.<app>.desktop` mapping and kept the generated `snap/gui/<desktopName>.desktop`. Its `Exec` uses the exported Snap command even when `executableName` contains uppercase letters.

## Verification

- `TEST_FILES=snapcraftTest pnpm ci:test`: 70 passed, two Multipass build cases skipped. The regression case fails with the old desktop mapping and with an unnormalized `Exec`.
- `pnpm compile`
- `pnpm ci:validate`


mmaietta (dismissed):

atirna:
the heavy snap suites keep their own snapshot file (test/snapshots/linux/snapHeavyTest.js.snap) and my first push only updated snapcraftTest's, so all four Test Snap jobs failed on 5 stale core24 entries.

pushed 1e30e57: drops the stale apps.se-wo-template.desktop lines from those 5 entries (snap full (core24), snap install+launch (core24), and the three core24 native cases). each failing diff in the job logs was exactly that one line, so no other expectations changed.

also merged current master (de70642c), the mac license-files PR, to build against the live base.

snapcraftTest: 70 passed, 2 skipped locally. the heavy suite needs linux+snapd so it only runs in CI here.

mmaietta (dismissed):

packages/app-builder-lib/src/targets/linux/snap/core24.ts
```diff
@@ -74,7 +74,7 @@ export class SnapCore24 extends SnapCore<SnapOptions24> {
     // Create desktop file in snap/gui/ directory
     // Snapcraft will automatically copy this to meta/gui/ in the final snap
     const desktopFilePath = path.join(guiOutput, `${this.helper.getDesktopFileName(snap.name)}.desktop`)
-    await this.helper.writeDesktopEntry(this.options, this.packager.executableName + " %U", desktopFilePath, desktopExtraProps)
+    await this.helper.writeDesktopEntry(this.options, snap.name + " %U", desktopFilePath, desktopExtraProps)
```
mmaietta, line 77: So `coreLegacy` has the same latent bug at [coreLegacy.ts:221](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/targets/linux/snap/coreLegacy.ts#L220-L221). It writes the same case-preserving Exec, and nothing rewrites it there either. So core24 did follow legacy, but legacy is wrong too. Can you please mirror the one-liner change here into `coreLegacy` and add an Exec= assertion to the legacy desktopFile test.
Note: The launcher scripts keep executableName, since that is the binary filename. Nothing else needs it.

I'm wondering if we also migrate the `" %U"` into the `writeDesktopEntry` function since it is replicated across both coreLeggacy and core24.

atirna:
pushed 797da4aa: coreLegacy writes the snap name in Exec now, and the heavy suite asserts Exec=se-wo-template %U for every core. left the %U move into writeDesktopEntry alone, all four callers append it so that felt like its own change.

mmaietta (approved):

Issue #3940, opened 2019-06-05
Executables within app.asar.unpacked are not being signed correctly in OSX 10.14.5


dboakes:
* **Version**: 20.43.0
* **Electron Updater Version**: 4.0.6

* **Target**: Mac OSX 10.14.5

I have implemented the fix mentioned in issue [#3828](https://github.com/electron-userland/electron-builder/issues/3828), also outlined [here.](https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/)

However, this isn't working for other binaries found within app.asar.unpacked. Before OSX 10.14.15, I had included a python binary which I had packed via pyinstaller, and this was then automatically signed by Electron Builder. 

The error output states "Code signature found in _[FILE_PATH]__ not valid for use in process using Library Validation: mapped file has no cdhash, completely unsigned? Code has to be at least ad-hoc signed."

I can see in the Contents folder of the app, that the CodeResources file contains keys for all the python binaries, but they don't seem to be recognised. 

I have tried signing the files manually with codesign, which allowed the app to work on my mac, but when trying to open the app on another mac, it would get an error, as the package had been amended after being signed.

yoannsark:
I had the same problem, that was fixed by adding `com.apple.security.cs.disable-library-validation` to plist file.

My electron builder configuration :
```
    "mac": {
      "hardenedRuntime": true,
      "gatekeeperAssess": false,
      "entitlements": "./mac_build_files/entitlements.mac.inherit.plist",
      "entitlementsInherit": "./mac_build_files/entitlements.mac.inherit.plist",
      "target": ["dmg"]
    },
```
My entitlements.mac.inherit.plist :
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>com.apple.security.cs.allow-jit</key>
    <true/>
    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
    <true/>
    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
    <true/>
    <key>com.apple.security.cs.disable-library-validation</key>
    <true/>
  </dict>
</plist>
```

2025 EDIT: can lead to security issues please check @balazserdeszlogmein [comment](https://github.com/electron-userland/electron-builder/issues/3940#issuecomment-2943380357)

[...]

PR #10181, opened 2026-09-07
feat(mac): tighten default macOS entitlements [BREAKING]

mmaietta:
_Requested by **Mike Maietta** · [Slack thread](https://electronhq.slack.com/archives/C0BGNPTEH8A/p1789681737078369)_

**Before:** Every macOS app built without its own entitlements file was signed with one blanket plist, applied to the app bundle and to every nested binary, that granted `com.apple.security.cs.allow-unsigned-executable-memory` and `com.apple.security.cs.disable-library-validation` to every process. MAS builds without `build/entitlements.mas.plist` got that same non-sandboxed mac template. Apps that loaded a native module signed by another Team ID (or unsigned) worked only because library validation was switched off for everyone, and nothing told you that was happening.

**After:** The default app entitlements grant only `com.apple.security.cs.allow-jit`, nested binaries get `@electron/osx-sign`'s Chromium-derived per-file defaults, MAS builds fall back to osx-sign's sandboxed `default.mas.plist`, and ad-hoc builds (`mac.sign.identity: "-"`) keep launching thanks to a dedicated ad-hoc template. After signing, electron-builder inspects the Mach-O binaries under `app.asar.unpacked` and warns when any of them is unsigned or signed by a different Team ID while `disable-library-validation` is absent, so the apps that genuinely need the entitlement find out at build time rather than from a launch crash.

Modern Electron does not need either loose entitlement in the main process, and both weaken the Hardened Runtime, so this stops granting them by default and reports the one case they actually cover.

**How:** `getOptionsForFile()` now goes through two resolvers, `getAppEntitlements()` and `getInheritEntitlements()`, which return `null` when nothing is configured so that `@electron/osx-sign` applies its own per-file default instead of a blanket plist. The same resolvers back `isLibraryValidationDisabled()`, so the signing path and the ad-hoc warning can no longer disagree. `warnAboutForeignSignedBinaries()` runs from `macPackager` after `doSign()` for non-custom signers; it reads the Team ID of the signed `.app` with `codesign -d --verbose=4`, walks `app.asar.unpacked`, identifies Mach-O files by magic (disambiguating the `0xcafebabe` fat magic from Java class files via `nfat_arch`), and compares each file's `TeamIdentifier` against the app's.

### What changed

- **Default templates tightened.** `templates/entitlements.mac.plist` grants `allow-jit` only, matching Chromium's `app-entitlements.plist`. New `templates/entitlements.mac.adhoc.plist` grants `allow-jit` + `disable-library-validation`.
- **Entitlement fallback chain** (`MacTargetHelper`):
  ```
  getAppEntitlements(targetPlatform, signOpts, adHoc)
    sign.entitlements → build/entitlements.{mac,mas}.plist
    → null for MAS (osx-sign default.mas.plist)
    → entitlements.mac.adhoc.plist when adHoc, else entitlements.mac.plist

  getInheritEntitlements(targetPlatform, signOpts, adHoc)
    sign.entitlementsInherit → build/entitlements.{mac,mas}.inherit.plist
    → entitlements.mac.adhoc.plist when adHoc and not MAS
    → null (osx-sign's per-file defaults)
  ```
  `null` becomes `entitlements: undefined` in the `PerFileSignOptions`, which osx-sign's `mergeOptionsForFile()` treats as "use my default for this file". `getOptionsForFile()` takes the resolved `Identity` so it can detect ad-hoc signing via the new `MacTargetHelper.isAdHocIdentity()`.
- **Post-sign diagnostic** `warnAboutForeignSignedBinaries(appPath, identity, targetPlatform, signOpts)`: skipped for MAS, ad-hoc identities, hardened runtime off, custom signers, apps whose entitlements already grant `disable-library-validation`, and apps with no `app.asar.unpacked`. Reads the app's Team ID from the signed bundle (not from the identity name, which need not carry one). Best-effort, never fails the build. Helpers `isMachOFile()` and `parseSigningTeamId()` are exported via `app-builder-lib/internal` for tests.
- **Ad-hoc warning** now fires only when a user-supplied entitlements file overrides the ad-hoc default without granting `disable-library-validation`.
- **`macOptions.ts`** / `scheme.json`: doc comments for `entitlements` and `entitlementsInherit` describe the new fallback chain.
- **Docs:** `website/docs/mac.md` (default entitlements table, "Supplying your own entitlements", "Loading third-party or unsigned binaries"), `website/docs/notarization.md`, `website/docs/features/code-signing/notarization.md`, `website/docs/features/code-signing/code-signing-mac.md`, plus `website/docs/migration/v27-breaking-changes.md` and the checklist in `website/docs/migration/v26-to-v27.md`.
- **Changeset:** `.changeset/tighten-mac-entitlement-defaults.md`, `app-builder-lib: major`.

### Breaking change

Apps that load frameworks or native modules signed by another Team ID, or unsigned, or that still rely on `allow-unsigned-executable-memory`, must add those keys to their own `build/entitlements.mac.plist` (and `build/entitlements.mac.inherit.plist` for nested binaries). The new post-sign warning surfaces exactly that case (the `#3940` failure mode). This is a runtime default, not a config-key rename, so `migrate-schema` does not touch it.

### Test coverage

`test/src/mac/MacTargetHelperTest.ts` grows from 48 to 115 tests: bundled template contents (with a regression guard that the mac default grants neither loose entitlement), both resolvers' precedence across `mac`/`mas`/`mas-dev`, the ad-hoc paths, `getOptionsForFile()` end to end, `isAdHocIdentity()`, the host-independent early returns of `warnAboutForeignSignedBinaries()`, `isMachOFile()` (thin LE/BE, fat, Java `.class` collision, script, short file, missing path) and `parseSigningTeamId()` (present, `not set`, absent, trailing whitespace).

Verified with `TEST_FILES=MacTargetHelperTest pnpm ci:test` (115 passed), `pnpm compile`, `pnpm typecheck:test`, `eslint` and `prettier --check` on the changed files.

### DIFF
--- a/packages/app-builder-lib/src/targets/mac/MacTargetHelper.ts
+++ b/packages/app-builder-lib/src/targets/mac/MacTargetHelper.ts
@@ -1,7 +1,8 @@
 import { notarize, type NotarizeOptions, type NotaryToolKeychainCredentials } from "@electron/notarize"
 import type { PerFileSignOptions, SigningDistributionType, SignOptions } from "@electron/osx-sign"
-import { Arch, InvalidConfigurationError, log, statOrNull } from "builder-util"
+import { Arch, InvalidConfigurationError, log, spawnAndWriteWithOutput, statOrNull, walk } from "builder-util"
 import { Nullish } from "builder-util-runtime"
+import { open, type FileHandle } from "fs/promises"
 import * as path from "path"
 import { CertType, findIdentity, Identity, reportError } from "../../codeSign/mac/macCodeSign.js"
 import { SigningResult } from "../../codeSign/signResult.js"
@@ -13,8 +14,23 @@ import { getTemplatePath } from "../../util/pathManager.js"
 export type MasPlatformType = "mas" | "mas-dev"
 export type PlatformType = MasPlatformType | "mac"
 
+const DISABLE_LIBRARY_VALIDATION = "com.apple.security.cs.disable-library-validation"
+
+// e.g. "Developer ID Application: Example Inc. (A1B2C3D4E5)"
+const TEAM_ID_IN_IDENTITY_NAME = /\(([A-Z0-9]{10})\)\s*$/
+
 export class MacTargetHelper {
   constructor(private packager: MacPackager) {}
+
+  /** Ad-hoc signing (`sign.identity: "-"`) produces a signature with no Team ID. */
+  static isAdHocIdentity(identity: Identity | Nullish): boolean {
+    return identity?.name === "-"
+  }
+
+  /** The Team ID embedded in a signing identity's common name, or `null` for ad-hoc/unnamed identities. */
+  static getTeamIdFromIdentity(identity: Identity | Nullish): string | null {
+    return (identity?.name && TEAM_ID_IN_IDENTITY_NAME.exec(identity.name)?.[1]) || null
+  }
 
   handleNullIdentity(): SigningResult {
     if (this.packager.forceCodeSigning) {
@@ -48,8 +64,9 @@ export class MacTargetHelper {
         if (MacTargetHelper.isHardenedRuntimeEnabledForSigning(targetPlatform, signOpts?.hardenedRuntime) && !(await this.isLibraryValidationDisabled(targetPlatform, signOpts))) {
           log.warn(
             null,
-            "ad-hoc signing with hardenedRuntime enabled requires the com.apple.security.cs.disable-library-validation entitlement " +
-              "to prevent app launch failures due to library validation. See https://electron.build/docs/features/code-signing for details."
+            `ad-hoc signing with hardenedRuntime enabled requires the ${DISABLE_LIBRARY_VALIDATION} entitlement ` +
+              "to prevent app launch failures due to library validation, but your entitlements file does not grant it. " +
+              "See https://electron.build/docs/features/code-signing for details."
           )
         }
         identity = new Identity("-", undefined)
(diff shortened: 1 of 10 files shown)


packages/app-builder-lib/src/targets/mac/MacTargetHelper.ts
```diff
@@ -63,25 +75,89 @@ export class MacTargetHelper {
...
+   * Nested binaries therefore do not inherit the app's entitlements, and apart from the plugin helper none of these
+   * defaults grant `disable-library-validation` or `allow-unsigned-executable-memory`.
+   */
+  async getInheritEntitlements(targetPlatform: PlatformType, signOpts: ElectronSignOptions | Nullish, adHoc: boolean): Promise<string | null> {
+    if (signOpts?.entitlementsInherit) {
+      return signOpts.entitlementsInherit
+    }
+    const isMas = MacTargetHelper.isMasTarget(targetPlatform)
+    const p = `entitlements.${isMas ? "mas" : "mac"}.inherit.plist`
+    if ((await this.packager.resourceList).includes(p)) {
```
mmaietta: @claude this should be using `this.packager.getResource(...)` function instead for accessing.

packages/app-builder-lib/src/targets/mac/MacTargetHelper.ts
```diff
@@ -63,25 +75,89 @@ export class MacTargetHelper {
...
+   * Returns `null` when no default applies, which lets `@electron/osx-sign` fall back to its own
+   * Chromium-derived defaults (`default.mas.plist` for MAS).
+   */
+  async getAppEntitlements(targetPlatform: PlatformType, signOpts: ElectronSignOptions | Nullish, adHoc: boolean): Promise<string | null> {
+    if (signOpts?.entitlements) {
+      return signOpts.entitlements
+    }
+    const isMas = MacTargetHelper.isMasTarget(targetPlatform)
+    const p = `entitlements.${isMas ? "mas" : "mac"}.plist`
+    if ((await this.packager.resourceList).includes(p)) {
```
mmaietta: @claude same comment here: https://github.com/electron-userland/electron-builder/pull/10181/changes#r4088837166

PR #10186, opened 2026-09-10
feat(nsis): "store-asar" mode for proportional differential updates

imlucas:
**Before:** The differential updater diffs the compressed NSIS package with a content-defined blockmap, and `resources/app.asar` is a single LZMA member. Any app-code change, even one line, rewrites the asar header and shifts every compressed block, so users re-download the entire asar on every release (measured: 536,260 B, 100% of the member, for a one-line renderer change on a ~32 MB asar).

**After:** With `nsis.differentialPackage: "store-asar"`, the asar is stored uncompressed (7-Zip `Copy`) inside the package. Unchanged regions stay byte-identical between releases, so the same one-line change costs 81,857 B (~3% of the asar). The trade-off, and why it is opt-in: the installer and full package grow by roughly what LZMA saved on the asar (30.8 MB to 33.1 MB here).

Apps whose releases are mostly app-code changes currently get little benefit from differential updates; this makes the delta proportional to what actually changed.

**How:** A generic `ArchiveOptions.storedPaths` in `archive()`. Listed paths are excluded from the compressed pass with exact `-x!` masks, then appended in a second 7za invocation with `-mx=0`, mirroring the first pass's `-snl` / `-mhc=off` flags so the final header matches. The append leaves the first pass's packed streams byte-identical. `ELECTRON_BUILDER_COMPRESSION_LEVEL` deliberately does not reach stored members. `Copy` needs no 7z filter, so members stay decodable by the install-time Nsis7z extractor. `NsisTarget` wires `resources/app.asar` through it when the package is differential-aware.

Config semantics (`nsis.differentialPackage`, default `true`):
- `false`: no differential download support (unchanged).
- `true` / `"compressed"` / unset: differential-aware, whole package compressed (today's behaviour).
- `"store-asar"`: differential-aware, asar stored uncompressed.

Misspelled values fail schema validation and the error lists the accepted values.

**Testing:** `archiveUtilTest`: `archive() storedPaths` (stored member appended as Copy while the rest stays compressed; raw bytes appear verbatim; exact differential-aware NSIS options; missing stored paths skipped; `..` throws; withoutDir prefixing; append pass leaves packed streams byte-identical; mirrors `isArchiveHeaderCompressed=false`; mirrors `preserveSymlinks`; `ELECTRON_BUILDER_COMPRESSION_LEVEL` does not reach stored members) and `archive() exclude masks`, all against real 7za.

**Changeset:** `app-builder-lib`: minor.

### DIFF
--- a/packages/app-builder-lib/src/targets/archive.ts
+++ b/packages/app-builder-lib/src/targets/archive.ts
@@ -109,7 +109,8 @@ export interface ArchiveOptions {
    * builds, so the delta stays proportional to what actually changed. The stored members are excluded
    * from the main (compressed) pass and appended in a second 7za invocation with `-mx=0`; the
    * `ELECTRON_BUILDER_COMPRESSION_LEVEL` override deliberately does not apply to them (byte-stability
-   * is the point). Paths that do not exist are skipped. Not supported by the native-zip NFD fallback.
+   * is the point), while `preserveSymlinks` and `isArchiveHeaderCompressed` are mirrored into the
+   * append pass. Paths that do not exist are skipped. Not supported by the native-zip NFD fallback.
    */
   storedPaths?: Array<string> | null
 
@@ -291,13 +292,21 @@ export async function archive(format: string, outFile: string, dirToArchive: str
 
     if (storedMembers.length > 0) {
       // Second pass: append the stored members with no compression. -mx=0 unconditionally — the
-      // compression-level env override must not reach these members (see storedPaths docs).
+      // compression-level env override must not reach these members (see storedPaths docs). The
+      // main pass's symlink, timestamp, and header-compression flags are mirrored so the option
+      // stays generic: the append rewrites the archive header and may itself store a symlink.
       const appendArgs = debug7zArgs("a")
       appendArgs.push("-mx=0")
+      if (options.preserveSymlinks) {
+        appendArgs.push("-snl")
+      }
       if (!options.isRegularFile) {
         appendArgs.push("-mtc=off")
       }
       if (format === "7z" || format.endsWith(".7z")) {
+        if (options.isArchiveHeaderCompressed === false) {
+          appendArgs.push("-mhc=off")
+        }
         appendArgs.push("-mtm=off", "-mta=off")
       } else if (format === "zip") {
         appendArgs.push("-mm=Copy", "-mcu")
--- a/packages/app-builder-lib/src/targets/archive.ts
+++ b/packages/app-builder-lib/src/targets/archive.ts
@@ -102,6 +102,18 @@ export interface ArchiveOptions {
   preserveSymlinks?: boolean
 
   /**
+   * Paths, relative to the archived directory, to append as uncompressed (`Copy`) members instead of
+   * compressing them with the rest of the archive. Differential updates diff the *compressed* archive
+   * with a content-defined blockmap, and a large compressed member (an asar) diverges wholesale from
+   * its previous version on any change — storing it keeps unchanged regions byte-identical between
+   * builds, so the delta stays proportional to what actually changed. The stored members are excluded
+   * from the main (compressed) pass and appended in a second 7za invocation with `-mx=0`; the
+   * `ELECTRON_BUILDER_COMPRESSION_LEVEL` override deliberately does not apply to them (byte-stability
+   * is the point). Paths that do not exist are skipped. Not supported by the native-zip NFD fallback.
+   */
+  storedPaths?: Array<string> | null
+
+  /**
    * Restrict the 7z filter to one the install-time extractor (the self-vendored Nsis7z plugin) can
    * decode. Modern 7za (24.09) auto-applies a CPU branch converter to executable content at
    * `-mx>=1` — `BCJ2` on x86/x64 and `ARM64` on arm64 — which that decoder silently skips, dropping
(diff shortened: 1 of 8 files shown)


packages/app-builder-lib/src/targets/archive.ts
```diff
@@ -101,6 +101,18 @@ export interface ArchiveOptions {
...
+   * Paths, relative to the archived directory, to append as uncompressed (`Copy`) members instead of
+   * compressing them with the rest of the archive. Differential updates diff the *compressed* archive
+   * with a content-defined blockmap, and a large compressed member (an asar) diverges wholesale from
+   * its previous version on any change — storing it keeps unchanged regions byte-identical between
+   * builds, so the delta stays proportional to what actually changed. The stored members are excluded
+   * from the main (compressed) pass and appended in a second 7za invocation with `-mx=0`; the
+   * `ELECTRON_BUILDER_COMPRESSION_LEVEL` override deliberately does not apply to them (byte-stability
+   * is the point). Paths that do not exist are skipped. Not supported by the native-zip NFD fallback.
+   */
+  storedPaths?: Array<string> | null
```
mmaietta, line 114: Some additional notes / things I'd like added:
- Test gap. Nothing asserts the second pass leaves the first pass's compressed streams byte-identical. One test comparing packed streams with and without the stored asar would close it.
- Append pass drops flags. It ignores `preserveSymlinks` and header compression. Note, this seems effectively moot for NSIS, but this option is documented as generic, so we should either mirror the flags or narrow tsdoc.

packages/app-builder-lib/src/targets/win/nsis/nsisOptions.ts
```diff
@@ -211,6 +211,21 @@ export interface NsisOptions extends CommonNsisOptions, CommonWindowsInstallerCo
...
+   * single compressed member — so any change to app code re-downloads the entire compressed asar (~100% of the
+   * member; its header rewrite alone diverges every block). Storing it keeps unchanged regions byte-identical
+   * between releases, making the delta proportional to what actually changed (measured on a ~32 MB asar: a
+   * one-line source change cost 0.2% instead of 100%).
+   *
+   * Trade-off: the installer grows by roughly what compressing the asar saved. Has no effect when
+   * `differentialPackage` is `false`.
+   * @default false
+   */
+  readonly differentialPackageStoreAsar?: boolean
```
mmaietta, line 227: Since this only has an effect when `differentialPackage` is not `false`, I'm thinking it would be better to extend the functionality of that property instead of adding a new one. e.g.
```
readonly differentialPackage?: boolean | "compressed" | "store-asar" | null
```
Mapping: `false` is off, "store-asar", everything else (true, "compressed", null, unset) is compressed.
Note: `Troubleshooting.md` will also need an update to state `non-false` if we update the type definition specificed here: https://github.com/electron-userland/electron-builder/blob/6be279576bf22a6f7521b146f89fd7501839bc94/website/docs/troubleshooting.md?plain=1#L123

Thoughts?

mmaietta (commented):
Thanks for the PR!
Left two comments. Let me know if you're good to push the updates from your side or if you'd like me to make the additions/mods via some Claude handiwork.

packages/app-builder-lib/src/targets/archive.ts
imlucas: Both addressed in the latest push:
Test gap — added "append pass leaves the compressed pass's packed streams byte-identical": it builds the archive with the stored asar and a baseline archive from the same compressed-pass member set (same files, asar absent), then compares the packed-stream regions byte-for-byte. One subtlety worth noting: the region can't be bounded by NextHeaderOffset, because that span also includes the LZMA-compressed archive header, which legitimately differs between the two archives (different member list). The file packed streams sit back-to-back right after the 32-byte signature header, so the test bounds the comparison with the sum of the entries' Packed Size from the -slt listing (new listArchiveEntryPackedSizes helper). It also asserts the appended Copy member's verbatim bytes start exactly where the compressed streams end.
Dropped flags — went with mirroring rather than narrowing the tsdoc, so the option stays generic as documented: the append pass now passes -snl when preserveSymlinks is set and -mhc=off when isArchiveHeaderCompressed === false (7z only). Both are load-bearing there — the append rewrites the archive header, and a stored path can itself be a symlink. Each has a test: the end-header marker check (kHeader 0x01 with -mhc=off vs kEncodedHeader 0x17 by default, with the default case guarding against a vacuous pass), and a stored symlink member surviving extraction as a link.

packages/app-builder-lib/src/targets/win/nsis/nsisOptions.ts
imlucas: Adopted exactly as proposed: differentialPackage is now boolean | "compressed" | "store-asar" | null and the separate differentialPackageStoreAsar option is gone. false disables differential support, "store-asar" opts into the stored asar, and everything else (true, "compressed", null, unset) keeps the fully compressed differential package. isBuildDifferentialAware (!== false) needed no change for the union. scheme.json is regenerated, and Troubleshooting.md now reads "must not be false" per your note.

imlucas:
@mmaietta thanks! Updated per feedback and replied.

> Thanks for the PR! Left two comments. Let me know if you're good to push the updates from your side or if you'd like me to make the additions/mods via some Claude handiwork.

mmaietta (approved):
Updated the PR title+description to match the changes we discussed. Also removed `null` as a follow-up since it improves the schema generation results.

LGTM. Kicking off CI now.

PR #10192, opened 2026-09-12
fix(updater): make multi-range differential downloads work on real servers

yi-ge:
**Before:** `downloadUpdate` regularly abandons the differential path and re-downloads the whole installer with

```
Cannot download differentially, fallback to full download: Error: Response ends without calling any handlers
```

Three independent defects end in that same message.

**1. `DataSplitter` only recognises CRLF.** RFC 2046 requires CRLF, but real servers do not always send it. Qiniu CDN (`Server: openresty`, `Content-Type: multipart/byteranges; boundary="web cache:<hash>"`) answers multi-range requests with bare LF — the body starts `0a 2d 2d` (`\n--`), not `0d 0a 2d 2d`. `searchHeaderListEnd` looks for `\r\n\r\n` only, so not a single part is ever recognised: every chunk is appended to `headerListBuffer` (a `Buffer.concat` per chunk), the response ends without one `finishHandler` call, and the watchdog rejects 10s later. Measured on a 397 MB plan: ~1 GB RSS in the main process and zero bytes written before it gave up.

**2. A header-list terminator split across two chunks is never found.** `searchHeaderListEnd` searches only the new chunk; the bytes saved into `headerListBuffer` are never searched again. When a chunk boundary happens to fall inside `\r\n\r\n`, that part's header end is missed, the parser locks onto the next part's header, the output is misaligned and the final sha512 check fails — a full re-download, on a spec-compliant server. Tests that feed the whole body in one chunk cannot see this.

**3. The 10s watchdog outlives a successful batch.** With more than 1000 operations the plan is split into several range requests. `response.on("end")` arms `setTimeout(reject, 10000)` unconditionally and nothing clears it when that batch finishes, so a download whose *remaining* batches take longer than the grace period is failed while it is still making progress. The next batch reuses the same `reject`, so the stale timer takes the whole download down.

**After:**

- header lists end at whichever of `\r\n\r\n` / `\n\n` comes first, and the `<EOL>--boundary` separator length follows the line ending the server actually uses (`\r\n--` vs `\n--`);
- only the last few bytes of an unfinished header list are carried over and searched together with the next chunk — which also removes the unbounded buffering of the whole response;
- the watchdog is cleared once the batch has been fully handled.

**Tests** — `test/src/updater/differentialMultipartTest.ts`:

- CRLF and bare-LF bodies, with and without a leading line break, including a quoted boundary containing a space and a colon, each fed at chunk sizes 1, 2, 3, 5, 7, 16, 64, 1000 and all-at-once (chunk size 1 exercises every possible split of a terminator);
- a two-batch download against a local, spec-compliant HTTP server where the second batch outlives the grace period (fake timers advance 11s while it is still in flight).

All five fail on `master` today: the four splitter cases time out with `splitter did not finish (chunk size 1)`, the batch case fails with `Response ends without calling any handlers`.

**Real-world verification** (Windows, Electron 43.6.0 via `ElectronHttpExecutor`, 543 MB NSIS installer, Qiniu CDN, blockmap plan = 41% of the file):

| | upstream 6.8.9 | with this change |
| --- | --- | --- |
| differential download | `Response ends without calling any handlers` after 68s, then a full 543 MB download | OK in 41s; sha512 of the reconstructed installer matches the published one |

Also verified in the shipping app on a plan that needed two range requests (151 MB + 70 MB, the second one ~21s): it completes now and failed on the same build before the change.

A changeset is included (`electron-updater`, patch).

### DIFF
--- a/.changeset/differential-multipart-parsing.md
+++ b/.changeset/differential-multipart-parsing.md
@@ -1,0 +1,9 @@
+---
+"electron-updater": patch
+---
+
+fix: make multi-range differential downloads work on real servers. Three independent defects made `downloadUpdate` fall back to a full download with `Response ends without calling any handlers`:
+
+- `DataSplitter` only recognised CRLF. Some CDNs answer `multipart/byteranges` with bare LF line endings, so no part was ever split and the whole response accumulated in memory. Header lists now end at whichever of `\r\n\r\n` / `\n\n` comes first, and the `<EOL>--boundary` separator size follows the line ending the server actually uses.
+- A header-list terminator split across two chunks was never found: only the new chunk was searched, the buffered bytes never were, so the parser locked onto the next part's header instead. Only the last few bytes of an unfinished header list are now carried over and searched together with the next chunk, instead of accumulating the whole list.
+- The 10s watchdog armed when a batch response ends was never disarmed after that batch succeeded. With more than 1000 operations (several range requests) it failed the whole download whenever a later batch took longer than the grace period.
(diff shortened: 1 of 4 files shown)


mmaietta:
@claude review

mmaietta (approved):

PR #10193, opened 2026-09-12
fix(pkg): add rootVolumeOnly to <options> to avoid the macOS 26 Insta…

### LINKED ISSUE #10179: pkg: unconditional `<domains>` in distribution.xml triggers "Installer would like to access data from other apps" TCC prompt on macOS 26
* **Electron-Builder Version**: 26.8.1 (the code path is identical in 26.15.3 `latest`, 26.16.0 `v26`, 27.0.0-alpha.8 `next`, and `master` — see below)
* **Electron Updater (as-needed)**: 6.8.3 (not involved)
* **Node Version**: v22.23.1
* **Electron Version**: 39.8.10
* **Platform & Target**: macOS 26.5.2 (25F84), Apple silicon → `mac.target: pkg`
* **Debug Logs**: not a build failure — the build succeeds and the `.pkg` installs. The relevant log is the `tccd` excerpt below, captured on the machine running the installer.

### Summary

On macOS 26, every `.pkg` produced by the `pkg` target makes Installer.app show the TCC prompt **"Installer would like to access data from other apps"** as soon as the user clicks *Continue* on the Introduction pane — before any file is written and before `preinstall` runs.

The trigger is the `<domains>` element that `PkgTarget.customizeDistributionConfiguration` unconditionally appends to `distribution.xml`:

```xml
<domains enable_anywhere="false" enable_currentUserHome="false" enable_localSystem="true"/>
```

`<domains>` is a deprecated Distribution element. While building the install plan, macOS 26's Installer probes the current-user-home domain when it sees this element — even with `enable_currentUserHome="false"` — and because Installer.app is sandboxed, that probe hits `kTCCServiceSystemPolicyAppData` and the prompt appears. Adding `rootVolumeOnly="true"` to `<options>` (the current spelling for the same restriction) makes the prompt go away with no change in install behaviour.

There is no configuration escape hatch: the element is inserted regardless of `allowAnywhere` / `allowCurrentUserHome` / `allowRootDirectory`, and `pkg.productbuild` can only append arguments after the `--distribution` that is already fixed.

### Environment

- electron-builder / app-builder-lib **26.8.1**. The same unconditional insertion is present in 26.15.3 (`latest`), 26.16.0 (`v26`), 27.0.0-alpha.8 (`next`) and `master` (`packages/app-builder-lib/src/targets/mac/pkg.ts`), so upgrading does not help.
- macOS 26.5.2 (25F84), Installer.app `Installer-1359.120.1`, Apple silicon.
- Config that reproduces it (nothing unusual):

```yaml
mac:
  target: [pkg]
pkg:
  installLocation: /Applications
  allowAnywhere: false
  allowCurrentUserHome: false
  isRelocatable: false
```

### Steps to reproduce

1. Build any app with the `pkg` target on macOS 26.
2. Double-click the `.pkg`, click **Continue** on the Introduction pane.
3. The prompt *"Installer" would like to access data from other apps* appears.

The prompt is a one-time TCC decision per machine, so it only shows once per user until the record is reset. To reproduce repeatedly:

```sh
tccutil reset SystemPolicyAppData com.apple.installer
open -a Installer MyApp.pkg   # then click Continue
```

### Evidence

`tccd` attributes the request to Installer.app itself, not to any package script (scripts have not run yet at this point):

```
tccd: AUTHREQ_PROMPTING: service=kTCCServiceSystemPolicyAppData,
      subject=Sub:{com.apple.installer}
      binary_path=/System/Library/CoreServices/Installer.app/Contents/MacOS/Installer
```

Bisected by rebuilding the same component package with single-variable changes to `distribution.xml` and counting Installer's `kTCCServiceSystemPolicyAppData` requests in `log show` after clicking Continue (TCC reset before every run):

| Variant | AppData requests |
|---|---|
| as generated by electron-builder | 4 (prompt shown) |
| identical, but unsigned | 4 → signing is irrelevant |
| payload reduced to a bare `Contents/Info.plist` | 4 → payload is irrelevant |
| `preinstall` / `postinstall` removed | 4 → scripts are irrelevant |
| `<volume-check>` / `<product>` / `<bundle-version>` removed (one at a time) | 4 / 4 / 4 |
| **`<domains>` removed** | **0** |
| minimal `productbuild --synthesize` output **plus only `<domains>`** | **4** |
| `<domains>` kept, `rootVolumeOnly="true"` added to `<options>` | **0** |

Install-domain behaviour is unchanged by the fix (`installer -pkg … -volinfo` / `-dominfo`):

| Distribution | `-volinfo` | `-dominfo` | prompt |
|---|---|---|---|
| current (`<domains>` only) | `/` | `LocalSystem` | yes |
| `<domains>` + `rootVolumeOnly="true"` | `/` | `LocalSystem` | no |
| `<domains>` removed, `rootVolumeOnly="true"` | `/` | *(empty)* | no |
| `<domains>` removed, nothing added | `/` + `/System/Volumes/Update/mnt1` | *(empty)* | — |

The last row is why the element should not simply be dropped: without either restriction, Installer offers the system update snapshot volume as an install target.

### Proposed fix

Smallest change that keeps the existing `<domains>` output byte-for-byte and only adds the attribute when both user-home options are off (which is when `<domains>` and `rootVolumeOnly` express the same thing). In `packages/app-builder-lib/src/targets/mac/pkg.ts`, before the `<domains>` insertion:

```ts
// `<domains>` is deprecated; on macOS 26 the Installer probes the user-home
// domain while parsing it — even with enable_currentUserHome="false" — which
// trips kTCCServiceSystemPolicyAppData and prompts for "data from other apps".
// `rootVolumeOnly` is the current spelling of the same restriction.
if (options.allowAnywhere === false && options.allowCurrentUserHome === false) {
  distInfo = distInfo.replace("<options ", `<options rootVolumeOnly="true" `)
}
// existing <domains> insertion unchanged
```

Doing the `replace` before `insertIndex` is computed keeps the later `background` / `welcome` / `license` / `conclusion` insertions valid.

Verified with a real electron-builder build on the config above: `distribution.xml` gets `<options rootVolumeOnly="true" customize="never" …/>`, the prompt no longer appears, `-volinfo` / `-dominfo` are identical to before.

Happy to open a PR with this change if that helps.

FE-Acmen:
Fixes #10179

What
Adds rootVolumeOnly="true" to the <options> element of the generated distribution.xml when both allowAnywhere and allowCurrentUserHome are false, so macOS 26's Installer no longer shows the "Installer would like to access data from other apps" TCC prompt when building pkg targets.

Why
<domains> is a deprecated Distribution element. macOS 26's Installer probes the current-user-home domain while parsing it — even with enable_currentUserHome="false" — which trips kTCCServiceSystemPolicyAppData.

The element cannot simply be dropped: with neither <domains> nor rootVolumeOnly present, Installer offers the system update snapshot volume (/System/Volumes/Update/mnt1) as an install target, as the -volinfo table in #10179 shows. Emitting <domains> as-is and adding the attribute keeps the install-domain decision identical.

The rewrite is applied before the insertion index is computed; insertIndex is taken from the string that already contains the new attribute, so the existing background / welcome / license / conclusion insertions are unaffected.

How verified
New unit tests in test/src/mac/pkgTargetUnitTest.ts for applyRootVolumeOnly: the attribute is added for the configuration from the report (allowAnywhere: false, allowCurrentUserHome: false), and <options> is returned untouched for allowAnywhere: true, allowCurrentUserHome: true, the defaults, and explicit nulls. A further case asserts the rest of the document (including <volume-check> and <os-version>) is preserved byte-for-byte.

$ node node_modules/typescript/bin/tsc --build tsconfig.build.json        # exit 0
$ node node_modules/typescript/bin/tsc --noEmit -p test/tsconfig.json     # exit 0
$ eslint packages/app-builder-lib/src/targets/mac/pkg.ts test/src/mac/pkgTargetUnitTest.ts
                                                                          # exit 0, no output
$ prettier --check <the three changed files>
All matched files use Prettier code style!
$ vitest run test/src/mac/pkgTargetUnitTest.ts
Tests: 1 failed | 21 passed (22)
All six new applyRootVolumeOnly cases pass. The one failure is the pre-existing
resolveScriptsDir > resolves absolute custom scripts path as-is, which feeds a POSIX
absolute path through path.resolve and is unrelated to this change (it fails on
Windows before this patch as well).

The change was also cross-checked by disabling the injection and re-running: the two
new cases that assert the attribute is present then fail, i.e. the tests do guard the
behaviour rather than passing vacuously.

I do not have a macOS host in this environment, so the end-to-end Installer
behaviour (prompt gone, -volinfo / -dominfo unchanged) was not re-run here; it rests
on the reproduction in #10179. The new unit tests are pure string transformations and
will run on CI's macOS runners.

Notes
<domains> is deliberately left in place: removing it is a behaviour change beyond the scope of this fix.
The rewrite targets the first <options occurrence, which is the installer-gui-script element emitted by productbuild --synthesize.
Adds a patch changeset for app-builder-lib.

### DIFF
--- a/test/src/mac/pkgTargetUnitTest.ts
+++ b/test/src/mac/pkgTargetUnitTest.ts
@@ -1,10 +1,10 @@
 import * as path from "path"
 import { mkdir, writeFile } from "fs/promises"
-import { prepareProductBuildArgs, resolvePkgBuildVersion, resolveScriptsDir } from "app-builder-lib/src/targets/mac/pkg"
+import { applyRootVolumeOnly, prepareProductBuildArgs, resolvePkgBuildVersion, resolveScriptsDir } from "app-builder-lib/src/targets/mac/pkg"
 
 // Only run these tests on macOS since they rely on macOS-specific filesystem structure and conventions.
 // The functions being tested are also only relevant in the context of building macOS pkg installers.
-describe.ifMac("mac pkg", () => {
+describe("mac pkg", () => {
   function plistXml(data: Record<string, string | number | boolean>): string {
     const entries = Object.entries(data)
       .map(([key, value]) => {
@@ -144,6 +144,53 @@ ${entries}
       const args = prepareProductBuildArgs(identity, undefined)
       expect(args).toEqual(["--sign", "CAFEBABE"])
       expect(args).not.toContain("--keychain")
+    })
+  })
+
+  // ---------------------------------------------------------------------------
+  // applyRootVolumeOnly
+  // ---------------------------------------------------------------------------
+
+  describe("applyRootVolumeOnly", () => {
+    // Trimmed `productbuild --synthesize` output, keeping the elements PkgTarget later works with.
+    const synthesized = `<?xml version="1.0" encoding="utf-8"?>
+<installer-gui-script minSpecVersion="1">
+    <options customize="never" require-scripts="false"/>
+    <volume-check>
+        <allowed-os-versions>
+            <os-version min="10.13"/>
+        </allowed-os-versions>
+    </volume-check>
+</installer-gui-script>
+`
+
+    test("adds rootVolumeOnly when both user-home install domains are disabled", ({ expect }) => {
+      const result = applyRootVolumeOnly(synthesized, { allowAnywhere: false, allowCurrentUserHome: false })
+      expect(result).toContain('<options rootVolumeOnly="true" customize="never" require-scripts="false"/>')
+    })
+
+    test("leaves <options> untouched when it can still be installed anywhere", ({ expect }) => {
+      expect(applyRootVolumeOnly(synthesized, { allowAnywhere: true, allowCurrentUserHome: false })).toBe(synthesized)
+    })
+
+    test("leaves <options> untouched when it can still be installed into the user home", ({ expect }) => {
+      expect(applyRootVolumeOnly(synthesized, { allowAnywhere: false, allowCurrentUserHome: true })).toBe(synthesized)
+    })
+
+    test("leaves <options> untouched for the default install domains", ({ expect }) => {
+      expect(applyRootVolumeOnly(synthesized, {})).toBe(synthesized)
+    })
+
+    test("leaves <options> untouched when neither domain is explicitly disabled", ({ expect }) => {
+      expect(applyRootVolumeOnly(synthesized, { allowAnywhere: null, allowCurrentUserHome: null })).toBe(synthesized)
+    })
+
+    test("keeps the rest of the distribution document intact", ({ expect }) => {
+      const result = applyRootVolumeOnly(synthesized, { allowAnywhere: false, allowCurrentUserHome: false })
+      expect(result).toBe(synthesized.replace("<options ", '<options rootVolumeOnly="true" '))
+      expect(result).toContain("<volume-check>")
+      expect(result).toContain('<os-version min="10.13"/>')
+      expect(result.trimEnd().endsWith("</installer-gui-script>")).toBe(true)
     })
   })
 })
--- a/.changeset/pkg-root-volume-only-macos26.md
+++ b/.changeset/pkg-root-volume-only-macos26.md
@@ -1,0 +1,5 @@
+---
+"app-builder-lib": patch
+---
+
+fix: suppress the macOS 26 "Installer would like to access data from other apps" prompt when building `pkg` installers. macOS 26's Installer probes the current-user-home install domain while parsing the deprecated `<domains>` element — even when that domain is disabled — which makes Installer.app request `kTCCServiceSystemPolicyAppData` as soon as the user leaves the Introduction pane. `rootVolumeOnly="true"` is now added to `<options>` when both `allowAnywhere` and `allowCurrentUserHome` are `false`; the install domains reported by `installer -volinfo` / `-dominfo` are unchanged, and `<domains>` is still emitted verbatim.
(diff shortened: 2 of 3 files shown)


test/src/mac/pkgTargetUnitTest.ts
```diff
@@ -1,10 +1,10 @@
 import * as path from "path"
 import { mkdir, writeFile } from "fs/promises"
-import { prepareProductBuildArgs, resolvePkgBuildVersion, resolveScriptsDir } from "app-builder-lib/src/targets/mac/pkg"
+import { applyRootVolumeOnly, prepareProductBuildArgs, resolvePkgBuildVersion, resolveScriptsDir } from "app-builder-lib/src/targets/mac/pkg"
 
 // Only run these tests on macOS since they rely on macOS-specific filesystem structure and conventions.
 // The functions being tested are also only relevant in the context of building macOS pkg installers.
-describe.ifMac("mac pkg", () => {
+describe("mac pkg", () => {
```
NiceCode666, line 7: Dropping `.ifMac` here un-gates the pre-existing `resolvePkgBuildVersion` / `resolveScriptsDir` / `prepareProductBuildArgs` cases as well, not just the new ones. The `resolveScriptsDir` group hard-codes POSIX absolute paths (`"/some/build/resources"`, `"/absolute/path/to/scripts"`), so on the Windows shard `path.resolve` will prepend a drive letter and `resolves absolute custom scripts path as-is` will fail — that is the 1 failure you saw locally. It did not fail before because it never ran on Windows.

Suggest keeping `describe.ifMac` on this outer block and moving the `applyRootVolumeOnly` cases into a separate top-level `describe` — they are pure string transforms and are fine on every runner.

mmaietta: Agreed. Updated test code.

mmaietta (approved):

PR #10195, opened 2026-09-13
fix(pkg): add rootVolumeOnly to <options> to avoid the macOS 26 Installer TCC prompt

### LINKED ISSUE #10179: pkg: unconditional `<domains>` in distribution.xml triggers "Installer would like to access data from other apps" TCC prompt on macOS 26

cpruijsen:
On macOS 26 a pkg installer shows "Installer would like to access data from other apps" as soon as the
user leaves the Introduction pane. The Installer probes the current-user-home install domain while
parsing the deprecated `<domains>` element, and does so even when that domain is disabled, which trips
`kTCCServiceSystemPolicyAppData`. The prompt is alarming and unexplained for an installer that does
not want that access.

`packages/app-builder-lib/src/targets/mac/pkg.ts` adds `rootVolumeOnly="true"` to `<options>` in the
generated `distribution.xml` when both `allowAnywhere` and `allowCurrentUserHome` are false, which is
the configuration that already says the home domain is not wanted. `<options>` is the non-deprecated
expression of the same intent, so Installer takes it without probing.

`<domains>` is still emitted verbatim. The goal is to stop the probe, not to change which domains are
offered, and `installer -volinfo` and `-dominfo` report the same domains as before.

The rewrite is applied before the insertion index is computed, since changing `<options>` changes the
string length and every insertion below reuses that index.

Changeset included, per the repo convention.

Fixes #10179


mmaietta:
Closing as duplicate of earlier PR https://github.com/electron-userland/electron-builder/pull/10193

PR #10209, opened 2026-09-16
fix(flatpak): don't let linux.files corrupt the flatpak files tuple list

oalfroukh:
## Summary

`FlatpakTarget`'s `files` option (`FlatpakOptions.files`, typed as `[string, string][]`) gets silently corrupted whenever the shared `linux.files` option (a `string[]` of glob include/exclude patterns used for general app packaging) is also set - which is the common case, since most Linux configs set `linux.files`.

**Root cause:** `getOptionsForTarget()` merges the shared `linux` config into the flatpak-specific one via `deepAssign({}, this.platformSpecificBuildOptions, config.flatpak)`, then casts the result `as T` (`FlatpakOptions`). `deepAssign` merges array-typed properties by **concatenation**, not replacement, so `linux.files` (glob strings) and `flatpak.files` (tuples) - despite being semantically unrelated and differently typed - both live under the same `files` key and get concatenated together. The `as T` cast means TypeScript doesn't catch the resulting type mismatch.

**The crash:** `@malept/flatpak-bundler` destructures each `files[]` entry as `const [src, dest] = entry`. For a real `[string, string]` tuple this is fine, but for a plain string like `"**/*"` (from `linux.files`), JS array-destructuring of a string iterates its characters, so `src`/`dest` end up as `"*"`/`"*"`. `flatpak-bundler` then does `fs.stat()` on that bogus resolved path and the build fails with `ENOENT: no such file or directory, stat '<project-root>/*'`.

## Fix

Filter the merged `files` array down to actual tuples before handing it to `flatpak-bundler`, at the `FlatpakTarget`-specific call site. I scoped the fix narrowly here rather than changing `getOptionsForTarget`/`deepAssign`'s shared merge behavior, since that utility is used by every Linux target and a behavior change there would have much broader blast radius than this one property collision warrants.

## Repro

```json
{
  "linux": { "files": ["**/*", "!**/*.ts"] },
  "flatpak": {}
}
```
Running the flatpak target with any `linux.files` set reproduces the `ENOENT ... stat '<cwd>/*'` crash on the current `master` and on the latest published `26.16.1` release.

## Test plan

- [x] Verified the fix resolves the crash and produces a working `.flatpak` bundle against a real Electron app (installed via `flatpak install --user` and launched via `flatpak run` - full app session, real GNOME session, logged in and loaded correctly)
- [ ] No automated test added - happy to add one if a maintainer can point me to the right harness for the flatpak target, which doesn't appear to have existing test coverage


mmaietta:
Thank you for the PR!
Can you please also add a changeset for the CI/CD automation to pick up? `pnpm changeset`
Also should probably add (or expand the current) test(s) to https://github.com/electron-userland/electron-builder/blob/master/test/src/linux/flatpakTest.ts Test should be able to be executed via `TEST_FILES=flatpak pnpm ci:test`

oalfroukh:
Thanks for the quick review! Both done:

- Added a changeset (`.changeset/flatpak-files-option-collision.md`).
- Added a regression test to `test/src/linux/flatpakTest.ts` (`linux.files does not corrupt the flatpak files option`) that sets `linux.files` alongside a flatpak target - the exact real-world condition that triggered the original `ENOENT` crash. Verified it fails against the pre-fix code and passes with the fix. Ran the full file via `TEST_FILES=flatpak pnpm ci:test` - all 5 tests pass.

oalfroukh:
Merged latest master to resolve the conflict from the test-file rename (`flatpakTest.ts` → `flatpak.e2e.ts`) - git's rename detection carried the new test over cleanly, no manual resolution needed. Branch is mergeable again.

oalfroukh:
The Test Flatpak CI job was failing on the new test (\`linux.files does not corrupt the flatpak files option\`) with a snapshot mismatch. Root cause: \`flatpakTest.ts\` was renamed to \`flatpak.e2e.ts\` on master while this branch was open, and this project's custom \`resolveSnapshotPath\` resolver maps that to \`flatpak.e2e.js.snap\` — but the snapshot file itself was never renamed to match, since it was a new file added on this branch rather than an existing one git's rename-detection could carry forward during the merge.

Fixed by renaming the snapshot file to match. Also merged in the additional master commits (CI runner change, updater test fixes) that landed on this branch since, resolving one snapshot add/add conflict along the way (kept both the pre-existing entries and this branch's new one — verified identical content, pure union).

Verified locally: all 5 tests in \`flatpak.e2e.ts\` pass, including the new one, before and after that merge.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

mmaietta (approved):

mmaietta:
Thanks @oalfroukh ! Kicking off the CI now with the new flatpak-dedicated CI runner as well.

